PR #2, opened 2017-11-15
Strong-mode analysis_options.yaml

isoos:
Use `dartanalyzer .` to check for errors.


luanpotter:
Great addition, @isoos. Thanks to @feroult commit, I think it should now pass this lint without errors!

PR #14, opened 2018-07-16
Initial Gamepad support

erickzanardo:
This PR adds an initial gamepad support on Flame.

Only Android is supported on this PR, I don`t think iOs support will be possible only using the resources provided by flutter, therefore some native code will be needed.

On Android all keys except the axes are supported, which I guess will also require native code.


luanpotter:
Thanks!!!

PR #19, opened 2018-09-27
Fix widget change

bparrishMines:
I had an issue where when switching between two different game widgets, the widget would not change. e.g.

* _onGameOver is a method called when the user finishes that minigame
```dart
class GamePageState extends State<GamePage> {
  int _nextGame = 0;
  MyBaseGame _currentGame;
  List<String> games = <String>['Game1', 'Game2'];

  @override
  void initState() {
    super.initState();
    _currentGame = _genGame(games[_nextGame++]);
  }

  void _onGameOver(double seconds) {
    if (_nextGame >= games.length) {
      // Do something
    } else {
      setState(() {
        _currentGame = _genGame(games[_nextGame++]);
      });
    }
  }

  MyBaseGame _genGame(String gameName) {
    switch (gameName) {
      case 'Game1':
        return Game1(onGameOver: _onGameOver);
      case 'Game2':
        return Game2(onGameOver: _onGameOver);
      default:
        throw UnimplementedError();
    }
  }

  @override
    return _currentGame.widget;
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [0.9.3]
+
+- Fixes issue when switching between games where new game would not attach
+
 ## [0.9.2]
 - Fixes to work with Dart 2.1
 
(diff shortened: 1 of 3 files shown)


bparrishMines:
I added a method that was included in other RenderObjects and it solved the problem for me.

luanpotter:
Excellent! Thanks for the contribution!

bparrishMines:
No problem! Thanks for being so quick!

PR #23, opened 2018-10-19
Add elapsed time to Animation

ianliu:
Disclaimer: Hello! This is my first PR to this repository, hope it is useful.

I was implementing [flame-example](https://github.com/luanpotter/flame-example) with the latest release of Flame (0.9.4), and the `Animation.lifeTime` property doesn't exists anymore. I tried swapping with the `.clock` property, but I think it changed semantics to count the elapsed time since last frame, so I ended up creating another property, `.ellapsed`, to indicate how long (in seconds) an animation has been running. This allows flame-example to work again.

### DIFF
--- a/lib/animation.dart
+++ b/lib/animation.dart
@@ -25,6 +25,9 @@ class Animation {
   /// It's ticked by the update method. It's reset every frame change.
   double clock = 0.0;
 
+  /// Total elapsed time of this animation, in seconds, since start or a reset.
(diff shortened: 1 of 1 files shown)


luanpotter:
That's a great addition, thanks a lot! Makes perfect sense to have that property.

Also, if you have any progress on updating flame-example to the latest version, please let me know or open a PR there, it's long overdue to update that repository :))

PR #37, opened 2019-01-23
Updating Text docs

### LINKED ISSUE #36: Update text apis docs
Docs for the Text APIs is outdated, we must update it to the latest implementation

erickzanardo:
Fixes #36

### DIFF
--- a/doc/text.md
+++ b/doc/text.md
@@ -36,35 +36,11 @@ config.render(canvas, Offset(size.width - 10, size.height - 10);
 
 ## Text Components
 
-Flame provides two text components that make it even easier to render text in your game: `TextBoxComponent` and `TextComponent`.
-
-
-### TextBoxComponent
-
-`TextBoxComponent` like its name suggest, is used to render a text inside a bounding box, creating line breaks according to the provided box size.
-
-Example usage:
-
-```dart
-class MyTextBox extends TextBoxComponent {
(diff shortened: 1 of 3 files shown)


doc/text.md
```diff
@@ -0,0 +1,90 @@
...
+    Color color = material.Colors.black;
+```
+
+After the creation of the config you can use its `render` method to draw some string on a canvas:
+
+```dart
+config.render(canvas, Offset(size.width - 10, size.height - 10);
+```
+
+## Text Components
```
luanpotter: I would change the order, as TextComponent is simpler and easier to grasp before TextBoxComponent.

doc/text.md
```diff
@@ -0,0 +1,90 @@
...
+
+You can use Dart's hexadecimal notation to make it really easy; for instance: `0xFF00FF00` is fully opaque green (the 'mask' would be `0xAARRGGBB`).
+
+There is a color enum to make it easy to use common colors; it is in the material flutter package:
+
+```dart
+    import 'package:flutter/material.dart' as material;
+
+    Color color = material.Colors.black;
+```
```
luanpotter: I like this explanation, but we could take the opportunity to link to a Color Palette documentation to instruct to better using colors and paints :)

erickzanardo: I am not very aware of the palette APIs, will merge this right now and we do the palette docs later. This way at least we remove the old and removed part of the text doc :D

erickzanardo:
Updating text documentation

PR #42, opened 2019-02-20
Adding patreon page

erickzanardo:


luanpotter:
What about putting a discord link in the patreon page? :)

erickzanardo:
Will do that, Actually I thought Patreon would do this automatically when I added the Patreon Bot on our discord, but I guess this did no happen. But this don't prevents us for already merging this right?

luanpotter:
Agreed!

Issue #45, opened 2019-02-21
Project startup doc


erickzanardo:
Create documentation on how to start a new game using Flame.

There could be two tutorials here
1- A classic game only application
2- Adding a game as a widget inside an existing Flutter app

erickzanardo:
Added a Getting started section on README with great articles: https://github.com/luanpotter/flame#getting-started

PR #50, opened 2019-03-05
Game as a Widget

feroult:
Deprecating the `new Game().widget` syntax.

### DIFF
--- a/lib/game.dart
+++ b/lib/game.dart
@@ -1,13 +1,12 @@
-import 'dart:ui';
 import 'dart:math' as math;
+import 'dart:ui';
 
 import 'package:flutter/gestures.dart';
 import 'package:flutter/rendering.dart';
 import 'package:flutter/scheduler.dart';
 import 'package:flutter/widgets.dart';
-
-import 'package:ordered_set/ordered_set.dart';
 import 'package:ordered_set/comparing.dart';
+import 'package:ordered_set/ordered_set.dart';
 
 import 'components/component.dart';
 import 'position.dart';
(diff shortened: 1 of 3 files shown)


lib/game.dart
```diff
@@ -36,23 +38,24 @@ abstract class Game {
...
-  Widget get widget {
-    if (_widget == null) {
-      _widget = new Center(
-          child: new Directionality(
-              textDirection: TextDirection.ltr,
-              child: new _GameRenderObjectWidget(this)));
-    }
-    return _widget;
-  }
+  @Deprecated('Use Game is a Widget')
```
erickzanardo: Use game as a Widget?

feroult: FIxed :)

luanpotter:
This breaks (or rather deprecates) all the examples and docs, we have to fix those!

feroult:
Ok! Let's do that. But what do you think? Is it better to rollback to the `.widget` api?

luanpotter:
No, let's leave it as that, LGTM. But keeping the old deprecated for now, when we release 0.11 we can remove the deprecated version :)

erickzanardo:
Should web release `0.10.3` soon? I am thinking on resolving #45 and I think would be good to create this doc already contemplating this changes. What do you think?

feroult:
Agree.

mtcliatt:
What was the motivation for this change? (just curious)

erickzanardo:
I think it is because it is more semantic correctly. I mean, the way you add a Flame game was by adding a widget in the Flutter app. A GameWidget seems more easy to understand, than a property widget stored inside Game class.

PR #53, opened 2019-03-07
Improving readme

erickzanardo:
This PR does a couple of things

- Removing the Paypall button since it is not working, let put it back when we fix it
- Separate the documentation and getting started into sections so it draws more attention
- Adding Alekhin's articles as tutorials since it very complete and updated


erickzanardo:
I think this also resolves #45 , what do you think?

luanpotter:
LGTM, also, agree on fixing #45. Thanks!

PR #68, opened 2019-03-24
New audio API

erickzanardo:
This PRs creates a new audio API for Flame, this will provide a new method to play sound effects, which will take advantages of the new player mode low latency from the audioplayers lib.


lib/flame_audio.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:audioplayers/audioplayers.dart';
+import 'package:audioplayers/audio_cache.dart';
+
+class FlameAudio {
+  AudioCache audioCache = AudioCache(prefix: 'audio/');
+
+  Future<AudioPlayer> play(String file) {
```
luanpotter: What about playMusic? to make a pair with playSfx?

lib/flame_audio.dart
```diff
@@ -0,0 +1,18 @@
...
+    return audioCache.play(file);
+  }
+
+  Future<AudioPlayer> loop(String file) {
```
luanpotter: for example, the loop is not clear to me which api it will use without checking the code. what about loopMusic?

lib/flame_audio.dart
erickzanardo: So your suggestion is to replace `play` by  `playMusic`?

luanpotter: Yes, that's my suggestion, what do you think?

erickzanardo: Agreed!

doc/audio.md
```diff
@@ -28,7 +30,7 @@ This uses the [audioplayers](https://github.com/luanpotter/audioplayer) lib, in
 If you want to play indefinitely, just use `loop` function:
 
 ```dart
-    Flame.audio.loop('music.mp3');
+    Flame.audio.loopMusic('music.mp3');
```
feroult: Is this using the low latency api, right? Wouldn't it be better to use `loopSfx`?

Because it will have the same limits of the `playSfx` in terms of file size...

luanpotter: also using loopMusic won't fix the gap between iterations, now that I think about it...

erickzanardo:
@luanpotter @feroult I've update this PR with the changes that we discussed today, let me know what you think please.

doc/audio.md
```diff
@@ -28,9 +31,12 @@ This uses the [audioplayers](https://github.com/luanpotter/audioplayer) lib, in
 If you want to play indefinitely, just use `loop` function:
 
 ```dart
-    Flame.audio.loop('music.mp3');
+    Flame.audio.loop('music.mp3'); // Or
+    Flame.audio.loopLongAudio('music.mp3');
 ```
 
+The difference between the `play/loop` and `playLongAudio/loopLongAudio` is that `play/loop` makes uses of optimezed features that allow sounds to be looped without gaps between their iterations, and no drop on the game frame rate will happen, you should whenever possible, prefer these methods.  `playLongAudio/loopLongAudio` allows any audio of any length to be played, but they do create frame rate drop, and the looped audio will feature a small gap between iterations.
```
luanpotter: `and no drop on the game frame` that's a bold overstatement there! hahaha maybe `and almost no drop`, or `smaller drop`;
frame rate will happen, you should whenever... => frame rate will happen. You should, whenever possible, prefer `play/loop` methods.
ds.  `playLongAudio/loopLongAudio` => double spaces there.
allows any audio of any length  => allows for audios of any length

feroult: Agree. :)

erickzanardo:
@luanpotter fixed the docs improvement and also changed this PR to be merged into `develop` branch.

PR #84, opened 2019-05-03
add part 5 from alekhin series to readme

impulse:
As suggested in https://github.com/luanpotter/flame/pull/83


erickzanardo:
Thanks @impulse :D

PR #103, opened 2019-06-09
Extra functionality in position class for vector rotation, normalization, degrees, and common operators

illiapoplawski:
Added operator support for common math functions +,-,*,/.  Added utility method to normalize a vector.  Added ability to get angle of vector from origin or from another vector.  Added ability to rotate a vector by a delta angle rather than just the absolute angle from the origin.  Added support for degrees on top of radians for angle methods.

I wasn't sure what was best way to name some of the methods using degrees instead of radians so any feedback is welcome.  Also, I like using degrees in certain cases but if you don't find it necessary it's something that I'm willing to remove from this pull request.  This is mostly just a collection of methods to simplify more common tasks.

### DIFF
--- a/lib/position.dart
+++ b/lib/position.dart
@@ -87,6 +87,49 @@ class Position {
     return this;
   }
 
+  /// Rotate around origin; [angle] in degrees.
+  Position rotateDeg(double angle) {
+    return rotate(angle * math.pi / 180);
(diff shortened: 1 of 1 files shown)


illiapoplawski:
Accidentally put it in master.

PR #104, opened 2019-06-09
Extra functionality in position class for vector rotation, normalization, degrees, and common operators

illiapoplawski:
Added operator support for common math functions +,-,*,/. Added utility method to normalize a vector. Added ability to get angle of vector from origin or from another vector. Added ability to rotate a vector by a delta angle rather than just the absolute angle from the origin. Added support for degrees on top of radians for angle methods.

I wasn't sure what was best way to name some of the methods using degrees instead of radians so any feedback is welcome. Also, I like using degrees in certain cases but if you don't find it necessary it's something that I'm willing to remove from this pull request. This is mostly just a collection of methods to simplify more common tasks.

### DIFF
--- a/lib/position.dart
+++ b/lib/position.dart
@@ -87,6 +87,49 @@ class Position {
     return this;
   }
 
+  /// Rotate around origin; [angle] in degrees.
+  Position rotateDeg(double angle) {
+    return rotate(angle * math.pi / 180);
+  }
+
+  /// Change current rotation by delta angle; [angle] in radians.
+  Position rotateDelta(double radians) {
+    return rotate(angle() + radians);
+  }
+  
+  /// Change current rotation by delta angle; [angle] in degrees.
+  Position rotateDeltaDeg(double degrees) {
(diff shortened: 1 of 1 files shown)


lib/position.dart
```diff
@@ -87,6 +87,49 @@ class Position {
...
+
+  /// Returns the angle of this vector from another [Position] in radians.
+  double angleFrom(Position other) {
+    //Technically the origin is not a vector so you cannot find the angle between
+    //itself and another vector so just default to calculating the angle between
+    //the non-origin point and the origin to avoid division by zero
+    if(x == 0 && y == 0)
+      return other.angle();
+    if(other.x == 0 && other.y == 0)
+      return angle();
```
luanpotter: just for linter compliance:
1. comments `// xxx` (space)
2. `if (xxx) {` (space & add braces)

illiapoplawski: Done!

luanpotter:
Thanks, @illiapoplawski ;)

PR #112, opened 2019-06-25
Adding Timer Class

erickzanardo:
This PR adds a very simple utility class, to help handling timers and interval events.

Right now I am only adding the concrete Timer class because I am needing this on a project, but this could eventually evolve to be a mixin/component.

### DIFF
--- a/lib/time.dart
+++ b/lib/time.dart
@@ -1,19 +1,36 @@
 
+/// Simple utility class that helps handling time counting and implementing interval like events.
+///
 class Timer {
   final double _limit;
   void Function() _callback;
-  double _current;
+  bool _repeat;
+  double _current = 0;
   bool _running = false;
 
-  Timer(this._limit, this._callback);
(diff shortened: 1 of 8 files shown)


lib/time.dart
```diff
@@ -0,0 +1,51 @@
...
+/// Simple utility class that helps handling time counting and implementing interval like events.
+///
+class Timer {
+  final double _limit;
+  void Function() _callback;
+  bool _repeat;
+  double _current = 0;
+  bool _running = false;
+
+  Timer(this._limit, { repeat = false, callback }) {
```
luanpotter: add types to repeat and callback on method signatures, please

lib/time.dart
```diff
@@ -0,0 +1,51 @@
...
+    _repeat = repeat;
+    _callback = callback;
+  }
+
+  double get current {
```
luanpotter: double get current => _current;

luanpotter: simpler

PR #130, opened 2019-07-12
Sprite as widget

erickzanardo:
I found no gain creating something similar for animations, what we have right now seems to be the best for animations already.

But for Sprites, I found that this is a good solution.

TODO: Still need to update docs.

### DIFF
--- a/lib/util.dart
+++ b/lib/util.dart
@@ -6,6 +6,7 @@ import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart' as widgets;
 
 import 'animation.dart';
+import 'sprite.dart';
 import 'components/animation_component.dart';
 import 'game.dart';
 import 'position.dart';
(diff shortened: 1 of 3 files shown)


lib/util.dart
```diff
@@ -88,4 +89,27 @@ class Util {
...
+
+  @override
+  void paint(Canvas canvas, Size size) {
+    if (_sprite.loaded()) {
+      _sprite.render(canvas, size.width, size.height);
+    }
+  }
+
+  @override
+  bool shouldRepaint(widgets.CustomPainter old) => false;
```
luanpotter: if shouldRepaint returns false always, _sprite should be final... what do you think?

erickzanardo: Sounds good, I will change that later

erickzanardo: Done!

erickzanardo:
I have also added some info on docs about sprite and animations as widgets.

PR #135, opened 2019-07-25
Adding override paint method to Sprite

erickzanardo:
This is a simple PR that adds a new parameter to the renders methods of the Sprite class.

Although the Sprite class already have a paint attribute, sometimes you may want to change it just for a single render, in my opinion having a parameter for that is more convenient than setting the attribute, rendering the sprite and reseting the attribute.

You can see an usage of this need on this feature: https://github.com/fireslime/bounce_box/commit/a638e6f1bfb4722f062dea09e0c8a6c6cd2570a9#diff-d02dcf32571d773232d836d3c98753e1R46

I wanted to use a color filter to make some sprites darker like you can see on the following image:

Screen Shot 2019-07-24 at 23 11 24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## next
 - Refactoring ParallaxComponent (thanks @spydon)
 - Fixing flare animation with embed images
+- Adding override paint parameter to Sprite, and refactoring it have named optional parameters
 
 ## 0.14.2
 - Refactoring BaseGame debugMode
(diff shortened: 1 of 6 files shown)


erickzanardo:
@luanpotter I have made the changes you asked, take a look please.

We need to remember to increase flame to 1.5.0 on the next version, because this refactor adds breaking changes. But I think it is ok, since we plan on adding more parameters here, having then named is the better  approach for a health API

luanpotter:
However, please fix the build!

PR #138, opened 2019-08-02
ParallaxComponent that can move layers in 2D

spydon:
This rewritten parallax component supports scrolling the layers of the parallax in both dimensions.
It also has extensive configurability of how the images should align, fill and repeat on the screen to fit all different types of games.


doc/components.md
```diff
@@ -116,22 +116,55 @@ class GameOverPanel extends PositionComponent with Resizable, ComposedComponent
...
+  final images = [
+    ParallaxImage("mountains.jpg"),
+    ParallaxImage("forest.jpg"),
+    ParallaxImage("city.jpg"),
+  ];
+  this.bg = ParallaxComponent(images);
 ```
 
-Then, render it as any other component.
+This creates a static background, if your want it to move you have to set the named optional parameters `baseSpeed` and `layerDelta`. For example if you want to move your background images along the X-axis and have the images further away you would do the following:
```
erickzanardo: There is a little typo here: `if your want` -> `if you want`

doc/components.md
```diff
@@ -116,22 +116,55 @@ class GameOverPanel extends PositionComponent with Resizable, ComposedComponent
...
+  this.bg = ParallaxComponent(images, baseSpeed: Offset(50, 0), layerDelta: Offset(20, 0));
+```
+You can set the baseSpeed and layerDelta at any time, for example if your character jumps or your game speeds up.
+
+```dart
+  this.bg.baseSpeed = Offset(100, 0);
+  this.bg.layerDelta = Offset(40, 0);
+```
+
+By default the images are aligned to the bottom left, repeated along the X-axis and scaled proportionally so that the image covers the height of the screen. If you want to change this behaviour, for example if you are not making a side scrolling game, you set the `repeat`, `alignment` and `fill` parameters for each ParallaxImage.
```
erickzanardo: I think you could change`you set` to `you can set`

PR #140, opened 2019-08-08
Exposing tile objects layer 

renatoferreira656:
Inside `tiled_component.dart` is being used the lib from `https://github.com/feroult/tiled.dart` and currentely the objects layer was not being expose, I created this a code to expose it and  I updated the tile example to show how to use it


doc/examples/tiled/lib/main.dart
```diff
@@ -1,14 +1,43 @@
...
 
 void main() {
+  Flame.images.load('coins.png');
   final TiledGame game = TiledGame();
   runApp(game.widget);
 }
 
 class TiledGame extends BaseGame {
+  TiledComponent tiledMap;
+  Animation a;
```
erickzanardo: Since you are using `BaseGame` here, you could use an instance of `AnimationComponent` for each object instead of manual drawing the coins on the game.

lib/components/tiled_component.dart
```diff
@@ -69,15 +69,19 @@ class TiledComponent extends Component {
...
+      final src = Rect.fromLTWH(rect.left.toDouble(), rect.top.toDouble(), rect.width.toDouble(), rect.height.toDouble());
+      final dst = Rect.fromLTWH(tile.x.toDouble(), tile.y.toDouble(), rect.width.toDouble(), rect.height.toDouble());
 
       c.drawImageRect(image, src, dst, paint);
     });
   }
 
   @override
   void update(double t) {}
+  ObjectGroup getObjectLayerByName(String name) {
```
erickzanardo: Insert a line above here please

erickzanardo: Also this method name is a little confusing for me, what do you think of `getObjectGroupFromLayer`?

lib/components/tiled_component.dart
```diff
@@ -69,15 +69,19 @@ class TiledComponent extends Component {
       final image = images[tile.image.source];
 
       final rect = tile.computeDrawRect();
-      final src = Rect.fromLTWH(rect.left.toDouble(), rect.top.toDouble(),
-          rect.width.toDouble(), rect.height.toDouble());
-      final dst = Rect.fromLTWH(tile.x.toDouble(), tile.y.toDouble(),
-          rect.width.toDouble(), rect.height.toDouble());
+      final src = Rect.fromLTWH(rect.left.toDouble(), rect.top.toDouble(), rect.width.toDouble(), rect.height.toDouble());
+      final dst = Rect.fromLTWH(tile.x.toDouble(), tile.y.toDouble(), rect.width.toDouble(), rect.height.toDouble());
```
erickzanardo: I think the only thing that changed here was the formatting right? This is what probably broke the build.

erickzanardo:
Also 2, your PR seems to have some commits from Luan, I think you will need to rebase your branch with the develop.

lib/components/tiled_component.dart
renatoferreira656: I was trying to dettach from the lib, actually I think the ObjectGroup isnt a good name, despite  objectgroup is the name of the tag in the xml, in the UI they use object layer, so I tried to look more like the UI than the XML, but I can change, no problem.

doc/examples/tiled/lib/main.dart
renatoferreira656: Huummm.. I didnt get that, but I will see how it done.

lib/components/tiled_component.dart
renatoferreira656: okk.. I will see... to be honest I have to look how can I run the full build in my PC. :(

erickzanardo: I just looked on travis, build failed because of files formatted outside our format, the problem is probably that code section where I commented.

Even so, to format the files just run `flutter format .`

doc/examples/tiled/lib/main.dart
```diff
@@ -1,14 +1,43 @@
...
   TiledGame() {
-    add(TiledComponent('map.tmx'));
+    tiledMap = TiledComponent('map.tmx');
+    add(tiledMap);
+    a = Animation.sequenced('coins.png', 8, textureWidth: 20);
```
erickzanardo: To use the animations components I said, just put the following code here on the constructor:

```dart
    final ObjectGroup obj = tiledMap.getObjectLayerByName("AnimatedCoins");
    if (obj == null) {
      return;
    }
    for (TmxObject obj in obj.tmxObjects) {
      final comp = AnimationComponent(30.0, 30.0, a);
      comp.x = obj.x.toDouble();
      comp.y = obj.y.toDouble();
      add(comp);
}      
```

Then you will not need to override the render method.

erickzanardo:
Looks good!

PR #162, opened 2019-09-24
Tapeable to Tapable

spydon:
Tapeable would be something that you can tape, like with sticky tape, so I suggest we rename it to Tapable. Which also follows the convention for how it is named in other languages and libraries.


erickzanardo:
Thanks @spydon

PR #163, opened 2019-09-27
Looping background music management class

japalekhin:
~~This is a work in progress and not ready to be merged. Just putting it out here for review, testing, and further discussion.~~

Edit: Just finished testing and it seems to work. Needs more testing before it's considered ready.

**Edit: I'll mark this one as ready**. Even though I've had enough testing, I encourage people to do more tests though.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -42,9 +42,10 @@ And start using it!
    - [Structure](structure.md)
    - [Game Loop](game.md)
    - [Components](components.md)
-   - [Audio](audio.md)
-     - [Looping Background Music](bgm.md)
    - [Input](input.md)
+ * Audio
+   - [General Audio](audio.md)
+   - [Looping Background Music](bgm.md)
  * Rendering
    - [Images, Sprites and Animations](images.md)
    - [Text Rendering](text.md)
--- a/doc/README.md
+++ b/doc/README.md
@@ -43,6 +43,7 @@ And start using it!
    - [Game Loop](game.md)
    - [Components](components.md)
    - [Audio](audio.md)
+     - [Looping Background Music](bgm.md)
    - [Input](input.md)
  * Rendering
    - [Images, Sprites and Animations](images.md)
--- a/doc/bgm.md
+++ b/doc/bgm.md
@@ -1,0 +1,48 @@
+# Looping Background Music
+
+To play a looping background music track, run:
+
+```dart
+import 'package:flame/flame.dart';
+
+Flame.bgm.play('adventure-track.mp3');
+```
+
+Or, if you prefer:
+
+```dart
+import 'package:flame/bgm.dart';
+
+Bgm audio = Bgm();
+audio.play('adventure-track.mp3');
+```
+
+**Note:** The `Bgm` class will always use the static instance of `FlameAudio` in `Flame.audio` for storing cached music files.
+
+You must have an appropriate folder structure and add the files to the `pubspec.yaml` file, as explained in [Flame Audio documentation](audio.md).
+
+The following functions can be used to preload (and unload) music files into the cache. These functions are just aliases for the ones in `Flame.audio` with the same names.
+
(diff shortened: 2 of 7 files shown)


lib/bgm.dart
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /// Plays and loops a background music file specified by [filename].
+  ///
+  /// The volume can be specified in the optional named parameter [volume]
+  /// where `0` means off and `1` means max.
+  ///
+  /// It is safe to call this function even when a current BGM track is
+  /// playing.
+  void play(String filename, {double volume}) async {
```
erickzanardo: why not just set the default value here? Like

```dart
  void play(String filename, {double volume = 1}) async {
```

lib/bgm.dart
```diff
@@ -0,0 +1,101 @@
...
+    isPlaying = true;
+    audioPlayer = await Flame.audio.loopLongAudio(
+      filename,
+      volume: volume,
+    );
+  }
+
+  /// Stops the currently playing background music track (if any).
+  void stop() {
+
```
erickzanardo: audioPlayer.stop() ?

japalekhin: yup added :D (and converted to async)

lib/bgm.dart
japalekhin: the first reason is that i didn't know i can do that :D
but i'm keeping it as is because of the second reason, which is: during development time, passing `null` to a `double` parameter is perfectly valid but could cause a run-time error. the null-aware assignment is a safer "catch code" making the default value in the parameter definition unnecessary.

doc/bgm.md
```diff
@@ -0,0 +1,96 @@
+# Looping Background Music
```
luanpotter: add this file to the doc/REAME.md list under Topics. That will add it to the flame docs site! You can use nesting if desired

lib/flame.dart
```diff
@@ -23,6 +24,13 @@ class Flame {
   /// Access a shared instance of the [FlameAudio] class.
   static FlameAudio audio = FlameAudio();
 
+  /// Access a shared instance of the [Bgm] class.
+  static Bgm _bgm;
+  static Bgm get bgm {
```
luanpotter: I might be wrong (please test carefully) but I think you can return the assignament and then you can just d
`static Bgm get bgm => _gm ??= Bgm();`

japalekhin: checking it out

japalekhin: i tested it and it seems to work:
```dart
int asdf;
int get aassddff => (asdf ??= 5);

print(asdf);
print('------------');
print(aassddff);
print('------------');
print(asdf);
```

got (as expected):
```dart
I/flutter ( 4439): null
I/flutter ( 4439): ------------
I/flutter ( 4439): 5
I/flutter ( 4439): ------------
I/flutter ( 4439): 5
```

any problem with multi-line getter functions though?

doc/bgm.md
japalekhin: added!

lib/flame.dart
luanpotter: absolutely no problem, but if we can make any code more concise and legible, I prefer that :) but nothing specific to multi-line getter functions

doc/README.md
```diff
@@ -43,6 +43,7 @@ And start using it!
    - [Game Loop](game.md)
    - [Components](components.md)
    - [Audio](audio.md)
+     - [Looping Background Music](bgm.md)
```
luanpotter: I think you can't have sub-categories like that, I should have explained better (this being parsed for the html on the flame website is the reason).
You can have
```md
 * Audio
  - [General Audio/or Audio/or something](audio.md)
  - [Looping Background Music](bgm.md)
```

japalekhin: fixed like your suggestion https://github.com/flame-engine/flame/pull/163/commits/544fdb72496d3f39470ab1c79c1269a24da750e1

Issue #64, opened 2019-03-21
Add option to flip position components on prepareCanvas


luanpotter:
This is a very useful override of the prepareCanvas method that I've been using:

```dart
  // TODO pull up to flame! this is **very** useful
  @override
  void prepareCanvas(Canvas canvas) {
    canvas.translate(x, y);
    canvas.translate(width/2, height/2);
    canvas.scale(inverted ? 1.0 : -1.0, 1.0);
    canvas.translate(-width/2, -height/2);
  }
```

This flips horizontally. PositionComponent should have a field to control that and allow for easy flipping of sprites, both horizontally and vertically (or both).

It would be a really good to have feature.

erickzanardo:
Fixed on #166

erickzanardo:
Thanks @kurtome

PR #166, opened 2019-10-06
Add render flip X or Y axis to position component

kurtome:
This is based on https://github.com/flame-engine/flame/issues/64, I'm using a hack to do the same thing based on the code from the issue already :)


luanpotter:
@kurtome thanks for your contribution! I don't think it's a hack at all, though it does look like it haha

Can you add an example app & docs (couple of lines under the PositionComponent should suffice) for that new feature?

Code looks good to me :)

lib/components/component.dart
```diff
@@ -128,6 +130,11 @@ abstract class PositionComponent extends Component {
     final double dy = -anchor.relativePosition.dy * height;
     canvas.translate(dx, dy);
 
+    // Handle inverted rendering by moving center and flipping.
+    canvas.translate(width / 2, height / 2);
```
renancaraujo: If both `renderFlipX` and `renderFlipY` are false we should avoid executing all of this code.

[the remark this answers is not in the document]
kurtome: Sounds good, 👍

kurtome: Sounds good, 👍

kurtome:
Updated according to comments, added example, doc and only doing the canvas calculations if flipping is necessary.

erickzanardo:
@kurtome I just noticed that you created this PR from the master branch, would you please rebase your PR with the develop branch? You doing that we can merge it!

kurtome:
@erickzanardo oops, I had the wrong base again, but this time i remembered to branch off of `develop`. Base for this PR updated to `develop` and only seeing my commits.

erickzanardo:
Thanks!

PR #172, opened 2019-10-14
Fixes and nit-picks

luanpotter:
This fixes a lot of warnings and small problems throughout the code.

Let's always keep it clean! :)

### DIFF
--- a/doc/examples/debug/lib/main.dart
+++ b/doc/examples/debug/lib/main.dart
@@ -46,7 +46,8 @@ class AndroidComponent extends SvgComponent with Resizable {
 }
 
 class MyGame extends BaseGame {
-  final TextConfig fpsTextConfig = const TextConfig(color: const Color(0xFFFFFFFF));
+  final TextConfig fpsTextConfig =
+      const TextConfig(color: const Color(0xFFFFFFFF));
 
   MyGame() {
     _start();
--- a/doc/examples/debug/lib/main.dart
+++ b/doc/examples/debug/lib/main.dart
@@ -11,7 +11,9 @@ import 'package:flutter/material.dart';
 void main() async {
   await Flame.util.initialDimensions();
 
-  runApp(MyGame().widget);
+  final myGame = MyGame();
+  runApp(myGame.widget);
+  myGame.start();
 }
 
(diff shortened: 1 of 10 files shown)


doc/examples/debug/lib/main.dart
```diff
@@ -54,16 +57,16 @@ class MyGame extends BaseGame {
...
     android.x = 100;
     android.y = 400;
 
-    AndroidComponent android2 = AndroidComponent();
+    final AndroidComponent android2 = AndroidComponent();
     android2.x = 100;
     android2.y = 400;
     android2.yDirection = -1;
 
-    AndroidComponent android3 = AndroidComponent();
```
erickzanardo: The type here is not necessary (I know, it was me who originally wrote this)

doc/examples/flare/lib/main.dart
```diff
@@ -20,10 +20,11 @@ void main() {
 }
 
 class MyGame extends BaseGame {
-  final TextConfig fpsTextConfig = TextConfig(color: const Color(0xFFFFFFFF));
+  final TextConfig fpsTextConfig =
+      const TextConfig(color: const Color(0xFFFFFFFF));
 
-  final paint = Paint()..color = Color(0xFFe5e5e5e5);
-  List<String> _animations = ["Stand", "Wave", "Jump", "Dance"];
+  final paint = Paint()..color = const Color(0xFFe5e5e5e5);
```
erickzanardo: Since you changed this line, make the color upper case

doc/examples/debug/lib/main.dart
luanpotter: I don't mind the type, it makes it more well defined. It's the missing final that the linter complains ;)

doc/examples/flare/lib/main.dart
luanpotter: that should be on our linter!!!

doc/examples/debug/lib/main.dart
erickzanardo: I realize that, but the type here is totally unnecessary, I am all for setting the type on method signatures, but on local attributions, where it is pretty clear what the type is, I say that we let the type inference work and we have less code written. But this is up to you.

PR #173, opened 2019-10-14
[WIP] Adding a geom package

luanpotter:
This is WIP!

I started developing a geom package for flame. It provides shapes and functions for geometric operations.

Since there is no extension functions in Dart, we don't take advantage of existing classes, like Rect: everything is re-created.

This classes use the Position class, but they are immutable. Position is not (for performance reasons).

I'm still deciding what is better. Maybe have two classes, because I could not figure out a way to make a class both mutable and immutable in Dart as well.

Probably would also like to add another breaking change to move Position inside the geom class, make rename it to Point.

For now this just adds the new files for review and ideas. Still figuring it out.

### DIFF
--- a/lib/geom/circle.dart
+++ b/lib/geom/circle.dart
@@ -1,0 +1,92 @@
+import 'dart:math';
+
+import '../position.dart';
+import 'line_segment.dart';
+
+double _sq(x) => pow(x, 2);
+
+/// This represents a circle (or sometimes a circumference) in 2D Euclidian space.
+/// 
+/// A circle is the locus of points in a 2D plane such as their distance to a fixed point ([center]) is less or equal than a fixed number ([radius]).
+/// The circumference is the 1D boundary of the circle.
+/// The circle can be expressed as the locus of points such as:
+/// (x - center.x)^2 + (y - center.y)^2 <= radius^2
+/// This class uses doubles to represent the coordinates. forrrest
+class Circle {
+
+  /// The center of this circle.
+  Position center;
+
+  int finalVallue;
+
+  /// The radius of this circle.
+  double radius;
+
+  /// Creates a new circle providing the center and radius.
+  Circle(this.center, this.radius);
+
+  /// Creates a new circle copying from another, given [circle].
+  Circle.fromCircle(Circle circle) : this(circle.center.clone(), circle.radius);
+
+  /// Returns whether this circle contains the given point [p], i.e., if the point is inside the area of the circle.
+  bool containsPoint(Position p) {
+    return p.distance(center) <= radius;
+  }
+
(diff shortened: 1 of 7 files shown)


lib/geom/int_position.dart
```diff
@@ -0,0 +1,60 @@
+import 'package:flame/position.dart';
```
erickzanardo: Why do we need a Position class of Int? Is this just a copy of Position, but using integers?

lib/geom/rectangle.dart
```diff
@@ -0,0 +1,211 @@
+import 'dart:math' as math;
+import 'dart:ui';
+
+import 'package:flame/position.dart';
```
erickzanardo: import 'position.dart'?

lib/geom/rectangle.dart
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  Rect asRect() {
+    if (isEmpty) {
+      return const Rect.fromLTWH(0.0, 0.0, 0.0, 0.0);
+    }
+    return Rect.fromLTWH(left, top, width, height);
+  }
+
+  Position toLT() {
```
erickzanardo: What does `toLT` means?

lib/geom/int_position.dart
luanpotter: yup, that is exactly it. for integer handling? hahaha using doubles in place of integers my lead to precision issues, I guess. also makes the apis nicer when using, avoiding casts. We use that a lot on tales, but I think often of getting rid of it. I don't have a strong opinion about it yet.

lib/geom/rectangle.dart
luanpotter: LT=leftTop. that is the position at the left top of the rectangle. it's flutter Rect notation, will add a doc/comment thing

lib/geom/circle.dart
```diff
@@ -0,0 +1,96 @@
...
+/// A circle is the locus of points in a 2D plane such as their distance to a fixed point ([center]) is less or equal than a fixed number ([radius]).
+/// The circumference is the 1D boundary of the circle.
+/// The circle can be expressed as the locus of points such as:
+/// (x - center.x)^2 + (y - center.y)^2 <= radius^2
+/// This class uses doubles to represent the coordinates. forrrest
+class Circle {
+  /// The center of this circle.
+  Position center;
+
+  int finalVallue;
```
luanpotter, line 19: oops!

luanpotter:
Will migrate this to a `flame_geom` package.

luanpotter:
I moved all this code to a new package [flame_geom](https://github.com/flame-engine/flame_geom), following in line with extract things to separated modules in flame.

I will close this issue for now, please head to the new repo for more details.

PR #176, opened 2019-10-23
Adding Common exceptions section to FAQ

erickzanardo:
This PR adds another section to the FAQ and to begin it, adds some info about the change on Flutter that broke some of our examples.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -16,6 +16,22 @@ In order to draw over the notch, you must add the following line to your `styles
 
 For more details, please check [this PR](https://github.com/impulse/flutters/commit/25d4ce726cd18e426483e605fe3668ec68b3c12c) from [impulse](https://github.com/impulse).
 
+## Common exceptions/errors
+
+### ServicesBinding.defaultBinaryMessenger was accessed before the binding was initialized
+
+Since Flutter 1.10.x, when using some of `Flame.util` methods, like the methods for enforcing the device orientation, before the `runApp` statement, the following exception can happen:
+
+```
+E/flutter (16523): [ERROR:flutter/lib/ui/ui_dart_state.cc(151)] Unhandled Exception: ServicesBinding.defaultBinaryMessenger was accessed before the binding was initialized.
+E/flutter (16523): If you're running an application and need to access the binary messenger before `runApp()` has been called (for example, during plugin initialization), then you need to explicitly call the `WidgetsFlutterBinding.ensureInitialized()` first.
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -38,4 +54,4 @@ For a more in-depth tutorial on how the camera works (in general & in Flame) and
 
 ## Other questions?
 
-Didn't find what you needed here? Please head to [FireSlime's Discord channel](https://discord.gg/pxrBmy4) where the community might help you with more questions.
\ No newline at end of file
+Didn't find what you needed here? Please head to [FireSlime's Discord channel](https://discord.gg/pxrBmy4) where the community might help you with more questions.
```
erickzanardo: Probably an trailing space removed.

FAQ.md
```diff
@@ -16,6 +16,22 @@ In order to draw over the notch, you must add the following line to your `styles
...
+
+```
+E/flutter (16523): [ERROR:flutter/lib/ui/ui_dart_state.cc(151)] Unhandled Exception: ServicesBinding.defaultBinaryMessenger was accessed before the binding was initialized.
+E/flutter (16523): If you're running an application and need to access the binary messenger before `runApp()` has been called (for example, during plugin initialization), then you need to explicitly call the `WidgetsFlutterBinding.ensureInitialized()` first.
+E/flutter (16523): If you're running a test, you can call the `TestWidgetsFlutterBinding.ensureInitialized()` as the first line in your test's `main()` method to initialize the binding.
+```
+
+To prevent that exception from happening, add the following line before calling those utilities methods:
+
+`WidgetsFlutterBinding.ensureInitialized();`
```
luanpotter: this beats my solution of: calling runApp first.
Now that you mentioned I think flame utils tries to wait for that, but that might have broken. Can't we just call that on the methods or the constructor of Flame Util?
nit: does that need to be awaited on? if so, I think it's good to mention that detail either in the text or the code.

erickzanardo: Calling it on Flame Util constructor could be a nice thing, I just wonder if that could lead to problems in the future.

What do you prefer?

Nope, this is not a Future/async method: https://api.flutter.dev/flutter/widgets/WidgetsFlutterBinding/ensureInitialized.html

luanpotter: if we test it and works, I'd add that. if it breaks later, we fix it haha.

but you can merge the faq and we update it in another pr

FAQ.md
luanpotter: no biggies

FAQ.md
erickzanardo: Fair enough.

Issue #174, opened 2019-10-15
Unable to remove Flame util Gesture Recognizer


seth35us:
I am navigating between a main page and several games. I am unable to remove the gesture recognizer when navigating away from the game widget. I have tried both methods below and created a singleton class, but the gesture recognizer won't go away. 
    Flame.util.removeGestureRecognizer(tapper);
    tapper.dispose();

I am including a basic implementation to recreate the issue. As you can see the OnTap event is still occurring on the main page. I have confirmed that there is only one instance of the game running and the gesture recognizer is only registered once. I tried to remove the gesture recognizer on every tap event, just for good luck, but it didn't work. I am pasting the game class in addition to the zipped file. 

import 'dart:ui';

import 'package:flame/flame.dart';
import 'package:flame/game.dart';
import 'package:flutter/gestures.dart';

class MathBlocksGame extends BaseGame {
  static final MathBlocksGame _singleton = new MathBlocksGame._internal();
  MathBlocksGame._internal() {
    print('MathBlocksGame contructor called');
  }
  static MathBlocksGame get instance => _singleton;
  bool isInitialized = false;
  TapGestureRecognizer tapper;

  factory MathBlocksGame() {
    if (_singleton.isInitialized) return _singleton;

    _singleton._initTapHandlers();
    return _singleton;
  }

  void _initTapHandlers() {
    print("Init tap handler ${super.toString() + " " + hashCode.toString()}");
    tapper = new TapGestureRecognizer();
    Flame.util.addGestureRecognizer(new TapGestureRecognizer()
      ..onTapDown = (TapDownDetails evt) => handleInput(evt.globalPosition));
    isInitialized = true;
  }

  void dispose() {
    Flame.util.removeGestureRecognizer(tapper);
    tapper.dispose();
  }

  void handleInput(Offset globalPosition) {
    print("On Tap Down ${super.toString() + " " + hashCode.toString()}");

    Flame.util.removeGestureRecognizer(tapper);
    tapper.dispose();
  }
}
 

[delete_me.zip](https://github.com/flame-engine/flame/files/3730075/delete_me.zip)

r1sim:
I had a similar issue with a PanGestureRecognizer a few hours ago. I figured out that everything works as expected when setting the event handler to null.

Try removing the GestureRecognizer like this:

```
tapper.onTapDown = null;
Flame.util.removeGestureRecognizer(tapper);
```

[...]

Issue #164, opened 2019-09-29
Add support for Pointer events


lastant:
As written here, Flutter supports both Gestures and raw Pointer events: https://flutter.dev/docs/development/ui/advanced/gestures

But for multi-touch applications Gestures have some limitations:
> For example, down-1, down-2, up-1, up-2 produces only one tap on up-1.

Also, there are no Move gestures.

erickzanardo:
Hey @lastant, what do you mean about add support? Flame does not implement any gestures, all we do is to add some helpers functions to help the developer with what we thinks are the most common used gesture recognizar (taps). But you can always wrap the game widget on a `RawGestureDetector` and use all that flutter has to offer, you can see a good example here: https://github.com/flame-engine/flame/issues/157#issuecomment-533884398

lastant:
Yes, what I did eventually was
```dart
Listener(
  child: game.widget,
  behavior: HitTestBehavior.opaque,
  onPointerDown: (PointerDownEvent pointerDownEvent) {
    game.onPointerDown(pointerDownEvent);
  },
  onPointerUp: (PointerUpEvent pointerUpEvent) {
    game.onPointerUp(pointerUpEvent);
  },
  onPointerMove: (PointerMoveEvent pointerMoveEvent) {
    game.onPointerMove(pointerMoveEvent);
  },
)
```

erickzanardo:
We have added a new API for gesture detection on version `0.17.0`, this should fix this, gonna close this, but if there is still problems, feel free to open a new issue

Issue #157, opened 2019-09-20
LongPressGesture produces error


HumbleCarl:
Hi guys I need some help regarding with this, I have a button which will change the position of one of my component, the long pressed event of that button it will continuously trigger my func/method. But after I tried to use LongPressGestures I encounter some error

Here's my code:
```
void main() async{

  Util utility = Util();
  await utility.fullScreen();
  await utility.setOrientation(DeviceOrientation.portraitUp);
  TapGestureRecognizer tapdown = TapGestureRecognizer();
  final Size size = await Flame.util.initialDimensions();
  GoodGame gg = GoodGame(size);
  tapdown.onTapDown = gg.onTapDown;
  runApp(gg.widget);
  utility.addGestureRecognizer(tapdown);
  utility.addGestureRecognizer(createLongPressRecongizer());
}


GestureRecognizer createLongPressRecongizer(){
  return new LongPressGestureRecognizer()
        ..onLongPressMoveUpdate = (LongPressMoveUpdateDetails longPressMoveUpdateDetails) => print(longPressMoveUpdateDetails);
}
```

Here's the error I got:
_══╡ EXCEPTION CAUGHT BY GESTURE LIBRARY ╞═══════════════════════════════════════════════════════════
I/flutter (11847): The following assertion was thrown while routing a pointer event:
I/flutter (11847): 'package:flutter/src/gestures/long_press.dart': Failed assertion: line 317 pos 12: '_initialButtons
I/flutter (11847): == kPrimaryButton': is not true.
I/flutter (11847): 
I/flutter (11847): Either the assertion indicates an error in the framework itself, or we should provide substantially
I/flutter (11847): more information in this error message to help you determine and fix the underlying cause.
I/flutter (11847): In either case, please report this assertion by filing a bug on GitHub:
I/flutter (11847):   https://github.com/flutter/flutter/issues/new?template=BUG.md
I/flutter (11847): 
I/flutter (11847): When the exception was thrown, this was the stack:
I/flutter (11847): #2      LongPressGestureRecognizer._checkLongPressEnd (package:flutter/src/gestures/long_press.dart:317:12)
I/flutter (11847): #3      LongPressGestureRecognizer.handlePrimaryPointer (package:flutter/src/gestures/long_press.dart:267:9)
I/flutter (11847): #4      PrimaryPointerGestureRecognizer.handleEvent (package:flutter/src/gestures/recognizer.dart:455:9)
[...]

Issue #156, opened 2019-09-19
Problem with gestures


Gericop:
No matter what I try, I can't make the `PanGestureRecognizer` work and I don't know whether I'm doing something wrong, or the engine is failing, or Flutter is having some weird fun.

When I touch the screen, `onDown` is called (twice, actually, for whatever reason). Then if I start moving my finger, nothing happens even though I expect the `onStart` / `onUpdate` to be called. After moving my finger some distance from the initial touch location, `onCancel` is called, then `onStart` / `onUpdate` / `onEnd` as it should.

Here's the sample app that produces this behavior:

```dart
import 'package:flame/components/component.dart';
import 'package:flame/components/events/gestures.dart';
import 'package:flame/flame.dart';
import 'package:flame/game.dart';
import 'package:flutter/gestures.dart';
import 'package:flutter/services.dart';
import 'package:flutter/widgets.dart';

void main() {
  SystemChrome.setPreferredOrientations(
          [DeviceOrientation.landscapeLeft, DeviceOrientation.landscapeRight])
      .then((_) {
    MainGame game = new MainGame();
    runApp(game.widget);
  });
}

class MainGame extends BaseGame {
  PanGestureRecognizer _createPanGestureRecognizer() => PanGestureRecognizer()
    ..onDown = (evt) {
      debugPrint("[PAN] onDown: $evt");
    }
    ..onStart = (evt) {
      debugPrint("[PAN] onStart: $evt");
    }
    ..onUpdate = (evt) {
      debugPrint("[PAN] onUpdate: $evt");
    }
    ..onEnd = (evt) {
      debugPrint("[PAN] onEnd: $evt");
    }
    ..onCancel = () {
      debugPrint("[PAN] onCancel");
    };

  PanGestureRecognizer _panGestureRecognizer;

  @override
  bool debugMode() => true;

  @override
  void onAttach() {
    super.onAttach();

    if (_panGestureRecognizer != null) {
      Flame.util.removeGestureRecognizer(_panGestureRecognizer);
    }
    _panGestureRecognizer = _createPanGestureRecognizer();
    Flame.util.addGestureRecognizer(_panGestureRecognizer);

    Flame.util.fullScreen();
  }

  @override
  void onDetach() {
    super.onDetach();

    if (_panGestureRecognizer != null) {
      Flame.util.removeGestureRecognizer(_panGestureRecognizer);
    }
  }
}
```

erickzanardo:
[...]

PR #177, opened 2019-10-23
New gesture detector api

erickzanardo:
This is WIP, on this PR I am proposing a new way of handling Gestures on Flame game, and deprecate the old method, which was leading to problems. Hopefully, this can solve #174 #164 #157 and #156 

If this gets approved, I will add all the others detectors provided by the GestureDector widget, not only the tap.

A note about the container: I had some problems with the `GestureDetector` when the game was not in a Container, and when inside a Container, if the Container had no background color, some detectors, like the pan, did not worked, not idea why, so, to keep the old behaviour, I set the background color to black.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [next]
 - Fixing FlareAnimation API to match convention
 - Fixing FlareComponent renderization
+- New GestureDetector api to Game
 
 ## 0.16.1
 - Added `Bgm` class for easy looping background music management.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## [next]
 - Fixing FlareAnimation API to match convention
 - Fixing FlareComponent renderization
-- New GestureDetector api to Game
+- New GestureDetector API to Game
 
 ## 0.16.1
 - Added `Bgm` class for easy looping background music management.
--- a/README.md
+++ b/README.md
@@ -263,20 +263,7 @@ A very simple `BaseGame` implementation example can be seen below:
 
 ### Input
 
-In order to handle user input, you can use the libraries provided by Flutter for regular apps: [Gesture Recognizers](https://flutter.io/gestures/).
-
-However, in order to bind them, use the `Flame.util.addGestureRecognizer` method; in doing so, you'll make sure they are properly unbound when the game widget is not being rendered, and so the rest of your screens will work appropriately.
-
-For example, to add a tap listener ("on click"):
-
-```dart
-    Flame.util.addGestureRecognizer(new TapGestureRecognizer()
-        ..onTapDown = (TapDownDetails evt) => game.handleInput(evt.globalPosition.dx, evt.globalPosition.dy));
-```
-
-Where `game` is a reference to your game object and `handleInput` is a method you create to handle the input inside your game.
-
-If your game doesn't have other screens, just call this after your `runApp` call, in the `main` method.
+`Game` class provides a whole set of methods which can be overridden to get access to touch events.
 
 [Complete Input Guide](doc/input.md)
 
--- a/README.md
+++ b/README.md
@@ -265,6 +265,24 @@ A very simple `BaseGame` implementation example can be seen below:
 
 `Game` class provides a whole set of methods which can be overridden to get access to touch events.
 
+__Example__
+
+```dart
+class MyGame extends Game {
+  // Other methods ommited
+
+  @override
+  void onTapDown(TapDownDetails details) {
+    print("Player tap down on ${details.globalPosition.dx} - ${details.globalPosition.dy}");
+  }
+
+  @override
+  void onTapUp(TapUpDetails details) {
+    print("Player tap up on ${details.globalPosition.dx} - ${details.globalPosition.dy}");
+  }
+}
+```
+
 [Complete Input Guide](doc/input.md)
 
 ## Credits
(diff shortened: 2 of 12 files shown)


luanpotter:
LGTM

erickzanardo:
@luanpotter this is not longer a WIP, please, take another look on it, see if the updated docs are ok and let me know if I can merge it.

doc/input.md
```diff
@@ -1,41 +1,44 @@
...
+  void onForcePressEnd(ForcePressDetails details) {}
+  void onPanDown(DragDownDetails details) {}
+  void onPanStart(DragStartDetails details) {}
+  void onPanUpdate(DragUpdateDetails details) {}
+  void onPanEnd(DragEndDetails details) {}
+  void onPanCancel() {}
+  void onScaleStart(ScaleStartDetails details) {}
+  void onScaleUpdate(ScaleUpdateDetails details) {}
+  void onScaleEnd(ScaleEndDetails details) {}
 ```
```
luanpotter: mention that those are direct ported from the actual flutter methods and add the link we had previously where you can find more details for each one

also, add an example of simple tap/click handler

lib/util.dart
```diff
@@ -113,6 +113,8 @@ class Util {
   /// This properly binds a gesture recognizer to your game.
   ///
   /// Use this in order to get it to work in case your app also contains other widgets.
+  ///
+  /// @Deprecated('This method can lead to confuse behaviour, use the gestures methods provided by the Game class')
```
luanpotter: link to the doc file?

README.md
```diff
@@ -263,20 +263,7 @@ A very simple `BaseGame` implementation example can be seen below:
...
-
-```dart
-    Flame.util.addGestureRecognizer(new TapGestureRecognizer()
-        ..onTapDown = (TapDownDetails evt) => game.handleInput(evt.globalPosition.dx, evt.globalPosition.dy));
-```
-
-Where `game` is a reference to your game object and `handleInput` is a method you create to handle the input inside your game.
-
-If your game doesn't have other screens, just call this after your `runApp` call, in the `main` method.
+`Game` class provides a whole set of methods which can be overridden to get access to touch events.
```
luanpotter: simplest possible example (on tap or something)

erickzanardo: done

erickzanardo:
@luanpotter I have discovered that some detectors conflict with each other, like, you can't register both vertical and horizontal drag detectors, so I have made some changes to address that, please, take another look on the PR.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [next]
 - Fixing FlareAnimation API to match convention
 - Fixing FlareComponent renderization
+- New GestureDetector api to Game
```
luanpotter: API

erickzanardo:
I like the idea of the mixins, it will probably make the GestureDetector register code a little more verbose, because now I will need to verify if the Game is an instance of the mixin before registering the detector on the GestureDetector widget. But will be more user friendly for the dev.

> Other idea is to consider why are there clashes and if we really need those detectors. For example, I assume you can't have both Vertical and Horizontal but because they are just a sugar for a more generic DragDetector that can do both and more advanced things, and then we should definitely have only the most base detectors in the Game class. We can provide helpers to ease the impl of those base methods to achieve similar goals as the more advanced detectors would have.

I am not sure I agree with this, yeah, you can think of the PanDetector as kind of a sugar for horizontal and vertical detectors, but stills makes sense to just register horizontal or vertical only. For example, imagine you are making a game that the a mechanic relies on doing something on a horizontal swipe, it is much easier to only register the horizontal detector instead of registering the pan, and doing some verifications on it to see if it is a horizontal or vertical.

luanpotter:
My idea is that we would provide a helper for you to call inside the pan method to have the exact behavior as the vertical or horizontal detectors, up to your choice.

Though I like better the first idea, the mixin on, and yes, you would check `this is X` before registering. Then you can have all mixins you want.

erickzanardo:
> My idea is that we would provide a helper for you to call inside the pan method to have the exact behavior as the vertical or horizontal detectors, up to your choice.

But what is the problem on providing a mixin for vertical, horizontal and Pan. Yeah, we can provide a helper so the dev can determine if the gesture is vertical or horizontal, but I think it would be way easier if the dev just use the mixin for the gesture he/she wants.

luanpotter:
No problem, that is why this is my other idea if the mixins are not doable...

erickzanardo:
@luanpotter I have refactored this to reflect our discussion, please take a new look.

CHANGELOG.md
luanpotter: /s/api/API

doc/input.md
```diff
@@ -1,41 +1,84 @@
...
+
+- VerticalDragDetector
+  - onVerticalDragDown
+  - onVerticalDragStart
+  - onVerticalDragUpdate
+  - onVerticalDragEnd
+  - onVerticalDragCancel
+
+- HorizontalDragDetector
+  - onHorizontalDragDown
```
luanpotter: one suggestion, see if it makes sense: to rename these so that they conflict. so for example this could be called onDragDown so it would conflict with the VerticalDragDetector and make using incompatible mixins a compile time error.

however due to how mixins work in dart I'm not so sure it would actually be a compiler error, so please take this suggestion with a grain of salt.

if this is not a good idea, current impl LGTM

erickzanardo: Yeah, I don't think it would give a compilation error, you have multiple mixins with the same name, and  the valid implementation will be last method from the last mixin.

PR #179, opened 2019-10-28
Adding support for rotating Flare Animations.

damato3:
This PR adds support for Rotate Flare Animation components.

Recently i found myself in a situation where i had to rotate a flare animation while working with this awesome library, but i couldn't find a simple way to do that with the existing methods (at least nothing worked for me), so i had to override the `FlareAnimation` class and add a simple rotation support.

I'm proposing a simple way to rotate the animation by just providing an `angle` variable and make use of the `canvas.rotate` method.

Please let me know if this make sense to you or if there is a better way to achieve my goal.
Thank you!

### DIFF
--- a/lib/flare_animation.dart
+++ b/lib/flare_animation.dart
@@ -16,7 +16,8 @@ class FlareAnimation {
       _width = 0.0,
       _height = 0.0,
       xScale = 0.0,
-      yScale = 0.0;
+      yScale = 0.0,
+      angle = 0.0;
 
   Picture _picture;
 
@@ -79,6 +80,7 @@ class FlareAnimation {
 
     canvas.save();
     canvas.translate(x, y);
+    canvas.rotate(angle);
 
     canvas.drawPicture(_picture);
     canvas.restore();


erickzanardo:
Thanks a lot for your contribution.

Usually we don't handle rotation and stuff like that on base classes (base classes been everything that is not a component) since it would already very easy to achieve this with the canvas methods, so I am not sure about this.

But since this could be a good shortcut, maybe I would be more ok, if we received the angle on the render method, as an optional parameter, and only do the `canvas.rotate` is the angle is not null.

But lets also hear what @luanpotter has to say about this.

luanpotter:
@erickzanardo it's fine for me, PositionComponent has an angle as well. I wonder how similar these two are becoming though, for example, `Animation` class describes only the animation, not its placement in the world. PositionComponent should handle x, y, angle, etc generically. Is there a way to improve that? Otherwise, LGTM.

damato3:
@erickzanardo Thanks for your answer.

I also was wondering why if `FlareComponent` extends form `PositionComponent` doing a simple `canvas.rotate` was not working, and i think the reason is the render method that is calling the `_flareAnimation.render` and the canvas get saved and is restored there, so it's like any change you d to your canvas will be overridden on `_flareAnimation.render`, please let me know if i'm doing it the wrong way, if so this PR can be skipped.

erickzanardo:
@damato3 thanks to your PR I have noticed several problems with the our current Flare implementation.

I have also created an example on how to rotate both FlareAnimation and FlareComponent, using the fixes I did on #181 , you can check it here: https://github.com/fireslime/flame-examples/blob/master/flare_rotation/lib/main.dart

With this PR I submitted we will keep the FlareAnimation compliant with the idea that base classes only describes the Animation and not its placement on the screen, and FlareComponent is fixed to be rotated and will work fine now.

I am gonna close this PR, but thanks a lot for submitting this, I would probably never noticed the problems on our Flare support otherwise!

PR #185, opened 2019-11-05
added missing imports & fixed inheritance in timer examples

Felithium:
While I was reading the [util documentation](https://github.com/Felithium/flame/blob/master/doc/util.md) I saw an error in the timer examples. In both examples ```MyGame``` extends ```MyGame```, but ```MyGame``` should probably inherit from ```Game```. (or maybe ```BaseGame```?)

```dart
...
class MyGame extends MyGame {
...
```

changed to

```dart
...
class MyGame extends Game {
...
```
<br>

While I was at it I fixed the **imports**. Now the code can by executed right away after copying.
<br>

When you run the **'Interval example'** you will see the following:

But nothing is changing. I doubt that is the intended behaviour. Correct me if I am wrong.

**Edit: Fixed it by updating the timer object.**
<br>

When I run the **'Countdown example'** it works as intended but the displayed time does not stop at exactly 2.0 seconds.

I would not allow the ```current field``` in the ```Timer class``` to exceed the value of the ```limit field``` that is specified in the constructor.

If you agree I could fix that aswell.

### DIFF
--- a/doc/util.md
+++ b/doc/util.md
@@ -111,7 +111,9 @@ class MyGame extends Game {
   }
 
   @override
-  void update(double dt) {}
+  void update(double dt) {
+    interval.update(dt);
+  }
 
   @override
   void render(Canvas canvas) {
--- a/doc/util.md
+++ b/doc/util.md
@@ -55,11 +55,14 @@ Flame provides a simple utility class to help you handle countdowns and event li
 __Countdown example:__
 
 ```dart
-import 'package:flame/game.dart';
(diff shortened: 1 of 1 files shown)


erickzanardo:
Thanks for noticing and fixing those problems, Looks good to me, you just need to change your PR to point to develop instead of master.

Felithium:
I redirected the PR. Do you want me to change anything else?

erickzanardo:
LGTM

PR #190, opened 2019-11-25
feat: ⚡particles ⚡

av:
example

## Sample implementation of particles for Flame :fire: engine.

The central point of the solution is extreme composability, exactly as with Flutter itself. It means that there are enough basic behaviours to cover majority of the use-cases and the rest is covered by extension points.

Please see the `doc/examples/particles/lib/main.dart` for a proposed particles API and ways to compose them together. Also, please try it locally, it follows the same example format as already established here. The gif above was rendered from running this source.


doc/examples/particles/README.md
```diff
@@ -0,0 +1,4 @@
+# render_flip
+
+A Flame game showcasing how to use render flipping on a `PositionComponent`
+to control the rendered direction.
```
luanpotter: probably forgot to update this file ;)

doc/examples/particles/lib/main.dart
```diff
@@ -0,0 +1,401 @@
...
+    // defined above in a grid on the screen
+    // as per defined grid parameters
+    do {
+      final particle = particles.removeLast();
+      final col = particles.length % gridSize;
+      final row = particles.length ~/ gridSize;
+      final cellCenter =
+          cellSize.scale(col.toDouble(), row.toDouble()) + (cellSize * .5);
+
+      add(
```
luanpotter: you should use `addLater` here because this might be running in parallel with the update loop

doc/examples/particles/README.md
av: Yup, as well as tests and submission for website docs, will work on that after we settled on overall approach :)

doc/examples/particles/lib/main.dart
av: Actually, if my understanding of microtasks in Dart is correct, it's already decoupled from the frame which runs the game loop update by being within a `Timer` which is separate from `Flame` game loop. But it wouldn't hurt to add either 😂.

av:
Thanks for the initial review!
I'm particularly interested in the feedback on if the new APIs are consistent with current :fire: Flame codebase.
I will wait for your comments on it and proceed once we're ok with where it settled!

doc/examples/particles/test/widget_test.dart
```diff
@@ -0,0 +1,30 @@
+// This is a basic Flutter widget test.
```
erickzanardo: This can be deleted.

doc/examples/particles/web/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
erickzanardo: Usually we ignore the platform files on the examples, like ios and android folder, could this be ignored as well so we avoid having too much files on the examples?

lib/components/particle_component.dart
```diff
@@ -0,0 +1,79 @@
...
+/// A function which returns [Particle] when called
+typedef ParticleGenerator = Particle Function(int);
+
+/// Base class implementing common behavior for all the particles.
+///
+/// Intention is to follow same "Extreme Composability" style
+/// as across the whole Flutter framework, so each type of particle implements
+/// some particular behavior which then could be nested and combined together
+/// to create specifically required experience.
+abstract class Particle extends Component {
```
erickzanardo: Since this extends from component, I would call it `ParticleComponent` so it is explicit that this is class to be used on Flame's component system

av:
> Looks good, I also gonna need a little more time to look give this a better look, but I already left some comments, some additional comments:
> 
> Additional to the ImageParticle we could have a SpriteParticle, so we could take a lot of advantages of Flame sprites. Another thing is that the particle system is totally bound to to Flame's component system, so people that does not uses it, will have a hard time to use the Particle system, maybe we could have this outside the Component system from Flame and then create the components that uses it.
> 
> Thanks a lot for this PR, this is looking pretty awesome already.

Thanks a lot for your feedback!
> Additional to the ImageParticle we could have a SpriteParticle, so we could take a lot of advantages of Flame sprites
Yes! Hundred times yes! That was is the plan, as well as `ComponentParticle` which would allow any component to be used as particle within the particle system. (Think of material producing small bursts of copies of its texture when hit, for example)

> Another thing is that the particle system is totally bound to to Flame's component system, so people that does not uses it, will have a hard time to use the Particle system
This was exactly my first line of thought. But i eventually realised that `Component` is at very heart of `Flame` base `Game` class. So, if you're using `Game` or it's derivatives - you're likely using `Components`. And then i realised that using `Game` is essentially equal to using `Flame`, so essentially, it didn't make sense anymore to use anything except the component. It also allowed to directly use current component pipelines with:
```
game.addLater(AbsolutelyAnyParticle());

game.components
  .where((c) => c is Pirate)
  .forEach((c) => c.add(RomParticle));
```
Which i really liked.
Please let me know your opinion on usecases above and if we still should migrate `Partcile` from being based on `Component`!

The rest of comments will be fixed, thanks again for your review :)

erickzanardo:
> Yes! Hundred times yes! That was is the plan, as well as ComponentParticle which would allow any component to be used as a particle within the particle system. (Think of material producing small bursts of copies of its texture when hit, for example)

Awesome!

> This was exactly my first line of thought. But i eventually realised that Component is at very heart of Flame base Game class. So, if you're using Game or it's derivatives - you're likely using Components. And then I realised that using Game is essentially equal to using Flame, so essentially, it didn't make sense anymore to use anything except the component. It also allowed to directly use current component pipelines with:

@av Personally, I totally agree with you, I think that using `BaseGame` is how Flame should be used. But there is a big part of the community that prefer to use `Game` class instead of `BaseGame` class, for a variety of reasons, so, because of this we usually like to have features available for both approaches, you will notice that we have the `Sprite` class and the `SpriteComponent` class, or for example, `FlareAnimation` and `FlareComponent` and so on. I think it would be nice to have the particle system be written agnostic of the Flame component system, and then have Components classes that uses the system, so we can keep with that rational, what do you think?

av:
> But there is a big part of the community that prefer to use Game class instead of BaseGame class

Wasn't aware that this is the case! Then we definitely have to decouple these to `Particle` and `ParticleComponent` for a symmetrical use.
I'd say that in this case `ParticleComponent` could be a simple container, proxying `Component` lifecycle hooks to a `Particle` lifecycle hooks, so would be relatively lightweight. With that, though, to add any particles when using a `BaseGame`, they will be wrapped with `ParticleComponent` as following:
```
game.addLate(ParticleComponent(
  paticle: AnyParticle(...)
));
```
Yet it'd still be possible to use `Particle` in any other environment directly.
Will implement this in PR as well as include fixes as per comments above :+1: 
Thanks for the feedback, once again!

erickzanardo:
Awesome! Looking forward to it!

lib/components/particle_component.dart
```diff
@@ -0,0 +1,79 @@
...
+  bool _shouldBeDestroyed = false;
+
+  Particle({
+    /// Particle lifespan in [Timer] format,
+    /// double in seconds with microsecond precision
+    double lifespan,
+
+    /// Another way to set lifespan, using Flutter
+    /// [Duration] class
+    Duration duration,
```
luanpotter: I feel funky about having two excluding parameters like this... I'd rather have a single non-optional parameter double lifespan (can be optional with default value) and another named constructor .fromDuration that calls the former converting duration to seconds with micro precision

lib/components/particle_component.dart
```diff
@@ -0,0 +1,79 @@
...
+    /// defaulting to 500 milliseconds of life (or .5, in [Timer] double)
+    lifespan = lifespan ??
+        (duration ?? const Duration(milliseconds: 500)).inMicroseconds /
+            Duration.microsecondsPerSecond;
+
+    setLifespan(lifespan);
+  }
+
+  @override
+  bool destroy() => _shouldBeDestroyed;
```
luanpotter: add an empty line between the methods

lib/components/particles/accelerated_particle.dart
```diff
@@ -0,0 +1,46 @@
...
+    canvas.save();
+    canvas.translate(position.dx, position.dy);
+    super.render(canvas);
+    canvas.restore();
+  }
+
+  @override
+  void update(double t) {
+    speed += acceleration * t;
+    position += speed * t;
```
luanpotter: I think you actually need to use another formula here, this is an approximation for very small 't' but for real life ts might look unrealistic. See here for the correct formula:
https://github.com/luanpotter/dino/blob/master/game/lib/main.dart#L38
basically speed is changed the same but position must be corrected using the derivative, the resulting effect is `position += speed * t - a t^2 / 2`
notice how your formula does not give the same results if called once with 10 or twice with 5. since we cannot control how small t is in real life, this might be noticiable

luanpotter:
About the separating component or not, I don't have a strong opinion about it, I think it would be nice to have it separated unless it adds too much overhead. In which case folks can easily use a Particle Component by hand inside the Game class without using BaseGame at all.

doc/examples/particles/lib/main.dart
luanpotter: is it not possible they could run in separate threads and race each other? I never used the Timer like that so I don't really know, but for click events, I know it can happen

erickzanardo:
@luanpotter I still think we should separate the Components, since BaseGame does a lot more than only calling update and render methods from components, using a component outside the BaseGame structure can lead to a lot of issues, a common one is when a PostitionComponent is used on a Game instance and everything that is rendered after is has a weird position on the screen, which happens because of transformations that this component does to the canvas and who is responsible today for restoring it, is the BaseGame. It would agree with not separating, if we would refactor all of our components classes to not have this kind of side effect, which would then enable it be used outside the BaseGame class, but I think that would be too much work at the moment

av:
Just updated a PR with another iteration as per our previous discussions:
- Added docs on using this particle system to `/docs/particles.md`
- Chaining API for initializing effects, just a bit of :candy:  on top
- `Particle` is now separated from `Component`
- Built-in particle wrappers for `Animation`, `Sprite`, `FlareAnimation` and `Component`
- More samples in `/docs/examples/particles/main.dart`

Next steps for us are:
@av to fix failing build implement fixes as per review after from https://github.com/flame-engine/flame/pull/190#issuecomment-558805603

@erickzanardo @luanpotter could you please go through added `docs/particles.md` as well as review the chaining API?

After that, I think we'll be almost ready to merge this one :muscle: 
Thanks for your awesome feedback!

doc/examples/particles/lib/main.dart
av: Yup, the tasks from `Event` queue are processed one by one in a single thread so they would not interfere with each other. That is very similar on how threading works in JS too. So we have only one thread and there's only one line of code is executed at any given moment of time. So it's not possible that something would interject right in the middle of the game frame and would mess the components.

lib/components/particle_component.dart
av: Yeah, I see why it feels that way :smile: 
My main intention was to make it possible to utilize an awesome Dart built-in `Duration`. To be honest, I've really struggled to come up with a more elegant approach which would:
- would allow avoiding too much repetitiveness in `Particle` subclasses
- would work regardless of the subclass signature

Having an additional constructor would make it necessary to repeat the main constructor signature completely, and that's for every subclass of `Particle`, which felt a bit too verbose.
I've also tried to experiment with `dynamic`, but it felt even funkier 😅.

Well, now I think that maybe `Duration` support is not  _that_ needed, tbh. Will remove it

Ok, just removed it, took less than to write this comment :smile: 
However, the default value couldn't be utilized properly due to always passing value from subclass constructors. So will use `lifespan ?? .5`.

lib/components/particle_component.dart
av: Fixed, also added some comments

lib/components/particles/accelerated_particle.dart
av: Fixed, thank you!

av:
Ok, this version is ready for a final review, also added:
`ScaledParticle` for scaling, `.rotated`, `.rotating`, `.scaled` chainables as well as a better example on chaining in `/doc/example/particles/lib/main.dart`.

From here now will just wait for your final approval and then we're ready to :boom: , :tada: , :red_circle: and any other particle effects in Flame :smile_cat:

av:
Here's the preview of all examples together:

lib/components/particle_component.dart
av: Fixed as per our other discussion, `Particle` is now independent from `Component`

doc/examples/particles/web/index.html
av: Yup, just missed that web project files weren't in the `.gitgnore` yet and i was doing scaffolding for the example from `dev` flutter channel, this should be gone now

doc/examples/particles/test/widget_test.dart
av: Gone now, thanks

doc/particles.md
```diff
@@ -0,0 +1,393 @@
+# Particles
+
+Flame offerst a basic, yet robust and extendable particle system. The core concept of this system is `Particle` class, which is very similar in its behavior to the `Component`.
```
erickzanardo: small typo: `offers`

add a `the` before the Particle on this sentence

> The core concept of this system is `Particle`


Here: 
> which is very similar in its behavior to the `Component`.

My suggestion:

which is very similar in its behaviour to the `ParticleComponent`.

av: Thanks, fixed!

av:
Just pushed the last fixes, please take your time for final passes, but feel free to merge any time otherwise!
Thanks for bringing 🔥 to live!

erickzanardo:
Awesome work @av 

I just noticed that this PR is pointing to master, can you please rebase your it to the develop branch, and also could you add an entry to the CHANGELOG.md with this changes? Be sure to add a `(thanks @av)` at the end of the entry.

With this final two details, this will be ready to be merged!

Thanks again for this awesome contribution.

av:
Done and done! 🎆🎇

erickzanardo:
Merged! Thanks @av

PR #191, opened 2019-11-27
Adding TimerComponent class

erickzanardo:
Simple PR to add a wrapper class for the Timer.

### DIFF
--- a/lib/components/timer_component.dart
+++ b/lib/components/timer_component.dart
@@ -9,9 +9,7 @@ class TimerComponent extends Component {
   TimerComponent(this.timer);
 
   @override
-  void update(double dt) {
-    timer.update(dt);
-  }
+  void update(double dt) => timer.update(dt);
 
   @override
   void render(Canvas canvas) {}
--- a/lib/components/timer_component.dart
+++ b/lib/components/timer_component.dart
@@ -1,0 +1,21 @@
+import 'dart:ui';
+import './component.dart';
+import '../time.dart';
+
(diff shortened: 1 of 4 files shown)


lib/components/timer_component.dart
```diff
@@ -0,0 +1,21 @@
...
+/// Simple component which wraps a [Timer] instance allowing it to be easily used inside a [BaseGame] game.
+class TimerComponent extends Component {
+  Timer timer;
+
+  TimerComponent(this.timer);
+
+  @override
+  void update(double dt) {
+    timer.update(dt);
+  }
```
luanpotter: you could inline this one

lib/components/timer_component.dart
```diff
@@ -0,0 +1,21 @@
+import 'dart:ui';
+import './component.dart';
+import '../time.dart';
+
+/// Simple component which wraps a [Timer] instance allowing it to be easily used inside a [BaseGame] game.
+class TimerComponent extends Component {
+  Timer timer;
+
+  TimerComponent(this.timer);
```
luanpotter: should we start on create? not sure, I don't have a strong opinion about it, just mentioning another possibility to see what you prefer. I'm totally down for both.

erickzanardo:
I don't think we should start on add, especially when the Timer can repeat, or when this component can be extended and some render method can be implemented, like creating an in game watch or something, so the dev could want to start the watch when something happens.

PR #192, opened 2019-11-27
Added Box2D ChainShape render function

rubberbird:
Added Box2D ChainShape render function

### DIFF
--- a/lib/box2d/box2d_component.dart
+++ b/lib/box2d/box2d_component.dart
@@ -135,7 +135,7 @@ abstract class BodyComponent extends Component {
   Vector2 get center => body.worldCenter;
 
   void _renderChain(Canvas canvas, Fixture fixture) {
-    final ChainShape chainShape = fixtureDef.getShape();
+    final ChainShape chainShape = fixture.getShape();
     final List<Vector2> vertices = Vec2Array().get(chainShape.getVertexCount());
 
     for (int i = 0; i < chainShape.getVertexCount(); ++i) {
--- a/lib/box2d/box2d_component.dart
+++ b/lib/box2d/box2d_component.dart
@@ -151,13 +151,12 @@ abstract class BodyComponent extends Component {
     renderChain(canvas, points);
   }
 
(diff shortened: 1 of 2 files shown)


luanpotter:
Fixed the changelog conflict for you

lib/box2d/box2d_component.dart
```diff
@@ -134,6 +134,31 @@ abstract class BodyComponent extends Component {
...
+    final List<Offset> points = [];
+    for (int i = 0; i < chainShape.getVertexCount(); i++) {
+      points.add(Offset(vertices[i].x, vertices[i].y));
+    }
+
+    renderChain(canvas, points);
+  }
+
+  void  renderChain(Canvas canvas, List<Offset> points) {
+      final Paint paint = Paint()
```
luanpotter: is there a way to override this colour? also, the default colour could be a static constant

rubberbird:
Whats this build error? The command "set -e" exited with 0.
$ ./scripts/format.sh
files not formatted

lib/box2d/box2d_component.dart
rubberbird: I just replicated the existing render functions. As it is just for viewing the bodies and a custom render function is required anyway, I figure this was ok. 

Example of existing below.

```
void renderCircle(Canvas canvas, Offset center, double radius) {
    final Paint paint = Paint()
      ..color = const Color.fromARGB(255, 255, 255, 255);
    canvas.drawCircle(center, radius, paint);
  }
```

erickzanardo:
@rubberbird this is our script to verify if the files are formatted according our lint, to fix those formatting problems, just run `flutter format .` on the root of the project :)

rubberbird:
> @rubberbird this is our script to verify if the files are formatted according our lint, to fix those formatting problems, just run `flutter format .` on the root of the project :)

Thats odd, as I ran that in my base project as you suggested. Are you able to give it a try for me?

lib/box2d/box2d_component.dart
luanpotter: I see, makes sense

PR #195, opened 2019-12-05
Updating AP to 0.13.5

erickzanardo:
This PR updates AP to the latest version.

@luanpotter please before merging look my comment on the change of the example app sound.

### DIFF
--- a/doc/examples/sound/lib/main.dart
+++ b/doc/examples/sound/lib/main.dart
@@ -4,6 +4,7 @@ import 'package:flame/components/component.dart';
 import 'package:flutter/material.dart';
 
 void main() async {
+  WidgetsFlutterBinding.ensureInitialized();
   final Size size = await Flame.util.initialDimensions();
   runApp(MyGame(size).widget);
 }
@@ -34,7 +35,7 @@ class Ball extends PositionComponent {
 
       forward = !forward;
       print('boin');
-      Flame.audio.play('boin.mp3', volume: 1.2);
+      Flame.audio.play('boin.mp3', volume: 1.0);
     }
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## [next]
-- Added backgroundColor method for overriding the game background (thanks @wolfenrain) 
+- Added backgroundColor method for overriding the game background (thanks @wolfenrain)
+- Update AudioPlayers version to 0.13.5
 
 ## 0.17.1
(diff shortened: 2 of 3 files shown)


doc/examples/sound/lib/main.dart
```diff
@@ -34,7 +35,7 @@ class Ball extends PositionComponent {
 
       forward = !forward;
       print('boin');
-      Flame.audio.play('boin.mp3', volume: 1.2);
+      Flame.audio.play('boin.mp3', volume: 1.0);
```
erickzanardo: For some reason that I haven't yet been able to take a closer look, when using a volume bigger than `1.0` on Android (works fine on iOS), the sound plays only once, then no more until the end of the execution of the app.

It looks to me that is a problem with AP LOW_LATENCY implementation.

Not sure if this is something related to this specific version of AP because there were a lot of changes on AP since this was implemented.

Please take this into consideration before merging and let me know what do you think. If you could help debugging AP LOW_LATENCY would be great too.

luanpotter: volume must be between 0 and 1, right? what does volume 1.2 even mean? It's certainly a but we must tackle but it should not be very impactful at all...

erickzanardo: Nope, at least from what I understand you can play the file with a bigger volume than its original value. On iOS for example the `1.2` works fine

luanpotter: From what I understand `1` is the maximum volume the device can reach, right? The docs imply you should not go over one:
```
  /// Sets the volume (amplitude).
  ///
  /// 0 is mute and 1 is the max volume. The values between 0 and 1 are linearly
  /// interpolated.
```

erickzanardo: Maybe this working on iOS made me confused then. So we probably would need to make an assertion on AP to prevent this been bigger than 1.

luanpotter: Or maybe make it work for >1 in android and change the docs to reflect that. We should ideally figure out what is this fraction in relation to (original audio volume or device max volume) and make sure both impl have the same interpretation, also reflecting it in the docs.

luanpotter: @erickzanardo I think for now we can merge this an if you please open an issue on AP to investigate that :) (currently version of the ap on flame has a way more severe bug anyway)

erickzanardo: Done: https://github.com/luanpotter/audioplayers/issues/351

PR #196, opened 2019-12-06
WIP: Implementing sprite stacking

wolfenrain:
Implementing sprite stacking

### DIFF
--- a/lib/components/sprite_stack_component.dart
+++ b/lib/components/sprite_stack_component.dart
@@ -1,0 +1,111 @@
+import 'dart:ui';
+
+import '../sprite.dart';
+import '../spritesheet.dart';
+import 'component.dart';
+
+class SpriteStackComponent extends PositionComponent {
+  SpriteSheet spriteSheet;
+
+  static final _sprites = <String, Map<double, Sprite>>{};
+
+  @override
+  set angle(double value) {
+    if (_sprites[spriteSheet.imageName][value] == null) {
+      _generated = false;
+    }
+    super.angle = value;
+  }
+
+  SpriteStackComponent(
+      {String imageName,
+      int textureWidth,
+      int textureHeight,
+      int columns,
+      int rows}) {
+    if (_sprites[imageName] == null) {
+      _sprites[imageName] = new Map();
+    }
+    x = 100;
+    y = 100;
+    width = textureWidth.toDouble();
+    height = textureHeight.toDouble();
+    spriteSheet = SpriteSheet(
+        imageName: imageName,
+        textureWidth: textureWidth,
+        textureHeight: textureHeight,
+        columns: columns,
+        rows: rows);
+  }
+
+  @override
+  void render(Canvas canvas) {
+    if (!loaded()) {
+      return;
+    }
+    prepareCanvas(canvas);
+    _sprites[spriteSheet.imageName][angle]
+        .render(canvas, width: width, height: height);
+  }
+
+  bool _generated = false;
+  bool _generating = false;
+
+  @override
+  bool loaded() {
+    for (var row = 0; row < spriteSheet.rows; row++) {
+      for (var column = 0; column < spriteSheet.columns; column++) {
+        var sprite = spriteSheet.getSprite(row, column);
+        if (!sprite.loaded()) {
+          return false;
+        }
+      }
+    }
+    generate();
+    return _generated;
+  }
+
+  void generate() async {
+    if (_generated || _generating) {
+      return;
+    }
+
+    _generating = true;
+    var pictureRecorder = PictureRecorder();
+    var canvas =
+        Canvas(pictureRecorder, Rect.fromLTWH(0.0, 0.0, width, height));
+
+    double index = 0;
+    canvas.rotate(-angle);
+    for (var row = 0; row < spriteSheet.rows; row++) {
+      for (var column = 0; column < spriteSheet.columns; column++) {
+        var spriteComponent = SpriteComponent.fromSprite(
+            spriteSheet.textureWidth.toDouble(),
+            spriteSheet.textureHeight.toDouble(),
+            spriteSheet.getSprite(row, column));
+        canvas.save();
+        spriteComponent.debugMode = false;
+        spriteComponent.angle = angle;
+        spriteComponent.y -= index++ - 5;
+        spriteComponent.render(canvas);
+        canvas.restore();
+      }
+    }
+
+    var pic = pictureRecorder.endRecording();
+    var image = await pic.toImage(width.toInt(), height.toInt());
+    var sprite = Sprite.fromImage(
+      image,
+      x: 0,
+      y: 0,
+      width: width,
+      height: height,
+    );
+    _sprites[spriteSheet.imageName][angle] = sprite;
+    _generating = false;
+    _generated = true;
+  }
+
+  @override
+  void update(double t) {}
+}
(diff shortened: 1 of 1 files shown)


lib/components/sprite_stack_component.dart
```diff
@@ -0,0 +1,111 @@
+import 'dart:ui';
+
+import '../sprite.dart';
+import '../spritesheet.dart';
+import 'component.dart';
+
+class SpriteStackComponent extends PositionComponent {
+  SpriteSheet spriteSheet;
+
+  static final _sprites = <String, Map<double, Sprite>>{};
```
erickzanardo: I don't think this should be `static` right? Making this static means that every instance of SpriteStackComponent will use this map.

lib/components/sprite_stack_component.dart
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  SpriteStackComponent(
+      {String imageName,
+      int textureWidth,
+      int textureHeight,
+      int columns,
+      int rows}) {
+    if (_sprites[imageName] == null) {
+      _sprites[imageName] = new Map();
```
erickzanardo: `new` keyword is not necessary on recent dart version

lib/components/sprite_stack_component.dart
```diff
@@ -0,0 +1,111 @@
...
+    _sprites[spriteSheet.imageName][angle]
+        .render(canvas, width: width, height: height);
+  }
+
+  bool _generated = false;
+  bool _generating = false;
+
+  @override
+  bool loaded() {
+    for (var row = 0; row < spriteSheet.rows; row++) {
```
erickzanardo: You should cache this, so when this is already done loading, the code will node run those fors anymore. I know these fors should be small and will not hurt the CPU very much, but if have a lot of SpriteStackComponent instances on the game, this can make a difference.

lib/components/sprite_stack_component.dart
```diff
@@ -0,0 +1,111 @@
...
+        canvas.save();
+        spriteComponent.debugMode = false;
+        spriteComponent.angle = angle;
+        spriteComponent.y -= index++ - 5;
+        spriteComponent.render(canvas);
+        canvas.restore();
+      }
+    }
+
+    var pic = pictureRecorder.endRecording();
```
erickzanardo: since you don't re assigns these variables, you can make them final

lib/components/sprite_stack_component.dart
wolfenrain: The reason why this is static at the moment is because it generates a Sprite for the given angle for the given image. If we did that at instance level, every instance would generate a sprite for the given angle for that image.

So if we have two instances that have the same image, and the same angle, they will both generate exactly the same Sprite but twice. Meaning twice the memory and CPU.

But to be honest I am not sure if generating the sprites on the go is the best way to go for us, maybe generate a set of sprites at certain angles, but then those angles are determined by giving a `step` value to the constructor? So that we can generate sprites beforehand instead of on the go? What do you think

lib/components/sprite_stack_component.dart
wolfenrain: Ah yes that is correct. Sorry old habits die hard, will fix it in the next push

lib/components/sprite_stack_component.dart
wolfenrain: Yea you are right, if the `_generated` is already true there is no need to go through the images. Will fix it in the next commit

lib/components/sprite_stack_component.dart
wolfenrain: On a side note, the `_generating` and `_generated` are not static so they lose all meaning if we keep the sprites cache on static level.

lib/components/sprite_stack_component.dart
wolfenrain: Ah yes, my IDE was already complaining, consider it done

wolfenrain:
> Another thing, we usually have Flame's feature independent of the component system, so there is always the feature class and its component, which uses that class. For example you will find that there is the `Sprite` class and the `SpriteComponent` component class, so people who don't want to use Flame's component system, still have easy access to the feature. So my suggestion is to have this `SpriteStack` class, which will be totally agnostic of Flame's component system, and the `SpriteStackComponent` class which will use the previous class.
> 
> What do you think?

I havent thought about it like that, it does make sense. But I am not sure if it is doable. Considering we need an `y` offset for the effect and a way to set an `angle` value to ensure the right rotation and stacking. If that makes sense?

erickzanardo:
> I havent thought about it like that, it does make sense. But I am not sure if it is doable. Considering we need an y offset for the effect and a way to set an angle value to ensure the right rotation and stacking. If that makes sense?

It does, but to solve this we can use the same rationale that we do on Sprite class, where it receives the coordinates on the render method, the same thought  could apply here

lib/components/sprite_stack_component.dart
```diff
@@ -0,0 +1,117 @@
...
+    canvas.rotate(-angle);
+    for (var row = 0; row < spriteSheet.rows; row++) {
+      for (var column = 0; column < spriteSheet.columns; column++) {
+        var spriteComponent = SpriteComponent.fromSprite(
+            spriteSheet.textureWidth.toDouble(),
+            spriteSheet.textureHeight.toDouble(),
+            spriteSheet.getSprite(row, column));
+        canvas.save();
+        spriteComponent.debugMode = false;
+        spriteComponent.angle = angle;
```
luanpotter: instead of creating different sprites for every pair of angle/offset, you can translate and rotate the canvas while you draw on the render method, what do you think?

lib/components/sprite_stack_component.dart
```diff
@@ -0,0 +1,117 @@
...
+    }
+    super.angle = value;
+  }
+
+  SpriteStackComponent(
+      {String imageName,
+      int textureWidth,
+      int textureHeight,
+      int columns,
+      int rows}) {
```
luanpotter: you could take a `SpriteSheet` instance here instead, right?

lib/components/sprite_stack_component.dart
```diff
@@ -0,0 +1,117 @@
...
+    if (_sprites[imageName] == null) {
+      _sprites[imageName] = Map();
+    }
+    x = 100;
+    y = 100;
```
luanpotter: why is this a default value?

lib/components/sprite_stack_component.dart
```diff
@@ -0,0 +1,117 @@
...
+    for (var row = 0; row < spriteSheet.rows; row++) {
+      for (var column = 0; column < spriteSheet.columns; column++) {
+        var spriteComponent = SpriteComponent.fromSprite(
+            spriteSheet.textureWidth.toDouble(),
+            spriteSheet.textureHeight.toDouble(),
+            spriteSheet.getSprite(row, column));
+        canvas.save();
+        spriteComponent.debugMode = false;
+        spriteComponent.angle = angle;
+        spriteComponent.y -= index++ - 5;
```
luanpotter: shouldn't this be something the user of the class would like to customize?

wolfenrain: These were just some test values to see how it responed. In my current (unstable) version they have all been removed

lib/components/sprite_stack_component.dart
wolfenrain: These were just some test values to see how it responded. In my current (unstable) version they have all been removed

erf:
This is batched rendering of sprites right? This seem like an important feature.. Is it still being worked on?

erickzanardo:
@wolfenrain This PR has been open for a long time now and has been behind the latests commits on master.

I am gonna close this for the moment, but feel free to return to work on this anytime.

Thanks

PR #201, opened 2020-01-03
Tweaking text box rendering to reduce pixelated text

kurtome:
Changing the image rendering of text to high quality to ensure sharp edges on rendering

### DIFF
--- a/lib/components/text_box_component.dart
+++ b/lib/components/text_box_component.dart
@@ -24,6 +24,9 @@ class TextBoxConfig {
 }
 
 class TextBoxComponent extends PositionComponent with Resizable {
+  static final Paint _imagePaint =
+      BasicPalette.white.paint..filterQuality = FilterQuality.high;
+
   Position p = Position.empty();
 
   String _text;
@@ -132,9 +135,7 @@ class TextBoxComponent extends PositionComponent with Resizable {
       return;
     }
     prepareCanvas(c);
-    c.drawImage(_cache,
-        Offset.zero,
-        BasicPalette.white.paint..filterQuality = FilterQuality.high);
+    c.drawImage(_cache, Offset.zero, _imagePaint);
   }
 
   Future<Image> _redrawCache() {
--- a/lib/components/text_box_component.dart
+++ b/lib/components/text_box_component.dart
@@ -24,8 +24,8 @@ class TextBoxConfig {
 }
 
 class TextBoxComponent extends PositionComponent with Resizable {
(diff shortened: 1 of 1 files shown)


lib/components/text_box_component.dart
```diff
@@ -132,7 +132,9 @@ class TextBoxComponent extends PositionComponent with Resizable {
       return;
     }
     prepareCanvas(c);
-    c.drawImage(_cache, Offset.zero, BasicPalette.white.paint);
+    c.drawImage(_cache,
+        Offset.zero,
+        BasicPalette.white.paint..filterQuality = FilterQuality.high);
```
luanpotter: This is an amazing finding! Thanks for this :)

Two points:
1. can we format it like so?
```dart
    c.drawImage(
        arg1,
        arg2,
        arg3,
    );
```
2. I wonder if this is also required in `text_config` and `text_component`? I looks like it's only here because it's rendering a pre-rendered image, or is it good to add everywhere?

erickzanardo: Also, maybe this should be a parameter? Even if with a default value

kurtome: 1. yup
2. from the docs, `filterQuality` effects image rendering, so I don't think it matters in any other scenario. I think generally it's good to add if it doesn't hurt performance, which in this case shouldn't be a problem since it's only redrawing on some frames and the image itself is small / not complicated.

I think ideally it should be possible to refactor `TextBoxComponent` to not need to cache an `Image`. That way there would be no quality loss. Is this caching just happening to avoid all the object allocations in each loop? If so, you could just cache `TextPainter` and use it on every loop.

kurtome: @erickzanardo Maybe, but I don't know when you'd want to draw text as a low quality image, I don't think this will have a big performance impact (I haven't done extensive testing though)

luanpotter: @kurtome makes sense, we can roll out this as is for me :)

the cache is intended to increase performance, specially once the text is all drawn, to avoid computing everything

kurtome:
@luanpotter thanks for the review! I'm not sure why the Travis build is failing, is that something from this PR?

erickzanardo:
@kurtome there are files not formatted according to our linter, to fix those problems, just run the following command on the terminal:

`flutter format .`

kurtome:
@erickzanardo Thanks! I started using AndroidStudio for Flutter code and apparently that doesn't automatically run the formatter like VSCode did :(

PR #204, opened 2020-01-11
adding ninebox class and example

luanpotter:


lib/nine_box.dart
```diff
@@ -0,0 +1,86 @@
...
+  ///
+  /// If [destTileSize] is not provided, the evaluated [tileSize] is used instead
+  /// (so no scaling happens).
+  NineBox(this.sprite, {int tileSize, int destTileSize}) {
+    this.tileSize = tileSize ?? sprite.src.width.toInt();
+    this.destTileSize = destTileSize ?? tileSize;
+  }
+
+  /// Renders this nine box as a rectangle of coordinates ([x], [y]) and size ([width], [height]).
+  void draw(Canvas c, double x, double y, double width, double height) {
```
erickzanardo: We should probably have a drawRect as well, what do you think?

lib/nine_box.dart
```diff
@@ -0,0 +1,86 @@
...
+  ///
+  /// [srite] is the 3x3 grid and [tileSize] is the size of each tile.
+  /// The src sprite must a square of size 3*[tileSize].
+  ///
+  /// If [tileSize] is not provided, the width of the sprite is assumed as the size.
+  /// Otherwise the width and height properties of the sprite are ignored.
+  ///
+  /// If [destTileSize] is not provided, the evaluated [tileSize] is used instead
+  /// (so no scaling happens).
+  NineBox(this.sprite, {int tileSize, int destTileSize}) {
```
erickzanardo: I am not sure about the name, NineBox seems a little unclear on what this is, maybe something like NineTilePanel, or something like this?

Also we should provide a component for this as well, to make it easy to use on basegames

lib/nine_box.dart
luanpotter: excellent idea

lib/nine_box.dart
luanpotter: I thought about adding a component, but I then realized you would probably never have a component with just this? like I use it in other components as composable piece

luanpotter: about the name, I can change to NineTilePanel

I think no name is totally clear but yours seems better

erickzanardo: I can see come good usages for a component like this, for example, if you override this component to draw some text inside it, add the Tapable mixin, you got yourself an easy way to make buttons inside Flame.

luanpotter: ok, I will add one

CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 ## [next]
+- Tweaking text box rendering to reduce pixelated text (thanks, @kurtome)
+- Adding nine_box component
```
erickzanardo: Update here too, when you rename the class

luanpotter:
@erickzanardo AFAICS every public api is document

PR #206, opened 2020-01-22
Nine tile box widget

erickzanardo:
This PR adds the NineTileBox widget, the same thing as the recently added NineTileBox and NineTileBoxComponent by @luanpotter , but the widget version.


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ## next
```
luanpotter: [next]

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
...
+- Adding NineTileBox widget
 - Fixing compilations errors regarding changes on `box2_flame`
```
luanpotter: this was deployed, right? this diff is weird, please take a look

doc/examples/widgets.md
```diff
@@ -0,0 +1,28 @@
+# Widgets
+
+One cool feature when developing games with Flutter is the ability to use Flutter's extensive toolset for building UIs, Flame tries to expand that introducing widgets which are specially made with games in mind.
+
+Here you can find all the available widgets provided by Flame.
+
+You can also see all the widgets showcased inside a Dashbook sandbox [here](/doc/examples/widgets)
```
luanpotter: link to dashbook github on the word Dashbook

doc/examples/widgets.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Nine Tile Box
+
+A Nine Tile Box is a rectangle drawn using a grid sprite.
+
+The grid sprite is a 3x3 grid and with 9 blocks, representing the 4 courners, the 4 sides and the middle.
+
+The corners are drawn at the same size, the sides are streched on the side direction and the middle is expanded both ways.
```
luanpotter: Please link to the in-game version here and briefly explain the difference

doc/examples/widgets/.flutter-plugins-dependencies
```diff
@@ -0,0 +1 @@
+{"_info":"// This is a generated file; do not edit or check into version control.","dependencyGraph":[{"name":"audioplayers","dependencies":["path_provider"]},{"name":"path_provider","dependencies":[]}]}
```
luanpotter: add this to gitignore as per requested in the file

doc/examples/widgets/lib/main.dart
```diff
@@ -0,0 +1,29 @@
...
+      'default',
+      (ctx) => Container(
+            width: ctx.numberProperty('width', 200),
+            height: ctx.numberProperty('height', 200),
+            child: NineTileBox(
+              image: nineTileBoxImage,
+              tileSize: 16,
+              destTileSize: 50,
+              child: const Center(
+                  child: const Text('Cool label',
```
luanpotter: please format like this:
```
    child: Foo(
         a: b,
         cccccc,
        // ...
    ),
```

lib/widgets/nine_tile_box.dart
```diff
@@ -0,0 +1,138 @@
...
+import '../sprite.dart';
+import 'dart:ui';
+
+class _Painter extends widgets.CustomPainter {
+  final Image image;
+  final double tileSize;
+  final double destTileSize;
+
+  _Painter(
+      {@required this.image,
```
luanpotter: can we have this formatted like so?
```
_Painter({
    @required ...,
    @required ...,
    @required ...,
});
```

lib/widgets/nine_tile_box.dart
```diff
@@ -0,0 +1,138 @@
...
+      @required this.tileSize,
+      @required this.destTileSize});
+
+  @override
+  void paint(Canvas canvas, Size size) {
+    final topLeftCorner =
+        Sprite.fromImage(image, x: 0, y: 0, width: tileSize, height: tileSize);
```
luanpotter: how about a getImageTime method or something like that that takes two ints x and y, multiply them by tileSize and do the rest and returns the Sprite?

lib/widgets/nine_tile_box.dart
```diff
@@ -0,0 +1,138 @@
...
+
+    // Middle
+    for (var y = destTileSize;
+        y < size.height - destTileSize;
+        y = y + destTileSize) {
+      for (var x = destTileSize;
+          x < size.width - destTileSize;
+          x = x + destTileSize) {
+        middle.renderRect(
+            canvas, Rect.fromLTWH(x, y, destTileSize, destTileSize));
```
luanpotter: you definitely don't need to do this... I thought of this at first, but please take a look at my implementation. you can expand the middle block and render it once. each 9 tile is rendered once, expand either: not at all, only in one direction and, the middle, on both directions. it's more efficient.

also this doesn't seem to cover if the width is not a multiple of destTileSize

lib/widgets/nine_tile_box.dart
```diff
@@ -0,0 +1,138 @@
...
+        Rect.fromLTWH(
+            0, size.height - destTileSize, destTileSize, destTileSize));
+    bottomRightCorner.renderRect(
+        canvas,
+        Rect.fromLTWH(size.width - destTileSize, size.height - destTileSize,
+            destTileSize, destTileSize));
+  }
+
+  @override
+  bool shouldRepaint(_) => false;
```
luanpotter: hum, I wonder if you need to check if your props changed here? not sure, did you read the docs/tried changing props to see if it works? maybe check similar implementations on the flutter sdk?

lib/widgets/nine_tile_box.dart
```diff
@@ -0,0 +1,138 @@
...
+
+  NineTileBox({
+    @required this.image,
+    @required this.tileSize,
+    @required this.destTileSize,
+    this.child,
+    this.width,
+    this.height,
+    this.padding,
+  });
```
luanpotter: this is perfectly formatted! please use this everywhere

lib/widgets/nine_tile_box.dart
```diff
@@ -0,0 +1,138 @@
...
+  final double tileSize;
+  final double destTileSize;
+  final double width;
+  final double height;
+
+  final widgets.Widget child;
+
+  final widgets.EdgeInsetsGeometry padding;
+
+  NineTileBox({
```
luanpotter: every widget is supposed to take and pass along the key

pubspec.yaml
```diff
@@ -15,6 +15,7 @@ dependencies:
   convert: ^2.0.1
   flutter_svg: ^0.15.0
   flare_flutter: ^1.5.4
+  meta: ^1.1.8
```
luanpotter: why?

erickzanardo: This is the package responsible for the `@required` annotation.

luanpotter: is it the one we normally use in flutter? because I don't remember to have to import a new package to use it in flutter regular app projects

erickzanardo: Yep, I did some research, and this package is also used by the flutter lint, I also thought at first that it was already bundled on flutter, but in the end I discovered that it isn't.

lib/widgets/nine_tile_box.dart
erickzanardo: Not really, I did thought on expanding the tile initially, but if I expand, a nine tile box sprite like this will not work (I know this sprite is ugly, I made it very fast, just to make my point :) )

ntb

If the in game implementation of the nin tile box is expanding, I suggest changing it to work with repeating tiles, otherwise we are excluding the support for sprites with details on the middles, which is something very cool that I think we should support.

> also this doesn't seem to cover if the width is not a multiple of destTileSize
Abou that, I agree with you, but I have made several tests on the dashbook story of this widgets, and it worked everytime, but I will give a second check.

CHANGELOG.md
erickzanardo: yep, I may have messed up when re basing my branch.

doc/examples/widgets.md
erickzanardo: Done, let me know if it is ok

doc/examples/widgets/lib/main.dart
erickzanardo: this is the result of the `flutter format` command, I think we should keep to it, so at least we have the same formatting base, even though I prefer the style suggested by you.

lib/widgets/nine_tile_box.dart
erickzanardo: same thing as above, `flutter format` result

lib/widgets/nine_tile_box.dart
erickzanardo: Fair enought, done!

lib/widgets/nine_tile_box.dart
erickzanardo: From my understanding, been an Stateless component, its attributes should never change, hence the warning immutable warning we receive on attributes on widgets, so I would say that the way this is implemented is correct, we should never repaint this, as the attributes should never change.

lib/widgets/nine_tile_box.dart
erickzanardo: I always use the flutter formater, this is not my fault/doing hahaha

lib/widgets/nine_tile_box.dart
erickzanardo: good catch, done!

erickzanardo:
@luanpotter resolved all the comments, and the ones that I didn't solved, I have added some comments as well, let me what you think.

doc/examples/widgets/lib/main.dart
luanpotter: it only tells you to format like that because you are missing a trailing comma. my formatting suggestion is formatter valid :)

lib/widgets/nine_tile_box.dart
luanpotter: because you lack trailing comma, please take a look again at my suggestion, it is totally formatter valid.

lib/widgets/nine_tile_box.dart
luanpotter: makes sense

doc/examples/widgets.md
luanpotter: it's good!

lib/widgets/nine_tile_box.dart
luanpotter: I see your use case, however, it makes it much more complex to handle the case I just mentioned. like if the container is 2.5 tiles of width, which part of the tiles will you show? I believe your "tiling" idea would only work well for perfectly tile multiples, but it's better to have something that works on every sizing, right? not sure, but I think we should have both game and widget impl. consistent

doc/examples/widgets/lib/main.dart
erickzanardo: fair enough, I didn't knew about the trailing comma thing, but my desire for this kind of thing be on our lint persists, it waster a lot of our time this not been there, as is very hard to remember all these stuff, the trailing coma should have been on the formatter, do you have any idea on how to set it up @luanpotter ?

lib/widgets/nine_tile_box.dart
erickzanardo: Fair enough, to be honest, I would like to keep my implementation, to enable that kind of sprite I mentioned before, but yeah, I can't find a good solution for cases when the dimensions are not multiplier of the tile size, I am gonna change to expand the middle.

erickzanardo:
@luanpotter please, take another look, I have finished all your comments.

PR #207, opened 2020-01-22
Fixing compilation errors on box 2d

erickzanardo:


lib/box2d/box2d_component.dart
```diff
@@ -159,10 +159,10 @@ abstract class BodyComponent extends Component {
   }
 
   void _renderCircle(Canvas canvas, Fixture fixture) {
-    final Vector2 center = Vector2.zero();
+    var center = Vector2.zero();
     final CircleShape circle = fixture.getShape();
     body.getWorldPointToOut(circle.p, center);
-    viewport.getWorldToScreen(center, center);
+    center = viewport.getWorldToScreen(center);
```
luanpotter: Just declare the variable here with final

luanpotter: And the same on several other places

erickzanardo: I can't, as it is used before, on line 164. But I have changed on the other places.

Issue #202, opened 2020-01-07
Cannnot set width or height of FlareComponent


stefankrosell:
I assume width and height should be passed on to internal _flareAnimation

FrankenApps:
Not sure if I understand correctly, but I am able to set width and height like this:
`FlareAnimation myAnimation  = await FlareAnimation.load("assets/myAnimation.flr");
myAnimation.width = 50;
myAnimation.height = 30;`

However the location seems to be influenced by this action, because
`myAnimation.x = 300;
myAnimation.y = 80`
moves in a strange way, if the size is changed.

It's even worse, when you have two FlareComponents in your view, as they seem to influence each others position.

stefankrosell:
FlareAnimation can be rescaled, but FlareComponent does not respond to new values for width or height.

luanpotter:
@erickzanardo wanna take a look?

erickzanardo:
I should be looking into this tonight.

stefankrosell:
Great! It would also be very nice to have "updateAnimation" available in the component.

erickzanardo:
This should be fixed by #210

erickzanardo:
Fixed on #210, a new version should be shipped soon with this fix, stay tuned for the next release.

PR #210, opened 2020-01-22
Improving FlareComponent API and updating FlareFlutter dependency

erickzanardo:
Solves #202 as well as updates the FlareFlutter dependency.

### DIFF
--- a/doc/examples/flare/lib/main_component.dart
+++ b/doc/examples/flare/lib/main_component.dart
@@ -1,10 +1,10 @@
 import 'dart:ui';
 import 'package:flame/gestures.dart';
-import 'package:flame/flame.dart';
 import 'package:flame/game.dart';
 import 'package:flame/components/flare_component.dart';
 import 'package:flame/text_config.dart';
 import 'package:flame/position.dart';
+import 'package:flame/palette.dart';
 
 import 'package:flutter/material.dart';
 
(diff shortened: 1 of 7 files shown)


doc/examples/flare/lib/main_component.dart
```diff
@@ -0,0 +1,74 @@
...
+void main() {
+  WidgetsFlutterBinding.ensureInitialized();
+
+  final game = MyGame();
+  runApp(game.widget);
+}
+
+class MyGame extends BaseGame with TapDetector, DoubleTapDetector {
+  final TextConfig fpsTextConfig =
+      const TextConfig(color: const Color(0xFFFFFFFF));
```
luanpotter: `BasicPalette.white.color`

doc/examples/flare/lib/main_component.dart
```diff
@@ -0,0 +1,74 @@
...
+
+  final paint = Paint()..color = const Color(0xFFE5E5E5E5);
+  final List<String> _animations = ["Stand", "Wave", "Jump", "Dance"];
```
luanpotter: use single quotes

lib/components/flare_component.dart
```diff
@@ -20,6 +20,12 @@ class FlareComponent extends PositionComponent {
     });
   }
 
+  void updateAnimation(String animation) {
```
luanpotter: should we add to the docs that you now need to call this method? or at least a dartdoc here explaining why we need to call it and where briefly, because looking at the example it's not clear to me where it's needed or not

PR #215, opened 2020-02-04
Updated flutter_svg and other packages

flowhorn:
In order to support the newest versions and have improved macos support, I incremented the flutter_svg version + others.


erickzanardo:
Please rebase your PR to the develop branch, and also, add an entry to the CHANGELOG.md describing the change. Other than that, looks good to me.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 0.18.1
```
luanpotter: we use [next] until we decide the next version (if its patch or minor), but you probably guessed right here haha

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## next
+- Updated versions for bugfixes + improved macOS support.
+- Update flutter_svg to 0.17.1 [Breaking Change]
```
erickzanardo: Is this really a breaking change? I mean, for people who uses Flame, there is no changes required right?

flowhorn: Yes, this is correct. There is no change for using flame, I just wanted to show that I had to change a method in flutter svg while the other plugins were updated without any changes

PR #216, opened 2020-02-05
Expose stepTime paramter from the Animation class to the animation component

luanpotter:


lib/components/animation_component.dart
```diff
@@ -35,6 +36,7 @@ class AnimationComponent extends PositionComponent {
       textureY: textureY,
       textureWidth: textureWidth,
       textureHeight: textureHeight,
+      stepTime: stepTime ?? 0.1,
```
erickzanardo: Why don't use default parameters instead of using `??`?

PR #228, opened 2020-03-06
Adding Component#onMount

erickzanardo:
This PR adds a simple utility method to enable using things like `gameRef` on components during its initialization.

This can supply for the limitation where you weren't able to use `gameRef` on the component construtor.

### DIFF
--- a/doc/components.md
+++ b/doc/components.md
@@ -37,6 +37,8 @@ The `destroy` method can be implemented to return true and warn the `BaseGame` t
 
 The `isHUD` method can be implemented to return true (default false) to make the `BaseGame` ignore the `camera` for this element.
 
+The `onMount` method can be overridden to run initializations code for the component, when this method is called, BaseGame has make sure that all the mixin which would change this component behaviour, is already resolved.
+
 There are also other implementations:
 
 * The `AnimationComponent` takes an `Animation` object and renders a cyclic animated sprite (more details about Animations [here](/doc/images.md#Animation))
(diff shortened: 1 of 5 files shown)


doc/components.md
```diff
@@ -37,6 +37,8 @@ The `destroy` method can be implemented to return true and warn the `BaseGame` t
 
 The `isHUD` method can be implemented to return true (default false) to make the `BaseGame` ignore the `camera` for this element.
 
+The `onMount` method can be overridden to run initializations code for the component, when this method is called, BaseGame has make sure that all the mixin which would change this component behaviour, is already resolved.
```
luanpotter:
```suggestion
The `onMount` method can be overridden to run initialization code for the component. When this method is called, BaseGame ensures that all the mixins which would change this component behaviour are already resolved.
```

lib/components/component.dart
```diff
@@ -54,6 +54,12 @@ abstract class Component {
   /// It can be any integer (negative, zero, or positive).
   /// If two components share the same priority, they will probably be drawn in the order they were added.
   int priority() => 0;
+
+  /// Called when the component has been added and preperad by the game instance
```
luanpotter:
```suggestion
  /// Called when the component has been added and preperad by the game instance.
```

lib/components/component.dart
```diff
@@ -54,6 +54,12 @@ abstract class Component {
...
+  ///
+  /// This can be used to make initializations on your component as when this method is called
+  /// things like resize, and mixins are already set for the component.
```
luanpotter:
```suggestion
  /// This can be used to make initializations on your component as, when this method is called,
  /// things like resize (and other mixins) are already set and usable.
```

PR #235, opened 2020-03-15
amend type for width and height property

wenxiangjiang:
Because the default type of a property is dynamic, an error will be prompted when assigning an int type.

```
type 'int' is not a subtype of type 'double'
```


renancaraujo:
lgtm

PR #240, opened 2020-03-20
Support line feed when create animation from a single image source

wenxiangjiang:
I want to create a series of animations through `AnimationComponent.sequenced`. Since there are 60 frames in a image source, the total size of the image source is `20370px * 356px`, but the size of the image finally loaded into memory is `16384px * 267px`.

I guess this is because `flutter` limits the maximum width of the image to 16384px, so I added some code to support line feed when create a sequenced animation from a single image source.


lib/animation.dart
```diff
@@ -55,6 +55,7 @@ class Animation {
   ///
   /// From a single image source, it creates multiple sprites based on the parameters:
   /// [amount]: how many sprites this animation is composed of
+  /// [amountPerRow]: how many sprites per row
```
luanpotter: I like this feature, but since it's a bit more complicated to understand, can we have a bit more text on the docs explaining what this means? maybe with an example...

luanpotter:
@wenxiangjiang I'm surprised to see you have such a huge animation haha I think I never saw a 60 frames game animation before. This LGTM but I have one nit about the doc. Also please add the change to the CHANGELOG file under [next], so it gets picked up on the next release. Thanks!

wenxiangjiang:
@luanpotter I also made the game for the first time, and the designer gave me a picture containing 60 frames. Although I have deleted some pictures, I think this feature may still be needed.

I just added some documentation descriptions and updated an example, I hope users can understand what I mean.

But I found `Travis CI` prompt an error...

erickzanardo:
@wenxiangjiang To fix the CI, you can just run a `flutter format .` on the terminal, it is just a linting problem

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Check if chidren are loaded before rendering on ComposedComponent (thanks @wenxiangjiang)
 - Amend type for width and height properties on Animation.sequenced (thanks @wenxiangjiang)
 - Fixing Tapable position checking
+- Support line feed when create animation from a single image source
```
erickzanardo: Add a `(thanks @wenxiangjiang)` after the entry to identify as your contribution :)

lib/animation.dart
```diff
@@ -55,6 +55,9 @@ class Animation {
   ///
   /// From a single image source, it creates multiple sprites based on the parameters:
   /// [amount]: how many sprites this animation is composed of
+  /// [amountPerRow]: If the sprites used to create a animation are not on the same row,
```
erickzanardo: `an animation`

wenxiangjiang:
I really made a low-level mistake. :(

@erickzanardo I learned something new from you, Thanks!

erickzanardo:
> I really made a low-level mistake. :(
> 
> @erickzanardo I learned something new from you, Thanks!

Hahaha, don't worry, I always forget about the linting as well, it is quite and easy thing to forget.

PR #242, opened 2020-03-24
Fixing edge cases with text box component.

kurtome:
Cases this fixes:
 - One word only, "hello" would display as "ello"
 - Single line, "hello world" would display as "hello worl"


kurtome:
Thanks for the review, updated the CHANGELOG

PR #253, opened 2020-03-29
BaseGame extension with Box2D support

spydon:
To simplify control over box2d components and handle them in the same way as other components are handled in the BaseGame.


spydon:
Realized this won't work, since the update functions will be called two times for the BodyComponents.
Will rethink this a bit and update the PR.

spydon:
There we go, now it can be reviewed.

PR #259, opened 2020-04-02
Adding Animation#paint

erickzanardo:
Adding paint to AnimationComponent, so it can be repassed to the sprite, useful for doing something like the following red blink effect:

blink

### DIFF
--- a/lib/components/animation_component.dart
+++ b/lib/components/animation_component.dart
@@ -54,7 +54,9 @@ class AnimationComponent extends PositionComponent {
   @override
   void render(Canvas canvas) {
     prepareCanvas(canvas);
-    animation.getSprite().render(canvas, width: width, height: height, overridePaint: paint);
+    animation
+        .getSprite()
+        .render(canvas, width: width, height: height, overridePaint: paint);
   }
 
   @override
--- a/lib/components/animation_component.dart
+++ b/lib/components/animation_component.dart
@@ -5,7 +5,7 @@ import 'package:flame/animation.dart';
 
 class AnimationComponent extends PositionComponent {
   Animation animation;
-  Paint paint;
+  Paint overridePaint;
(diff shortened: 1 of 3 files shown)


lib/components/animation_component.dart
```diff
@@ -53,7 +54,9 @@ class AnimationComponent extends PositionComponent {
   @override
   void render(Canvas canvas) {
     prepareCanvas(canvas);
-    animation.getSprite().render(canvas, width: width, height: height);
+    animation
+        .getSprite()
+        .render(canvas, width: width, height: height, overridePaint: paint);
```
luanpotter, line 59: is a default value not necessary?

lib/components/animation_component.dart
```diff
@@ -5,6 +5,7 @@ import 'package:flame/animation.dart';
 
 class AnimationComponent extends PositionComponent {
   Animation animation;
+  Paint paint;
```
luanpotter, line 8: should we call it overridePaint to be consistent with Sprite?

lib/components/animation_component.dart
erickzanardo: Nope, Sprite already has a default to handle that.

lib/components/animation_component.dart
erickzanardo: overridePaint is the name only on Sprite's render methods, the Sprite also has a property called paint. I think we should keep the paint name, as this is a property of the AnimationComponent.

I can change to overridePaint if you prefer, but in this case I think it will be a little confusing.

erickzanardo: Changed my mind, it is overridePaint now.

erickzanardo:
@luanpotter I took the opportunity of this PR to add an `overridePaint` to `SpriteComponent` as well.

PR #264, opened 2020-04-05
Fixing addLater component lifecycle

erickzanardo:
When using `BaseGame#addLater`, `onMount` on Component was not called, causing problematic behaviours.

### DIFF
--- a/lib/game/game.dart
+++ b/lib/game/game.dart
@@ -176,6 +176,8 @@ abstract class BaseGame extends Game with TapDetector {
     if (c is ComposedComponent) {
       c.components.forEach(preAdd);
     }
+
+    c.onMount();
   }
 
   /// Adds a new component to the components list.
@@ -184,7 +186,6 @@ abstract class BaseGame extends Game with TapDetector {
   void add(Component c) {
     preAdd(c);
     components.add(c);
-    c.onMount();
   }
 
   /// Registers a component to be added on the components on the next tick.
@@ -229,7 +230,7 @@ abstract class BaseGame extends Game with TapDetector {
   /// You can override it further to add more custom behaviour.
   @override
   void update(double t) {
-    _addLater.forEach(add);
+    components.addAll(_addLater);
     _addLater.clear();
 
     components.forEach((c) => c.update(t));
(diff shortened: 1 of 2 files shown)


lib/game/game.dart
```diff
@@ -229,7 +229,7 @@ abstract class BaseGame extends Game with TapDetector {
   /// You can override it further to add more custom behaviour.
   @override
   void update(double t) {
-    components.addAll(_addLater);
+    _addLater.forEach(add);
```
luanpotter, line 232: this doesn't look right to me. preAdd is called immediately when the component is added to the add later list. you are calling it twice by calling add here. I separated all handling to `preAdd` so we didn't have this problem.

erickzanardo: This makes sense, I don't know how I missed the `predAdd` method...

erickzanardo:
Updated @luanpotter

spydon:
Wouldn't it be cleaner to iterate over the _addLater list in update and send in the add function?
So that we don't have to have to separate code paths.
Something like:
`_addLater.foreach(add)`

erickzanardo:
> Wouldn't it be cleaner to iterate over the _addLater list in update and send in the add function?
> So that we don't have to have to separate code paths.
> Something like:
> `_addLater.foreach(add)`
> 
> EDIT: Oh, that's what it used to do. Can't you just move out preAdd out of the addLater() logic to avoid the double call?

@luanpotter I kind of like this idea, what do you think?

spydon:
> LGTM

What do you think about removing the preAdd call from the addLater() logic and just do _addLater.foreach(add) in update? So that we can have two slightly less different different code paths for the adding logic.

luanpotter:
@spydon that is a a good idea, IIRC I started doing that but I changed to call preAdd prematurely because one of my components was expecting things to be initialized right away, even before being actually added to the game. I don't recall the details rn but as long as our components keep working even with the delayed call to preAdd, I think it's ok to change.

luanpotter:
@spydon @erickzanardo let's merge this as is and we can reconsider whether to join preAdd to add or not on a followup. I actually think it might be valuable to be able to override each one individually as you might want to hook on preAdd (maybe should be called prepare or something) and also independently be able to replace the component list with another impl of ordered set by override add.

PR #266, opened 2020-04-07
Fixes onDestroy been called multiple times.

erickzanardo:


erickzanardo:
> LGTM, beware this is a breaking change on flame as we expose ordered_set api

Yeah, but I don't think this is problematic. We can just make this `0.19.0` imo

Issue #110, opened 2019-06-17
Support `shouldRepaint` method


truongsinh:
In my game, sometimes there is no pixel change at all, it's kind of waste to repaint if I know Skia should not.

erickzanardo:
I am not sure about this, in a Game it is very unusual not needing to re draw something, that is why game usually uses active rendering instead os passive rendering, like normal/simpler applications do.

Also, I think this would change basically nothing on perfomance, since Flutter always strives for, and achieves 60fps on apps, we would only add unnecessary complexity with this method.

That said, are you having any perfomance issues @truongsinh ?

And @luanpotter what are your thoughts on this?

truongsinh:
Just to clarify, the performance is great for my game (just a simple clone of Flappy Bird), even on a crappy Samsung A3 2014.

That being said, when I turn on performance overlay, I see it's always repainting (as you describee as active rendering). I wonder whether having `shouldRepaint` could even further improve performance for some complicated games with hundreds of sprites (both rendering and battery consumption)

erickzanardo:
@truongsinh I am still not sure about this, nevertheless, I created a branch named `should-repaint` which adds a method `shouldRepaint` on the Game class, that you can override and controls its, its default implementation always returns true.

Would you mind making a test and see how this improves the perfomance, battery consumption? To make sure that this is worth the small added complexity?

truongsinh:
Hi, I just tried the `shouldRepaint` method, but the performance overlay is still running (the expectation is that the green indicator not moving anymore when `shouldRepaint` is false).

<img width="400" alt="Screen Shot 2019-06-26 at 6 36 03 PM" src="https://user-images.githubusercontent.com/358585/60176725-59e53300-9841-11e9-8a23-130f38ad0732.png">

erickzanardo:
hm, weird, I have added another commit to the branch. if there any other repaints now, it should be because of stuff of flutter, Flame should not paint anything when the method returns false now. Can't you give your test another try?

ProgrammingLife:
I also do not need to repaint every frame. There are many small games that don't need it. Such as: Sudoku, simple games for kids, educational tests, puzzles/jigsaw (this should be repainted only when we dragging some piece and the rest of the time player just thinks what to do next) and other educational stuff. We don't need aggressive repainting. It would be cool to stop it from repainting when we don't need it.

erickzanardo:
Hey @truongsinh have you had any chance to test this?

[...]

PR #272, opened 2020-04-10
Pause and resume engine

erickzanardo:
Solves #110 .

On the example the engine don't start right on, and upon tapping the screen, it toggles the engine execution.

I can see two motivations for this:

1- Pause the engine when the game is paused to save battery.

2- Not starting the engine upon creation can be very helpful, as we can for example leave the game rendered on the background while we show some tutorials or something like that.

Let me know what you think.


lib/game/game.dart
```diff
@@ -366,10 +391,12 @@ class GameRenderBox extends RenderBox with WidgetsBindingObserver {
   }
 
   void _scheduleTick() {
+    _running = true;
     _frameCallbackId = SchedulerBinding.instance.scheduleFrameCallback(_tick);
   }
 
   void _unscheduleTick() {
+    _running = false;
     SchedulerBinding.instance.cancelFrameCallbackWithId(_frameCallbackId);
```
luanpotter, line 400: could _frameCallbackId be null?

erickzanardo: I don't think so, I haven't changed this, and this seems to be working so far...

lib/game/game.dart
```diff
@@ -366,11 +391,14 @@ class GameRenderBox extends RenderBox with WidgetsBindingObserver {
...
 
   void _scheduleTick() {
+    _running = true;
     _frameCallbackId = SchedulerBinding.instance.scheduleFrameCallback(_tick);
   }
 
   void _unscheduleTick() {
-    SchedulerBinding.instance.cancelFrameCallbackWithId(_frameCallbackId);
+    _running = false;
+    if (_frameCallbackId != null)
```
luanpotter, line 400: please add braces to the if statement

PR #276, opened 2020-04-12
FlameBinding has invalid inheritance

spydon:
With the new flutter, FlameBinding had an invalid inheritance (and it has a typo in the class name).
So I was going to fix that, but it seems like it is not needed.
I don't exactly understand what it was for, so please verify.

### DIFF
--- a/lib/flame.dart
+++ b/lib/flame.dart
@@ -41,6 +41,8 @@ class Flame {
       {AssetBundle bundle,
       bool fullScreen = true,
       DeviceOrientation orientation}) async {
+    initializeWidget();
+
     if (fullScreen) {
       await util.fullScreen();
     }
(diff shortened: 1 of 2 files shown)


lib/flame.dart
```diff
@@ -53,24 +51,5 @@ class Flame {
 
     _bundle = bundle;
   }
-
-  /// TODO verify if this is still needed (I don't think so)
-  static void initializeWidget() {
```
erickzanardo, line 58: I think this is useful still, tbh I didn't even remember that we had a init method, I usually always do this `WidgetsFlutterBinding.ensureInitialized();` manually.

spydon: Alright, should I remove the TODO comment?

spydon: From the comment for ensureInitialized:
```
  /// You only need to call this method if you need the binding to be
  /// initialized before calling [runApp].
```
Do we need the binding initialized before runApp?

erickzanardo: yeah, the ensureInitialized is needed when you need to use some APIs to make full screen or things like that before the runApp.

I think we could remove the TODO

spydon: Done

PR #281, opened 2020-04-15
Added Sprites example

erf:
I added a `sprites.dart` example which renders 500 random positioned sprites.


spydon:
Nice! This is very focused on sprites, so maybe it would fit better as `flame/doc/examples/sprite` for example?

erf:
Aha! I was not aware that examples were kept in the doc folder. I will move it there. It might be a bit redundant since you have a particle project, but i was thinking of doing some benchmarking and comparing with a new sprite batch im working on. Is there anyting else i need to consider? I should perhaps do some commenting.

erickzanardo:
We need to organise better our examples, but atm @spydon suggestion would the best place to but your test. The `example` folder is just a quick simple example that we use when want to do a very quick test.

erf:
@spydon i moved the example to the `docs/examples/sprites` folder. rebased and forced pushed.

erf:
So i dont need to update my `sprite-example` branch from the `develop` branch?

spydon:
@erf you don't have any overlapping changes with what's in the develop branch now since you are only doing changes to that example folder, so it should be fine.

doc/examples/sprites/lib/main.dart
```diff
@@ -0,0 +1,39 @@
...
+    final r = Random();
+    try {
+      List.generate(500, (i) => SpriteComponent.square(32, 'test.png')).forEach((sprite) {
+        sprite.x = r.nextInt(size.width.toInt()).toDouble();
+        sprite.y = r.nextInt(size.height.toInt()).toDouble();
+        add(sprite);
+      });
+    } catch (e) {
+      debugPrint(e.toString());
+    }
```
luanpotter, line 37: is there any specific exception you were looking for here? overall I'd rather not have the try catch and let it blow (if anything ever happens). this is just an example app, but always good to fail fast and fix. also makes the code a tad bit cleaner

luanpotter:
Just commenting on why we have this weird example structure. It's because of how pub/pana linters except the project to be. They expect that there is a folder called example on root with a single flutter project. and we want to have multiple examples to showcase different features. I tried leaving then inside the example root folder, creating a examples folder on root, this all upsets pub because they want a single example folder with a single project. so I had to move it to the docs. I do like having a "main" example witch is very bare bones and shows how flame works, and then having examples specific to each feature so like, I wanna add text, go to the text example, etc. but I do think the folder structure is terrible, it was just the best I could create that was compatible with pub. Please feel free to suggest changes as long as they are compatible with (the current) pub (possibly it could've changed ever since and is more lenient more).

doc/examples/sprites/lib/main.dart
erf: i think i had some initial problem loading assets in the example when it was in the Example folder, but i'll remove it

erf:
Pushed the change. Maybe add a link to the `examples` on the front README?

erf:
Maybe we should add some spesific performance examples .. not sure if it possbile to have benchmarks test for graphics rendering, but we should run the profiler and watch the FPS and memory usage for various scenarious

erickzanardo:
> Pushed the change. Maybe add a link to the `examples` on the front README?

That is a good idea IMO

Issue #282, opened 2020-04-15
Sprite batch


erf:
As per my discussion with @erickzanardo in #268, i would like to add a new `SpriteBatch` class, which will use the [Canvas.drawAtlas](https://api.flutter.dev/flutter/dart-ui/Canvas/drawAtlas.html) call to render a a set of sprites with transforms. This will have performance benefits when drawing sprite atlases, and will not break existing functionality. Next we could create a component for this, and look into refactoring `SpriteSheet` and perhaps `Animation` to use this.

erf:
@erickzanardo i've added a PR for this now in #283 

There is is improvements to be made, but should be a start. Let me know what you think.

erickzanardo:
Awesome, will look when I get some time

erf:
Thank you for the merge!

I think next would be interesting to implement `SpriteSheet`and/or `Animation` using this, and add to `SpriteBatch` as needed.

PR #283, opened 2020-04-15
Add SpriteBatch API

erf:
This is an PR for #282.

- added a `SpriteBatch` class which buffers rects, transforms and optional colors, and renders using `Canvas.drawAtlas`
- added a `SpriteBatchComponent` which accepts a `SpriteBatch`
- added an Example which render a bunch of sprites in various ways ( using `SpriteBatch` )

EDIT: 
- added new doc entry `SpriteBatch` under `images.md`
- updated CHANGELOG


erf:
Ok, i will look into those changes. 

Here is a screenshot:
Simulator Screen Shot - iPhone 11 Pro - 2020-04-16 at 14 32 01

erf:
I added a doc entry and updated changelog.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [next]
 
+## 0.20.0
```
erickzanardo, line 5: Leave this on the `[next]` entry, which version will be is defined when we release the version.

doc/images.md
```diff
@@ -45,6 +45,18 @@ Sprites can also be used as widgets, to do so, just use `Flame.util.spriteAsWidg
...
+
+If you have an image atlas ( an image with smaller images inside ), and would like to render it effectively - `SpriteBatch` does the job.
+
+Give it the filename of the image, and then add rectangles which describes various part of the image, in addition to transforms ( position, scale and rotation ) and optional colors.
+
+You render it with an `Canvas` and an optional `Paint`, `BlendMode` and `CullRect`.
+
+An `SpriteBatchComponent` is also available for your convenience.
+
+See example [here](examples/sprite_batch).
```
erickzanardo, line 58: Does this link work? I think it should be `docs/example/sprite_batch` doesn't it?

lib/sprite_batch.dart
```diff
@@ -0,0 +1,82 @@
...
+  static final defaultTransform = RSTransform(1, 0, 0, 0);
+  static const defaultColor = const Color(0x00000000); // transparent
+
+  SpriteBatch(String fileName) {
+    Flame.images.load(fileName).then((image) => atlas = image);
+  }
+
+  SpriteBatch.fromImage(this.atlas);
+
+  int get width => atlas.width;
```
erickzanardo, line 25: Just to make sure, I would verify for null here

Example
```dart
  int get width => atlas?.width;
```

Same thing for `height` and `size`

doc/images.md
erf: I think you are right. I looked at `examples/animation_widget` which is also wrong then? Will update that also.

lib/sprite_batch.dart
erf: do `atlas?.height ?? 0` or `atlas?.height` and return `null`

for size do:
```
  Size get size => atlas == null ? null : Size(width.toDouble(), height.toDouble());
```
or return size with 0 values ?

btw: i think i would prefer to pass an image to the constructor, which is not null, so we dont have to do async load stuff and null checks there. assets could be loaded outside and prepared beforehand

erf:
Should i amend and force push changes as i fix them here?

erf:
Rebased and force pushed changes with fixes.

doc/examples/sprite_batch/lib/main.dart
```diff
@@ -0,0 +1,65 @@
...
+      final y = r.nextInt(size.height ~/ 2).toDouble() + size.height / 2.0;
+      spriteBatch.add(
+        rect: Rect.fromLTWH(sx, sy, 128, 128),
+        offset: Offset(x - 64, y - 64),
+      );
+    }
+
+    add(SpriteBatchComponent.fromSpriteBatch(
+      spriteBatch,
+      //paint: Paint()..invertColors = true,
```
luanpotter, line 61: please either uncomment or remove

doc/images.md
```diff
@@ -43,7 +43,19 @@ All render methods from the Sprite class can receive a `Paint` instance on the o
 
 Sprites can also be used as widgets, to do so, just use `Flame.util.spriteAsWidget`
 
-A complete example of using sprite as widgets can be found [here](examples/animation_widget).
+A complete example of using sprite as widgets can be found [here](/doc/examples/animation_widget).
+
+## SpriteBatch
+
+If you have an image atlas ( an image with smaller images inside ), and would like to render it effectively - `SpriteBatch` does the job.
```
luanpotter, line 50: 1. please don't add spaces inside parenthesis (like this).
2. we pretty much call "sprite atlas" spritesheets everywhere else. AFAICS this is the same thing just being rendered more efficiently. so maybe I'd change the text to be
> If you have a sprite sheet (also called an image atlas, which is ...), ...
just to make it more clear that spritesheet and atlas from the flame perspective are the same thing, this is just a new way to render it (correct me if I'm wrong)

doc/images.md
```diff
@@ -43,7 +43,19 @@ All render methods from the Sprite class can receive a `Paint` instance on the o
...
+
+Give it the filename of the image, and then add rectangles which describes various part of the image, in addition to transforms ( position, scale and rotation ) and optional colors.
```
luanpotter, line 52: (spaces in parenthesis)

lib/components/sprite_batch_component.dart
```diff
@@ -0,0 +1,36 @@
...
+      canvas,
+      blendMode: blendMode,
+      cullRect: cullRect,
+      paint: paint,
+    );
+  }
+
+  @override
+  bool loaded() {
+    return spriteBatch != null && spriteBatch.atlas != null;
```
luanpotter, line 31: you can do `spriteBatch?.atlas != null`. and then you can inline this function like this
```
bool loaded() => spriteBatch?.atlas != null;
```

doc/images.md
erf: Yeah, they are essentially doing the same thing, only SpriteSheet now has a bit more functionality. Should replace "SpriteSheet" with "SpriteBatch" later ( and perhaps rename SpriteBatch to SpriteSheet again )

lib/components/sprite_batch_component.dart
erf: could `loaded` be a `get` property here?

erf:
pushed changes

erf:
I added a new commit where i changed the constructor, given the image asset, and created an async factory method which creates and return a new `SpriteBatch` with a preloaded asses, as i commented in https://github.com/flame-engine/flame/issues/289#issuecomment-615374515 
In this way we always guarantee that the atlas image is loaded and we dont have to do null checks.

erf:
Btw i wanted to pass optional named parameters for rects, transforms and colors to the spritebatch, but they become null, even if i declared them as empty lists in the class. Not sure why this is. I added a question on stackoverflow here:
https://stackoverflow.com/questions/61303278/optional-constructor-parameter-set-the-list-to-null-in-dart/61303489?noredirect=1#comment108449447_61303489

but i think it might be better to use named constructors at a later point, in order to create sprite sheets etc.

lib/sprite_batch.dart
erickzanardo: I think we should just return null, not 0 as default, just to avoid a NPE inside the method.

erf: In the latest commit you dont have to check for null

lib/components/sprite_batch_component.dart
```diff
@@ -0,0 +1,34 @@
...
+    spriteBatch.render(
+      canvas,
+      blendMode: blendMode,
+      cullRect: cullRect,
+      paint: paint,
+    );
+  }
+
+  @override
+  bool loaded() => true;
```
luanpotter, line 30: this returns true by default, no need to override it

erf:
Ok, should be good to go.

renancaraujo:
This is amazing. Good to go!

PR #287, opened 2020-04-16
Added game loop abstraction for GameRenderBox

erf:
I added a `GameLoop` class which abstracts the game loop from `GameRenderBox`, to clean things up a bit.

Functionwise it should be the same. I do now check for `running` instead of `attached` in `GameLoop._tick`, but i think it should work the same.

A callback is passed for updates.


erf:
It was mostly for cleaning up, but if you need to use it somewhere else, it's nice that its a separate class.

PR #297, opened 2020-04-23
I added support for rotated tiles in Tiled Component

Arne1303:
Tiles now get drawn like their supposed to.
This does only work together with my PR on flame-engine/tiled.dart which adds the Backbone for this.

(also these are my first PR so tell me if I do something wrong)

### DIFF
--- a/lib/components/tiled_component.dart
+++ b/lib/components/tiled_component.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:ui';
+import 'dart:math' as math;
 
 import 'package:flame/components/component.dart';
 import 'package:flame/flame.dart';
@@ -60,7 +61,8 @@ class TiledComponent extends Component {
     });
   }
 
-  void _renderLayer(Canvas c, Layer layer) {
+
+  void _renderLayer(Canvas c, Layer layer){
     layer.tiles.forEach((tile) {
       if (tile.gid == 0) {
         return;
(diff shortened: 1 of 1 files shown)


lib/components/tiled_component.dart
```diff
@@ -60,7 +61,8 @@ class TiledComponent extends Component {
     });
   }
 
-  void _renderLayer(Canvas c, Layer layer) {
+
```
erickzanardo, line 64: No need for this blank line

lib/components/tiled_component.dart
```diff
@@ -60,7 +61,8 @@ class TiledComponent extends Component {
...
+  void _renderLayer(Canvas c, Layer layer){
```
erickzanardo, line 65: keep the space beetwen the `)` and `{`

lib/components/tiled_component.dart
```diff
@@ -74,10 +76,84 @@ class TiledComponent extends Component {
       final dst = Rect.fromLTWH(tile.x.toDouble(), tile.y.toDouble(),
           rect.width.toDouble(), rect.height.toDouble());
 
-      c.drawImageRect(image, src, dst, paint);
+      //Rotating & Drawing our Tiles
+      switch(tile.rotation){
```
erickzanardo, line 80: Leave a space between the `)` and `{`

lib/components/tiled_component.dart
```diff
@@ -74,10 +76,84 @@ class TiledComponent extends Component {
...
+        case 0:
+          //0 ° Rotation
```
erickzanardo, line 82: You already handle this on the `default`, no need to handle this case.

lib/components/tiled_component.dart
```diff
@@ -74,10 +76,84 @@ class TiledComponent extends Component {
...
+          c.drawImageRect(image, src, dst, paint);
+          break;
+        case 1:
+          //90 ° Rotation
+          c.save();
+          c.translate( dst.center.dx, dst.center.dy );
+          c.rotate((3 * math.pi/2));
+          c.translate( -dst.center.dx, -dst.center.dy );
+          c.drawImageRect(image, src, dst, paint);
+          c.restore();
```
erickzanardo, line 92: These lines are repeated on all the cases, the only thing that changes is the radians passed to the rotate method, you should encapsulate this in a method receiving the radian as a parameter, so we don't have duplicated code.

PR #302, opened 2020-04-27
Simplify GameLoop using a Ticker

erf:
Simplify the `GameLoop` class using a [Ticker](https://api.flutter.dev/flutter/scheduler/Ticker-class.html).

I believe it's essentially doing the same as before as it is an abstraction on `SchedulerBinding.scheduleFrameCallback`.

But please do some testing..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [next]
 - Adding SpriteBatch API, which renders sprites effectively using Canvas.drawAtlas
 - Refactor game.dart classes into separate files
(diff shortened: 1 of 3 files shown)


luanpotter:
This is a major infrastructural change. Have you tested this well and how it performs compared to the old method?

erf:
I've only tested it briefly, but performance seem on par, and pausing, seem to work as before.

luanpotter:
I tested this on BGUG and it seems to be working well! I think we can merge it and move forward with it :)

PR #304, opened 2020-04-30
An easier way to handle contacts with Box2DGame

spydon:
Currently it is quite tedious and creates a lot of confusion of how to handle when two objects in Box2D come in contact. With this PR you can register ContactCallbacks which contains which objects that should come in contact and what should happen when they do and Box2DGame handles the rest.


doc/examples/box2d/contact_callbacks/lib/boundaries.dart
```diff
@@ -0,0 +1,57 @@
...
+    Offset start = coordinates[0];
+    Offset end = coordinates[1];
+    canvas.drawLine(start, end, paint);
+  }
+
+  void _createBody(Vector2 start, Vector2 end) {
+    final PolygonShape shape = PolygonShape();
+    shape.setAsEdge(start, end);
+
+    final fixtureDef = FixtureDef();
```
erickzanardo, line 45: Suggestion: Use the cascade notation here:

```dart
    final fixtureDef = FixtureDef()
       ..setUserDate(this);
...
```

doc/examples/box2d/contact_callbacks/lib/boundaries.dart
```diff
@@ -0,0 +1,57 @@
...
+    fixtureDef.setUserData(this); // To be able to determine object in collision
+    fixtureDef.shape = shape;
+    fixtureDef.restitution = 0.0;
+    fixtureDef.friction = 0.1;
+
+    final bodyDef = BodyDef();
```
erickzanardo, line 51: Suggestion: Use the cascade notation here

doc/examples/box2d/contact_callbacks/lib/main.dart
```diff
@@ -0,0 +1,123 @@
...
+    return PaletteEntry(Color.fromARGB(rng.nextInt(255), rng.nextInt(255),
+            rng.nextInt(255), rng.nextInt(255)))
+        .paint;
+  }
+
+  void _createBody(double radius, Vector2 position) {
+    final CircleShape shape = CircleShape();
+    shape.radius = radius;
+
+    final fixtureDef = FixtureDef();
```
erickzanardo, line 38: Suggestion: Use the cascade notation here

doc/examples/box2d/contact_callbacks/lib/main.dart
```diff
@@ -0,0 +1,123 @@
...
+    // To be able to determine object in collision
+    fixtureDef.setUserData(this);
+    fixtureDef.shape = shape;
+    fixtureDef.restitution = 1.0;
+    fixtureDef.density = 1.0;
+    fixtureDef.friction = 0.1;
+
+    final bodyDef = BodyDef();
```
erickzanardo, line 46: Suggestion: Use the cascade notation here

spydon:
Fixed that cascade style comments. Just going to add some docs.

spydon:
Added some docs and refactored how the ContactCallback mixin works.

lib/box2d/contact_callbacks.dart
```diff
@@ -0,0 +1,72 @@
...
+  void _maybeCallback(Contact contact, ContactCallback callback, Function f) {
+    final Object a = contact.fixtureA.userData;
+    final Object b = contact.fixtureB.userData;
+    final ContactTypes current = ContactTypes(a.runtimeType, b.runtimeType);
+    final ContactTypes wanted = callback.types;
+
+    if (current.equals(wanted)) {
+      _inOrder(callback, a, b) ? f(a, b) : f(b, a);
+    } else if (wanted.has(AnyObject) && current.hasOneIn(wanted)) {
+      !_inOrder(callback, a, b) ? f(a, b) : f(b, a);
```
luanpotter, line 52: is there a value providing the `contact` object as the last parameter? that way implementers of ContactCallback can access more details if desired, like the contact region and such (might be useful for some collision handling). like maybe I wanna know how "hard" they hit or something. I don't know a lot about box2d or what the contact class provides other than the bodies themselves

lib/box2d/contact_callbacks.dart
```diff
@@ -0,0 +1,72 @@
...
+
+  @override
+  void postSolve(Contact contact, ContactImpulse impulse) {}
+
+  @override
+  void preSolve(Contact contact, Manifold oldManifold) {}
+}
+
+// Used to handle collision with bodies of any other type
+class AnyObject {}
```
luanpotter, line 72: can't we use BodyComponent for this? all objects inherit from it right?

lib/box2d/contact_callbacks.dart
```diff
@@ -0,0 +1,72 @@
...
+
+  @override
+  void endContact(Contact contact) =>
+      _callbacks.forEach((c) => _maybeCallback(contact, c, c.end));
+
+  @override
+  void postSolve(Contact contact, ContactImpulse impulse) {}
+
+  @override
+  void preSolve(Contact contact, Manifold oldManifold) {}
```
luanpotter, line 68: we might want to similarly expose these functions (but that can be a followup, no need to bundle it all together). they can be empty implementations on the ContactCallback so that you don't need to implement, but you can if you want. like optional callbacks.

lib/box2d/contact_callbacks.dart
spydon: Very true, I'll add that!

spydon:
> I really like this! awesome contribution. this makes stuff much more easy to use. I have some comments for some suggestions (let me know if it makes sense as I'm not 100% familiar with box2d)

Thanks! Very good comments, updated accordingly (will add pre/postsolve in a later PR I think).
I just realized I have to fix one more thing before it is merged and that is to check inheritance properly.

lib/box2d/contact_callbacks.dart
spydon: Added these two, optionally like you suggested too, because I don't think they are used in the common usecase.

lib/box2d/contact_callbacks.dart
```diff
@@ -1,26 +1,27 @@
...
-  // If it contains at least one type from the other ContactTypes
-  bool hasOneIn(ContactTypes other) => has(other.type1) || has(other.type2);
-
-  bool equals(ContactTypes other) =>
-      (type1 == other.type1 && type2 == other.type2) ||
-          (type2 == other.type1 && type1 == other.type2);
+class ContactTypes<T1, T2> {
+  // If o1 is, or inherits from, T1 or T2
+  bool has(Object o1) => o1 is T1 || o1 is T2;
+  bool hasOne(Object o1, Object o2) => has(o1) || has(o2);
```
erickzanardo, line 7: I found the `has` confusing here, what about: `is` instead of `has`?

spydon: `Is` would be confusing since that lingo is already used for types in Dart, and since this checks two objects and not only one. Would `contains` make it less confusing maybe?

erickzanardo: Idk, contains seems weird to me as well hahahahaha, I think it is ok to leave a `has` then, it was a minor suggestion in the end.

spydon: It's only used internally anyways. :)

spydon:
I think #311 should be merged first and then I'll rebase this branch.

spydon:
Now everything is done on this branch, I just rebased on develop.

doc/box2d.md
```diff
@@ -4,13 +4,59 @@ Although Flame does not implements Box2d itself, it bundles a forked port of the
...
+
+Now you have to make an implementation of `ContactCallback` where you set which two types that it should react when they come in contact.
+If you have two `BodyComponent`s named Ball and Wall and you want to do something when they come in contact you would do like this:
+
+```dart
+class BallWallCallback implements ContactCallback<Ball, Wall> {
+  BallWallCallback();
+
+  @override
+  void begin(Ball ball, Wall wall) {
```
luanpotter, line 37: you need to add the new param Contact contact here :)

spydon: Good catch! Fixed now :)

PR #310, opened 2020-05-03
Adding Images#fromBase64

erickzanardo:
Adding a way to load images from base64 string images.


doc/examples/sprites/lib/generated_plugin_registrant.dart
```diff
@@ -0,0 +1,16 @@
+//
+// Generated file. Do not edit.
```
spydon, line 2: What is this file?

luanpotter: I think this is a new file that all flutter projects have now but these old examples built with older flutter didn't. not sure if we should version or not but it doesn't say on the file it shouldn't be. I would look at the other newer examples to see what we do. maybe the new flutter create adds it to the gitignore.

erickzanardo: I just checked, it is on git ignore, I am going to remove it

PR #311, opened 2020-05-04
Fix render priority

### LINKED ISSUE #306: Render priority doesn't seems to be respected
I have tried to set the priority to first render the HUD and then the world components, but it seems that no matter what the world is always rendered after the HUD. Here's the link for the project: [repository](https://github.com/rezendegc/vertical-scape-game/tree/bugs).

I don't think that there's a small piece of code to share but i've tried putting the priority as 0 on the world (box2dcomponent), and 100 on the score (lib->components->score), but still the winzone (lib->components->winzone) gets in front of the score...

spydon:
Fixes #306 and does it how it should have been made from the beginning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 ## [next]
-- Fix Box2DGame to follow render prio
+- Fix Box2DGame to follow render priority
 
 ## 0.20.0
 - Refactor game.dart classes into separate files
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## [next]
+- Fix Box2DGame to follow render prio
```
erickzanardo, line 4: shouldn't be `priority` here?

spydon: True, since that is what it is called in the code. Fixed.

luanpotter:
@spydon will add this so you can continue working on the other PR

PR #315, opened 2020-05-07
Fix the text-component position is changed by anchor

sam3ha2:
Component location is determined by 'prepareCanvas()'
In v0.20.0, it is moved additional as much as '_translatedOffset'.
When the component is created(constructor), the anchor is 'topleft' and does not affect '_translatedOffset'. However, when text or config are changed, '_updateBox()' is called, and '_translatedOffset' is affected by anchor.
There is no problem at first, but the problem occurs when the anchor is set and '_translatedOffset' is changed.


spydon:
By the way, could you add a CHANGELOG entry too?

erickzanardo:
Please rebase your PR to the develop branch.

@luanpotter can you take a look on this please?

lib/components/text_component.dart
```diff
@@ -38,19 +36,11 @@ class TextComponent extends PositionComponent {
     _tp = config.toTextPainter(_text);
     width = _tp.width;
     height = _tp.height;
-
-    _updateOffset(Position.empty());
-  }
-
-  void _updateOffset(Position p) {
```
luanpotter, line 45: so basically you are saying this is doing twice the translation because PositionComponent is already translating when we set width/height rignt? let me just test this on the example app

luanpotter: gotcha, that makes sense!

PR #316, opened 2020-05-07
GitHub's proposed Code of Conduct

spydon:
This is GitHub's proposed Code of Conduct, I think it sounded reasonable.
(And it brings us up one level on the community standard)
What do you think?


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at contact@flame-engine.org. All
```
erickzanardo, line 58: We may need to change this email, as it doesn't exists.

spydon: We could just write that they should contact some admin on the discord maybe?

erickzanardo: That could work, maybe instead of admins, we could say `Flame Staff`, and then we could create that role on the discord.

spydon: Good idea, updated accordingly.

PR #337, opened 2020-05-15
Splitting record fps mode from debug mode

erickzanardo:
# Description

The mode to count the fps is locked to the mode on Flame, but they are different things, on debug mode, all the position components renders is bounding box, with its coordinates and dimensions on the screen, which really can affect FPS.

This PR proposes a split on these two modes.

## Type of change

# Checklist:

I am gonna update the examples and docs after this proposal have been accepted


lib/game/base_game.dart
```diff
@@ -153,22 +153,27 @@ class BaseGame extends Game with TapDetector {
...
+  /// Returns `false` by default. Override to use the `fps` counter method.
+  /// In recording fps, the [recordDt] method actually records every `dt` for statistics.
+  /// Then, you can use the [fps] method to check the game FPS.
+  bool recordFps() => false;
+
   /// This is a hook that comes from the RenderBox to allow recording of render times and statistics.
   @override
   void recordDt(double dt) {
-    if (debugMode()) {
+    if (recordFps()) {
```
luanpotter, line 169: since we are changing this, I'm having some thoughts:

1. should this be a property, for performance reasons? not sure how expensive function invocation is but this is run every frame regardless of the result
2. does debug=true and record=false make sense? if not, this could be an enum (REGULAR, FPS, DEBUG). not sure if it's better or not, just a thought.

erickzanardo: Normally I would agree with you, but I am thinking that record fps is not an subset of a debug mode, so I wouldn't make it be a parameter. And I don't know if we would have much more modes than these two. One other reason to keep like this, is that this will be less a breaking change than would be to have an enum.

luanpotter: ok, let's keep as is and rethink on v 1.0 (specially my first bullet)

Issue #336, opened 2020-05-15
Effects improvements


spydon:
Things we could add to Effects:
- [x] Rotational effect
- [x] Sequenced effects (@erickzanardo's idea)
- [ ] Effect that can move along path, possibly just an extra parameter to MoveEffect
- [x] CombinedEffect, combine two effects into one, with potential offset in time to start
- [ ] ShakeEffect, an effect that inserts some noise to shake the component
- [ ] Elasticity
- [ ] Have an argument which decides between relative and absolute
- [x] onComplete callback
- [x] ~~Anchor for effects, right now the anchor is topLeft but it would be nice to be able to choose any anchor.~~

erickzanardo:
I like the idea of both rotation and sequence effects.

But the Effect don't need to have an anchor, the component already has a property for that, and the effects are working fin with it.

spydon:
Good point!

spydon:
@erickzanardo what do you think about a CombinedEffect that combines two (or multiple maybe) effects with an optional offset in time between them?

erickzanardo:
> @erickzanardo what do you think about a CombinedEffect that combines two (or multiple maybe) effects with an optional offset in time between them?

This seems like a good idea, but I don't know how easy would it be to implement this, imagine a MoveEffect together a ScaleEffect, wouldn't the scaling mess with the repositioning? 

Anyway, I think it is nice, I just worry that this would be a complex take.

spydon:
It would depend on the anchor I think, on a centered anchor it shouldn't look too weird.
Not sure how bad it would be on other anchors, I'll do some testing.

erickzanardo:
> It would depend on the anchor I think, on a centered anchor it shouldn't look too weird.
> Not sure how bad it would be on other anchors, I'll do some testing.

Actually, since we rely only on the curve progress, maybe we will not have any trouble... Anyhow, it would be nice to do some testing, let me know how it turned out.

spydon:
It worked fine. :)
I think it would also make sense to have an optional anchor on the effects which can override the component anchor.

spydon:
CombinedEffect is almost done so I think that one will be aimed for 0.23.0 along with a fix for a drifting bug that I found.

For 1.0.0 I will propose a general update to the effects where you can choose if you want them to be relative or absolute.

[...]

PR #329, opened 2020-05-13
Infinite and alternating effects

spydon:
# Description

Create infinite and alternating effects.

Fixes #324 

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# Checklist:

If something is unclear, please submit the PR anyways and ask about what you thought was unclear.

- [x] This branch is based on `develop`
- [x] This PR is targeted to merge into `develop` (not `master`)
- [x] I have formatted my code with `flutter format`
- [x] I have made corresponding changes to the documentation
- [x] I have added examples for new features in `doc/examples`
- [x] The continuous integration (CI) is passing

PR #338, opened 2020-05-17
Rotational effect

spydon:
# Description

Adds rotational effect as mentioned in #336.
It is based on the infinite effects branch (#329), so that should be merged first.

## Type of change

# Checklist:

### DIFF
--- a/doc/effects.md
+++ b/doc/effects.md
@@ -46,6 +46,24 @@ square.addEffect(ScaleEffect(
 ));
 ```
 
+## RotationalEffect
+
+Applied to `PositionComponent`s, this effect can be used to rotate the component, using an [animation curve](https://api.flutter.dev/flutter/animation/Curves-class.html).
+
+Usage example:
+```dart
+import 'dart:math';
+
+import 'package:flame/effects/effects.dart';
+
+// Square is a PositionComponent
+square.addEffect(RotationalEffect(
+  rotation: 2 * pi, // In radians
+  speed: 1.0,
+  curve: Curves.easeInOut,
+));
+```
+ 
 # Examples
 
 Full examples can be found [here](/doc/examples/effects/simple) and [here](/doc/examples/effects/infinite_effects).
--- a/doc/effects.md
+++ b/doc/effects.md
@@ -46,9 +46,11 @@ square.addEffect(ScaleEffect(
 ));
 ```
 
-## RotationalEffect
+## RotateEffect
 
 Applied to `PositionComponent`s, this effect can be used to rotate the component, using an [animation curve](https://api.flutter.dev/flutter/animation/Curves-class.html).
+
+The angle (`radians`) is in radians and the speed is in radians per second, so if you for example want to turn 180° in 2 seconds you set `radians: pi` and `speed: 0.25`.
 
 Usage example:
 ```dart
@@ -57,9 +59,9 @@ import 'dart:math';
 import 'package:flame/effects/effects.dart';
 
 // Square is a PositionComponent
-square.addEffect(RotationalEffect(
-  rotation: 2 * pi, // In radians
-  speed: 1.0,
+square.addEffect(RotateEffect(
+  radians: 2 * pi, // In radians
+  speed: 1.0, // Radians per second
   curve: Curves.easeInOut,
 ));
 ```
--- a/doc/examples/effects/infinite_effects/lib/square.dart
+++ b/doc/examples/effects/infinite_effects/lib/square.dart
@@ -6,16 +6,20 @@ import 'dart:ui';
 class Square extends PositionComponent {
   final Paint _paint;
 
-  Square(this._paint, double x, double y) {
+  Square(this._paint, double x, double y, {double angle = 0.0}) {
     width = 100;
     height = 100;
     this.x = x;
     this.y = y;
+    this.angle = angle;
     anchor = Anchor.center;
   }
 
   @override
   void render(Canvas canvas) {
+    canvas.translate(x, y);
+    canvas.rotate(angle);
+    canvas.translate(-x, -y);
     canvas.drawRect(toRect(), _paint);
   }
 }
--- a/lib/effects/rotational_effect.dart
+++ b/lib/effects/rotational_effect.dart
@@ -1,0 +1,38 @@
+import 'package:flutter/animation.dart';
+import 'package:meta/meta.dart';
+
+import './effects.dart';
+
+class RotationalEffect extends PositionComponentEffect {
+  double rotation;
+  double speed;
+  Curve curve;
+
+  double _originalAngle;
+  double _peakAngle;
+  double _direction;
+
+  RotationalEffect({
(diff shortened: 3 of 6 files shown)


erickzanardo:
Is this PR based on your branch adding the alternating and infinite effects?

spydon:
> Is this PR based on your branch adding the alternating and infinite effects?

It is, like it says in the PR description. ;)

erickzanardo:
> > Is this PR based on your branch adding the alternating and infinite effects?
> 
> It is, like it says in the PR description. ;)

Shame on me for not reading the description...

spydon:
Now this is rebased on develop since the infinite-effects branch is in.

doc/effects.md
```diff
@@ -46,6 +46,24 @@ square.addEffect(ScaleEffect(
...
+
+Usage example:
+```dart
+import 'dart:math';
+
+import 'package:flame/effects/effects.dart';
+
+// Square is a PositionComponent
+square.addEffect(RotationalEffect(
+  rotation: 2 * pi, // In radians
```
erickzanardo, line 61: Just a thought, I always find it easier to think in angle instead of radians, I would leave for the effect to converte the angle to radians, what do you think?

doc/effects.md
```diff
@@ -46,6 +46,24 @@ square.addEffect(ScaleEffect(
 ));
 ```
 
+## RotationalEffect
```
erickzanardo, line 49: The other effects are using the verb instead of the noun, I think we should name it RotateEffect, to keep on the same pattern, what do you think?

doc/examples/effects/infinite_effects/lib/square.dart
```diff
@@ -6,16 +6,20 @@ import 'dart:ui';
...
     height = 100;
     this.x = x;
     this.y = y;
+    this.angle = angle;
     anchor = Anchor.center;
   }
 
   @override
   void render(Canvas canvas) {
+    canvas.translate(x, y);
```
erickzanardo, line 20: This should not be needed, `PositionComponent` already has an angle that is already makes the component rotate.

lib/effects/rotational_effect.dart
```diff
@@ -0,0 +1,38 @@
...
+  double rotation;
+  double speed;
+  Curve curve;
+
+  double _originalAngle;
+  double _peakAngle;
+  double _direction;
+
+  RotationalEffect({
+    @required this.rotation,
```
erickzanardo, line 16: Suggestion: maybe call this angle?

doc/effects.md
luanpotter: I actually find it much easier to use radians haha. I think we should first and foremost stick to conventions, angles are used in a lot of places already in flame and I don't recall now which unit is used (I probably copied flutter's default). So I think we should use that one, whichever it is :)

doc/effects.md
```diff
@@ -46,6 +46,24 @@ square.addEffect(ScaleEffect(
...
+Usage example:
+```dart
+import 'dart:math';
+
+import 'package:flame/effects/effects.dart';
+
+// Square is a PositionComponent
+square.addEffect(RotationalEffect(
+  rotation: 2 * pi, // In radians
+  speed: 1.0,
```
luanpotter, line 62: this is rad / s  recond? would be nice to add the unit in the example as well

doc/examples/effects/infinite_effects/lib/square.dart
luanpotter: you need to call prepareCanvas to use that (and we should!)

in v1.0 it will be super.render :p (spoilers)

doc/effects.md
erickzanardo: PositionComponent#angle is in radians.

I think angles is way more human friendly, as I can easily visualize what 90° degres mean, when any radian value means nothing to me haha

but if we are keeping with the standards, which I think we should, lets keep in radians.

@spydon lets use this PR to add a comment on the angle property of PositionComponent making it clear that that property is in radians.

lib/effects/rotational_effect.dart
erickzanardo: Or even better, radians?

doc/examples/effects/infinite_effects/lib/square.dart
erickzanardo: Hmm, that is true, somehow I thought PositionComponent already did that.

I wonder if shouldn't we add a render method on PositionComponent, with this call and a @mustCallSuper already on the 0.22.0 version as we don't know when 1.0 will land, and this is not a very big breaking change.

Thoughts @luanpotter ?

doc/effects.md
spydon: I think it is much better to teach people radians if they haven't already been taught it in school. Angle has no real basis, meanwhile radians do.
It also feels like radians are the standard for most programming libraries, that I have used at least.
But you are correct Erick, it is very much a thing to get used to. :)

doc/effects.md
spydon: Fixed

lib/effects/rotational_effect.dart
spydon: Fixed

doc/effects.md
spydon: Fixed!

doc/examples/effects/infinite_effects/lib/square.dart
luanpotter: that is a VERY big breaking change! it will break a lot of games. that is why I packaged it onto v1.0.

erickzanardo: It will break ALL the games, but IMO, it is not common for people to use PositionComponent directly, and even people who uses it, will have to just add this line to fix: `super.render(canvas)`, so it is a very easy fix. That is why I consider this very problematic to add prior to v1, as it would be an easy fix, and probably not many people would be affected. But I am fine with not adding this prior to v1...

doc/examples/effects/infinite_effects/lib/square.dart
```diff
@@ -6,16 +6,20 @@ import 'dart:ui';
...
+    canvas.rotate(angle);
```
erickzanardo, line 21: @spydon lets use `prepareCanvas` here them, instead of handle rotation manually.

spydon: Oh I forgot to commit that, fixing when I get in from the park!

spydon: Fixed

spydon:
The example has to be updated after #349 is merged.

erickzanardo:
> The example has to be updated after #349 is merged.

@spydon #349 has been merged, wanna update the example and land this?

spydon:
Will fix tomorrow :)

spydon:
Now the example is updated according the new BaseGame structure.

PR #355, opened 2020-05-25
Added a virtual joystick component

yahel-ck:
# Description

I added a highly customizable, minimalistic (I think), virtual joystick component. 
I originaly created it for my own game but thought it could be usefull for other people too.
It's heavily documented and intuitive (imo) and I also added an example project in the doc folder.

This basically adds one file to the source code with one class in it (Joystick).

## Type of change

# Checklist:

- [ ] I have added an entry under `[next]` in `CHANGELOG.md` // I'm not sure how
- [ ] The continuous integration (CI) is passing // I'm not sure what this is :/ but I think it is passing


doc/examples/joystick/README.md
```diff
@@ -0,0 +1,3 @@
+# nine_tile_box
```
spydon, line 1: Readme for the wrong example

doc/examples/joystick/lib/static-joystick.dart
```diff
@@ -0,0 +1,67 @@
...
+
+import 'package:flame/game.dart';
+import 'package:flame/gestures.dart';
+import 'package:flame/joystick.dart';
+import 'package:flame/sprite.dart';
+import 'package:flutter/cupertino.dart';
+
+import 'player.dart';
+
+class StaticJoystick extends Joystick {
```
spydon, line 12: The filename should have an underscore, not a dash.

doc/examples/joystick/pubspec.yaml
```diff
@@ -0,0 +1,22 @@
...
+    path: ../../../
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+
+flutter:
+  uses-material-design: true
+  assets:
+    - assets/images/minotaur.png # thanks https://elthen.itch.io/2d-pixel-art-minotaur-sprites
```
spydon, line 22: How does that work, you don't seem to define the joystick assets here?

doc/examples/joystick/lib/static-joystick.dart
yahel-ck: Changed it, sorry.

doc/examples/joystick/README.md
yahel-ck: I commited and pushed again. Still new to Github, sorry :/

doc/examples/joystick/pubspec.yaml
yahel-ck: Same issue like the README file. It's fixed now I think.

spydon:
If you look in the root directory of flame you will find a file called CHANGELOG.md and in the top it will look something like this:
```
[next]
    Fixing BaseGame tap detectors issues
    Upgrade Flutter SVG to fix for flame web
    Add linting to all the examples
```
Just add a line there explaining your change, like "Add virtual joystick component".

doc/examples/joystick/README.md
spydon: No problem. :)

spydon:
`joystick.dart` should be called `joystick_component.dart` and like in the `lib/component` directory too.

doc/examples/joystick/lib/dynamic_joystick.dart
```diff
@@ -0,0 +1,69 @@
...
+  final joystick = DynamicJoystick();
+
+  DynamicJoystickGame() {
+    add(player);
+    add(joystick);
+  }
+
+  void _startInput(Offset pos) => joystick.showAt(pos);
+
+  void _updateInput(Offset pos) {
```
erickzanardo, line 43: I think we should make this transparent to the developer, as it can be kind boilerplate to use.

I think we can probably use a similar solution used on Tapables components, where the listener of that component was direct hooked on the widgets builder.

@luanpotter @spydon @renancaraujo your thoughts?

doc/examples/joystick/lib/static_joystick.dart
```diff
@@ -0,0 +1,67 @@
...
+import 'package:flame/gestures.dart';
+import 'package:flame/joystick.dart';
+import 'package:flame/sprite.dart';
+import 'package:flutter/cupertino.dart';
+
+import 'player.dart';
+
+class StaticJoystick extends Joystick {
+  StaticJoystick() {
+    visible = true;
```
erickzanardo, line 14: I think all these properties should be received by the Joystick constructor as named parameters, as it this is a great way to for people to understand your API

lib/joystick.dart
```diff
@@ -0,0 +1,125 @@
...
+
+class Joystick extends Component {
+  double _movementRadius = 200;
+  double _radiusSquare = 40000;
+  Offset _offset = Offset.zero;
+
+  /// The base part of the joystick. It's like the area in which the stick swivels.
+  ///
+  /// You shouldn't change it's position manually, but rather use the Joystick methods.
+  final SpriteComponent base = SpriteComponent()..anchor = Anchor.center;
```
erickzanardo, line 15: I am not sure by the way `SpriteComponent` is used here.

All components are supposed to be added to a base game class, and not be handled (update and render) manually.

Since you are handling it manually here, I think you should use the plain `Sprite` class, or add these `SpriteComponent` to the game, you just need to make sure that their destroy methods are tied to this component.

doc/examples/joystick/lib/dynamic_joystick.dart
yahel-ck: Just a clarification, the StaticJoystick and DynamicJoystick are just examples on how someone can use the Joystick class. The Joystick class itself is not abstract, but I think it's a good practice to extend it to setup everything and use the resize method to adjust it according to the screen.

The Joystick class doesn't handle input automatically, just provides usefull methods that can be fed input by the developer, like I did in this example.

erickzanardo: > The Joystick class doesn't handle input automatically, just provides usefull methods that can be fed input by the developer, like I did in this example.

indeed it can't handle, but my point is that I think that it should, as one of the main reasons of the components to exist on Flame, is to make it easy and transparent to use.

Of course that "auto handling" could be a thing to be added on the future and we can go like that atm, but still is something that I want to discuss about.

yahel-ck: Oh ok got it. It's just that there are a lot of ways (or more then one at least) that developers might want to configure the input, so I thought this way gives them more freedom.

erickzanardo:
Closing this in favor of #391 

Thanks for your contribution @yahel-ck, hopefully you will be able to contribute more in the future.

PR #357, opened 2020-05-25
Sequence effect

spydon:
# Description

This addition lets you sequence effects, either so that they run from start until end, or in a loop.

Related to #336

## Type of change

# Checklist:


doc/examples/effects/sequence_effect/lib/main.dart
```diff
@@ -0,0 +1,74 @@
...
+      speed: 2.0,
+      curve: Curves.decelerate,
+      isInfinite: false,
+      isAlternating: false,
+    );
+
+    final sequence = SequenceEffect(
+        effects: [move1, scale, move2, rotate],
+        isInfinite: true,
+        isAlternating: true);
```
luanpotter, line 71: trailing comma :)

lib/effects/rotate_effect.dart
```diff
@@ -26,7 +26,7 @@ class RotateEffect extends PositionComponentEffect {
     _originalAngle = component.angle;
     _peakAngle = _originalAngle + radians;
     _direction = _peakAngle.sign;
-    travelTime = (_peakAngle / speed).abs();
+    travelTime = (radians / speed).abs();
```
luanpotter, line 29: this was a bug before right? nice catch!

spydon: Hehe yeah, I was very confused why the times didn't add up for the sequence.

lib/effects/sequence_effect.dart
```diff
@@ -0,0 +1,85 @@
...
+    effects.forEach((effect) {
+      effect.reset();
+      _comp.setBySize(currentSize);
+      _comp.setByPosition(currentPosition);
+      effect.component = _comp;
+      if (effect is MoveEffect) {
+        currentPosition = effect.destination;
+      } else if (effect is ScaleEffect) {
+        currentSize = Position.fromSize(effect.size);
+      }
```
luanpotter, line 35: what if it's something else? should we throw an error in case we use an unsupported effect? I am also wondering if there is a way to make SequenceEffect generic so it supports any sort of PositionComponentEffect, even custom ones. eg can't we call `component` on the `currentEffect` instead of if/else casing all the possible instances?

lib/effects/sequence_effect.dart
```diff
@@ -0,0 +1,85 @@
...
+    });
+    travelTime = effects.fold(
+        0,
+        (time, effect) =>
+            time + effect.travelTime * (effect.isAlternating ? 2 : 1));
```
luanpotter, line 40: trailing comma :)

luanpotter: I think the base PositionComponentEffect would benefit from a totalTravelTime getter `= travelTime * (isAlternating ? 2 : 1)` with a nice comment distinguishing it from `travelTime`, what do you think?

lib/effects/sequence_effect.dart
```diff
@@ -0,0 +1,85 @@
...
+      currentEffect.isAlternating = _currentWasAlternating;
+      _currentIndex++;
+      final iterationSize = isAlternating ? effects.length * 2 : effects.length;
+      if (_currentIndex != 0 && _currentIndex % iterationSize == 0) {
+        isInfinite ? reset() : dispose();
+        return;
+      }
+      final orderedEffects =
+          curveDirection.isNegative ? effects.reversed.toList() : effects;
+      currentEffect = orderedEffects[_currentIndex % effects.length];
```
luanpotter, line 67: can currentEffect be a getter defined by `= orderedEffects[_currentIndex % effects.length]` instead of an additional property?

lib/effects/sequence_effect.dart
```diff
@@ -0,0 +1,85 @@
...
+        currentEffect.percentage = 1.0;
+      }
+    }
+  }
+
+  @override
+  void reset() {
+    super.reset();
+    _currentIndex = 0;
+    component = component;
```
luanpotter, line 83: this is a bit confusing, can we extract the relevant part of set component to a private method and call it on both places? Like `_initialSetup` or something...

spydon: Good idea

lib/effects/sequence_effect.dart
spydon: No, because when it goes backwards it would give the wrong effect.
_currentIndex could be refactored to point to the index for the reverse one too, but it becomes more complicated since it has to do the last effect twice.

lib/effects/sequence_effect.dart
spydon: That is a good idea. :)

lib/effects/sequence_effect.dart
spydon: It can support the rotate effect too which is not covered by that clause. The problem is that scale and move is absolute meanwhile the rotate is relative. That is probably a coming addition too, that all can be both relative and absolute. Not sure how to solve the casing though, open to suggestions.

doc/effects.md
```diff
@@ -65,6 +65,24 @@ square.addEffect(RotateEffect(
   curve: Curves.easeInOut,
 ));
 ```
+
+## SequenceEffect
+
+This effect is a combination of other effects. You provide it with a list of your predefined effects, that are not added to a component.
```
erickzanardo, line 71: The ending on this phrase is a little confusing for, what do you mean by not added to a component?

doc/examples/effects/sequence_effect/README.md
```diff
@@ -0,0 +1,3 @@
+# Infinite effects
+
+A Flame game showcasing how to create infinite alternating effects.
```
erickzanardo, line 3: Should be sequence effects right?

doc/examples/effects/sequence_effect/pubspec.yaml
```diff
@@ -0,0 +1,17 @@
+name: infinite_effects
```
erickzanardo, line 1: Sequence

lib/effects/rotate_effect.dart
erickzanardo: Probably could add an entry on the changelog about this fix, what do you think?

lib/effects/sequence_effect.dart
spydon: Trailing comma for the fold by the way? 
Isn't that very strange syntax since it will never have any more arguments?

lib/effects/sequence_effect.dart
```diff
@@ -0,0 +1,85 @@
...
+  int _currentIndex = 0;
+  PositionComponentEffect currentEffect;
+  bool _currentWasAlternating;
+  double _driftModifier = 0.0;
+
+  SequenceEffect({
+    @required this.effects,
+    isInfinite = false,
+    isAlternating = false,
+  }) : super(isInfinite, isAlternating);
```
erickzanardo, line 17: Add an assert here! Making sure that the effects are finite, what do you think?

spydon: That might be a good idea, there are probably not that many usecases where you want it to get stuck.

doc/effects.md
spydon: That they haven't called:
`myComponent.addEffect(effectThatIsPartOfSequence)`
Maybe I can clarify that.

erickzanardo: Maybe we can verify that on an assert as well? Not sure if is possible, just a thought

spydon: Tried to clarify it, is it more clear now? Or should I simply remove that sentence?

spydon: Added an assert for that too

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
  - Upgrade Flutter SVG to fix for flame web
  - Add linting to all the examples
  - Run linting only on affected and changed examples
+ - Add SequenceEffect
+ - Fixed but with travelTime in RotateEffect
```
erickzanardo, line 9: Is this written correctly? 🤔

spydon: Haha wtf have I written...

spydon: Aahh s/but/bug. Fixed.

lib/effects/sequence_effect.dart
spydon: Fixed this by moving the end properties to the abstract class so no more type specific initialization. Let me know what you think.

spydon:
Refactored a bit so that it doesn't have type specific initialization in the SequenceComponent.

doc/effects.md
```diff
@@ -65,6 +65,26 @@ square.addEffect(RotateEffect(
...
 ));
 ```
+
+## SequenceEffect
+
+This effect is a combination of other effects. You provide it with a list of your predefined effects.
+ 
+The effects in the list should only be passed to the SequenceEffect, never added to a PositionComponent with `addEffect`.
+
+Note that no effects added to the sequence can have their `isInfinite` property set to `true`, because then naturally the sequence will get stuck once it gets to that effect.
```
luanpotter, line 75: the last one could :) do we validate that or is this just a cautionary tale?

lib/effects/sequence_effect.dart
```diff
@@ -0,0 +1,90 @@
...
+  bool _currentWasAlternating;
+  double _driftModifier = 0.0;
+
+  SequenceEffect({
+    @required this.effects,
+    isInfinite = false,
+    isAlternating = false,
+  }) : super(isInfinite, isAlternating) {
+    /// All effects need to be finite, otherwise the sequence will get stuck
+    assert(effects.every((effect) => !effect.isInfinite));
```
luanpotter, line 19: the last one could :p

lib/effects/sequence_effect.dart
luanpotter: gotcha

lib/effects/sequence_effect.dart
```diff
@@ -0,0 +1,90 @@
...
+    super.initialize(_comp);
+    _currentIndex = 0;
+    final originalSize = _comp.toSize();
+    final originalPosition = _comp.toPosition();
+    final originalAngle = _comp.angle;
+    effects.forEach((effect) {
+      effect.reset();
+      _comp.setBySize(endSize);
+      _comp.setByPosition(endPosition);
+      _comp.angle = endAngle;
```
luanpotter, line 36: tbh I'm still a bit puzzled as to why you need to do this. but I don't think it's a merge blocker, after it's merged I can try to take a look to see if I understand it or see a way that doesn't require this.

I think this is at least better than SequenceEffect assuming the implementations of Effect are bounded.

doc/effects.md
spydon: It has an assert on if any is infininite now, but maybe we should just let the user choose if it wants to make it get stuck somewhere? And add that to the docs.

lib/effects/sequence_effect.dart
spydon: Because otherwise you can't calculate the traveltime for each individual effect, and then as a result not get the traveltime of the sequence. Since sometimes depending on where in the chain an effect is it will have different travel time.

doc/effects.md
spydon: Removed the assert and clarified the docs.

PR #358, opened 2020-05-25
Speed up linting

spydon:
# Description

Only run linting on changed and affected examples.

## Type of change

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
  - Fixing BaseGame tap detectors issues
  - Upgrade Flutter SVG to fix for flame web
  - Add linting to all the examples
+ - Run linting only on affected and changed examples
 
 ## 0.21.0
 - Adding AssetsCache.readBinaryFile
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,54 +1,54 @@
 #!/usr/bin/env bash -xe
 
 if [[ $(flutter format -n .) ]]; then
-    echo "flutter format issue"
-    exit 1
+  echo "flutter format issue"
+  exit 1
 fi
 
 flutter pub get
 result=$(dartanalyzer lib/)
 if ! echo "$result" | grep -q "No issues found!"; then
-    echo "$result"
-    echo "dartanalyzer issue: lib"
-    exit 1
+  echo "$result"
+  echo "dartanalyzer issue: lib"
+  exit 1
 fi
 
 analyzer() {
-    cd $1
-    flutter pub get
-    result=$(dartanalyzer .)
-    if ! echo "$result" | grep -q "No issues found!"; then
-        echo "$result"
-        echo "dartanalyzer issue: $1"
-        exit 1
-    fi
-    cd - > /dev/null
+  cd $1
+  flutter pub get
+  result=$(dartanalyzer .)
+  if ! echo "$result" | grep -q "No issues found!"; then
+    echo "$result"
+    echo "dartanalyzer issue: $1"
+    exit 1
+  fi
+  cd - > /dev/null
 }
 
 analyzer "example"
 
 #  Examples that are changed
 changed=$(git diff --name-only develop... doc/examples \
-    | xargs -I {} dirname {} | sed 's/\/lib$//' | uniq \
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
  - Fixing BaseGame tap detectors issues
  - Upgrade Flutter SVG to fix for flame web
  - Add linting to all the examples
+ - Run linting only on affected and changed examples
```
erickzanardo, line 7: I don't know if we should add those "internal" kind of stuff on changelog, what do you guys think?

spydon: I was thinking that too, doesn't really add any value. But I saw that it had been done before so I added mine too. Should I remove it?

erickzanardo: Yeah, it has been done before, but I always questioned in my head if we should continue doing it. Lets hear what @luanpotter thinks

luanpotter: I don't think we need to enforce it but I don't see why we can't have it (since it's already there).

The reason I added mine is because it was a required item on the PR checklist

erickzanardo: Maybe we could change the checklist to have an if :P

luanpotter: We could, sincerely I don't have a strong opinion either way. I was just following :p

erickzanardo:
> So I suggest we have two files, one to locally use and fast (your version) but on CI I think we can run from scratch.

My concern is taking too long on the CI, locally the hardware is way better and runs way faster, so if we can't find a alternative that runs good on the CI, I think we should just keep one script.

spydon:
> So basically the logic is: if I change a file that is explicitly imported in the example right? (if I understood your code right).
> 
> I think that is a very, very good approximation but it's not perfect (I can create a proper example if you want but I think you agree it's just a heuristic). So I suggest we have two files, one to locally use and fast (your version) but on CI I think we can run from scratch.

It should be possible to get the other ones too, but I think this will very rarely not catch everything so I didn't bother to write it even more complicated. And this will take maybe 20s instead of 7minutes, if you only have 2 affected examples.

luanpotter:
I don't see why taking a bit to run on the CI is that important, it's not like we are pumping PR's 10 a day. We manually review the PRs anyway, that is way longer than the CI would take.

I think perfect correctness is much more important than saving a few minutes, the reason the CI is there is to make sure we don't ever break anything, and it's doing that automatically for us.

What we could do to speed up w/o sacrificing correctness is to run then all in parallel in different subshells? that should be faster.

CHANGELOG.md
spydon: You can just remove things from the checklist per PR if you don't think they apply, like I did on this one. ;)

luanpotter:
@spydon I don't think it's worth to make the "detect what changes impact" thing perfect, it's way, way too hard. In fact it's so hard it would probably take just as long as running everything (it requires not only syntactical but semantical analysis so it would require interpreting the dart code).

I'm ok with having a quick and easy shortcut for the 99% but we need to have an automated runner that checks the 100%...

scripts/lint.sh
```diff
@@ -15,19 +15,31 @@ fi
 
 cd example/
 flutter pub get
-result=`dartanalyzer .`
+result=$(dartanalyzer .)
```
luanpotter, line 18: you could use a similar analysis to check this example as well (not sure if worth the effort)

erickzanardo:
> I think perfect correctness is much more important than saving a few minutes, the reason the CI is there is to make sure we don't ever break anything, and it's doing that automatically for us.

Of course, I don't want to lose that, I want to have both :P (if possible)

I just raised the question we could have perfect CI check, but making it a little faster.

spydon:
I'll try parallelizing it and running this check first and then the full check afterwards, then it will exit early almost always when there is an error at least.

luanpotter:
I'm also good for having two scripts, I would def. use the fast one locally before pushing if I was pretty confident on my change. And then let the CI handle the expensive checks.

spydon:
Flutter is using locking so there was no point in the parallelization, but the code is a bit cleaned up now at least and it will most of the time fail earlier if there is an error. :)

PR #367, opened 2020-06-01
Update `README-ZH` to 0.22.0

HarrisonQi:
# Description

Following `README.md`, updated `README-ZH.md` to 0.22.0

## Type of change

Changed some introduces in the README by comparing version differences.

## Why branch is master
I think that ` readme` is different from the other features of the project and may need to be kept in the latest version. If submit it to the develop branch, it may always be one version behind.


spydon:
I guess we should ping you every time we update the README and otherwise if there are no updates we could just bump the version in your version before doing the release.

HarrisonQi:
> I guess we should ping you every time we update the README and otherwise if there are no updates we could just bump the version in your version before doing the release.

Good idea

PR #412, opened 2020-07-03
[WIP] TiledComponent makes use of the SpriteBatch API

wolfenrain:
# Description
The current `TiledComponent` draws each tile individually on the canvas, switching to the `SpriteBatch` API will give a enormous performance gain because it will only do a single call to the canvas drawing all tiles using a single atlas.

**Note 1**: Because of how I implemented it, a developer wont be able to toggle the visibility of a layer. Because all the tiles are added to the `SpriteBatch` on creation, instead of drawn each frame, where the visibility check originally was.

**Note 2**: I also updated the constructor to accept a optional `Size` parameter for the destination instead of a required `double`. I made it optional so that it will grab the default tile size if non was given and so a developer could easily use irregular shaped tiles, which Tiled it self supports. But this has the side effect that the scaling wont properly work with irregular shaped tiles. But this will also be the case using a required `double` destination that does not fit with either the `width` or `height` depending on which value we use to calculate the scale.

## Type of change

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## [next]
+ - Updated the TiledComponent to use the SpriteBatch API for efficient drawing to canvas.
 
 ## 0.24.0
  - Outsourcing SVG support to an external package
--- a/doc/examples/tiled/lib/main.dart
+++ b/doc/examples/tiled/lib/main.dart
@@ -15,7 +15,10 @@ void main() {
 
 class TiledGame extends BaseGame {
   TiledGame() {
-    final TiledComponent tiledMap = TiledComponent('map.tmx', 16.0);
+    final TiledComponent tiledMap = TiledComponent(
+      'map.tmx',
+      destTileSize: const Size(16.0, 16.0),
+    );
     add(tiledMap);
     _addCoinsInMap(tiledMap);
   }
--- a/lib/components/tiled_component.dart
+++ b/lib/components/tiled_component.dart
@@ -4,6 +4,7 @@ import 'dart:ui';
 
 import 'package:flame/components/component.dart';
 import 'package:flame/flame.dart';
+import 'package:flame/sprite_batch.dart';
 import 'package:flutter/material.dart' show Colors;
 import 'package:tiled/tiled.dart' hide Image;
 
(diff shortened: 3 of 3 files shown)


wolfenrain:
> LGTM. Nice! A next improvement we could make is to use the new Layer API to optionally pre-render maps. I do that with all our maps in our games, because it's too expensive for big maps to render every frame.

That would be a good optimization indeed, did you also check out the notes I added to the description? Might be blocking for merging(and the reason I put the PR on draft)

erickzanardo:
About the first note, couldn't we just have a method that recreated the SpriteBatch so the visibility is check again?

The second note, if I understood correctly, the behaviour is the same as we have today right? So I don't see any problems.

wolfenrain:
> About the first note, couldn't we just have a method that recreated the SpriteBatch so the visibility is check again?
> 
> The second note, if I understood correctly, the behaviour is the same as we have today right? So I don't see any problems.

Hmm that would be indeed a solution, a sort of `recreate` method. I could add it to this PR.

For the second note, yes it is technically the same outcome, but the point I was trying to make, which I should have formatted beter, is that the `destination size` parameter is useless when you want to scale down irregular shaped tiles. But it also messes up your x positions or y positions, if you scale it down on either the width of height of the given destination size, then the tiles wont align up, on either the x or y axis.

That is because the SpriteBatch applies the scale after applying the transform I believe? So even if you pre-scale your positions so it would fit, it would still not align I believe? But I am currently not a 100% sure if I tested that properly....

wolfenrain:
I have finally gotten around to add a `generate` method so developers can generate the batches after messing with the visibility and such of each layer.

I have decided to keep the Size for the dstTileSize in, currently scaling does not work on both axis. But I do think it will be possible using transforms and such. But no clue on how to properly implement it right now. So I think it is ready for review.

erickzanardo:
@wolfenrain we are moving the tiled features to a separe package: https://github.com/flame-engine/flame/pull/430

If it is not much to ask, would you mind move this PR to that repository after we merge the initial PR that is open there?

wolfenrain:
Ofcourse. Just give me a pling either here or on discord and I will get right on it :+1:

erickzanardo:
> Ofcourse. Just give me a pling either here or on discord and I will get right on it 👍

Awesome!! Thanks!

erickzanardo:
@wolfenrain , when you get some time, here is the repository: https://github.com/flame-engine/flame_tiled

PR #434, opened 2020-08-03
Update v1 branch with current develop

luanpotter:


luanpotter:
You mean no PRs as in do it from local and push or no PRs as is in no approval needed?

I did the PR as I was not sure about the process. But I'm ok with either options.

erickzanardo:
> You mean no PRs as in do it from local and push or no PRs as is in no approval needed?
> 
> I did the PR as I was not sure about the process. But I'm ok with either options.

I think that when we are just updating the branch, there is no need for PR, maybe just a heads up on our telegram group is already enough.

PR #439, opened 2020-08-08
Move fps to mixin

wolfenrain:
# Description

Moved the FPS code to its own mixin so people can use it with the `Game` class instead of having to extend from `BaseGame` when they dont use components.

## Type of change

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
  - Externalizing Tiled support to its own package
  - Preventing some crashed that could happen on web when some methods were called
  - Add mustCallSuper to BaseGame `update` and `render` methods
- - Moved FPS code from BaseGame to a mixin, BaseGame used the new mixin.
+ - Moved FPS code from BaseGame to a mixin, BaseGame uses the new mixin.
 
 ## 0.24.0
  - Outsourcing SVG support to an external package
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
  - Externalizing Tiled support to its own package
  - Preventing some crashed that could happen on web when some methods were called
  - Add mustCallSuper to BaseGame `update` and `render` methods
+ - Moved FPS code from BaseGame to a mixin, BaseGame used the new mixin.
 
 ## 0.24.0
  - Outsourcing SVG support to an external package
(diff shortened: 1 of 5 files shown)


wolfenrain:
> Nice! Maybe add a line in the docs about it?

Wasn't entirely sure where to add it. So if you have a suggestion, please tell! :+1:

spydon:
> > Nice! Maybe add a line in the docs about it?
> 
> Wasn't entirely sure where to add it. So if you have a suggestion, do tell! :+1: 

In doc/debug.md I think it would fit nicely.

luanpotter:
I love this! One thing that crossed my mind is: what if instead of having a boolean to control whether recordDt or not we use the presence of the mixin? And then neither BaseGame nor Game extends the mixin by default but you can enable it by adding the mixing to your Game (or BaseGame)? I don't have a strong opinion either way, and I am ok with merging this as is, but this thought crossed my mind and I wanted to float it. Might lead to a simpler API but might be harder to understand and maybe harder to toggle, not sure.

wolfenrain:
> I love this! One thing that crossed my mind is: what if instead of having a boolean to control whether recordDt or not we use the presence of the mixin? And then neither BaseGame nor Game extends the mixin by default but you can enable it by adding the mixing to your Game (or BaseGame)? I don't have a strong opinion either way, and I am ok with merging this as is, but this thought crossed my mind and I wanted to float it. Might lead to a simpler API but might be harder to understand and maybe harder to toggle, not sure.

It would be a breaking changes in that case. Because currently only `BaseGame` has it implemented and people have to overwrite the `recordFps` functionality. If we base it on the presence of the mixin it would break backwards-compatibility.

But I do think it would be cleaner, so maybe for v1?

wolfenrain:
> > > Nice! Maybe add a line in the docs about it?
> > 
> > 
> > Wasn't entirely sure where to add it. So if you have a suggestion, do tell! +1
> 
> In doc/debug.md I think it would fit nicely.

Updated the `doc/debug.md` and slightly restructured it, to seperate the non `BaseGame` debug features from the `BaseGame`. If you have any objections or see it differently do tell :-)

wolfenrain:
> Hope I am not too late for a review, but I would like to suggest renaming the min to `FPSCounter`, What do you think?
> 
> Other than that, LGTM!

I personally dont mind, it is just a naming thing :-). Shall I supply a PR to change the name?

PR #466, opened 2020-09-19
Fix cleanup

luanpotter:
The cleanup script wasn't cleaning the nested examples we have.

Just the effects example folder is 1.2 GB w/o cleanup.

This might be a bit aggressive as it would delete folders called build, test, inside the srcs of examples but I think that will never be a problem since examples are very small.

### DIFF
--- a/doc/examples/cleanup.sh
+++ b/doc/examples/cleanup.sh
@@ -1,9 +1,11 @@
 #!/bin/bash -xe
 
-rm -rf */build
-rm -rf */android
-rm -rf */ios
-rm -rf */web
-rm -rf */macos
-rm -rf */test
-rm -rf */.dart_tool
+shopt -s globstar
+
+rm -rf **/build
+rm -rf **/android
+rm -rf **/ios
+rm -rf **/web
+rm -rf **/macos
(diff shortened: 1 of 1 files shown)


spydon (approved):

doc/examples/cleanup.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/bash -xe
 
-rm -rf */build
-rm -rf */android
-rm -rf */ios
-rm -rf */web
-rm -rf */macos
-rm -rf */test
-rm -rf */.dart_tool
+shopt -s globstar
```
erickzanardo, line 3: What is this?

luanpotter: enables **

erickzanardo: Is this a bash thing? Works on Mac as well?

luanpotter: it's bash. it should work on mac if it's compliant but wont work on sh/zsh/other shells I think. though i zsh ** works by default IIRC

erickzanardo (approved):

PR #467, opened 2020-09-20
Refactoring assets loading on Flame

erickzanardo:
# Description

Please include a summary of the change.

Refactor how assets are loaded and sprite and animations built on Flame.

## Type of change

# Checklist:


lib/main.dart
```diff
@@ -0,0 +1,117 @@
+import 'package:flutter/material.dart';
```
spydon, line 1: This was probably accidentally added

spydon (commented):
The code is nicer like this, but it feels like it will be more overall boilerplate for the user?

erickzanardo:
> The code is nicer like this, but it feels like it will be more overall boilerplate for the user?

A little bit, as now the user will have to load the images before adding them to the sprites and animations.

But this will be way less confusing and will solve a bunch of the problems with have been having so far regarding this.

Also I think with the onLoad method on game, it will be way more clear to the user how and where assets should be loaded. I am even thinking on adding some more features about this on the game class, to minimize the boilerplate issue

doc/examples/animation_widget/lib/main.dart
```diff
@@ -14,7 +14,8 @@ SpriteAnimation _animation;
 
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
-  _sprite = await Sprite.loadSprite('minotaur.png', width: 96, height: 96);
+  final image = await Flame.images.load('minotaur.png');
+  _sprite = Sprite(image, width: 96, height: 96);
```
luanpotter, line 18: we can have a `await Sprite.load` method that does this as well

luanpotter: a static method I mean. not a named constructor. because constructors cant be async

doc/examples/animations/lib/main.dart
```diff
@@ -16,14 +17,24 @@ void main() async {
...
+  Image chopper;
+  Image creature;
+  SpriteAnimation animation;
+
+  @override
+  Future<void> onLoad() async {
+    chopper = await Flame.images.load('chopper.png');
+    creature = await Flame.images.load('creature.png');
+
+    animation = SpriteAnimation.sequenced(
```
luanpotter, line 29: similarly, we can have a SpriteAnimation.loadSequenced static method. for every constructor overload we had, we can have a static async helper method

doc/examples/animations/lib/main.dart
```diff
@@ -32,7 +43,7 @@ class MyGame extends BaseGame with TapDetector {
     final animationComponent = SpriteAnimationComponent.sequenced(
       291,
       178,
-      'creature.png',
+      creature,
```
luanpotter, line 46: if we use the image twice though, we can use the syntax you used. or rely on flame.images.load caching stuff

doc/examples/aseprite/lib/main.dart
```diff
@@ -13,14 +13,16 @@ void main() async {
...
-    _start();
   }
 
-  void _start() async {
+  @override
+  Future<void> onLoad() async {
+    final image = await Flame.images.load('chopper.png');
     final animation = await SpriteAnimation.fromAsepriteData(
-      'chopper.png',
+      image,
```
luanpotter, line 22: this is already a static async method, so we can have both options here

lib/game/game.dart
```diff
@@ -91,6 +91,12 @@ abstract class Game {
 
   VoidCallback pauseEngineFn;
   VoidCallback resumeEngineFn;
+
+  /// Use this method to load the assets need for the game instance to run
+  Future<void> onLoad() async {}
+
+  /// Returns the widget which will be show while the instance is loading
+  Widget loadingWidget() => Container();
```
luanpotter, line 99: is this easily compatible with our splash screen? I really want to have a way to pass a Future for all my loads to splash screen. see my code on gravitational_waves for an example of how useful this would be!

lib/sprite.dart
```diff
@@ -11,67 +9,24 @@ class Sprite {
...
-  }
-
-  Sprite.fromImage(
     this.image, {
     double x = 0.0,
     double y = 0.0,
     double width,
     double height,
-  }) {
+  }) : assert(image != null, "image can't be null") {
```
luanpotter, line 17: oh how we need nullable type safety :cry:

doc/examples/animation_widget/lib/main.dart
erickzanardo: I am not sure about that, those helpers will need to duplicate the parameter structure from the constructor and we will need to repeat a bunch of code, on Sprite class that is not that bad, but on Animation for example, where we have different contractors that can lead to a bunch of repeated code.

Actually a have a nice idea to make up for not having those constructors, I will let you know when I have implemented then.

lib/sprite.dart
```diff
@@ -11,67 +9,24 @@ class Sprite {
...
     width ??= image.width.toDouble();
     height ??= image.height.toDouble();
```
luanpotter, line 19: SO much nicer!

lib/sprite.dart
```diff
@@ -114,19 +60,13 @@ class Sprite {
...
-      return;
-    }
     size ??= this.size;
     renderRect(canvas,
         Rect.fromLTWH(p.x - size.x / 2, p.y - size.y / 2, size.x, size.y),
         overridePaint: overridePaint);
   }
 
   void renderRect(Canvas canvas, Rect dst, {Paint overridePaint}) {
-    if (!loaded()) {
```
luanpotter, line 127: I think we can replace loaded with a `hidden` method/prop. that could be used for the same thing if the user wants but is much more widely useful. a lot of our components have a custom if on render to hide things

doc/examples/animations/lib/main.dart
erickzanardo: Like a mentioned before I have a nice idea for this, will implant it and will let you know

lib/sprite_animation.dart
```diff
@@ -72,33 +73,30 @@ class SpriteAnimation {
...
-        x: textureX + (i % amountPerRow) * textureWidth,
-        y: textureY + (i ~/ amountPerRow) * textureHeight,
-        width: textureWidth,
-        height: textureHeight,
-      );
-      frames[i] = SpriteAnimationFrame(sprite, stepTime);
-    }
-  }
+    bool loop = true,
+  }) : this.variableSequenced(
```
luanpotter, line 85: I thought this name to be a bit confusing but I don't have a better one

luanpotter:
This is great stuff @erickzanardo !!!

lib/game/game.dart
erickzanardo: I think this don't replace a splash screen, and for the most of the cases the default empty Container will suffice. A splash screen would be more of loading things for the entire game, like initializing IAP, Ads, services and etc. And this the assets needed for that Game instance.

lib/assets/images.dart
```diff
@@ -6,32 +6,39 @@ import 'dart:convert' show base64;
...
 
   void clearCache() {
-    loadedFiles.clear();
+    _loadedFiles.clear();
+  }
+
+  Image fromCache(String fileName) {
+    final image = _loadedFiles[fileName];
+    assert(image?.loadedImage != null,
+        'Tried to access an inexistent entry on cache "$fileName"');
```
luanpotter, line 22: 1. trailing commas :p 
2. I think this message could be more clear by mentioning you must call `.load` before using this.

lib/game/game.dart
luanpotter: I think we can load both on the splash screen. I mean, in general. of course we should be as flexible as possible.

doc/examples/animation_widget/lib/main.dart
luanpotter: I don't think replicating the parameters is a big deal... Animation class has like 100 constructors that are extremely similar. I don't even know what variableSequenced is for example haha. and it already is a copy of the Animation constructors. so one more is not a problem if it's going to simplify use case for some people

doc/examples/animations/lib/main.dart
```diff
@@ -16,14 +17,24 @@ void main() async {
...
-    stepTime: 0.15,
-    loop: true,
-  );
+  Image chopper;
+  Image creature;
+  SpriteAnimation animation;
+
+  @override
+  Future<void> onLoad() async {
+    chopper = await Flame.images.load('chopper.png');
```
luanpotter, line 26: should we update this example to use game.images as well?

doc/examples/animation_widget/lib/main.dart
erickzanardo: No sure I agree with that, on Animation class we have the variableSenquence and sequence, and that constructor was almost identical, and that lead us to so problems. Anyway, now with the asset loading on the game, the static method on the Sprite is not possible, at least not without rethinking the whole solution, so I would prefere to not add those. Let me know if you disagree

doc/examples/animations/lib/main.dart
erickzanardo: same thing here.

doc/examples/aseprite/lib/main.dart
erickzanardo: Didn't understood your suggestion.

lib/game/game.dart
erickzanardo: Exactly, we need to keep this flexible. On Snake Chef for example, I would have things being loaded on the splash, and things here on the game.

lib/sprite_animation.dart
erickzanardo: Me too, but alike you, I can't think on a better name, totally open to suggestions.

doc/examples/animations/lib/main.dart
erickzanardo: Sure thing! thanks for noticing that, done!

lib/sprite_animation.dart
```diff
@@ -72,33 +73,30 @@ class SpriteAnimation {
...
+          image,
+          amount,
+          List.generate(amount, (_) => stepTime),
```
spydon, line 88: List.filled is maybe more appropriate?

lib/sprite_animation.dart
spydon: Can you describe what it does and maybe we can come up with a new one together?

spydon (approved):
Looks good, mostly grammar comments

doc/examples/animation_widget/lib/main.dart
luanpotter: > the static method on the Sprite is not possible

I don't see why that is the case. but it's definitely a separated method we could add later, so let's discuss that in a followup

> we have the variableSenquence and sequence, and that constructor was almost identical

yes and I hate that. I have some ideas for improving Animation constructors. I will put a followup after this with better constructors now that the image is always loaded.

doc/examples/aseprite/lib/main.dart
luanpotter: exactly. but lets think about that on a followup.

lib/game/game.dart
luanpotter: yes, of course it is flexible, it can be any widget. I am just asking if it's easy to use the splash screen here if the user wants. they can provide sort of loading bar easily as well, of course.

lib/sprite_animation.dart
luanpotter: I will address that in a followup. i have some ideas to improve these constructors now that we did this refactor.

doc/examples/sprites/lib/main_base64.dart
```diff
@@ -1,25 +1,23 @@
...
 class MyGame extends Game {
   Sprite _sprite;
 
-  MyGame(Image image) {
-    _sprite = Sprite.fromImage(image);
+  @override
+  Future<void> onLoad() async {
+    const exampleUrl =
+        'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8/9hAAAAxElEQVQ4jYWTMQ7DIAxFIeoNuAGK1K1ISL0DMwOHzNC5p6iUPeoNOEM7GZnPJ/EUbP7Lx7KtIfH91B/L++gs5m5M9NreTN/dEZiVghatwbXvY68UlksyPjprRaxFGAJZg+uAuSSzzC7rEDirDYAz2wg0RjWRFa/EUwdnQnQ37QFe1Odjrw04AKTTaBXPAlx8dDaXdNk4rMsc0B7ge/UcYLTZxoFizxCQ/L0DMAhaX4Mzj/uzW6phu3AvtHUUU4BAWJ6t8x9N/HHcruXjwQAAAABJRU5ErkJggg==';
+    final image = await images.fromBase64('shield.png', exampleUrl);
```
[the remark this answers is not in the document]
luanpotter: yes, just answered. it had nothing to do with this constructor, rather he was using then inside the game loop, which doesn't make sense.

doc/images.md
```diff
@@ -13,29 +13,113 @@ flutter:
...
+    import 'package:flame/images.dart';
+    final imagesLoader = Images();
+    Image image = await imagesLoader.load('asd');
+```
+
+But flame also offers to ways of using this class wihtout instantiating it yourself.
+
+### Flame.images
+
+Simply singleton, provided byt the `Flame` class, is can be used as a global image caching.
```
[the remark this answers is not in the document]
luanpotter: I think he meant simple with simply. Also, `is can be` is non grammatical.

doc/images.md
```diff
@@ -13,29 +13,113 @@ flutter:
 
 It has to be a PNG file and it can have transparency.
 
+## Loading images
+
+Flame bundles an utility class called `Images` that allows you to easily load and cache images from the assets into the memory.
+
+Flutter has a collection of types related to images, and converting everything properly from a local asset to the Image that can be drawn on Canvas is a bit of a pain. This class allows you to obtain an Image that can be drawn on a Canvas using the `drawImageRect` method.
```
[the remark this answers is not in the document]
luanpotter: I might have originally wrote this, but since we are talking about the type system, I think "a collection of types" might be just too confusing for now reason, let's say "a handful of types" instead :)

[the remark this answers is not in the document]
luanpotter: let's also tone down a bit of a pain to a bit convoluted

doc/images.md
```diff
@@ -13,29 +13,113 @@ flutter:
...
+
+It automatically caches any image loaded by filename, so you can safely call it many times.
+
+The methods for loading and clearing the cache are: `load`, `loadAll`, `clear` and `clearAll`. They return a `Future` for the Image loaded.
+
+To retrieve a previously cached image, the `fromCache` method can be used.
```
luanpotter, line 26: 1. To synchronously retrieve.
2. If an image with that key was not loaded prior, it will throw an exception.

doc/images.md
```diff
@@ -13,29 +13,113 @@ flutter:
...
+
+### Manual use
```
luanpotter, line 28: Maybe a better name would be something like: Standalone usage

doc/images.md
```diff
@@ -13,29 +13,113 @@ flutter:
...
+
+It can be manually used by instantiating it, like:
+
+```dart
+    import 'package:flame/images.dart';
+    final imagesLoader = Images();
+    Image image = await imagesLoader.load('asd');
+```
+
+But flame also offers to ways of using this class wihtout instantiating it yourself.
```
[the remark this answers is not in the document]
luanpotter: he meant 2 ways
also, Flame

doc/images.md
```diff
@@ -13,29 +13,113 @@ flutter:
...
+
+    // inside an async context
+    Image image = await Flame.images.load('player.png');
+
+    final playerSprite = Sprite(image);
+```
+
+# Game.images
+
+The `Game` class offers some utility for handling images loading too, it bundles an instance of the `Images` class, that can be used to load images assets that will be used during the game. The game will automatically free the cache when the game widget is removed from the widget tree.
```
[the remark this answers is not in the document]
luanpotter: I would change
```
images loading too, it bundles an instance of the `Images` class, that
```
to
```
images loading too. It bundles an instance of the `Images` class that
```
for better pacing.

also, `images assets that will be used` => `image assets to be used`

doc/images.md
```diff
@@ -13,29 +13,113 @@ flutter:
...
+
+  @override
+  Future<void> onLoad() async {
+    final playerImage = await images.load('player.png');
+    player = Sprite(playerImage);
+  }
+}
+```
+
+Loaded assets can also be retrieved during the gampley by `images.fromCache`, for example:
```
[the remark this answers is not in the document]
luanpotter: I think gameplay doesn't fit so well here, maybe `retrieved while the game is running`?

doc/images.md
```diff
@@ -13,29 +13,113 @@ flutter:
...
+# Game.images
+
+The `Game` class offers some utility for handling images loading too, it bundles an instance of the `Images` class, that can be used to load images assets that will be used during the game. The game will automatically free the cache when the game widget is removed from the widget tree.
+
+The method `onLoad` from the `Game` class is a great place for the initial assets to be loaded.
+
+Example:
+
+```dart
+class MyGame extends Game {
```
luanpotter, line 64: BaseGame

doc/images.md
```diff
@@ -13,29 +13,113 @@ flutter:
...
+    final playerImage = await images.load('player.png');
+    player = Sprite(playerImage);
+  }
+}
+```
+
+Loaded assets can also be retrieved during the gampley by `images.fromCache`, for example:
+
+```dart
+class MyGame extends Game {
```
luanpotter, line 79: BaseGame

doc/images.md
```diff
@@ -13,29 +13,113 @@ flutter:
...
+  Future<void> onLoad() async {
+    // other loads omitted
+    await images.load('bullet.png');
+  }
+
+  void shoot() {
+    _shoots.add(Sprite(images.fromCache('bullet.png'));
+  }
+
+  // other methods omitted
```
luanpotter, line 93: I think these comments might be overkill, this is clearly implied by context

doc/images.md
```diff
@@ -13,29 +13,113 @@ flutter:
...
+}
+```
+
 ## Sprite
 
 Flame offers a `Sprite` class that represents a piece of an image (or the whole).
 
-You can create a `Sprite` giving it a pre-loaded `Image` via the `fromImage` constructor, or you can use the nameless constructor to pass a file name and have the image loaded asynchronously.
+You can create a `Sprite` by providing it an `Image` and coordinates that defines the piece of the image that that sprite represents.
```
luanpotter, line 101: I think portion is better than piece.

luanpotter: maybe region is even better

lib/components/isometric_tile_map_component.dart
```diff
@@ -46,13 +45,7 @@ class IsometricTileset {
     final i = tileId % columns;
     final j = tileId ~/ columns;
     final s = size.toDouble();
-    return Sprite.fromImage(tileset, x: s * i, y: s * j, width: s, height: s);
-  }
-
-  /// Load a tileset based on a file name.
-  static Future<IsometricTileset> load(String fileName, int size) async {
```
luanpotter, line 53: I don't think this method is useless, I think we can provide a different way of loading components without needing to care about first loading images. This is exactly the method I was suggesting to add to the others, and I don't see how leaving this here conflicts at all with the other method we are introducing.

lib/widgets/sprite_widget.dart
```diff
@@ -20,7 +20,7 @@ class SpriteWidget extends StatelessWidget {
   SpriteWidget({
     @required this.sprite,
     this.anchor = Anchor.topLeft,
-  }) : assert(sprite.loaded(), 'Sprite must be loaded');
+  });
```
luanpotter, line 23: nice side effect!

luanpotter (approved):
LGTM once all comments from spydon are done :P (and mine)

lib/game/game.dart
erickzanardo: Oh, alright, I think now I understood your suggestion, yep, would be cool if the splash screen could be integrated somehow, we could probably think on something when we update the splash screen module to be compatible with the v1

lib/components/isometric_tile_map_component.dart
erickzanardo: I removed to be in the same pattern as the others classes.

I really think that this can't exists anymore, I think that methods like these will confuse people and we will fall on the same problem that we are trying to avoid here.

This is a static method, so this will no have access to the `game.images` instance to make the load, so we will have to use `Flame.images` on those methods, and we will end up on a scenario where some assets are stored on the `game.images`, and some on `Flame.images`, and only `game.images` is automatically cleared. This seems rather confusing for me.

doc/images.md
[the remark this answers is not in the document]
erickzanardo: Lol, I am ashamed of that typo.

doc/images.md
erickzanardo: Not really, I didn't used `BaseGame` on this example snippet.

doc/images.md
erickzanardo: ditto

lib/components/isometric_tile_map_component.dart
erickzanardo: We can talk more about this later, gonna solve this for now.

PR #469, opened 2020-09-21
Fix all imports (relative & ordering)

luanpotter:
# Description

This is thanks to https://github.com/luanpotter/vscode-dart-import/pull/23 !

## Type of change

This is just a refactoring.

### DIFF
--- a/lib/components/timer_component.dart
+++ b/lib/components/timer_component.dart
@@ -1,4 +1,5 @@
 import 'dart:ui';
+
 import './component.dart';
 import '../time.dart';
 
--- a/lib/assets_cache.dart
+++ b/lib/assets_cache.dart
@@ -1,5 +1,6 @@
-import 'package:flutter/services.dart' show rootBundle;
 import 'dart:typed_data';
+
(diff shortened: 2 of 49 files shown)


erickzanardo (approved):
Conflicts...Conflicts everywhere.

Joking aside, LGTM

lib/components/timer_component.dart
```diff
@@ -1,4 +1,5 @@
 import 'dart:ui';
+
 import './component.dart';
```
spydon, line 3: Should we have a standard for whether they should have `./` maybe?
Because now there is a mix.

spydon (approved):
Lgtm, just one comment

lib/components/timer_component.dart
luanpotter: Good point, I will fix my plugin to unify that and fix in a followup.

luanpotter: Fixed here https://github.com/flame-engine/flame/pull/501

PR #472, opened 2020-09-22
Add isRelative to effects that it makes sense for

### LINKED ISSUE #398: Relative or absolute argument on effects
Have an argument which decides whether the effect should be relative or absolute.
For example, if a component's effect has size: 10 as an argument, should the component be resized to 10 (absolute) or add 10 to the size of the component (relative).

spydon:
# Description

Gives the user the ability to define whether an effect should be relative or not.

Fixes #398

## Type of change

# Checklist:

### DIFF
--- a/lib/effects/move_effect.dart
+++ b/lib/effects/move_effect.dart
@@ -17,7 +17,7 @@ class MoveEffect extends PositionComponentEffect {
     this.curve,
     isInfinite = false,
     isAlternating = false,
-    isRelative = true,
+    isRelative = false,
     Function onComplete,
   }) : super(isInfinite, isAlternating,
             isRelative: isRelative, onComplete: onComplete);
--- a/lib/effects/move_effect.dart
+++ b/lib/effects/move_effect.dart
@@ -1,26 +1,15 @@
 import 'package:flutter/animation.dart';
 import 'package:meta/meta.dart';
 
-import 'dart:math';
-
 import '../extensions/vector2.dart';
 import './effects.dart';
-
-double _direction(double p, double d) => (p - d).sign;
-double _distance(double a, double b) => (a - b).abs();
 
 class MoveEffect extends PositionComponentEffect {
   Vector2 destination;
   double speed;
   Curve curve;
-
-  double _xOriginal;
-  double _xDistance;
-  double _xDirection;
-
-  double _yOriginal;
-  double _yDistance;
-  double _yDirection;
+  Vector2 _startPosition;
+  Vector2 _delta;
 
   MoveEffect({
     @required this.destination,
@@ -28,8 +17,9 @@ class MoveEffect extends PositionComponentEffect {
     this.curve,
     isInfinite = false,
     isAlternating = false,
+    isRelative = true,
     Function onComplete,
-  }) : super(isInfinite, isAlternating, onComplete: onComplete);
+  }) : super(isInfinite, isAlternating, isRelative: isRelative, onComplete: onComplete);
 
   @override
   void initialize(_comp) {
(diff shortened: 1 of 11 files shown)


lib/effects/move_effect.dart
```diff
@@ -1,62 +1,42 @@
...
+  Vector2 _startPosition;
+  Vector2 _delta;
 
   MoveEffect({
     @required this.destination,
     @required this.speed,
     this.curve,
     isInfinite = false,
     isAlternating = false,
+    isRelative = true,
```
spydon, line 20: Should be false by default.

spydon: Thanks, fixed

lib/effects/scale_effect.dart
```diff
@@ -4,51 +4,39 @@ import 'package:meta/meta.dart';
...
   ScaleEffect({
     @required this.size,
     @required this.speed,
     this.curve,
     isInfinite = false,
     isAlternating = false,
+    isRelative = false,
     Function onComplete,
-  }) : super(isInfinite, isAlternating, onComplete: onComplete);
+  }) : super(isInfinite, isAlternating,
```
luanpotter, line 22: one per line with trailing commas or all in one line :p

lib/effects/rotate_effect.dart
```diff
@@ -7,36 +7,35 @@ class RotateEffect extends PositionComponentEffect {
...
   RotateEffect({
     @required this.radians, // As many radians as you want to rotate
     @required this.speed, // In radians per second
     this.curve,
     isInfinite = false,
     isAlternating = false,
+    isRelative = false,
     Function onComplete,
-  }) : super(isInfinite, isAlternating, onComplete: onComplete);
+  }) : super(isInfinite, isAlternating,
```
luanpotter, line 21: ditto

lib/effects/move_effect.dart
```diff
@@ -1,62 +1,42 @@
...
   MoveEffect({
     @required this.destination,
     @required this.speed,
     this.curve,
     isInfinite = false,
     isAlternating = false,
+    isRelative = false,
     Function onComplete,
-  }) : super(isInfinite, isAlternating, onComplete: onComplete);
+  }) : super(isInfinite, isAlternating,
```
luanpotter, line 22: ditto

luanpotter (approved):
LGTM with minor formatting nit. btw, can this be on v0? seems like a non-breaking change that we could just merge on develop instead of v1. but if there are complications I am ok with v1 as well, just curious.

spydon:
> LGTM with minor formatting nit. btw, can this be on v0? seems like a non-breaking change that we could just merge on develop instead of v1. but if there are complications I am ok with v1 as well, just curious. 

It is a breaking change unfortunately, because the rotational one used to be relative and now all are absolute by default. And they use the vector2 now.

lib/effects/scale_effect.dart
spydon: Good catch, fixed!

spydon:
> > LGTM with minor formatting nit. btw, can this be on v0? seems like a non-breaking change that we could just merge on develop instead of v1. but if there are complications I am ok with v1 as well, just curious.
> 
> It is a breaking change unfortunately, because the rotational one used to be relative and now all are absolute by default. And they use the vector2 now.

Scratch that, it is only internal and I think that the fix with relative -> absolute for the RotateEffect can be seen as a bug fix. So I'll point it to develop.

erickzanardo (approved):

renancaraujo (approved):

luanpotter (approved):
LGTM

PR #484, opened 2020-10-05
Generalize effects so that they can be used on all components

spydon:
# Description

Currently we only support effects for `PositionComponent`, but in theory an effect can run on any component, this PR generalizes the API so it opens up that possibility.

## Type of change

# Checklist:

### DIFF
--- a/lib/components/component.dart
+++ b/lib/components/component.dart
@@ -2,7 +2,6 @@ import 'dart:ui';
 
 import 'package:flame/effects/effects.dart';
 import 'package:flutter/painting.dart';
-import 'package:meta/meta.dart';
 
 import '../extensions/vector2.dart';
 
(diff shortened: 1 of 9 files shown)


lib/components/component.dart
```diff
@@ -1,5 +1,6 @@
 import 'dart:ui';
 
+import 'package:flame/effects/effects.dart';
```
erickzanardo, line 3: We have been using relative imports for local files.

lib/effects/effects.dart
```diff
@@ -1,5 +1,6 @@
 import 'dart:math';
 
+import 'package:flame/components/component.dart';
```
erickzanardo, line 3: ditto

lib/components/component.dart
spydon: Stupid IDE, I'll update it. :)

erickzanardo (approved):

PR #491, opened 2020-10-10
Removing audio stuff in favor of flame_audio

luanpotter:
# Description

Removing audio stuff in favor of flame_audio

Fixes: v1 milestone ticket

I haven't changed the docs yet because not sure where each part will live (flame_audio, flame, etc). Once we have a better idea of how the website will integrate with these artifacts I can refactor all the audio docs to this new structure.

## Type of change


pubspec.yaml
```diff
@@ -6,10 +6,7 @@ homepage: https://github.com/flame-engine/flame
 dependencies:
   flutter:
     sdk: flutter
-  audioplayers: ^0.15.1
   ordered_set: ^2.0.0
-  path_provider: ^1.6.0
```
luanpotter, line 11: wasn't even used! \o/

spydon (approved):
Nice, lgtm

lib/flame.dart
```diff
@@ -1,40 +1,30 @@
...
   static AssetBundle _bundle;
   static AssetBundle get bundle => _bundle == null ? rootBundle : _bundle;
 
-  /// Access a shared instance of the [FlameAudio] class.
-  static FlameAudio audio = FlameAudio();
-
-  /// Access a shared instance of the [Bgm] class.
-  static Bgm _bgm;
-  static Bgm get bgm => _bgm ??= Bgm();
+  /// Access a shard instance of [AssetsCache] class.
```
erickzanardo, line 19: I know this was not written by you, but there is a typo here: `shared`

pubspec.yaml
erickzanardo: \o/

erickzanardo (approved):
One nit, but LGTM already.

I think we could keep our doc here, and I would just point out that there is an underlying package that should be used to enable the audio features.

lib/flame.dart
luanpotter: boy scout's rule!

PR #496, opened 2020-10-12
Utilize Vector2 for Sprite

spydon:
# Description

Fixes the issue with Sprite crashing in the constructor if it doesn't get the optional position argument.
(Also converts size to Vector2 where it makes sense)

## Type of change

# Checklist:

### DIFF
--- a/lib/sprite.dart
+++ b/lib/sprite.dart
@@ -1,12 +1,13 @@
 import 'dart:ui';
 
+import 'extensions/offset.dart';
 import 'extensions/vector2.dart';
 import 'palette.dart';
 
 class Sprite {
   Paint paint = BasicPalette.white.paint;
   Image image;
-  Rect src;
+  Rect bounds;
 
   Sprite(
     this.image, {
@@ -14,7 +15,7 @@ class Sprite {
     Vector2 size,
   }) : assert(image != null, "image can't be null") {
     size ??= Vector2(image.width.toDouble(), image.height.toDouble());
-    src = position.toPositionedRect(size);
+    this.position = position;
   }
 
   double get _imageWidth => image.width.toDouble();
@@ -23,7 +24,13 @@ class Sprite {
 
   Vector2 get originalSize => Vector2(_imageWidth, _imageHeight);
 
-  Vector2 get size => Vector2(src.width, src.height);
+  Vector2 get size => Vector2(bounds.width, bounds.height);
+
+  Vector2 get position => bounds.topLeft.toVector2();
+
+  set position(Vector2 position) {
+    bounds = (position ?? Vector2.zero()).toPositionedRect(size);
+  }
 
   /// Renders this Sprite on the position [p], scaled by the [scale] factor provided.
   ///
@@ -51,14 +58,11 @@ class Sprite {
 
   void render(
     Canvas canvas, {
-    double width,
-    double height,
+    Vector2 size,
     Paint overridePaint,
   }) {
-    width ??= size.x;
-    height ??= size.y;
-    renderRect(canvas, Rect.fromLTWH(0.0, 0.0, width, height),
-        overridePaint: overridePaint);
+    size ??= this.size;
+    renderRect(canvas, size.toRect(), overridePaint: overridePaint);
   }
 
   /// Renders this sprite centered in the position [p], i.e., on [p] - [size] / 2.
@@ -72,9 +76,11 @@ class Sprite {
     Paint overridePaint,
   }) {
     size ??= this.size;
-    renderRect(canvas,
-        Rect.fromLTWH(p.x - size.x / 2, p.y - size.y / 2, size.x, size.y),
-        overridePaint: overridePaint);
+    renderRect(
+      canvas,
+      (p - size / 2).toPositionedRect(size),
+      overridePaint: overridePaint,
+    );
   }
 
   void renderRect(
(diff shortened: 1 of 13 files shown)


lib/sprite.dart
```diff
@@ -1,20 +1,21 @@
...
 class Sprite {
   Paint paint = BasicPalette.white.paint;
   Image image;
-  Rect src;
+  Rect bounds;
 
   Sprite(
     this.image, {
     Vector2 position,
     Vector2 size,
```
erickzanardo, line 15: Add @required here?

lib/widgets/sprite_widget.dart
```diff
@@ -45,20 +47,14 @@ class _SpritePainter extends CustomPainter {
...
-    final double dx = _anchor.relativePosition.x * size.width;
-    final double dy = _anchor.relativePosition.y * size.height;
-
-    canvas.translate(
-      dx - w * _anchor.relativePosition.x,
-      dy - h * _anchor.relativePosition.y,
-    );
+    final paintSize = _sprite.size * rate;
+    final anchorPosition = _anchor.relativePosition;
+    final anchoredPosition = size.toVector2()..multiply(anchorPosition);
```
erickzanardo, line 53: Why not make size be a Vector2 already? So we can avoid this conversion?

spydon: It is not under our control, it is the size in CustomPainter.

erickzanardo: Oh, didn't realized this was the widget class.

lib/sprite.dart
```diff
@@ -1,20 +1,21 @@
...
 
+import 'extensions/offset.dart';
 import 'extensions/vector2.dart';
 import 'palette.dart';
 
 class Sprite {
   Paint paint = BasicPalette.white.paint;
   Image image;
-  Rect src;
+  Rect bounds;
```
luanpotter, line 10: I actually think `src` is clearer. This is to distinguish between the target rect (which is where in the screen it will be drawn). The `src` is where from the texture it's extracted.

lib/sprite.dart
luanpotter: @spydon I totally 100% thought you were talking about SpriteComponent! only now I realized it was the actual Sprite class haha. sorry about that. I think this for sure could have a 0,0 default. then it doesn't need to be required. I was thinking spriteComponent.position might not be as clear with a default value but this one for sure can have one. In fact should be easy to create a sprite with no params other than the image: it should default to the whole image.

lib/sprite.dart
```diff
@@ -1,20 +1,21 @@
...
 
   Sprite(
     this.image, {
     Vector2 position,
     Vector2 size,
   }) : assert(image != null, "image can't be null") {
     size ??= Vector2(image.width.toDouble(), image.height.toDouble());
-    src = position.toPositionedRect(size);
+    this.position = position;
```
luanpotter, line 18: `?? Vector.empty()`? or you can add the default value to the constructor param, either works

lib/sprite.dart
```diff
@@ -23,7 +24,13 @@ class Sprite {
 
   Vector2 get originalSize => Vector2(_imageWidth, _imageHeight);
 
-  Vector2 get size => Vector2(src.width, src.height);
+  Vector2 get size => Vector2(bounds.width, bounds.height);
```
luanpotter, line 27: this is the same as originalSize. maybe could also be called srcSize. this method is not necessary I think.

lib/sprite.dart
spydon: It is done in the setter

lib/sprite.dart
```diff
@@ -23,7 +24,13 @@ class Sprite {
...
+
+  Vector2 get position => bounds.topLeft.toVector2();
```
luanpotter, line 29: similarly this could be srcPosition to be more clear?

lib/sprite.dart
```diff
@@ -23,7 +24,13 @@ class Sprite {
...
+
+  set position(Vector2 position) {
+    bounds = (position ?? Vector2.zero()).toPositionedRect(size);
```
luanpotter, line 32: oh I see what you did. I don't think the use case existing sprite.position = null throwing an error is a major problem as it's clearly a mistake but I'm ok with handling this here rather than in the constructor.

lib/sprite.dart
```diff
@@ -51,14 +58,11 @@ class Sprite {
...
-    double width,
-    double height,
+    Vector2 size,
     Paint overridePaint,
   }) {
-    width ??= size.x;
-    height ??= size.y;
-    renderRect(canvas, Rect.fromLTWH(0.0, 0.0, width, height),
-        overridePaint: overridePaint);
+    size ??= this.size;
```
luanpotter, line 64: nice!

lib/sprite.dart
spydon: Not necessarily, since the user can send in a size argument

lib/sprite.dart
spydon: Now you can move the sprite position afterwards too buy just setting the position was my thought

erickzanardo (approved):

lib/sprite.dart
```diff
@@ -10,11 +11,11 @@ class Sprite {
 
   Sprite(
     this.image, {
-    Vector2 position,
+    Vector2 srcPosition,
     Vector2 size,
```
luanpotter, line 15: since the other one is srcPosition, how about srcSize? to distinguish it from originalSize which is the whole image size? (just a thought, not sure if it's needed)

luanpotter (approved):
LGTM

lib/sprite.dart
spydon: Sounds reasonable, fixed

PR #500, opened 2020-10-14
Improve IsometricTileMap and Spritesheet classes

luanpotter:
Improve IsometricTileMap and Spritesheet classes


lib/spritesheet.dart
```diff
@@ -6,66 +6,95 @@ import 'sprite.dart';
...
-  int rows;
+  /// The src image from which each sprite will be generated.
+  final Image image;
 
-  List<List<Sprite>> _sprites;
+  /// The size of each rectangle within the image that define each sprite.
+  ///
+  /// For example, if this sprite sheet is a tile map, this would be the tile size.
+  /// If it's an animation sheet, this would be the frame size.
+  final Vector2 srcSize;
```
luanpotter, line 28: to be consistent with sprite I change textuteW/H to srcSize

lib/spritesheet.dart
```diff
@@ -6,66 +6,95 @@ import 'sprite.dart';
...
-  Sprite getSprite(int row, int column) {
-    final Sprite s = _sprites[row][column];
+  /// Compute the number of columns the image has
+  /// by using the image width and tile size.
+  int get columns => image.width ~/ srcSize.x;
 
-    assert(s != null, 'No sprite found for row $row and column $column');
+  /// Compute the number of rows the image has
+  /// by using the image height and tile size.
+  int get rows => image.height ~/ srcSize.y;
```
luanpotter, line 53: cols and rols can be getters, no need to have duplicated fields

lib/spritesheet.dart
```diff
@@ -6,66 +6,95 @@ import 'sprite.dart';
...
-    int y,
-  ) {
-    final size = Vector2(textureWidth.toDouble(), textureHeight.toDouble());
-    return Sprite(
-      image,
-      srcPosition: Vector2(x.toDouble(), y.toDouble())..multiply(size),
-      srcSize: size,
-    );
-  }
+  SpriteSheet.fromColsAndRows({
```
luanpotter, line 38: two easy ways to create the sheet. providing both tileSize and col/row is unnecessary

spydon (approved):
Nice, much cleaner

lib/spritesheet.dart
```diff
@@ -6,66 +6,95 @@ import 'sprite.dart';
...
+  /// Compute the number of rows the image has
+  /// by using the image height and tile size.
+  int get rows => image.height ~/ srcSize.y;
 
-    return s;
+  /// Gets the sprite in the position (row, column) on the sprite sheet grid.
+  ///
+  /// This is lazily computed and cached for your convenience.
+  Sprite getSprite(int row, int column) {
+    return getSpriteById(row * columns + column);
```
erickzanardo, line 59: Nit: This could be an onliner

lib/spritesheet.dart
```diff
@@ -6,66 +6,95 @@ import 'sprite.dart';
...
+  final Image image;
 
-  List<List<Sprite>> _sprites;
+  /// The size of each rectangle within the image that define each sprite.
+  ///
+  /// For example, if this sprite sheet is a tile map, this would be the tile size.
+  /// If it's an animation sheet, this would be the frame size.
+  final Vector2 srcSize;
+
+  final Map<int, Sprite> _spriteCache = {};
```
erickzanardo, line 30: You could use the MemoryCache class here: https://github.com/flame-engine/flame/blob/v1.0.0/lib/memory_cache.dart

lib/spritesheet.dart
erickzanardo: I wouldn’t abbreviate column here on the method name

lib/spritesheet.dart
```diff
@@ -6,66 +6,95 @@ import 'sprite.dart';
...
-    );
-  }
+  SpriteSheet.fromColsAndRows({
+    @required this.image,
+    @required int columns,
+    @required int rows,
+  }) : srcSize = Vector2(
+          image.width / columns,
+          image.height / rows,
+        );
```
erickzanardo, line 45: This indentation looks weird on github

erickzanardo (approved):
Some suggestions and nits but LGTM

lib/spritesheet.dart
luanpotter: sorry, old habits die hard but I agree

lib/spritesheet.dart
luanpotter: sadly it's from flutter format, the way they ident after the `:` on constructor declarations is not much of my liking but it is what it does

luanpotter: I can try to see if it fits in one line but IIRC it didn't. regardless, there are cases we will need to accept this eventually

lib/spritesheet.dart
luanpotter: doesn't fit by one character

lib/spritesheet.dart
luanpotter: I don't think it fits here because the only thing it adds over the map is eviction, that we don't want on this case.

PR #501, opened 2020-10-15
Followup fix imports with new version of vscode-dart-import

luanpotter:
This is a followup to @spydon 's comment here: https://github.com/flame-engine/flame/pull/469#discussion_r492371603

I updated my plugin to standardize `./foo` imports into `foo`.

Would be good to have a linter rule for imports so we don't get so much inconsistency, specially between relative/absolute imports that cause hard to debug problems.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -48,6 +48,7 @@ linter:
     - prefer_initializing_formals
     - prefer_is_empty
     - prefer_is_not_empty
+    - prefer_relative_imports
     - prefer_typing_uninitialized_variables
     - recursive_getters
     - slash_for_doc_comments
--- a/lib/components/joystick/joystick_action.dart
+++ b/lib/components/joystick/joystick_action.dart
@@ -1,15 +1,15 @@
 import 'dart:math';
 import 'dart:ui';
 
-import 'joystick_component.dart';
-import 'joystick_events.dart';
-import '../../gestures.dart';
-import '../../sprite.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/widgets.dart';
 
 import '../../extensions/offset.dart';
 import '../../extensions/vector2.dart';
+import '../../gestures.dart';
+import '../../sprite.dart';
+import 'joystick_component.dart';
+import 'joystick_events.dart';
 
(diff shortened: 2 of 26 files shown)


erickzanardo:
I think we need lint rules to enforce things like this. We can't rely on a plugin from a specific editor. Not everyone uses vs code.

luanpotter:
Yes, would be good to have a linter, but not sure how to do extend the dart analyze, it seems very limited (doesn't properly enforce trailing commas either, for example). But until we figure it out, we can be careful not to use any incorrect imports. Btw my plugin has been ported to [several other IDEs](https://github.com/luanpotter/vscode-dart-import/issues/24), and also it is open source :p

luanpotter:
Oh there is something I think https://dart-lang.github.io/linter/lints/prefer_relative_imports.html

erickzanardo (approved):
> Yes, would be good to have a linter, but not sure how to do extend the dart analyze, it seems very limited (doesn't properly enforce trailing commas either, for example). But until we figure it out, we can be careful not to use any incorrect imports. Btw my plugin has been ported to [several other IDEs](https://github.com/luanpotter/vscode-dart-import/issues/24), and also it is open source :p

My point is, this is the second PR open that do this. We need to figure it out how to enforce this, otherwise we will keep having this and will keep having possible conflicts due to changes like this.

We need also to consider that other people outside the team will eventually contribute, and we can't assume that people will have this plugin installed.

With all that said, LGTM :P

luanpotter:
It is the second one because I changed the plugin to ban the unnecessary `./` prefix as per requested on the other PR.

People don't need this or any plugin, they just need to write the import statements properly and we need to be thorough on our code reviews not to let any absolute imports (or `./` prefixes) get in.

That being said, I did find and add the linter rule. And also hopefully contributors will choose an editor/IDE that does automatic import for them (or fix it via a plugin), for their own sanity :p

spydon (approved):
Lgtm, is it possible to run the import formatter headless? Then we can just use that one as an extra linting script and compare the diffs.

erickzanardo:
> they just need to write the import statements properly and we need to be thorough on our code reviews not to let any absolute imports (or ./ prefixes) get in.

That is the part that I am uncomfortable with. Code style things should be enforced by automatic ways of checking, it is very easy to let these details pass. But good thing that you found a linter for that.

PR #502, opened 2020-10-15
Enable strong strict mode on dartalyzer

luanpotter:
So upon further studying dartalyzer turns out we can enable more strict type safety by adding some options to `analysis_options.yaml`. This basically bans implicit dynamics which in turn makes all type errors compile time (except for casts).

The only place we should genuinely use dynamic is on json parsing, which I fixed to use explicit cast. Everywhere else should be strongly type (explicitly for parameters/fields/function defs and implicitly for local variables).


doc/images.md
```diff
@@ -214,10 +214,9 @@ This constructor makes creating an Animation very easy using sprite sheets.
 If you use Aseprite for your animations, Flame does provide some support for Aseprite animation's JSON data. To use this feature you will need to export the Sprite Sheet's JSON data, and use something like the following snippet:
 
 ```dart
-    SpriteAnimation animation = await Animation.fromAsepriteData(
-      imageInstance, // Sprite Sheet image path
-      "./assets/chopper.json" // Sprite Sheet animation JSON data
-    );
+  final image = await images.load('chopper.png');
+  final jsonData = await assets.readJson('chopper.json');
+  final animation = SpriteAnimation.fromAsepriteData(image, jsonData);
```
luanpotter, line 219: this was super outdated

lib/components/position_component.dart
```diff
@@ -202,8 +202,8 @@ abstract class PositionComponent extends Component {
     return _children.remove(c);
   }
 
-  Iterable<PositionComponent> removeChildren(
-    bool Function(PositionComponent) test,
+  Iterable<Component> removeChildren(
+    bool Function(Component) test,
```
luanpotter, line 206: this is a good example of a bug I introduced that would only be catch at runtime (and this function was not tested). compile time errors are a blessing.

lib/particles/accelerated_particle.dart
```diff
@@ -41,7 +40,7 @@ class AcceleratedParticle extends CurvedParticle with SingleChildParticle {
   @override
   void update(double t) {
     speed += acceleration * t;
-    position += speed * t - (acceleration * pow(t, 2)) / 2;
+    position += speed * t - (acceleration * t * t) / 2;
```
luanpotter, line 43: so `pow` is declared for `num` instead of `int`/`double` and strong mode doesn't like it. probably dart needs to do some updates so it's compatible but for this I think multiplication is perfectly fine. other option is `pow(a, b).toDouble()`.

lib/sprite_batch.dart
```diff
@@ -13,10 +13,9 @@ class SpriteBatch {
   List<Color> colors = [];
 
   static const defaultBlendMode = BlendMode.srcOver;
-  static const defaultCullRect = null;
```
luanpotter, line 16: no need to `??` to `null`.

spydon (approved):

PR #504, opened 2020-10-17
Proposal: rename game.size to game.gameSize as well

luanpotter:
Proposal: Since we recently changed component.size to gameSize, though game.size is more clear this makes it more consistent. Not 100% convinced though, so thoughts welcome.

### DIFF
--- a/lib/game/base_game.dart
+++ b/lib/game/base_game.dart
@@ -31,7 +31,7 @@ class BaseGame extends Game with FPSCounter {
   final List<Component> _removeLater = [];
 
   /// Current game viewport size, updated every resize via the [resize] method hook
-  Vector2 size;
+  Vector2 gameSize;
 
   /// Camera position; every non-HUD component is translated so that the camera position is the top-left corner of the screen.
   Vector2 camera = Vector2.zero();
@@ -58,8 +58,8 @@ class BaseGame extends Game with FPSCounter {
     }
 
     // first time resize
-    if (size != null) {
-      c.onGameResize(size);
+    if (gameSize != null) {
+      c.onGameResize(gameSize);
     }
 
     c.onMount();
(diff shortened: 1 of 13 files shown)


erickzanardo (approved):
You just forget the changelog, other than that, LGTM

lib/game/base_game.dart
```diff
@@ -31,7 +31,7 @@ class BaseGame extends Game with FPSCounter {
   final List<Component> _removeLater = [];
 
   /// Current game viewport size, updated every resize via the [resize] method hook
-  Vector2 size;
+  Vector2 gameSize;
```
spydon, line 34: I think it makes sense on the component that it has gameSize named gameSize, but on the game which is another type of object I don't think we have to follow that convention, that the size field on game is the game size should be clear and just adds redundancy to call game.gameSize imho.

erickzanardo: On a second thought, I think I agree with @spydon !

renancaraujo (approved):

lib/game/game.dart
```diff
@@ -40,7 +40,7 @@ abstract class Game {
   /// This is the resize hook; every time the game widget is resized, this hook is called.
   ///
   /// The default implementation does nothing; override to use the hook.
-  void resize(Vector2 size) {}
+  void onGameResize(Vector2 gameSize) {}
```
erickzanardo, line 43: This could be just `onResize` what do you think?

erickzanardo (commented):
I think we should leave the property named size, but the `resize` method, could be improved, my suggestion is just `onResize`.

luanpotter:
@erickzanardo sgtm. I will close this and make a followup PR just with the method name change.

I will use this PR as an example on a small talk about contributing to open source next week, so please leave it this to me :)

PR #520, opened 2020-10-30
Pause and resume effects

spydon:
# Description

Pause and resume effects

## Type of change

# Checklist:

### DIFF
--- a/lib/components/component.dart
+++ b/lib/components/component.dart
@@ -23,7 +23,7 @@ abstract class Component {
   @mustCallSuper
   void update(double dt) {
     _effects.removeWhere((e) => e.hasFinished());
-    _effects.forEach((e) => e.update(dt));
+    _effects.where((e) => !e.isPaused).forEach((e) => e.update(dt));
   }
 
   /// Renders this component on the provided Canvas [c].
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
  - Refactor timer class
  - include all changed that are included on 0.28.0
  - Effects duration and test suite for basic effects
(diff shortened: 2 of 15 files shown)


lib/components/component.dart
```diff
@@ -23,7 +23,7 @@ abstract class Component {
   @mustCallSuper
   void update(double dt) {
     _effects.removeWhere((e) => e.hasFinished());
-    _effects.forEach((e) => e.update(dt));
+    _effects.where((e) => !e.isPaused).forEach((e) => e.update(dt));
```
erickzanardo, line 26: Minor suggestion, to avoid multiple iterations (a where and then a forEach). We could check on the update if the effect is paused. What do you think?

erickzanardo (approved):
One suggestion, but I don't have strong opinions about it, so it is up to you. LGTM!

lib/components/component.dart
spydon: Isn't the Dart compiler smart enough to bundle them? Before I had it like this:
`effect.forEach((e) => if(!e.isPaused) e.update(dt));`
but I thought the current solution was easier to read?

erickzanardo: I don't know if the dart compiler will be smart or not for this, but I would bet that it don't on this case.

Anyway, I liked the new approach 🎉

luanpotter: we had that discussion before, if dart optimizes that like java streams. my conclusion is: unless it's a core loop, let's go for readability. this is one components effects. it will likely never matter.

if there is something we can optimize in the for each component in the game loop, thats where we should invest

luanpotter (approved):
LGTM

PR #525, opened 2020-11-11
Fix parallax bug

spydon:
# Description

Fixing parallax bug due to missing parenthesis.

## Type of change

# Checklist:

### DIFF
--- a/lib/components/parallax_component.dart
+++ b/lib/components/parallax_component.dart
@@ -78,21 +78,20 @@ class ParallaxLayer {
 
     // Number of images that can fit on the canvas plus one
     // to have something to scroll to without leaving canvas empty
-    final count = Vector2.all(1) + size.clone()
-      ..divide(_imageSize);
+    final Vector2 count = Vector2.all(1) + (size.clone()..divide(_imageSize));
 
     // Percentage of the image size that will overflow
-    final overflow = ((_imageSize.clone()..multiply(count)) - size)
+    final Vector2 overflow = ((_imageSize.clone()..multiply(count)) - size)
       ..divide(_imageSize);
 
     // Align image to correct side of the screen
-    final alignment = parallaxImage.alignment;
-    final marginX = alignment.x == 0 ? overflow.x / 2 : alignment.x;
-    final marginY = alignment.y == 0 ? overflow.y / 2 : alignment.y;
+    final Alignment alignment = parallaxImage.alignment;
+    final double marginX = alignment.x == 0 ? overflow.x / 2 : alignment.x;
+    final double marginY = alignment.y == 0 ? overflow.y / 2 : alignment.y;
     _scroll ??= Vector2(marginX, marginY);
 
     // Size of the area to paint the images on
-    final paintSize = count..multiply(_imageSize);
+    final Vector2 paintSize = count..multiply(_imageSize);
     _paintArea = paintSize.toRect();
   }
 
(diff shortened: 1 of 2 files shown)


lib/components/parallax_component.dart
```diff
@@ -78,21 +78,20 @@ class ParallaxLayer {
 
     // Number of images that can fit on the canvas plus one
     // to have something to scroll to without leaving canvas empty
-    final count = Vector2.all(1) + size.clone()
-      ..divide(_imageSize);
+    final Vector2 count = Vector2.all(1) + (size.clone()..divide(_imageSize));
```
spydon, line 81: The parenthesis here is what was missing and which completely broke the positioning.

luanpotter: I think the parenthesis would only be necessary in the multiline case right? But either way lets keep it for clarity. A bit odd we are summing (1,1) to the vector? looks a bit unusual to me.

lib/components/parallax_component.dart
```diff
@@ -78,21 +78,20 @@ class ParallaxLayer {
...
-    final marginX = alignment.x == 0 ? overflow.x / 2 : alignment.x;
-    final marginY = alignment.y == 0 ? overflow.y / 2 : alignment.y;
+    final Alignment alignment = parallaxImage.alignment;
+    final double marginX = alignment.x == 0 ? overflow.x / 2 : alignment.x;
+    final double marginY = alignment.y == 0 ? overflow.y / 2 : alignment.y;
     _scroll ??= Vector2(marginX, marginY);
 
     // Size of the area to paint the images on
-    final paintSize = count..multiply(_imageSize);
+    final Vector2 paintSize = count..multiply(_imageSize);
```
luanpotter, line 94: do we need the explicit type for local variables? I think in general we dont do that

lib/components/parallax_component.dart
```diff
@@ -117,10 +116,13 @@ class ParallaxLayer {
         break;
     }
 
-    final dx = _scroll.x * _imageSize.x;
-    final dy = _scroll.y * _imageSize.y;
-
-    _paintArea = Rect.fromLTWH(-dx, -dy, _paintArea.width, _paintArea.height);
+    final Vector2 scrollPosition = Vector2.copy(_scroll)..multiply(_imageSize);
```
luanpotter, line 119: whats the difference between Vector2.copy and vector2.clone()?

lib/components/parallax_component.dart
```diff
@@ -201,12 +203,8 @@ class ParallaxComponent extends PositionComponent {
     if (!loaded()) {
       return;
     }
-
     super.render(canvas);
-
-    canvas.save();
     _layers.forEach((layer) => layer.render(canvas));
-    canvas.restore();
```
luanpotter, line 209: is this safe to remove?

lib/components/parallax_component.dart
spydon: It is explained in the comment above :)
Does dart make a difference of whitespace for ordering?!

lib/components/parallax_component.dart
spydon: It became clearer here since Offset is also used.

lib/components/parallax_component.dart
spydon: Nothing really, I guess we should settle for `clone` since that is used in the most places.

lib/components/parallax_component.dart
spydon: Maybe not? I removed it as part of finding where the bug was and forgot to add it again.

lib/components/parallax_component.dart
spydon: After reading the style guide again now it should be removed on the places where it doesn't clarify anything though, like the explicit doubles I added.

lib/components/parallax_component.dart
spydon: Added them back

erickzanardo (approved):
LGTM

PR #526, opened 2020-11-14
Simplify BaseGame interactions

spydon:
# Description

Suggested simplification of BaseGame.

I don't think the user needs to be aware from the names of the methods that the components aren't added/removed directly, and if they need to modify the components list directly they can still easily do so.

## Type of change

# Checklist:

### DIFF
--- a/lib/game/base_game.dart
+++ b/lib/game/base_game.dart
@@ -78,12 +78,12 @@ class BaseGame extends Game with FPSCounter {
     _addLater.addAll(components);
   }
 
-  /// Marks a component to be removed from the components list on the next game loop cycle
+  /// Marks a component to be removed from the components list on the next game tick
   void remove(Component c) {
     _removeLater.add(c);
   }
 
-  /// Marks a list of components to be removed from the components list on the next game loop cycle
+  /// Marks a list of components to be removed from the components list on the next game tick
   void removeAll(Iterable<Component> components) {
     _removeLater.addAll(components);
   }
--- a/lib/game/base_game.dart
+++ b/lib/game/base_game.dart
@@ -38,7 +38,11 @@ class BaseGame extends Game with FPSCounter {
   /// Camera position; every non-HUD component is translated so that the camera position is the top-left corner of the screen.
   Vector2 camera = Vector2.zero();
 
-  /// Does preparation on a component before any update or render method is called on it
+  /// This method is called for every component added.
+  /// It does preparation on a component before any update or render method is called on it.
+  ///
+  /// You can use this to setup your mixins, pre-calculate stuff on every component, or anything you desire.
+  /// By default, this calls the first time resize for every component, so don't forget to call super.preAdd when overriding.
   @mustCallSuper
   void prepare(Component c) {
     if (c is Tapable) {
--- a/lib/game/base_game.dart
+++ b/lib/game/base_game.dart
@@ -16,12 +16,13 @@ import 'game.dart';
 
 /// This is a more complete and opinionated implementation of Game.
 ///
-/// It still needs to be subclasses to add your game logic, but the [update], [render] and [resize] methods have default implementations.
+/// BaseGame should be extended to add your game logic.
+/// [update], [render] and [onResize] methods have default implementations.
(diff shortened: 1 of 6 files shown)


test/base_game_test.dart
```diff
@@ -0,0 +1,110 @@
...
+      final MyComponent component = MyComponent();
+
+      game.size = size;
+      game.add(component);
+      game.update(0.0);
+
+      expect(game.components.contains(component), true);
+    });
+
+    flutter.testWidgets('component render and update is called',
```
spydon, line 85: Any ideas of how to do this test better without calling update and render directly?

lib/game/base_game.dart
```diff
@@ -31,17 +32,15 @@ class BaseGame extends Game with FPSCounter {
...
-  Vector2 size;
+  final Vector2 size = Vector2.zero();
+  set size(Vector2 size) => this.size.setFrom(size);
 
   /// Camera position; every non-HUD component is translated so that the camera position is the top-left corner of the screen.
   Vector2 camera = Vector2.zero();
 
-  /// This method is called for every component added, both via [add] and [addLater] methods.
-  ///
-  /// You can use this to setup your mixins, pre-calculate stuff on every component, or anything you desire.
```
luanpotter, line 41: I think this phrase is still useful in the new name of the method

lib/game/base_game.dart
```diff
@@ -61,32 +60,30 @@ class BaseGame extends Game with FPSCounter {
...
-  void addLater(Component c) {
-    preAdd(c);
-    _addLater.add(c);
+  /// Prepares and registers a list of components to be added on the next game tick
+  void addAll(Iterable<Component> components) {
+    components.forEach(prepare);
+    _addLater.addAll(components);
   }
 
   /// Marks a component to be removed from the components list on the next game loop cycle
```
luanpotter, line 77: let's unify `on the next game tick` and `on the next game loop cycle` as they both mean the same thing. no preferences on which but we should use consistent language to avoid confusion

lib/game/base_game.dart
```diff
@@ -139,7 +137,7 @@ class BaseGame extends Game with FPSCounter {
   @override
   @mustCallSuper
   void onResize(Vector2 size) {
-    this.size = size;
+    this.size.setFrom(size);
```
luanpotter, line 140: the perks of mutability

test/base_game_test.dart
```diff
@@ -0,0 +1,110 @@
...
+      expect(component.gameRef, game);
+    });
+
+    test('component can be tapped', () {
+      final MyGame game = MyGame();
+      final MyComponent component = MyComponent();
+
+      game.size = size;
+      game.add(component);
+      game.update(0.0);
```
luanpotter, line 68: I think adding a comment here: components are only added on the next tick will make it super clear for new people reading it (similarly on the other test)

test/base_game_test.dart
luanpotter: I think for most logic tests it's fine to call update/render directly like you did above, and other tests we have do that. but I think it's super great to have this one test do it like this to make sure it's wired correctly. I don't think there is any other option other than use flutter.testWidgets. we could make it test more by using game.widget instead of rebuilding some bindings, not sure how that would work

luanpotter (approved):
I love these new apis! A great improvement and simplification to understanding I think

lib/game/base_game.dart
spydon: Yeah, I guess we should settle for what we want, it can be both a perk and a confusion I think.

lib/game/base_game.dart
spydon: Ah true, I removed preAdd/prepare for a while completely until I realized the composed component needs it. 
I'll re-add the comments.

lib/game/base_game.dart
spydon: I took game tick now, but if anyone prefers the other I can change to that.

erickzanardo (approved):
LGTM

PR #527, opened 2020-11-15
Unify naming standard for removal of components

spydon:
# Description

I think it is a bit confusing for the user to have `destroy` returning a boolean to mark it for removal, since destroy is a verb.
This change also unifies that there is no difference in whether `onRemove` (`onDestroy` before) is called if the component is removed from within or from the `BaseGame` instance.

## Type of change

# Checklist:

### DIFF
--- a/doc/components.md
+++ b/doc/components.md
@@ -32,13 +32,13 @@ Every `Component` has a few other methods that you can optionally implement, tha
 
 The `resize` method is called whenever the screen is resized, and in the beginning once when the component is added via the `add` method. You need to apply here any changes to the x, y, width and height of your component, or any other changes, due to the screen resizing. You can start these variables here, as the sprite won't be rendered until everything is set.
 
-The `destroy` method can be implemented to return true and warn the `BaseGame` that your object is marked for destruction, and it will be remove after the current update loop. It will then no longer be rendered or updated.
+The `shouldRemove` method can be implemented to return true and `BaseGame` will remove it before the next update loop. It will then no longer be rendered or updated. Note that `game.remove(Component c)` can also be used to remove components.
 
 The `isHUD` method can be implemented to return true (default false) to make the `BaseGame` ignore the `camera` for this element.
 
 The `onMount` method can be overridden to run initialization code for the component. When this method is called, BaseGame ensures that all the mixins which would change this component behaviour are already resolved.
 
-The `onDestroy` method can be overridden to run code before the component is removed from the game.
+The `onRemove` method can be overridden to run code before the component is removed from the game.
 
 There are also other implementations:
 
--- a/doc/components.md
+++ b/doc/components.md
@@ -38,7 +38,7 @@ The `isHUD` method can be implemented to return true (default false) to make the
 
 The `onMount` method can be overridden to run initialization code for the component. When this method is called, BaseGame ensures that all the mixins which would change this component behaviour are already resolved.
 
-The `onRemove` method can be overridden to run code before the component is removed from the game.
+The `onRemove` method can be overridden to run code before the component is removed from the game, it is only run once even if the component is removed both by using the `BaseGame` remove method and the ´Component´ remove method.
 
 There are also other implementations:
 
--- a/lib/components/component.dart
+++ b/lib/components/component.dart
@@ -32,12 +32,14 @@ abstract class Component {
   /// The smaller the priority, the sooner your component will be updated/rendered.
   /// It can be any integer (negative, zero, or positive).
   /// If two components share the same priority, they will probably be drawn in the order they were added.
-  final int priority = 0;
+  final int priority;
 
   /// Whether this component should be removed or not.
   ///
   /// It will be checked once per component per tick, and if it is true, [BaseGame] will remove it.
   bool shouldRemove = false;
+
+  Component({this.priority = 0});
 
   /// This method is called periodically by the game engine to request that your component updates itself.
   ///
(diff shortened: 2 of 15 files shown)


doc/components.md
```diff
@@ -32,13 +32,13 @@ Every `Component` has a few other methods that you can optionally implement, tha
...
 
-The `destroy` method can be implemented to return true and warn the `BaseGame` that your object is marked for destruction, and it will be remove after the current update loop. It will then no longer be rendered or updated.
+The `shouldRemove` method can be implemented to return true and `BaseGame` will remove it before the next update loop. It will then no longer be rendered or updated. Note that `game.remove(Component c)` can also be used to remove components.
 
 The `isHUD` method can be implemented to return true (default false) to make the `BaseGame` ignore the `camera` for this element.
 
 The `onMount` method can be overridden to run initialization code for the component. When this method is called, BaseGame ensures that all the mixins which would change this component behaviour are already resolved.
 
-The `onDestroy` method can be overridden to run code before the component is removed from the game.
+The `onRemove` method can be overridden to run code before the component is removed from the game.
```
luanpotter, line 41: I think this also fixes a bug where onRemove was called twice on v1. I pointed this out on telegram. I think we can make it explicit that is desired here? maybe can even add a test for that

luanpotter (approved):
LGTM + 1 suggestion

doc/components.md
spydon: Good idea, more tests to the people!

spydon: Done! Did quite a lot of other changes too, so would be good if you could re-review.

lib/components/component.dart
```diff
@@ -34,38 +34,41 @@ abstract class Component {
   /// Use [Resizable] to save the gameSize in a component.
   void onGameResize(Vector2 gameSize) {}
 
+  /// Whether this component should be removed or not.
+  ///
+  /// It will be checked once per component per tick, and if it is true, [BaseGame] will remove it.
+  bool shouldRemove = false;
+
+  /// Remove the component from the game it is added to in the next tick
+  void remove() => shouldRemove = true;
```
erickzanardo, line 43: Should we make `shouldRemove` private since we have a method for marking it for removal?

lib/components/component.dart
```diff
@@ -34,38 +34,41 @@ abstract class Component {
...
+  bool isHud = false;
 
   /// Render priority of this component. This allows you to control the order in which your components are rendered.
   ///
   /// Components are always updated and rendered in the order defined by this number.
   /// The smaller the priority, the sooner your component will be updated/rendered.
   /// It can be any integer (negative, zero, or positive).
   /// If two components share the same priority, they will probably be drawn in the order they were added.
-  int priority() => 0;
+  int priority = 0;
```
erickzanardo, line 62: We need to think about how to handle the changing of this.

Changing this variable will not change the render order. That is why this was a method.

Honestly I don't have any solution for this but we need to think on something hahahaha 😅

lib/components/component.dart
spydon: I used `shouldRemove` just as a variable with the intention to replace the old `destroy()` method, so it's that value that `BaseGame` looks at when it is removing components from the component list.

lib/components/component.dart
spydon: Isn't this doing that? :)
https://github.com/flame-engine/flame/pull/527/files#diff-39ef6a651074ea090af849d06cafd156e3e49ddc3bcde638bb83b9d3a8d7069eR26

erickzanardo: As far as I rememeber, not really, that only executes when the element is added

spydon: Ah, it doesn't resort it later?

erickzanardo-creditas: Nope :(

That is why we have this card on the board actually: https://github.com/flame-engine/flame/projects/1#card-40463639

erickzanardo: Nope :(

That is why we have this card on the board actually: https://github.com/flame-engine/flame/projects/1#card-40463639

spydon: I just tried, it is the same behavior having it as a method as having it as a field, both of them only puts them in the correct position on insertion.

lib/components/component.dart
```diff
@@ -15,6 +15,30 @@ import '../extensions/vector2.dart';
...
+  /// HUD objects ignore the [BaseGame.camera] when rendered (so their position coordinates are considered relative to the device screen).
+  bool isHud = false;
+
+  /// Render priority of this component. This allows you to control the order in which your components are rendered.
+  ///
+  /// Components are always updated and rendered in the order defined by what this number is when the component is added to the game.
+  /// The smaller the priority, the sooner your component will be updated/rendered.
+  /// It can be any integer (negative, zero, or positive).
+  /// If two components share the same priority, they will probably be drawn in the order they were added.
+  final int priority = 0;
```
erickzanardo, line 35: I think this can't be initialized here right? This should be received on the constructor, otherwise this will be always zero, on all components.

Or am I missing something?

spydon: I was thinking that it should be overridden by subclasses, but it is probably better to have it in the constructor like you say

spydon: Just pushed a change with it.

erickzanardo: Fair enough! I think it makes sense, overriding this will be the easier way, adding on the constructor is a good addition, but I think it could be done on a different PR maybe

erickzanardo (approved):

renancaraujo (approved):

PR #573, opened 2020-12-14
Fix ConcurrentModificationException on add in onMount

spydon:
# Description

If `game.add` is called in `onMount` you get a `ConcurrentModificationException`.

## Type of change

# Checklist:

### DIFF
--- a/lib/game/base_game.dart
+++ b/lib/game/base_game.dart
@@ -129,9 +129,10 @@ class BaseGame extends Game with FPSCounter {
     });
     _removeLater.clear();
 
-    components.addAll(_addLater);
-    _addLater.forEach((component) => component.onMount());
+    final addNow = _addLater.toList(growable: false);
     _addLater.clear();
+    components.addAll(addNow);
+    addNow.forEach((component) => component.onMount());
     components.forEach((c) => c.update(t));
   }
 
--- a/lib/game/base_game.dart
+++ b/lib/game/base_game.dart
@@ -129,10 +129,12 @@ class BaseGame extends Game with FPSCounter {
     });
     _removeLater.clear();
 
-    final addNow = _addLater.toList(growable: false);
-    _addLater.clear();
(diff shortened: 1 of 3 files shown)


lib/game/base_game.dart
```diff
@@ -129,9 +129,10 @@ class BaseGame extends Game with FPSCounter {
     });
     _removeLater.clear();
 
-    components.addAll(_addLater);
-    _addLater.forEach((component) => component.onMount());
+    final addNow = _addLater.toList(growable: false);
```
erickzanardo, line 132: I have one doubt, and one suggestion about this.

Doubt: Will this add any perfomance overhead? If this do, we could check the _addLater length, and only copy the list if length > 0

Suggestion `addNow` and `addLater` got a little confusing here? Maybe `toAddComponents` or something like that. Maybe I am the problem so feel free to ignore this suggestion.

spydon: A list of length 0 should be extremely cheap to copy, I think it is better to have it like this for readability.

On the topic of readability, you are correct that the naming might confusing, not sure if `toAddComponents` are less confusing though? Let's see what the other reviewers think.

erickzanardo: > A list of length 0 should be extremely cheap to copy, I think it is better to have it like this for readability.
Not sure about that, on a game engine I would like to squeeze every bit of perfomance possible, and I don't think wrapping this small section of code in an `if` would impact that much on the perfomance.

 > not sure if toAddComponents is less confusing though?
I think it is less confusing, but I didn't liked it very much, but couldn't think on anything better 😅 hahaha

spydon: You're right, it actually got better when wrapping it. :) 
I thought it was just going to wrap a line, but it could wrap the whole thing.

erickzanardo (approved):
LGTM

luanpotter (approved):
LGTM

PR #576, opened 2020-12-15
[Suggestion] Rename the layer folder to layers to be consistent with other similar packages

luanpotter:
`game` I think it makes sense. as you typically only have on instance of game. but layer felt inconsistent with all the others. we could also change all to singular if there is some dart / flutter convention on that.


erickzanardo (approved):
Isn't there any docs that needs to be update as well? Also changelog is missing

But LGTM

spydon (approved):

renancaraujo (approved):

PR #579, opened 2020-12-18
Properly propagate onMount and onRemove to children

spydon:
# Description

Properly propagate onMount and onRemove to children

## Type of change

## Checklist:


lib/components/base_component.dart
```diff
@@ -80,6 +80,16 @@ abstract class BaseComponent extends Component {
     _children.forEach((child) => child.onGameResize(gameSize));
   }
 
+  @override
+  void onMount() {
```
erickzanardo, line 84: Add must call super here

lib/components/base_component.dart
```diff
@@ -80,6 +80,16 @@ abstract class BaseComponent extends Component {
...
+    children.forEach((child) => child.onMount());
+  }
+
+  @override
+  void onRemove() {
+    children.forEach((child) => child.onRemove());
```
erickzanardo, line 90: Here too

lib/components/base_component.dart
```diff
@@ -104,27 +114,25 @@ abstract class BaseComponent extends Component {
...
     }
 
-    final childOnLoadFuture = c.onLoad();
+    final childOnLoadFuture = child.onLoad();
     if (childOnLoadFuture != null) {
       await childOnLoadFuture;
     }
-    _children.add(c);
+    _children.add(child);
+    if (gameRef?.components?.contains(this) ?? false) {
```
erickzanardo, line 133: I am wandering, how come this can be null? Maybe only when the mixin is not used?

lib/game/game.dart
```diff
@@ -23,6 +24,9 @@ abstract class Game {
   final assets = AssetsCache();
   BuildContext buildContext;
 
+  /// The list of component that is related to the game
+  final Iterable<Component> components = const Iterable.empty();
```
erickzanardo, line 28: Is this change correct? Only basegame has support for components right?

lib/components/base_component.dart
spydon: When the child is added before the parent is added to the game it will be null.

erickzanardo: But is that a behavior that we want to allow? Maybe we could add an assert to make sure that the parent already exists in the game before children can be added?

lib/game/game.dart
spydon: It is due to the line on the other place where you commented on, since `HasGameRef` looks like this `HasGameRef<T extends Game>`. But maybe there is a better way to know whether `onMount` has already been run on a component?

spydon: Or maybe we should add `bool isMounted`, that might be useful for other things too?

lib/components/base_component.dart
spydon: I did have an assert that checked that, but since the parent -> child relationship is so general now I think we can allow it not to need to be prepared by the game? It is also mentioned in the comment on line 116.

lib/game/game.dart
erickzanardo: Seems like a good idea, I think the `bool isMounted` could be better than to have this list here.

Or maybe, we should change the mixin? It could be `HasGameRef<T extends BaseGame>`?

lib/components/base_component.dart
erickzanardo: yeah, I think it could. I was worried that the onMount/onLoad on children would never be called, but I think those would end up been called when the parent is added to the game?

lib/game/game.dart
spydon: Updated to `isMounted`

lib/components/component.dart
```diff
@@ -16,6 +17,9 @@ abstract class Component {
   /// HUD objects ignore the [BaseGame.camera] when rendered (so their position coordinates are considered relative to the device screen).
   bool isHud = false;
 
+  /// Whether this component is currently mounted on a game or not
+  bool isMounted = false;
```
spydon, line 21: What do you think about this name? Is `isAttached` a better name, since this is set to false when it is removed from the game? This is however more clear that it is set by `onMount`...

lib/components/base_component.dart
spydon: Exactly, they would

doc/examples/composability/lib/main.dart
```diff
@@ -0,0 +1,64 @@
...
+  void createChildren() {
+    // All positions here are in relation to the parent's position
+    final children = [
+      Square(Vector2(100, 100), Vector2(50, 50), angle: 2),
+      Square(Vector2(160, 100), Vector2(50, 50), angle: 3),
+      Square(Vector2(170, 150), Vector2(50, 50), angle: 4),
+      Square(Vector2(70, 200), Vector2(50, 50), angle: 5),
+    ];
+
+    children.forEach((child) => addChild(child, gameRef: gameRef));
```
erickzanardo, line 43: Nit: `gameRef` here could be omitted right?

lib/components/component.dart
erickzanardo: tbh, IMO `isMounted` is fine, but I have nothing against `isAttached` as well hahha

erickzanardo (approved):
left one additional nit, but LGTM already

lib/components/base_component.dart
erickzanardo: Fair enough then

lib/components/component.dart
wolfenrain: I personally would say `isMounted`, considering that is what it is indicating. It is either mounted to a game or not. And the methods that can influence it are also using the same naming convention :-).

I do have a quick wonder, what would happen if I as a user update this value? Would that do anything at all? If it doesn't should this not be a `getter`, and only let the system(flame) control the value?

wolfenrain (approved):

lib/components/component.dart
spydon: @wolfenrain good point, did that change now.

lib/components/component.dart
```diff
@@ -16,6 +17,10 @@ abstract class Component {
   /// HUD objects ignore the [BaseGame.camera] when rendered (so their position coordinates are considered relative to the device screen).
   bool isHud = false;
 
+  /// Whether this component is currently mounted on a game or not
```
wolfenrain, line 20: I would move this dartdoc to the getter. Otherwise IDE's wont display the dartdoc comment considering this is a private property :-)

wolfenrain (changes requested):

lib/components/component.dart
spydon: Done!

wolfenrain (approved):

PR #583, opened 2020-12-20
Effect shouldn't affect non-used properties

spydon:
# Description

Right now an effect resets all the properties of a position component when it is done, it should only reset the once it has affected.
For example a ScaleEffect will reset the position of the PositionComponent when it is done.

## Type of change

## Checklist:

### DIFF
--- a/lib/effects/effects.dart
+++ b/lib/effects/effects.dart
@@ -168,7 +168,7 @@ abstract class PositionComponentEffect
       if (originalAngle != endAngle) {
         component?.angle = angle;
       }
-      if (originalSize != originalSize) {
+      if (originalSize != endSize) {
         component?.size?.setFrom(size);
       }
     }
--- a/lib/effects/effects.dart
+++ b/lib/effects/effects.dart
@@ -157,22 +157,31 @@ abstract class PositionComponentEffect
     endSize = component.size.clone();
   }
 
-  @override
-  void setComponentToOriginalState() {
+  /// Only change the parts of the component that is affected by the
+  /// effect, and only set the state if it is the root effect (not part of
+  /// another effect, like children of a CombinedEffect or SequenceEffect).
+  void _setComponentState(Vector2 position, double angle, Vector2 size) {
     if (isRootEffect()) {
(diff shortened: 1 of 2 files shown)


lib/effects/effects.dart
```diff
@@ -157,22 +157,31 @@ abstract class PositionComponentEffect
...
-  void setComponentToOriginalState() {
+  /// Only change the parts of the component that is affected by the
+  /// effect, and only set the state if it is the root effect (not part of
+  /// another effect, like children of a CombinedEffect or SequenceEffect).
+  void _setComponentState(Vector2 position, double angle, Vector2 size) {
     if (isRootEffect()) {
-      component?.position?.setFrom(originalPosition);
-      component?.angle = originalAngle;
-      component?.size?.setFrom(endSize);
+      if (originalPosition != endPosition) {
```
erickzanardo, line 165: This if shouldn't include the `position` variable?

lib/effects/effects.dart
```diff
@@ -157,22 +157,31 @@ abstract class PositionComponentEffect
...
+        component?.position?.setFrom(position);
+      }
+      if (originalAngle != endAngle) {
```
erickzanardo, line 168: Same question here

lib/effects/effects.dart
```diff
@@ -157,22 +157,31 @@ abstract class PositionComponentEffect
...
+        component?.angle = angle;
+      }
+      if (originalSize != originalSize) {
```
erickzanardo, line 171: This if will never be false :P

lib/effects/effects.dart
spydon: No, it is just checking whether this effect made any changes on the component.
The position variable will be either `endPosition` or `originalPosition`, but this way I don't have to duplicate the code.

lib/effects/effects.dart
spydon: Good catch, should be `endSize`!

erickzanardo (approved):
LGTM

luanpotter (approved):
LGTM

PR #593, opened 2020-12-28
Fix issue in BaseComponent where addChild results in a TypeError

fbcouch:
# Description

This adds the compare function to the children getter on `BaseComponent` as it throws a TypeError since Components are not themselves comparable without that. I didn't log an issue for it since this is a simple change.

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

If something is unclear, please submit the PR anyways and ask about what you thought was unclear.

- [x] This branch is based on the latest `master`
- [x] I have added an entry under `[next]` in `CHANGELOG.md`
- [x] I have formatted my code with `flutter format`
- [x] I have made corresponding changes to the documentation
- [x] I have added examples for new features in `doc/examples`
- [x] The continuous integration (CI) is passing

PR #594, opened 2020-12-28
BaseComponent should return immutable chlidren iterable

spydon:
# Description

The BaseComponent children getter should return an immutable data structure.

## Type of change

## Checklist:

### DIFF
--- a/lib/components/base_component.dart
+++ b/lib/components/base_component.dart
@@ -23,7 +23,7 @@ abstract class BaseComponent extends Component {
   final OrderedSet<Component> _children =
       OrderedSet(Comparing.on((c) => c.priority));
 
-  Iterator<Component> get children => _children.iterator;
+  List<Component> get children => _children.toList(growable: false);
 
   /// This is set by the BaseGame to tell this component to render additional debug information,
   /// like borders, coordinates, etc.
--- a/lib/components/base_component.dart
+++ b/lib/components/base_component.dart
@@ -1,3 +1,4 @@
+import 'dart:collection';
 import 'dart:ui';
 
 import 'package:meta/meta.dart';
@@ -23,7 +24,13 @@ abstract class BaseComponent extends Component {
   final OrderedSet<Component> _children =
       OrderedSet(Comparing.on((c) => c.priority));
 
-  List<Component> get children => _children.toList(growable: false);
+  /// The children list shouldn't be modified directly, that is why an
+  /// [UnmodifiableListView] is used. If you want to add children use the
+  /// addChild method, and if you want to propagate something to the children
+  /// use the propagateToChildren method.
+  UnmodifiableListView<Component> get children {
+    return UnmodifiableListView(_children);
+  }
 
   /// This is set by the BaseGame to tell this component to render additional debug information,
   /// like borders, coordinates, etc.
--- a/lib/components/base_component.dart
+++ b/lib/components/base_component.dart
@@ -23,7 +23,7 @@ abstract class BaseComponent extends Component {
   final OrderedSet<Component> _children =
       OrderedSet(Comparing.on((c) => c.priority));
 
-  List<Component> get children => _children.toList(growable: false);
+  Iterator<Component> get children => _children.iterator;
 
   /// This is set by the BaseGame to tell this component to render additional debug information,
   /// like borders, coordinates, etc.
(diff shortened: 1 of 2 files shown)


spydon:
Fixes the issue that was mentioned in #593

erickzanardo (approved):
LGTM

But I still kind of think that we should drop that getter and add a method to iterate over the children instead, I am a little worried about the perfomance hit that we could face if the programmer used that getter on an update method for example.

spydon:
> LGTM
> 
> But I still kind of think that we should drop that getter and add a method to iterate over the children instead, I am a little worried about the perfomance hit that we could face if the programmer used that getter on an update method for example.

What if we just return an iterator to the OrderedSet?

erickzanardo:
> What if we just return an iterator to the OrderedSet?

Hmmm, that is interesting. Do you know if it creates new instances when `.iterator` is called? If so, I think we will still have a perfomance hit (but maybe less). If not, thats brilliant,

spydon:
> > What if we just return an iterator to the OrderedSet?
> 
> Hmmm, that is interesting. Do you know if it creates new instances when `.iterator` is called? If so, I think we will still have a perfomance hit (but maybe less). If not, thats brilliant,

Will the user not be quite confused getting an iterator back though? 
It just creates a new iterator for the set, it should be a very cheap operation.

erickzanardo:
> Will the user not be quite confused getting an iterator back though?
Not sure I understood, you think this method returning an iterator will be confusing?

> It just creates a new iterator for the set, it should be a very cheap operation.
Indeed, SGTM then!

spydon:
> LGTM
> 
> But I still kind of think that we should drop that getter and add a method to iterate over the children instead, I am a little worried about the perfomance hit that we could face if the programmer used that getter on an update method for example.

We have the `propagateToChildren` method, I guess that could be used for the use-cases where you want to run something on the children. But I'm thinking that there are other use-cases, when you'd want to maybe filter your children and do something with them for example.

> Not sure I understood, you think this method returning an iterator will be confusing?

Yeah, exactly, but maybe they wouldn't? I don't think we use iterators anywhere else in the code base, but maybe we should in these kind of cases?

erickzanardo:
> Yeah, exactly, but maybe they wouldn't? I don't think we use iterators anywhere else in the code base, but maybe we should in these kind of cases?

I don't think it would be confusing, almost every method on collections like `where`, `map` returns iterators, so I think it will be pretty normal for a Flutter dev to use iterators, so I don't think I see any problems with that.

spydon:
> > Yeah, exactly, but maybe they wouldn't? I don't think we use iterators anywhere else in the code base, but maybe we should in these kind of cases?
> 
> I don't think it would be confusing, almost every method on collections like `where`, `map` returns iterators, so I think it will be pretty normal for a Flutter dev to use iterators, so I don't think I see any problems with that.

Heheh, this might not be a good idea, because I have already managed to confuse you. ;)
All those functions return and `Iterable`. not an iterator.

erickzanardo:
> Heheh, this might not be a good idea, because I have already managed to confuse you. ;)
> All those functions return and `Iterable`. not an iterator.

Oh no! hahahah indeed :(

wolfenrain (approved):
LGTM

lib/components/base_component.dart
```diff
@@ -23,7 +24,13 @@ abstract class BaseComponent extends Component {
   final OrderedSet<Component> _children =
       OrderedSet(Comparing.on((c) => c.priority));
 
-  List<Component> get children => _children.toList(growable: false);
+  /// The children list shouldn't be modified directly, that is why an
+  /// [UnmodifiableListView] is used. If you want to add children use the
+  /// addChild method, and if you want to propagate something to the children
+  /// use the propagateToChildren method.
```
erickzanardo, line 30: Wrap the propagateToChildren with [], so it will create a link on the docs

lib/components/base_component.dart
```diff
@@ -23,7 +24,13 @@ abstract class BaseComponent extends Component {
...
+  UnmodifiableListView<Component> get children {
+    return UnmodifiableListView<Component>(_children);
```
erickzanardo, line 32: Nit: I think this is small enough to be a one liner

erickzanardo (approved):
Some nits but LGTM

wolfenrain (approved):

lib/components/base_component.dart
spydon: It's a tiiiny bit too long unfortunately.

lib/components/base_component.dart
spydon: Fixed!

PR #598, opened 2020-12-29
Add a badge with our current RC as well alongside the current stable version

luanpotter:
This is what it would look like:

image

(just a suggestion, not sure if too cluttered)


erickzanardo (approved):

wolfenrain (commented):
The more the merrier

spydon (approved):
Nice!

PR #599, opened 2020-12-31
Bump versions and fix lint issues (v0.x)

luanpotter:
This is a new release for our v0 (stable) channel. This:

* updates audioplayers to fix some issues with notifications center
* fixes some lint issues on 0.28.0 that costs us points on pub
* I had to add `publish_to: none` to all our examples, that was being complained about on our local analyze

Remember that v0.x of Flame only works on Flutter stable because there was a Flutter breaking change (use Flame 1.0 for newer flutter channels).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
           java-version: '12.x'
       - uses: subosito/flutter-action@v1
         with:
-          channel: 'beta'
+          channel: 'stable'
       - run: flutter --version
       - run: flutter pub get
       - run: ./scripts/lint.sh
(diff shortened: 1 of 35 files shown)


.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
           java-version: '12.x'
       - uses: subosito/flutter-action@v1
         with:
-          channel: 'beta'
+          channel: 'stable'
```
luanpotter, line 16: changed this just for this branch. since pub uses stable we should make sure our build will be green on pub.

spydon (approved):

erickzanardo (approved):

PR #601, opened 2021-01-01
Improve error handling for onLoad

luanpotter:
I think errors should not be swallowed, that is a really bad practice. I was very against this when we first discussed, but now that I fell ill of the consequence myself (and took me longer than it should to realize what was going on), I have become vehemently against.

onLoad exceptions should be re-thrown.

This is what it looks like with this change:

image

You still get the option to provide an error handler but now the handler is aware of what the error was. If you do not, the exception is re-thrown as anyone would except.

Before if a handler was not provided (and it was optional) the exception was completely swallowed and not logged, seem, expressed anywhere. And you got a black screen from an empty game. Also even if you wanted to add a custom handler you would not know what went wrong, which made the custom error handler practically useless.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## [next]
+ - Improve error handling for the onLoad function
 
 ## 1.0.0-rc5
  - Option for overlays to be already visible on the GameWidget
--- a/lib/game/game_widget.dart
+++ b/lib/game/game_widget.dart
@@ -11,7 +11,11 @@ import 'game_render_box.dart';
 
 typedef GameLoadingWidgetBuilder = Widget Function(
   BuildContext,
-  bool error,
+);
+
+typedef GameErrorWidgetBuilder = Widget Function(
+  BuildContext,
+  Object error,
 );
 
 typedef OverlayWidgetBuilder<T extends Game> = Widget Function(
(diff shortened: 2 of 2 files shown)


wolfenrain (approved):
A CHANGELOG update is missing I believe, apart from that LGTM

luanpotter:
Good catch, fixed!

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## [next]
+ - Improve error handling for the onLoad functiomaya jama dioramabig fatn
```
wolfenrain, line 4: Either you just went "keyboard smash" or this is should be translated, either way I think it should be changed

wolfenrain (changes requested):

CHANGELOG.md
luanpotter: hahahaha that was an accidental control v from my google search :p

luanpotter: do recommend the big fat quiz of the year, 2020 edition just got out but they are all good

wolfenrain: Ah in that case, let me **not** pass that on my Google search haha

wolfenrain (approved):

spydon (approved):
LGTM

erickzanardo (approved):

PR #602, opened 2021-01-02
🚨 Make `scripts/lint.sh` work recursively in examples as intended

lig:


spydon (approved):
This is posix compliant, but the other one should work in most modern shells too right?
Probably worth merging this anyways!

lig:
> This is posix compliant, but the current syntax should work in most modern shells too right?
> Probably worth merging this anyways!

The script demands `bash` in the shebang.
 
In `bash` `**` means the same as `*` and there are examples on the second level already that `**` misses.

spydon:
> > This is posix compliant, but the current syntax should work in most modern shells too right?
> > Probably worth merging this anyways!
> 
> The script demands `bash` in the shebang.
>  
> In `bash` `**` means the same as `*` and there are examples on the second level already that `**` misses.

It works in bash if you have `shopt -s globstar ` activated, but since that is not default I'll go ahead and merge this.

spydon:
A changelog entry is missing though.

lig:
> A changelog entry is missing though.

Added a changelog entry

> It works in bash if you have `shopt -s globstar ` activated, but since that is not default I'll go ahead and merge this.

I've just tried to set the `globstar` option via the `BASHOPTS` var in the shebang. It doesn't work in Linux. Thus, that's not an option.

spydon:
> > A changelog entry is missing though.
> 
> Added a changelog entry
> 
> > It works in bash if you have `shopt -s globstar ` activated, but since that is not default I'll go ahead and merge this.
> 
> I've just tried to set the `globstar` option via the `BASHOPTS` var in the shebang. It doesn't work in Linux. Thus, that's not an option.

It works for me in bash 5.0.17, and it should work if you don't have an ancient version of bash. But that is just a side discussion, I think your solution is better anyways.

spydon (approved):

wolfenrain:
> A changelog entry is missing though.

Is a CHANGELOG entry needed for changes that are not relevant to the end user? Because the CHANGELOG is shown on pub.dev, and people tend to use that to see if there are breaking changes, bug fixes and that kind of stuff.

This is just an internal script we use for the CI/CD

spydon:
> > A changelog entry is missing though.
> 
> Is a CHANGELOG entry needed for changes that are not relevant to the end user? Because the CHANGELOG is shown on pub.dev, and people tend to use that to see if there are breaking changes, bug fixes and that kind of stuff.
> 
> This is just an internal script we use for the CI/CD

In my opinion everything should have a changelog entry, nothing is fully internal in an open source project. And it would be easier to track down regression on CI/CD if we also keep entries for that.

wolfenrain (approved):

luanpotter (approved):
LGTM

PR #549, opened 2020-11-28
null safety

lig:
This is a work in progress.

`ordered_set` uses override to a branch in my fork until this is merged https://github.com/luanpotter/ordered_set/pull/5.

As for now, nothing is done in the flame code itself.

Todo:
- [x] prepare main `pubspec.yaml` file
- [x] migrate main dependencies
- [x] migrate main code
- [x] prepare examples `pubspec.yaml` files
- [x] migrate examples dependencies
- [x] migrate examples code
- [x] ensure CI passes
- [x] changelog entry
- [x] rebase on `master` with null safety backports merged in 
- [x] Dashbook null safety release

PR #603, opened 2021-01-02
🏷 Backport some code improvements from the `null-safety` branch

lig:
As suggested by @spydon (https://github.com/flame-engine/flame/pull/549#issuecomment-753447386), backporting some code changes made in the null safety PR (https://github.com/flame-engine/flame/pull/549)

- [x] changelog
- [x] CI passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## [next]
+ - Code improvements and preparing APIs to null-safety
  - BaseComponent removes children marked as shouldRemove during update
  - Use `find` instead of `globstar` pattern in `scripts/lint.sh` as the later isn't enabled by default in bash
  - Fixes aseprite constructor bug
--- a/doc/examples/effects/infinite_effects/lib/main.dart
+++ b/doc/examples/effects/infinite_effects/lib/main.dart
@@ -21,18 +21,19 @@ void main() async {
   );
 }
 
+const green = Color(0xAA338833);
+const red = Color(0xAA883333);
+const orange = Color(0xAABB6633);
+
 class MyGame extends BaseGame with TapDetector {
   Square greenSquare;
   Square redSquare;
   Square orangeSquare;
 
-  MyGame() {
-    final green = Paint()..color = const Color(0xAA338833);
-    final red = Paint()..color = const Color(0xAA883333);
-    final orange = Paint()..color = const Color(0xAABB6633);
-    greenSquare = Square(green, Vector2.all(100));
-    redSquare = Square(red, Vector2.all(200));
-    orangeSquare = Square(orange, Vector2(200, 400));
+  MyGame()
+      : greenSquare = Square(Paint()..color = green, Vector2.all(100)),
+        redSquare = Square(Paint()..color = red, Vector2.all(200)),
+        orangeSquare = Square(Paint()..color = orange, Vector2(200, 400)) {
     add(greenSquare);
     add(redSquare);
     add(orangeSquare);
--- a/doc/examples/joystick/lib/player.dart
+++ b/doc/examples/joystick/lib/player.dart
@@ -25,14 +25,15 @@ class Player extends Component implements JoystickListener {
 
   @override
   void render(Canvas canvas) {
-    if (_rect != null) {
-      canvas.save();
-      canvas.translate(_rect.center.dx, _rect.center.dy);
-      canvas.rotate(radAngle == 0.0 ? 0.0 : radAngle + (pi / 2));
-      canvas.translate(-_rect.center.dx, -_rect.center.dy);
-      canvas.drawRect(_rect, _paint);
-      canvas.restore();
+    if (_rect == null) {
+      return;
     }
+    canvas.save();
+    canvas.translate(_rect.center.dx, _rect.center.dy);
+    canvas.rotate(radAngle == 0.0 ? 0.0 : radAngle + (pi / 2));
+    canvas.translate(-_rect.center.dx, -_rect.center.dy);
+    canvas.drawRect(_rect, _paint);
+    canvas.restore();
   }
 
   @override
@@ -79,12 +80,15 @@ class Player extends Component implements JoystickListener {
     final double nextX = (currentSpeed * dtUpdate) * cos(radAngle);
     final double nextY = (currentSpeed * dtUpdate) * sin(radAngle);
 
+    if (_rect == null) {
+      return;
+    }
+
     final Offset diffBase = Offset(
           _rect.center.dx + nextX,
           _rect.center.dy + nextY,
         ) -
         _rect.center;
-
     _rect = _rect.shift(diffBase);
   }
 }
--- a/doc/examples/widgets/lib/main.dart
+++ b/doc/examples/widgets/lib/main.dart
@@ -108,8 +108,9 @@ void main() async {
           child: SpriteWidget(
             sprite: shieldSprite,
             anchor: parseAnchor(
-              ctx.listProperty('anchor', 'Anchor.center', anchorOptions),
-            ),
+                  ctx.listProperty('anchor', 'Anchor.center', anchorOptions),
+                ) ??
+                Anchor.topLeft,
           ),
         ),
       );
@@ -134,8 +135,9 @@ void main() async {
             animation: _animation,
             playing: ctx.boolProperty('playing', true),
             anchor: parseAnchor(
-              ctx.listProperty('anchor', 'Anchor.center', anchorOptions),
-            ),
+                  ctx.listProperty('anchor', 'Anchor.center', anchorOptions),
+                ) ??
+                Anchor.topLeft,
           ),
         ),
       );
--- a/lib/components/joystick/joystick_action.dart
+++ b/lib/components/joystick/joystick_action.dart
@@ -34,11 +34,11 @@ class JoystickAction {
   bool _dragging = false;
   Sprite _spriteAction;
   Offset _dragPosition;
-  Paint _paintBackground;
-  Paint _paintAction;
-  Paint _paintActionPressed;
+  final Paint _paintBackground;
+  final Paint _paintAction;
+  final Paint _paintActionPressed;
   JoystickController _joystickController;
-  double _sizeBackgroundDirection;
+  final double _sizeBackgroundDirection;
   DragEvent _currentDragEvent;
   double _tileSize;
 
@@ -55,9 +55,17 @@ class JoystickAction {
     this.align = JoystickActionAlign.BOTTOM_RIGHT,
     this.opacityBackground = 0.5,
     this.opacityKnob = 0.8,
-  }) {
-    _spriteAction = sprite;
-    _sizeBackgroundDirection = sizeFactorBackgroundDirection * size;
+  })  : _spriteAction = sprite,
+        _sizeBackgroundDirection = sizeFactorBackgroundDirection * size,
+        _paintBackground = Paint()
+          ..color = color.withOpacity(opacityBackground)
+          ..style = PaintingStyle.fill,
+        _paintAction = Paint()
+          ..color = color.withOpacity(opacityKnob)
+          ..style = PaintingStyle.fill,
+        _paintActionPressed = Paint()
+          ..color = color.withOpacity(opacityBackground)
+          ..style = PaintingStyle.fill {
     _tileSize = _sizeBackgroundDirection / 2;
   }
 
@@ -91,25 +99,6 @@ class JoystickAction {
       center: Offset(dx, dy),
       radius: _sizeBackgroundDirection / 2,
     );
-
-    if (spriteBackgroundDirection == null) {
-      _paintBackground = Paint()
-        ..color = color.withOpacity(opacityBackground)
-        ..style = PaintingStyle.fill;
-    }
-
-    if (sprite == null) {
-      _paintAction = Paint()
-        ..color = color.withOpacity(opacityKnob)
-        ..style = PaintingStyle.fill;
-    }
-
-    if (spritePressed == null) {
-      _paintActionPressed = Paint()
-        ..color = color.withOpacity(opacityBackground)
-        ..style = PaintingStyle.fill;
-    }
-
     _dragPosition = _rectAction.center;
   }
 
@@ -128,7 +117,7 @@ class JoystickAction {
   }
 
   void update(double dt) {
-    if (_dragging) {
+    if (_rectBackgroundDirection != null && _dragging) {
       final double _radAngle = atan2(
         _dragPosition.dy - _rectBackgroundDirection.center.dy,
         _dragPosition.dx - _rectBackgroundDirection.center.dx,
@@ -149,13 +138,15 @@ class JoystickAction {
       final double nextY = dist * sin(_radAngle);
       final Offset nextPoint = Offset(nextX, nextY);
 
-      final Offset diff = Offset(
-            _rectBackgroundDirection.center.dx + nextPoint.dx,
-            _rectBackgroundDirection.center.dy + nextPoint.dy,
-          ) -
-          _rectAction.center;
+      if (_rectAction != null) {
+        final Offset diff = Offset(
+              _rectBackgroundDirection.center.dx + nextPoint.dx,
+              _rectBackgroundDirection.center.dy + nextPoint.dy,
+            ) -
+            _rectAction.center;
 
-      _rectAction = _rectAction.shift(diff);
+        _rectAction = _rectAction.shift(diff);
+      }
 
       final double _intensity = dist / _tileSize;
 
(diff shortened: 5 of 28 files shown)


lig:
I'm going to add some entries to the changelog for this one.

lig:
> I'm going to add some entries to the changelog for this one.

done

lib/effects/combined_effect.dart
```diff
@@ -47,8 +47,8 @@ class CombinedEffect extends PositionComponentEffect {
           effect.originalAngle != effect.endAngle ? effect.endAngle : endAngle;
       endSize =
           effect.originalSize != effect.endSize ? effect.endSize : endSize;
-      peakTime = max(peakTime ?? 0,
-          effect.iterationTime + offset * effects.indexOf(effect));
+      peakTime = max(
+          peakTime, effect.iterationTime + offset * effects.indexOf(effect));
```
luanpotter, line 51:
```suggestion
      peakTime = max(
          peakTime,
          effect.iterationTime + offset * effects.indexOf(effect),
       );
```

lib/sprite.dart
```diff
@@ -8,7 +8,7 @@ import 'palette.dart';
 class Sprite {
   Paint paint = BasicPalette.white.paint;
   Image image;
-  Rect src;
+  Rect src = Rect.zero;
```
luanpotter, line 11: hum should the default be the whole image?

lib/particles/computed_particle.dart
```diff
@@ -19,7 +19,6 @@ class ComputedParticle extends Particle {
   ComputedParticle({
     @required this.renderer,
     double lifespan,
-    Duration duration,
```
luanpotter, line 22: nice cleanup!

lib/gestures.dart
```diff
@@ -13,6 +13,8 @@ mixin MultiTouchTapDetector on Game {
 class DragEvent extends Drag {
   Offset initialPosition;
```
luanpotter, line 14: this can be final right?

lib/game/game_widget.dart
```diff
@@ -175,7 +176,7 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
 
   void _checkOverlays(Set<String> overlays) {
     overlays.forEach((overlayKey) {
-      assert(widget.overlayBuilderMap.containsKey(overlayKey),
+      assert(widget.overlayBuilderMap?.containsKey(overlayKey) ?? false,
           "A non mapped overlay has been added: $overlayKey");
     });
```
luanpotter, line 181:
```suggestion
      assert(
          widget.overlayBuilderMap?.containsKey(overlayKey) ?? false,
          'A non mapped overlay has been added: $overlayKey',
       );
```

lib/effects/sequence_effect.dart
```diff
@@ -36,9 +36,6 @@ class SequenceEffect extends PositionComponentEffect {
   @override
   void initialize(PositionComponent component) {
     super.initialize(component);
```
luanpotter, line 38: IIRC @spydon had a reason why this needed to be on initialize and not on the constructor/inits. not sure now what it was, just flagging for context

lib/effects/move_effect.dart
```diff
@@ -96,7 +96,11 @@ class MoveEffect extends SimplePositionComponentEffect {
     }
     final double totalPathLength = isAlternating ? pathLength * 2 : pathLength;
     speed ??= totalPathLength / duration;
+
+    /// `duration` is not null when speed is null
     duration ??= totalPathLength / speed;
+
+    /// `speed` is always not null here already
```
luanpotter, line 103: since these are within the method body, I think they can be regular (`//`) comments (not dartdoc)

lib/components/sprite_component.dart
```diff
@@ -34,6 +34,9 @@ class SpriteComponent extends PositionComponent {
   @override
   void render(Canvas canvas) {
     super.render(canvas);
+    if (sprite == null) {
+      return;
+    }
     sprite.render(
```
luanpotter, line 40: how about `sprite?.render`?

luanpotter: should we even allow null sprites here?

lib/components/parallax_component.dart
```diff
@@ -93,10 +94,12 @@ class ParallaxLayer {
     // Size of the area to paint the images on
     final Vector2 paintSize = count..multiply(_imageSize);
     _paintArea = paintSize.toRect();
+
+    sized = true;
   }
 
   void update(Vector2 delta) {
-    if (!loaded()) {
+    if (!sized) {
```
luanpotter, line 102: what was wrong with the loaded() method?

lib/components/joystick/joystick_action.dart
```diff
@@ -149,13 +137,15 @@ class JoystickAction {
...
-            _rectBackgroundDirection.center.dx + nextPoint.dx,
-            _rectBackgroundDirection.center.dy + nextPoint.dy,
-          ) -
-          _rectAction.center;
+      if (_rectAction != null) {
+        final Offset diff = Offset(
+              _rectBackgroundDirection.center.dx + nextPoint.dx,
+              _rectBackgroundDirection.center.dy + nextPoint.dy,
+            ) -
+            _rectAction.center;
```
luanpotter, line 145: existing code, but this could be:

```dart
final nextPoint = Vector2(nextX, nextY);
// ...
final Offset diff = _rectBackgroundDirection.center.toVector2() + nextPoint - _rectAction.center;
_rectAction = _rectAction.shift(diff.toOffset());
```

doc/examples/widgets/lib/main.dart
```diff
@@ -134,8 +135,9 @@ void main() async {
             animation: _animation,
             playing: ctx.boolProperty('playing', true),
             anchor: parseAnchor(
-              ctx.listProperty('anchor', 'Anchor.center', anchorOptions),
-            ),
+                  ctx.listProperty('anchor', 'Anchor.center', anchorOptions),
+                ) ??
+                Anchor.topLeft,
```
luanpotter, line 140: I think parseAnchor should just throw instead of returning null

luanpotter:
This is excellent, @lig ! Thanks for helping with this migration, hopefully we can get this merged so the diff for the actual non-null branch is minimal :)

Left some minor comments!

lib/sprite.dart
lig: It is. see `srcSize`/`srcPosition` getters and setters. This default allows those to work as intended while avoiding mind-bending null checks on src.

lib/gestures.dart
lig: You are right. The analyzer didn't see that as it is not private.

lib/components/sprite_component.dart
erickzanardo: I also prefer `sprite?.render`

> should we even allow null sprites here?

Since we have the `onLoad` method on components now we can allow sprite to be lazy loaded

lib/effects/sequence_effect.dart
lig: I would be happy to revert this adding an appropriate comment to the code.

lib/components/sprite_component.dart
lig: > should we even allow null sprites here?

`SpriteComponent` has a default empty constructor defined. It looks like it could be just a do-nothing sprite.

lib/components/parallax_component.dart
lig: See https://github.com/flame-engine/flame/pull/549#discussion_r550809473 and the corresponding changelog entry

lib/components/joystick/joystick_action.dart
lig: Let's change this afterward. Created #606.

doc/examples/widgets/lib/main.dart
lig: #607

lib/effects/sequence_effect.dart
lig: @spydon could you please elaborate on this? It's about `_currentIndex` and `_driftModifier` defaults in `SequenceEffect`. Thanks!

lib/effects/effects.dart
```diff
@@ -43,13 +43,11 @@ abstract class ComponentEffect<T extends BaseComponent> {
...
+    Curve curve,
     this.onComplete,
-  }) {
-    isInfinite = _initialIsInfinite;
-    isAlternating = _initialIsAlternating;
-    curve ??= Curves.linear;
-  }
+  })  : isInfinite = _initialIsInfinite,
+        isAlternating = _initialIsAlternating,
+        curve = curve ?? Curves.linear;
```
spydon, line 50: Is this line needed now when the default is moved up to the named arguments?

lib/effects/sequence_effect.dart
spydon: It needs to be there since `reset` is using `initialize` if I remember correctly.

lib/effects/effects.dart
lig: In the `null-safety` branch `curve` argument os nullable. This effectively means that it could receive `null` and default value won't be applied. Thus, we provide `Curves.linear` as a value to replace the passed `null` value. If the value of `curve` is not passed at all it defaults to `null` and the same code is in effect again.

lig: Actually, here is another issue. I would rather add an assert for `isRelative` here as it could be passed as `null` in this before-null-safety code and that disables the default value `false` defined for it.

lib/effects/sequence_effect.dart
lig: > It needs to be there since `reset` is using `initialize` if I remember correctly.

I've added `_initialIndex` and `_initialDriftModifier` constants. IMO, this way the code looks more clear in its intent now.

lib/effects/effects.dart
lig: > Actually, here is another issue. I would rather add an assert for `isRelative` here as it could be passed as `null` in this before-null-safety code and that disables the default value `false` defined for it.

Added the assert for `isRelative`

lig:
A couple more changes:

 - Never return `null` from `RotatingParticle.angle` getter
 - Provide a default value for `TextBoxComponent._gameSize` forcing it to never be `null`

CHANGELOG.md
```diff
@@ -1,6 +1,24 @@
 # CHANGELOG
 
 ## [next]
+ - Never return `null` from `RotatingParticle.angle` getter
```
erickzanardo, line 4: I am not sure if we should have some much entries on the CHANGELOG for this PR, from the use perspective, it can be quite boring to read the changes in such details. I think we could an entry which captures the essence of this PR, something like: `- Code improvements and preparing APIs to null-safety`

doc/examples/joystick/lib/player.dart
```diff
@@ -25,14 +25,15 @@ class Player extends Component implements JoystickListener {
...
   @override
   void render(Canvas canvas) {
-    if (_rect != null) {
-      canvas.save();
-      canvas.translate(_rect.center.dx, _rect.center.dy);
-      canvas.rotate(radAngle == 0.0 ? 0.0 : radAngle + (pi / 2));
-      canvas.translate(-_rect.center.dx, -_rect.center.dy);
-      canvas.drawRect(_rect, _paint);
-      canvas.restore();
+    if (_rect == null) {
```
erickzanardo, line 28: this isn't changing much, any reason for this?

doc/examples/widgets/lib/main.dart
```diff
@@ -39,7 +39,7 @@ Anchor parseAnchor(String name) {
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
 
-  final dashbook = Dashbook();
+  final Dashbook dashbook = Dashbook();
```
erickzanardo, line 42: We don't have any lint rule enforcing this, but we went on the past removing all types that weren't necessary, so here we should keep the former code.

doc/examples/widgets/lib/main.dart
```diff
@@ -108,8 +108,9 @@ void main() async {
           child: SpriteWidget(
             sprite: shieldSprite,
             anchor: parseAnchor(
-              ctx.listProperty('anchor', 'Anchor.center', anchorOptions),
-            ),
+                  ctx.listProperty('anchor', 'Anchor.center', anchorOptions),
+                ) ??
+                Anchor.topLeft,
```
erickzanardo, line 113: This isn't really necessary, the `Anchor.center` value on the listProperty method is already the default value, so there is no way this method would return null

doc/examples/widgets/lib/main.dart
```diff
@@ -134,8 +135,9 @@ void main() async {
             animation: _animation,
             playing: ctx.boolProperty('playing', true),
             anchor: parseAnchor(
-              ctx.listProperty('anchor', 'Anchor.center', anchorOptions),
-            ),
+                  ctx.listProperty('anchor', 'Anchor.center', anchorOptions),
```
erickzanardo, line 138: Same comment here, no way this can return null

lib/components/joystick/joystick_utils.dart
```diff
@@ -11,6 +11,8 @@ class JoystickUtils {
     }
 
     if (sprite == null) {
+      assert(paint != null);
```
erickzanardo, line 14: We should add a message here alongside the assert

lib/components/parallax_component.dart
```diff
@@ -50,8 +50,9 @@ class ParallaxLayer {
   }
 
   bool loaded() => _image != null;
+  bool sized = false;
```
erickzanardo, line 53: This whole component will be refactored to match the Standard for asset loading that we are implementing on other features, so this will probably not be necessary in the future.

lib/effects/effects.dart
```diff
@@ -170,13 +169,13 @@ abstract class PositionComponentEffect
   /// another effect, like children of a CombinedEffect or SequenceEffect).
   void _setComponentState(Vector2 position, double angle, Vector2 size) {
     if (isRootEffect()) {
-      if (modifiesPosition) {
+      if (modifiesPosition && position != null) {
```
erickzanardo, line 172: Maybe we should add assert, instead of just silently doing nothing?

lig: > Maybe we should add assert, instead of just silently doing nothing?

yeah, this makes sense

doc/examples/widgets/lib/main.dart
lig: This was different in the `null-safety` branch. Not sure why. Reverting in both.

doc/examples/widgets/lib/main.dart
lig: This is necessary because this addresses a possible `null` value returning from `parseAnchor`. See #607 which is extracted from here to change its logic in a separate changeset.

doc/examples/widgets/lib/main.dart
lig: Same reason here.

doc/examples/widgets/lib/main.dart
erickzanardo: Oh sorry, I missread this, I though the `??` was for `ctx.listProperty`, it is for `parseAnchor` so this makes sense.

doc/examples/joystick/lib/player.dart
lig: This looks more consistent with the check in `moveFromAngle` below and other code around the project.

And this follows the `return early` best practice.

wolfenrain (approved):

erickzanardo (approved):

luanpotter (approved):
LGTM, thanks @lig

lib/effects/sequence_effect.dart
spydon: I would rather see this as a comment, this feels a bit over-engineered, especially since those values aren't used anywhere else. If I didn't write this code and I got here I would have to look up what the initial values were, instead of seeing them directly.

spydon (commented):
A few more comments, but overall it looks good!

lib/components/joystick/joystick_action.dart
```diff
@@ -34,14 +34,21 @@ class JoystickAction {
...
 
+  Paint get _paintBackground => Paint()
+    ..color = color.withOpacity(opacityBackground)
+    ..style = PaintingStyle.fill;
+  Paint get _paintAction => Paint()
+    ..color = color.withOpacity(opacityKnob)
+    ..style = PaintingStyle.fill;
+  Paint get _paintActionPressed => Paint()
+    ..color = color.withOpacity(opacityBackground)
+    ..style = PaintingStyle.fill;
```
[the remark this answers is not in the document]
luanpotter: can we just make them properites?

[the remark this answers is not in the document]
luanpotter: (I mean fields not methods/getters)

renancaraujo (approved):

lib/components/joystick/joystick_action.dart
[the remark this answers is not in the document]
lig: I wasn't sure what impact it will cause. Each render tick sounds horrible.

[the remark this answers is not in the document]
lig: Actually,  `color`, `opacityBackground`, and `opacityKnob` are all final. It looks like there is no harm in making those fields and set them in the constructor. If I'm not missing something.

doc/examples/effects/infinite_effects/lib/main.dart
```diff
@@ -26,13 +26,13 @@ class MyGame extends BaseGame with TapDetector {
...
-    greenSquare = Square(green, Vector2.all(100));
-    redSquare = Square(red, Vector2.all(200));
-    orangeSquare = Square(orange, Vector2(200, 400));
+  MyGame()
+      : greenSquare =
+            Square(Paint()..color = const Color(0xAA338833), Vector2.all(100)),
+        redSquare =
+            Square(Paint()..color = const Color(0xAA883333), Vector2.all(200)),
+        orangeSquare = Square(
+            Paint()..color = const Color(0xAABB6633), Vector2(200, 400)) {
```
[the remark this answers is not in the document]
lig: Moved color constants to the top level.

lib/components/joystick/joystick_action.dart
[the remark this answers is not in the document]
lig: Reverted to `_paintBackground`, `_paintAction`, and `_paintActionPressed` being fields. Made them final. Moved calculation of them to the initializer.

spydon (approved):
Lgtm!

PR #612, opened 2021-01-06
Added fallback support for the web on the `SpriteBatch` API

### LINKED ISSUE #561: SpriteBatch not work in Chrome
## Description

doc/examples/sprite_batch not work in flutter web (Chrome). Crashed on render method of SpriteBatch

## Development environment

Flame version:

branch develop

```
======== Exception caught by rendering library =====================================================
The following UnimplementedError was thrown during paint():
UnimplementedError

dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 216:49  throw_
lib/_engine/engine/html/canvas.dart 521:5                                     drawAtlas
packages/flame/sprite_batch.dart 73:11                                        render
packages/flame/components/sprite_batch_component.dart 21:17                   render
packages/flame/game/base_game.dart 113:6                                      renderComponent
...
The following RenderObject was being processed when the exception was fired: GameRenderBox#4579d
...  parentData: <none> (can use size)
...  constraints: BoxConstraints(w=500.0, h=966.0)
...  size: Size(500.0, 966.0)
RenderObject: GameRenderBox#4579d
  parentData: <none> (can use size)
  constraints: BoxConstraints(w=500.0, h=966.0)
  size: Size(500.0, 966.0)
==========================================================================================
```

Flutter doctor:

```
[✓] Flutter (Channel beta, 1.24.0-10.2.pre, on Mac OS X 10.15.7 19H15 darwin-x64, locale ru)
[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
[✓] Xcode - develop for iOS and macOS (Xcode 12.1.1)
[✓] IntelliJ IDEA Ultimate Edition (version 2020.3)
[✓] VS Code (version 1.51.1)
```

## Runtime

This issue is related to running on which platform? (Select all that apply)

 - [ ] Android :android:
 - [ ] iOS :apple:
 - [x] Web :browser:

## Minimal reproducible code (Required for bugs)

https://github.com/flame-engine/flame/tree/develop/doc/examples/sprite_batch

wolfenrain:
# Description

Improved the documentation on the `SpriteBatch` class and added fallback support for the web using black magic.

Fixes #561

## Type of change

## Checklist:


lib/sprite_batch.dart
```diff
@@ -29,16 +29,64 @@ class SpriteBatch {
...
   }) {
+    transform ??= defaultTransform;
     rects.add(rect);
-    transforms.add(transform ?? defaultTransform);
+    transforms.add(transform);
     colors.add(color ?? defaultColor);
+
+    /// Because `drawAtlas` is not supported on the web we also
+    /// build a `Matrix4` based on the given `RSTransform`.
+    matrices.add(Matrix4(
```
erickzanardo, line 55: Any reason for not doing that only on web?

wolfenrain: There is no computation taking place so adding an extra if statement would "impact" the performance more than what we would benefit from it IMHO, and this also ensures those lists will never be out of sync.

erickzanardo: I am wondering now which one would impact more, an `if` or instantiating something, but the last statement convinced me already.

wolfenrain: Well an `if` would have a lesser impact I guess, but it would be extra for the web then :-)

lib/sprite_batch.dart
```diff
@@ -70,14 +121,37 @@ class SpriteBatch {
...
+      for (var i = 0; i < matrices.length; i++) {
+        final matrix = matrices[i];
+        final rect = rects[i];
+        final color = colors[i];
+        paint..blendMode = blendMode ?? paint.blendMode ?? defaultBlendMode;
+
+        canvas
+          ..save()
+          ..transform(matrix.storage)
+          ..drawRect(Offset.zero & rect.size, Paint()..color = color)
```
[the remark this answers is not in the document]
wolfenrain: Refactored the whole API so it is not as messy, and all the important objects are now in one place. So consider it memoized.

lib/game/game.dart
```diff
@@ -130,6 +131,21 @@ abstract class Game {
     );
   }
 
+  /// Utility method to load and cache the image for a [SpriteBatch] based on its options
+  Future<SpriteBatch> loadSpriteBatch(
+    String path, {
+    Color defaultColor = const Color(0x00000000),
```
erickzanardo, line 137: Shouldn't this be opaque white?

lib/game/game.dart
```diff
@@ -130,6 +131,21 @@ abstract class Game {
     );
   }
 
+  /// Utility method to load and cache the image for a [SpriteBatch] based on its options
+  Future<SpriteBatch> loadSpriteBatch(
```
erickzanardo, line 135: Since SpriteBatch is not so common like Sprite and SpriteAnimations are, I think this could be into an extension inside the spritebatch.dart, just like it was done on Parallax.

Also, we should have the game cache being used here, with a fallback to the global cache when omitted.

wolfenrain: Let me update that :-)

wolfenrain: > Also, we should have the game cache being used here, with a fallback to the global cache when omitted.

Not sure what you meant with this, but I updated the PR to use the extension, so if you could have another look.

erickzanardo (approved):
LGTM

lib/game/game.dart
wolfenrain: This was the value it already was, fully transparent, because it is used as the background color for the sprites.

spydon (approved):

lib/sprite_batch.dart
```diff
@@ -2,66 +2,244 @@ import 'dart:ui';
...
+/// might be a performance hit on web when working with many batch items.
 class SpriteBatch {
+  /// List of all the existing batch items.
+  final _batchItems = <BatchItem>[];
+
+  /// The sources to use on the [atlas].
+  final _sources = <Rect>[];
+
+  /// The transforms that should be applied on the [_sources].
+  final _transforms = <RSTransform>[];
```
wolfenrain, line 112: After working a bit more with the API, I came to the conclusion that exposing both the `_transforms` and `_sources` list to the public can be quite useful for optimizing rendering.

My current usecase is that I reuse the transforms modify them slightly and then pass them back to the batch through `addTransform`. That way it won't have to do any kind of calculation, as my modification is just a simple `+` and `-`. 

And if we don't expose these lists then a user will have to precreate his own `RSTransform`s instead of using our helper method `add`.

erickzanardo: But is it safe to directly change the transformations?

wolfenrain: It is as safe as passing the wrong values to the addTransform method ;-). So I would not recommend it, but it does allow people to use the SpriteBatch more efficiently when they know what they are doing

Otherwise you'll always have the calculation overhead. And you do notice it with many transforms

erickzanardo: Sounds good to me then!

wolfenrain: Then I'll refactor this PR a little bit more soon

lib/sprite_batch.dart
```diff
@@ -1,67 +1,267 @@
...
-      translateX: offset.dx,
-      translateY: offset.dy,
-    );
-    addTransform(rect: rect, transform: transform, color: color);
+    RSTransform transform;
+
+    // If any of the transform arguments is different from the defaults,
+    // then we create one. This is to prevent unnecessary computations
+    // of the sine and cosine of the rotation.
+    if (scale != 1.0 ||
```
erickzanardo, line 242: This will add an empty transform when this if is false, is that desired?

erickzanardo (approved):
One small comment, but LGTM

lib/sprite_batch.dart
wolfenrain: It calls the `addTransform` and the `addTransform` already checks if the given transform(which is optional) is null, if so it will set a default one that does not do any calculations.

spydon (approved):
Nice!

doc/examples/sprite_batch/lib/main.dart
```diff
@@ -20,19 +21,19 @@ class MyGame extends BaseGame {
...
     spriteBatch.add(
-      rect: const Rect.fromLTWH(128 * 4.0, 128 * 4.0, 64, 128),
-      offset: const Offset(200, 200),
+      source: const Rect.fromLTWH(128 * 4.0, 128 * 4.0, 64, 128),
+      offset: Vector2(200, 200),
       color: Colors.greenAccent,
       scale: 2,
       rotation: pi / 9.0,
-      anchor: const Offset(64, 64),
+      anchor: Vector2(64, 64),
```
luanpotter, line 32: nit: Vector2.all ? (2x)

lib/sprite_batch.dart
```diff
@@ -70,14 +272,35 @@ class SpriteBatch {
...
-      colors,
-      blendMode ?? defaultBlendMode,
-      cullRect,
-      paint ?? defaultPaint,
-    );
+    paint ??= Paint();
+
+    if (kIsWeb) {
+      for (var i = 0; i < _batchItems.length; i++) {
+        final batchItem = _batchItems[i];
```
luanpotter, line 279: for each?

lib/sprite_batch.dart
```diff
@@ -1,67 +1,269 @@
...
+  final Paint paint;
+
+  BatchItem({
+    @required this.source,
+    @required this.transform,
+    @required this.color,
+  })  : assert(source != null),
+        assert(transform != null),
+        assert(color != null),
+        matrix = Matrix4(
```
luanpotter, line 68: despite our styleguide rule of one param per line, since this is a 4x4 matrix, this is a place I would personally open an exception and allow it to be grouped 4 per line.
```dart
matrix = Matrix4(
    1, 2, 3, 4,
    5, 6, 7, 8,
    9, 10, 11, 12,
    13, 14, 15, 16,
)
```
something like this.

not sure if it will look prettier though, might be worth a shot just to see

luanpotter (approved):
I don't know much about impl details but code overall LG with comments.

lib/sprite_batch.dart
wolfenrain: Will the formatter not mess with it? Also not sure How we can fit in the `scale` comment then ;(

lib/sprite_batch.dart
wolfenrain: I chose `for(var i < length)` syntax because on web that is the quicker variant, and well we need all the performance we can get when it comes to this workaround.

lib/sprite_batch.dart
wolfenrain: Ah the formatter does not mess with it, never mind. I'll look into making this work with the comment :-)

wolfenrain: So `flutter format` does mess with it, but I was able to "hack" it by adding a `//` behind each line.

lib/sprite_batch.dart
luanpotter: you mean compared to forEach right? you can do `for (final el in list)` which at least doesn't have you concerning about counters (since you don't otherwise use i).

lib/sprite_batch.dart
luanpotter: I like that!

but if you think the previous is clearer bc of the comment that is good as well

wolfenrain: I moved the comment to a const variable, IMHO still readable so I'll resolve it as is

lib/sprite_batch.dart
wolfenrain: Switched to the `for in` syntax

PR #613, opened 2021-01-06
Refactor ParallaxComponent

spydon:
# Description

Adding loading methods for the different `ParallaxComponent` parts and refactor how the delta velocity works

Fixes # (issue number, if there is one)

## Type of change

## Checklist:


doc/components.md
```diff
@@ -185,57 +185,76 @@ For a working example, check this [source file](/doc/examples/flare/lib/main_com
...
 ```dart
-  final images = [
-    ParallaxImage('mountains.jpg'),
-    ParallaxImage('forest.jpg'),
-    ParallaxImage('city.jpg'),
-  ];
-  this.bg = ParallaxComponent(images);
+  @override
+    final parallaxComponent = await ParallaxComponent.load(
```
erickzanardo, line 200: I feel that this will encourage people to use the global cache instead of the local game cache.

I think we could do two things to make that point clearer, let me list them here and we can discuss.

 - We could add the images parameter here
 - (My preferred) Since the Parallax isn't usual as Sprite/SpriteAnimation are I don't think we should have a `loadParallax` directly on the game class. But we could do that method as an extension on the parallax file, and here on the docs we could use that method instead.
 - (I am ok with this too) A mix of both, add the images parameter on the doc, and implement the extension method.

spydon: Is it maybe better to have those extension methods on `ParallaxComponent`, since that would be the one that would most commonly be imported when used with the game?

erickzanardo: Hmm, good question, I think it should be on Parallax, for the cases where the component system is not used...maybe we could have both? `loadParallax` and `loadParallaxComponent` so we have all covered?

spydon: Alright, like this? (Updated)

erickzanardo (approved):
LGTM

doc/components.md
erickzanardo: excellent

wolfenrain (approved):
LGTM

PR #634, opened 2021-01-17
Initial implementation of the Composition class.

wolfenrain:
# Description

Allowing for composing multiple images into a single composition. Allows for creating new images at runtime.

Also added fallback functionality for `decodeImageFromPixels`  for the web.

Fixes # (issue number, if there is one)

## Type of change

## Checklist:


lib/image_merger.dart
```diff
@@ -0,0 +1,142 @@
...
+      );
+
+      // If the rect is not fully contained inside the current outputRect,
+      // we have to resize the data list.
+      if (!(output.contains(destination.topLeft) &&
+          output.contains(destination.topRight) &&
+          output.contains(destination.bottomLeft) &&
+          output.contains(destination.bottomRight))) {
+        final newOutput = output.expandToInclude(destination);
+        final newData = Uint8List(
```
erickzanardo, line 82: Last time I needed to concatenate images I used PictureRecorder, which was way easier than manipulating the bytes themselves, any reason for you to have used bytes directly?

lib/util.dart
```diff
@@ -119,4 +121,62 @@ class Util {
...
+
+  /// Convert an array of pixel values into an [Image] object.
+  ///
+  /// The [pixels] parameter is the pixel data in the encoding described by
+  /// [PixelFormat.rgba8888], the encoding can't be changed to allow for web support.
+  ///
+  /// If you want the image to be decoded as it would be on the web you can set
+  /// [runAsWeb] to `true`. Keep in mind that it is slightly slower than the native
+  /// [ui.decodeImageFromPixels]. By default it is set to [kIsWeb].
+  Future<Image> decodeImageFromPixels(
```
erickzanardo, line 133: I think this method could be in the `Images` class instead, what do you think?

doc/util.md
```diff
@@ -25,6 +25,18 @@ This method sets the orientation of the whole application (effectively, also the
 
 If a finer control of the allowed orientations is required (without having to deal with `SystemChrome` directly), `setOrientation` (accepts a single `DeviceOrientation` as a parameter) and `setOrientations` (accepts a `List<DeviceOrientation>` for possible orientations) can be used.
 
+### `Flame.util.decodeImageFromPixels()`
+
+The [dart-ui decodeImageFromPixels](https://api.flutter.dev/flutter/dart-ui/decodeImageFromPixels.html) currently does not support the web platform. So if you are looking for a way to manipulate pixel data on the web this method can be used as a replacement for `dart-ui decodeImageFromPixels`:
+
+```dart
+final image = await Flame.util.decodeImageFromPixels(
```
erickzanardo, line 33: Just for making it easy to  understand, I would put the type here, so it is clear that this method returns an Image.

lib/image_merger.dart
wolfenrain: I originally wrote this as a command line tool, I already refactored it to the PictureRecorder in my local branch👌🏻

lib/util.dart
wolfenrain: Felt more a utility kind of method as we are basically wrapping existing functionality to allow for more possibilities, while the Images class is more meant to facilitate image loading stuff for Flame.

erickzanardo: I still think that this would fits better the images.dart, at the moment, I agree that our images.dart is only useful for loading images, but I think this could be easily expanded, and this method seems like a good one to start that expansion.

Also, we are doing some work for removing this file and moving the methods to places where makes more sense, so adding this method here seems like a move contrary to that work.

wolfenrain: You convinced me :-) I have updated the code.

doc/util.md
wolfenrain: Agreed.

erickzanardo (approved):
LGTM

doc/images.md
```diff
@@ -148,6 +148,24 @@ A `SpriteBatchComponent` is also available for your convenience.
 
 See example [here](https://github.com/flame-engine/flame/tree/master/doc/examples/sprite_batch).
 
+## Composition
+
+In some cases you may want to merge multiple images into a single image, this is called [Compositing](https://en.wikipedia.org/wiki/Compositing). This can be useful for example when working with the [SpriteBatch](#spritebatch) API to optimize your drawing calls.
```
luanpotter, line 153:
```suggestion
In some cases you may want to merge multiple images into a single image; this is called [Compositing](https://en.wikipedia.org/wiki/Compositing). This can be useful for example when working with the [SpriteBatch](#spritebatch) API to optimize your drawing calls. 
```

doc/images.md
```diff
@@ -148,6 +148,24 @@ A `SpriteBatchComponent` is also available for your convenience.
...
+final composition = Composition()
+  ..add(image1, Vector2(0, 0))
+  ..add(image2, Vector2(64, 0));
+  ..add(image3, 
+    Vector2(128, 0), 
+    source: Rect.fromLTWH(32, 32, 64, 64),
+  );
+
+Image image = await composition.compose();
+```
```
luanpotter, line 167: should we mention that this is too expensive to do every tick, and therefore compositions must be pre-rendered?

lib/composition.dart
```diff
@@ -0,0 +1,103 @@
...
+  final List<Image> _images = [];
+
+  /// The positions where each [_images] will be added.
+  final List<Vector2> _positions = [];
+
+  /// The sources that should be used on each [_images].
+  final List<Rect> _sources = [];
+
+  /// The [BlendMode] to be used for each [_images].
+  final List<BlendMode> _blendModes = [];
```
luanpotter, line 17: would it be clearer to have a single list of a data class (Fragment or something) that has all these fields?

luanpotter (approved):
LGTM

lib/composition.dart
wolfenrain: I didn't see a need to introduce another class just for that. But atm you also can't clear the composition, so if we introduce that then it would be smarter to have a single list.

wolfenrain: I added a `clear` method and refactored it to use a single list.

PR #653, opened 2021-02-03
Added Color and Image extensions

wolfenrain:
# Description

Follow up PR from https://github.com/flame-engine/flame/pull/646#discussion_r569535816, introducing an image extension and color extension.

I also added a few methods that I personally use, so if any of those are deemed unnecessary in the context of Flame then I am okay with removing them.

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
  - Adapting SpriteBatchComponent constructors to match the pattern used on other components
  - Renamed `Composition` to `ImageComposition` to prevent confusion with the composition component
  - Added `rotation` and `anchor` arguments to `ImageComposition.add`
+ - Added `Image` extensions
+ - Added `Color` extensions
 
 ## 1.0.0-rc6
  - Use `Offset` type directly in `JoystickAction.update` calculations
--- a/lib/src/extensions/color.dart
+++ b/lib/src/extensions/color.dart
@@ -1,0 +1,38 @@
+import 'dart:ui';
+
+export 'dart:ui' show Color;
+
+extension ColorExtension on Color {
+  /// Darken the shade of the color by the [amount].
+  ///
+  /// [amount] is a double between 0 and 1.
+  ///
+  /// Based on: https://stackoverflow.com/a/60191441.
+  Color darken(double amount) {
+    assert(amount >= 0 && amount <= 1);
+
+    final f = 1 - amount;
+    return Color.fromARGB(
+      alpha,
+      (red * f).round(),
+      (green * f).round(),
+      (blue * f).round(),
+    );
+  }
+
+  /// Brighten the shade of the color by the [amount].
+  ///
+  /// [amount] is a double between 0 and 1.
+  ///
+  /// Based on: https://stackoverflow.com/a/60191441.
+  Color brighten(double amount) {
+    assert(amount >= 0 && amount <= 1);
+
+    return Color.fromARGB(
+      alpha,
+      red + ((255 - red) * amount).round(),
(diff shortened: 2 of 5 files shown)


lib/src/extensions/color.dart
```diff
@@ -0,0 +1,38 @@
...
+
+export 'dart:ui' show Color;
+
+extension ColorExtension on Color {
+  /// Darken the shade of the color by the [amount].
+  ///
+  /// [amount] is a double between 0 and 1.
+  ///
+  /// Based on: https://stackoverflow.com/a/60191441.
+  Color darken(double amount) {
```
erickzanardo, line 11: Oh, this is excellent!

lib/src/extensions/image.dart
```diff
@@ -0,0 +1,73 @@
+import 'dart:typed_data';
+import 'dart:ui';
+
+import '../flame.dart';
+import 'color.dart';
+
+export 'dart:ui' show Image;
+
+extension ImageExtension on Image {
```
erickzanardo, line 9: Since you created this extension, there is a method that I think that would be cool to be here. We could have a `fromSprite` method, which would take a `Sprite` instance and return an `Image` instance which represents the "cropped" image that the sprite maps. What do you think?

Motivation: This would help on a new feature that I am planning on creating on fire atlas

erickzanardo (approved):
LGTM

lib/src/extensions/image.dart
erickzanardo: But in my use case, I would like to get the image from a Sprite, yeah, I could use composition, but this seems weird to me to use a class that compose an image, when I want to do the opposite operation. I don't think it would be such a redundant API since both classes have places which they fit better.

We could also put this on the Sprite class maybe, having a `toImage` method there.

spydon (approved):
Good stuff!

lib/src/extensions/image.dart
luanpotter: how about getCroppedImage() ? we can use ImageComposition underneath to impl it but I think having a method is useful!

lib/src/extensions/image.dart
```diff
@@ -0,0 +1,73 @@
+import 'dart:typed_data';
+import 'dart:ui';
+
+import '../flame.dart';
+import 'color.dart';
+
+export 'dart:ui' show Image;
+
```
luanpotter, line 8: I would add some comments here regarding:

this is *very* slow. do **not** do this every tick in your game, or your device, emulator, computer, city's power grid and the www will explode.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
  - Adapting SpriteBatchComponent constructors to match the pattern used on other components
  - Renamed `Composition` to `ImageComposition` to prevent confusion with the composition component
  - Added `rotation` and `anchor` arguments to `ImageComposition.add`
+ - Added `Image` extensions
+ - Added `Color` extensions
```
luanpotter, line 11: I think an example could be useful? but it can be a followup on the new repo instead of adding examples on the old structure

luanpotter (approved):
LGTM but I defer to @renancaraujo as the image manipulation expert

lib/src/extensions/image.dart
wolfenrain: I have discussed this with @erickzanardo  privately as well and we concluded a method on the `Sprite` class would be useful indeed, and I'll be following up on that after this PR is merged.

renancaraujo (approved):

PR #656, opened 2021-02-10
Remove old examples that have already been ported

luanpotter:
All these now live on: https://github.com/flame-engine/flame_example


spydon:
Maybe we should move them and fix the catalogue structure in the same PR? Now when we decided(?) not to have them in a separate repo?

luanpotter:
@spydon we can easily change from/to mono repo later; the actual code/structure won't change with that. I will keep porting on the current repo so at least we have the code ready. I am also taking a look at potential api changes.

PR #660, opened 2021-02-14
Draggables Overhaul

luanpotter:
Part 1. Fix the local vs global position bug
Part 2. Refactor onReceiveDrag to {onDragStarted,onDragUpdated,onDragEnded,onDragCancelled}

This could be easier to review analyzing the parts separately. Or not, go with your heart.

Part 1 is auto sufficient if we don't want to go for the whole refactor.
The refactor was bigger than I thought and midway through I was starting to wonder if it was worth it, but after having it done I now think it is.
I got really sad dart doesn't have interfaces while doing this. I dislike the duplicated methods over our mixins, but I don't think there is anything that can be done about that.

Changing joystick would be a mess because it was wiring stuff over several layers so I made use of composability to replace all the "passing things around" code. In fact, that made it much simpler as a side effect. I think there is more we can improve on that code to promote reuse. I tried not to mess too much past what I need but I will do a followup.

I ported all the gestures examples, in particular the draggable and joystick examples (that were not working on the example app because of the padding issue). I tested everything a lot on the flame_example branch `luan.next-release` and all seems good.


doc/input.md
```diff
@@ -179,22 +188,34 @@ class DraggableComponent extends PositionComponent with Draggable {
...
+  bool onDragStarted(int pointerId, Vector2 startPosition) {
+    dragDeltaPosition = startPosition - position;
+    return false;
+  }
+
+  @override
+  bool onDragUpdated(int pointerId, DragUpdateDetails details) {
+    final localCoords = gameRef.convertGlobalToLocalCoordinate(
+      details.globalPosition.toVector2(),
+    );
```
luanpotter, line 204: due to the way we bind this, anything coming from flutter is global. localPosition on the events is set to be == global. The reason and documentation is widely explained on the flutter docs in the code, so I don't think this is the end of the world. Exposing these methods from the game is a great thing to give users flexibility. I am considering monkey patching the local field since with this refactor we have full control, but if I do that will be a followup, this is already too big.

lib/src/components/base_component.dart
```diff
@@ -181,14 +181,14 @@ abstract class BaseComponent extends Component {
...
-        if (!shouldContinue) {
-          break;
-        }
+      }
+      if (shouldContinue && child is T) {
+        shouldContinue = handler(child);
+      }
+      if (!shouldContinue) {
+        break;
       }
```
luanpotter, line 192: this is unrelated but I think it was just wrong before

lib/src/game/game_widget.dart
```diff
@@ -423,65 +425,72 @@ Widget _applyBasicGesturesDetectors(Game game, Widget child) {
 
 Widget _applyAdvancedGesturesDetectors(Game game, Widget child) {
   final Map<Type, GestureRecognizerFactory> gestures = {};
```
luanpotter, line 427: I dont know why I decided to refactor this, I was not fully understanding it and I really disliked the fact that I had to declare yet another "fake interface" with the whole set of onX methods. In the end I think this is not particularly better than the version before but I don't think it's worse either so I left. I like that this version makes it more clear that there can only really be one handler. the list might trick you into thinking there might be two.

lib/src/game/game_widget.dart
```diff
@@ -528,3 +526,24 @@ class _GameRenderObjectWidget extends LeafRenderObjectWidget {
     );
   }
 }
+
+class _DragEvent extends Drag {
```
luanpotter, line 530: this is 100% internal now.

luanpotter: I would use an anonymous class but dart apparently lacks that as well.

lib/src/components/mixins/draggable.dart
```diff
@@ -1,36 +1,93 @@
...
+    return true;
+  }
+
+  bool onDragEnded(int pointerId, DragEndDetails details) {
+    return true;
+  }
+
+  bool onDragCanceled(int pointerId) {
+    return true;
+  }
```
luanpotter, line 25: technically I could have made these void. since we check for pointers at this stage. but I decided to keep consistent, if a user wants to do something where the drag has multiple owners or change owners during the drag they can.

lib/src/components/joystick/joystick_component.dart
```diff
@@ -29,58 +28,35 @@ abstract class JoystickController extends Component with HasGameRef<BaseGame> {
     _observers.add(listener);
   }
 
-  void onReceiveDrag(DragEvent drag) {}
-
   @override
   bool isHud = true;
 }
 
 class JoystickComponent extends JoystickController {
```
luanpotter, line 35: this just became a Parent

lib/src/components/joystick/joystick_component.dart
```diff
@@ -29,58 +28,35 @@ abstract class JoystickController extends Component with HasGameRef<BaseGame> {
...
-  final JoystickDirectional directional;
   @override
   int priority;
 
   JoystickComponent({
-    this.actions,
-    this.directional,
+    @required BaseGame gameRef,
+    List<JoystickAction> actions = const [],
+    JoystickDirectional directional,
```
luanpotter, line 42: I was unhappy but I think this is the only way

lib/src/components/joystick/joystick_action.dart
```diff
@@ -144,45 +159,47 @@ class JoystickAction {
...
+        _rectAction = _rectAction.shift(diff.toOffset());
       }
     }
   }
 
-  void onReceiveDrag(DragEvent event) {
-    if (_dragging || !(_rectAction?.contains(event.initialPosition) ?? false)) {
-      return;
+  @override
+  bool containsPoint(Vector2 point) {
```
luanpotter, line 181: these guys should really all be position components. but that is too much for this PR

lib/src/game/game_widget.dart
```diff
@@ -423,65 +425,72 @@ Widget _applyBasicGesturesDetectors(Game game, Widget child) {
...
 
-  if (game is MultiTouchDragDetector) {
-    dragHandlers
-        .add(_GenericDragEventHandler()..onReceiveDrag = game.onReceiveDrag);
+  void addDragRecognizer(Drag Function(int, Vector2) config) {
+    addAndConfigureRecognizer(
+      () => ImmediateMultiDragGestureRecognizer(),
+      (ImmediateMultiDragGestureRecognizer instance) {
+        instance.onStart = (Offset o) {
+          final pointerId = lastGenDragId++;
```
luanpotter, line 452: https://stackoverflow.com/questions/51187030/atomiclong-equivalent-in-dart

lib/src/game/game_widget.dart
```diff
@@ -423,65 +425,72 @@ Widget _applyBasicGesturesDetectors(Game game, Widget child) {
...
-      ..onTapUp = game.onTapUp
-      ..onTapCancel = game.onTapCancel);
-  }
-
-  if (game is MultiTouchTapDetector) {
-    tapHandlers.add(_GenericTapEventHandler()
-      ..onTapDown = game.onTapDown
-      ..onTapUp = game.onTapUp
-      ..onTapCancel = game.onTapCancel);
+  int lastGenDragId = 0;
```
luanpotter, line 428: if the tree gets rebuilt everything gets restarted so I think it's ok to reset this. but I can move it to the global scope if needed.

doc/examples/gestures/lib/main_multitap_advanced.dart
```diff
@@ -49,37 +50,31 @@ class MyGame extends BaseGame
...
-  void onPanCancel() {
+  void onDragCanceled(int pointerId) {
     _end = null;
     _start = null;
     _panRect = null;
   }
 
-  void onPanStart(Offset position) {
+  @override
+  void onDragStarted(int pointerId, Vector2 initialPosition) {
```
spydon, line 60: Isn't it better to name it `onDragStart` to follow the Flutter gesture convention with `onPanStart` etc? And the same for the other methods.

doc/examples/gestures/lib/main_multitap_advanced.dart
```diff
@@ -49,37 +50,31 @@ class MyGame extends BaseGame
...
-  void onPanEnd(DragEndDetails details) {
+  @override
+  void onDragEnded(int pointerId, DragEndDetails details) {
     if (_start != null && _end != null) {
       _panRect = Rect.fromLTRB(
-        _start.dx,
-        _start.dy,
-        _end.dx,
-        _end.dy,
+        _start.x,
```
spydon, line 74: Could be `_start.toPositionedRect(_end-_start)` too, not sure what is the easiest to read.

doc/input.md
```diff
@@ -161,15 +161,24 @@ Warning: `HasTapableComponents` uses an advanced gesture detector under the hood
...
 ```dart
-  void onReceiveDrag(DragEvent details) {}
+  void onDragStarted(int pointerId, Vector2 startPosition) {}
+  void onDragUpdated(int pointerId, DragUpdateDetails details) {}
+  void onDragEnded(int pointerId, DragEndDetails details) {}
+  void onDragCanceled(int pointerId) {}
 ```
 
-Minimal component example:
+Note that all events that a uniquely generated pointer id so you can, if desired, distinguish between different simultaneous drags.
```
spydon, line 173: "that a uniquely generated pointer id" -> "receives a uniquely generated pointer id"

doc/input.md
```diff
@@ -161,15 +161,24 @@ Warning: `HasTapableComponents` uses an advanced gesture detector under the hood
...
+
+The default implementation provided by `Draggable` will already check:
+
+* upon drag start only accept if the position is within the bounds of the Component; keep track of pointerId.
```
spydon, line 177: "upon drag start the component only receives the event if the position ..."

doc/input.md
```diff
@@ -161,15 +161,24 @@ Warning: `HasTapableComponents` uses an advanced gesture detector under the hood
...
+* when handling updates/end/cancel only accept if the pointerId was tracked (regardless of position).
```
spydon, line 178: "only accept" -> "the component receives the event if..."

doc/input.md
```diff
@@ -161,15 +161,24 @@ Warning: `HasTapableComponents` uses an advanced gesture detector under the hood
...
+* on end/cancel, un-track pointerId
```
spydon, line 179: untrack the pointerId

lib/src/components/base_component.dart
spydon: Aah true, so that it doesn't break the tree if there is an unrelated component in between.

lib/src/components/joystick/joystick_action.dart
```diff
@@ -116,26 +132,25 @@ class JoystickAction {
...
 
+  @override
   void update(double dt) {
+    super.update(dt);
     if (_rectBackgroundDirection != null && _dragging) {
-      final double _radAngle = atan2(
-        _dragPosition.dy - _rectBackgroundDirection.center.dy,
-        _dragPosition.dx - _rectBackgroundDirection.center.dx,
-      );
+      final diff = _dragPosition - _rectBackgroundDirection.center.toVector2();
```
spydon, line 139: Not related to this PR, but couldn't it be refactored so that `_rectBackgroundDirection` be a Vector2 for what it is used for?

lib/src/components/mixins/draggable.dart
spydon: But one pointer could propagate to several children still?

lib/src/extensions/rect.dart
```diff
@@ -11,17 +11,21 @@ extension RectExtension on Rect {
 
   /// Creates a [Vector2] starting in top left and going to [width, height].
   Vector2 toVector2() => Vector2(width, height);
+
+  bool containsVector2(Vector2 vector) {
```
spydon, line 15: Maybe have it consistent with the others and name it `containsPoint`?

lib/src/game/game_widget.dart
spydon: lastGeneratedDragId

spydon (commented):
Good stuff!!
Mostly I think the methods should be named:
`onDragUpdate`, `onDragEnd`, `onDragCancel`
To keep it both consistent with Flutter and with our other gestures.

lib/src/components/mixins/draggable.dart
luanpotter: it could if you like, return false on Start and then return true on Update. that is the reason I considered removing the bool from anything other the Start. but I though this was more consistent

lib/src/components/joystick/joystick_action.dart
luanpotter: I will do further improvements to the joystick classes in a followup, I think converting them all to PosComp will make it much simpler. also I dont understand why there is JoyAct and JoyDir they seem to be the same

doc/examples/gestures/lib/main_multitap_advanced.dart
luanpotter: much better, I will fix on the other repo

lib/src/components/mixins/draggable.dart
spydon: Right, makes sense

spydon (approved):
Grand!

PR #664, opened 2021-02-15
Removes all the examples, for now

luanpotter:
Yeah this will break a lot of links (for now).


erickzanardo (approved):

spydon:
I don't really get why we want to break all the links, when we can just keep the examples until the PR where we move in the flame_example into the root of this repo and fix the links in the docs?

erickzanardo:
Since all the examples are migrated now, I suggest that this PR do the following:

 - remove the examples from doc
 - make this repo a mono repo like we talking before, where the root would have the examples and flame folder, and maybe the doc folder could be on the root as well?
 - fix all the links

But I would wait for the PRs, especially @spydon's collision detection one, to be merged before, to avoid conflicts.

luanpotter:
Agreed, let me close this for now.

Issue #657, opened 2021-02-10
AudioCache prefix is changed since `audioplayers` 0.16.0


mpugach:
## Description

Since `audioplayers` version 0.16.0 the mandatory non-configurable prefix is changed
https://github.com/luanpotter/audioplayers/blob/0.15.1/lib/audio_cache.dart#L138
https://github.com/luanpotter/audioplayers/blob/0.16.0/lib/audio_cache.dart#L147

This causes `Unable to load asset` exception for assets in `./assets/audio`

## Development environment

Flame version:

0.29.0
0.29.1-beta

0.28.0 is not affected since it depends on `audioplayers: ^0.15.1`

Flutter doctor:

```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 1.22.6, on macOS 11.2.1 20D74 darwin-x64, locale en-UA)

[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.2)
[✓] Xcode - develop for iOS and macOS (Xcode 12.4)
[!] Android Studio (version 4.1)
    ✗ Flutter plugin not installed; this adds Flutter specific functionality.
    ✗ Dart plugin not installed; this adds Dart specific functionality.
[✓] VS Code (version 1.53.1)
[✓] Connected device (1 available)
```

## Runtime

This issue is related to running on which platform? (Select all that apply)

 - [x] Android :android:
 - [x] iOS :apple:
 - [ ] Web :browser: **(Did not tried)**

## Minimal reproducible code (Required for bugs)

https://github.com/japalekhin/langaw/tree/22f064e91f42f88198ab3473718662a778b828b5 just change it to 0.29.0 in the pubspec.yaml

luanpotter:
Excellent catch, @mpugach, I indeed forgot about this small breaking change I did on AP.

Fix: https://github.com/flame-engine/flame/pull/665

erickzanardo:
Fixed on #665

PR #665, opened 2021-02-16
Fix flame to be compatible with audioplayers breaking change, also re…

luanpotter:
This fixes https://github.com/flame-engine/flame/issues/657 for v0.

I will fix the analogous issue on flame_audio once we release rc7 and I can add an example to the new app.

Note: I am also removing path provider so that pub accepts that we support web (as we discussed).

After merging this I will release a next patch version.

### DIFF
--- a/lib/audio_pool.dart
+++ b/lib/audio_pool.dart
@@ -25,7 +25,7 @@ class AudioPool {
       {this.repeating = false,
       this.maxPlayers = 1,
       this.minPlayers = 1,
-      String prefix = 'audio/sfx/'}) {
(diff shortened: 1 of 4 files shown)


erickzanardo (approved):
LGTM, just be aware that this branch has the dry layout fix so if published we may take a hit on the pub score.

I suggest that we create a patch based on the tag too

luanpotter:
> this branch has the dry layout fix so if published we may take a hit on the pub score.

this branch is supposed to be whatever we have currently published right? why does it have code that is not equivalent to pub?

PR #667, opened 2021-02-16
Improve default constructor for positioned components

pedia:
Since `position` and `size` is most important property of PositionComponent,
so the default constructor is very important. We should be able to write:

```dart
  SpriteComponent(position: Vector2(x, y));
```

Normally? user define a component like
```dart
class SomeComponent extends SpriteComponent {
  Future<void> onLoad() async {
    // load image, set size, ...
  }
}
```

Add some instances to `Game` like:
```dart
game.add(SomeComponent(position: ...));
```

It's reasonable to split sync initial and async load in different routine.
In library(flame), it is better to keep same style.

This PR also changed all derived components and never break old code.


spydon (commented):
I think the default constructor is nice, if there wasn't a reason that we removed it, maybe it had to do with loading things @erickzanardo? The *.position factories I think will just make the code less uniform, so those I think should be removed.

lib/src/components/sprite_component.dart
```diff
@@ -24,15 +24,36 @@ class SpriteComponent extends PositionComponent {
...
+  @Deprecated('Use SpriteComponent instead')
+  factory SpriteComponent.fromImage(Vector2 size, Image image) =>
+      SpriteComponent(size: size, sprite: Sprite(image));
 
+  @Deprecated('Use SpriteComponent instead')
   SpriteComponent.fromSprite(Vector2 size, this.sprite) {
     super.size.setFrom(size);
   }
 
+  factory SpriteComponent.position(
```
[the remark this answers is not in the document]
pedia: because:
- better performance than `SomeComponent()..x=3 ..y=4`
- user always define `AnyComponent` with `onLoad`, and add into game like:
```
game.add(AnyComponent(x, y))
```
it's better to keep same style in library.

lib/src/components/position_component.dart
```diff
@@ -108,6 +108,13 @@ abstract class PositionComponent extends BaseComponent {
     return point.clone()..rotate(angle, center: position);
   }
 
+  PositionComponent({Vector2 position, Vector2 size})
+      : position = position ?? Vector2.zero(),
+        size = size ?? Vector2.zero();
+
+  factory PositionComponent.position({double x = 0.0, double y = 0.0}) =>
```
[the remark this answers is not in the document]
pedia: Added already.
```dart
PositionComponent({Vector2 position, Vector2 size});
```

lib/src/components/sprite_component.dart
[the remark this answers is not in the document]
pedia: trivial *.position removed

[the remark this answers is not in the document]
spydon: Great, please update the PR description too :)
(and for future PRs, please don't remove the template)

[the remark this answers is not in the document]
pedia: oh, forgot this. fixed.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
  - Change RaisedButton to ElevatedButton in timer example
  - Overhaul the draggables api to fix issues relating to local vs global positions
  - Preventing errors caused by the premature use of size property on game
+ - `PositionComponent` remove abstract
+ - Component constrctor unified
```
luanpotter, line 18: nit: typo

luanpotter:
@pedia I am not against this in broad lines, but I am trying to understand why this would be the case?

> better performance than SomeComponent()..x=3 ..y=4

For me:

```
  SpriteComponent(position: Vector2(10, 20))
```

and

```
  SpriteComponent()..position = Vector2(10, 20)
```

are pretty much the same. I guess that is the reason we didn't add these kinds of constructors

erickzanardo (commented):
> I think the default constructor is nice, if there wasn't a reason that we removed it, maybe it had to do with loading things @erickzanardo? The *.position factories I think will just make the code less uniform, so those I think should be removed.

It don't have direct relation to the loading things, but was on that refactor that we simplified our components constructors.

I don't really believe that having this on the constructor will improve our perfomance, at most, even if there is indeed perfomance improve, that would be impossible to notice. But if there any benchmarks on that, I would love to see them.

Before, we had a bunch of constructors, and that was quite confusing and hard to manage, I still prefer to keep then at minimal and take advantage of Dart's cascading operator, otherwise, we will need to have optional parameters for all attributes.

luanpotter:
> Before, we had a bunch of constructors, and that was quite confusing and hard to manage

I agree we should only merge something that does not change the number of constructors we have. I would be ok with adding extra optional parameters to all constructors if there is a benefit (which I am trying to understand). we should strive to not go back to constructor hell with half a dozen overloads per class

erickzanardo:
> I would be ok with adding extra optional parameters

The problem with that is that all child classes of PositionComponent would need to call super on that, otherwise we would have some inconsistencies. And since those parameters are all optional already, it is way more manageable to let then be set by the cascading operator IMO

pedia:
Cascading operator vs Constructor

Without optional argument, we can't create instance with only constructor.
Fully optional constructor will reduce other factories.
Performance is very very little but they are different.
Create with constructor is more dart/flutter stylish.

erickzanardo:
> * Without optional argument, we can't create instance with only constructor.

True, but there enters cascading operator, which solves that easily

> * Fully optional arguments constructor will reduce other factories

True, but it creates more complicated constructors for us to maintain

> * Performance is very very little but they are different.

I still think the perfomance, even if it is better, it would never me noticed. But again, if you have any benchmark to share about this, that would be cool

> * Create with constructor is more dart/flutter stylish.

Not sure I fully agree with you on that one. If we were talking about Widgets then I would 100% agree, but components are not Widgets. Initialization of optional properties is even used as example on the [dart guide](https://dart.dev/guides/language/language-tour#cascade-notation-)

This is a major change in our API regarding components, so this need to be carefully thought, I still prefer that we keep the use of cascading operator for initializing optional props because that will make this part of the API easy to maintain.

Changing to an optional parameter approach, means that we will start to replicating these parameters on all available factories, of all of our components (for example, on your PR,  you didn't changed the SpriteAnimationComponent.fromFrameData).

Since this is a major a change in our API I think we need to get on a common decision as a team, what do you think @spydon @luanpotter @renancaraujo @wolfenrain ?

spydon:
> > * Without optional argument, we can't create instance with only constructor.
> 
> True, but there enters cascading operator, which solves that easily
> 
> > * Fully optional arguments constructor will reduce other factories
> 
> True, but it creates more complicated constructors for us to maintain
> 
> > * Performance is very very little but they are different.
> 
> I still think the perfomance, even if it is better, it would never me noticed. But again, if you have any benchmark to share about this, that would be cool
> 
> > * Create with constructor is more dart/flutter stylish.
> 
> Not sure I fully agree with you on that one. If we were talking about Widgets then I would 100% agree, but components are not Widgets. Initialization of optional properties is even used as example on the [dart guide](https://dart.dev/guides/language/language-tour#cascade-notation-)
> 
> This is a major change in our API regarding components, so this need to be carefully thought, I still prefer that we keep the use of cascading operator for initializing optional props because that will make this part of the API easy to maintain.
> 
> Changing to an optional parameter approach, means that we will start to replicating these parameters on all available factories, of all of our components (for example, on your PR, you didn't changed the SpriteAnimationComponent.fromFrameData).
> 
> Since this is a major a change in our API I think we need to get on a common decision as a team, what do you think @spydon @luanpotter @renancaraujo @wolfenrain ?

I see benefits with both. For these values that are almost always needed I'd find it easier as a user to find them if they were optional arguments, but then we need to have a strict definition of what should be optional arguments and what should be assigned with the cascading operator to keep our code base consistent. So if it is not easy to make such a definition it is easier to go with the cascading operator.
I don't think that it would be that hard to maintain the `super` calls from subclasses, if the number of optional arguments is low, like here.

pedia:
I write a simple [benchmark](https://gist.github.com/pedia/f13576387ef40a29c223c9c742bde565)

It seems that constructor is *bit litter* faster than assign after construct.

@erickzanardo

pedia:
In C++, construct and later assignment are much different and significant. That makes sense too in Dart. It’s a little bit weird to assign after construct. 

Maybe the library should not prevent the optional arguments in the constructor.

luanpotter:
I'm really over the fence, but I don't think the harm in adding these is that big (should be fairly simple to maintain the chain an pass everything along AS LONG AS the number of constructors we offer do not change), so I think I would lean towards giving two options to the user.

wolfenrain:
> I'm really over the fence, but I don't think the harm in adding these is that big (should be fairly simple to maintain the chain an pass everything along AS LONG AS the number of constructors we offer do not change), so I think I would lean towards giving two options to the user.

I agree with Luan's vision of this. The only pitfall I can see is that we can lose sight of maintaining either of these options, which can result in a sloppy API feeling.

lib/src/components/position_component.dart
```diff
@@ -20,9 +20,9 @@ import 'mixins/hitbox.dart';
 /// rendered automatically when this is updated and rendered.
 /// They are translated by this component's (x,y). They do not need to fit
 /// within this component's (width, height).
-abstract class PositionComponent extends BaseComponent {
+class PositionComponent extends BaseComponent {
```
erickzanardo, line 23: I don't see any reason for making this a concrete class, PositionComponent don't implement a render method and I think it should continue to remain as abstract.

lib/src/components/position_component.dart
```diff
@@ -108,6 +108,10 @@ abstract class PositionComponent extends BaseComponent {
     return point.clone()..rotate(angle, center: position);
   }
 
+  PositionComponent({Vector2 position, Vector2 size})
```
erickzanardo, line 111: Nit: Add some empty spaces between the curly braces and the parameters please

lib/src/components/sprite_animation_component.dart
```diff
@@ -14,18 +14,24 @@ class SpriteAnimationComponent extends PositionComponent {
...
+    Vector2 size,
+    this.animation,
+    this.overridePaint,
+    this.removeOnFinish = false,
+  }) : super(position: position, size: size);
 
   /// Creates an [SpriteAnimationComponent] from an [animation] and a [size]
   ///
   /// Optionally [removeOnFinish] can be set to true to have this component be auto removed from the [BaseGame] when the animation is finished.
+  @Deprecated('Use SpriteAnimationComponent instead')
```
erickzanardo, line 28: On our current phase of the project (improving the APIs to release v1) it is ok to do breaking changes, so this constructor can be removed

lib/src/components/sprite_animation_component.dart
```diff
@@ -14,18 +14,24 @@ class SpriteAnimationComponent extends PositionComponent {
...
   SpriteAnimationComponent.fromSpriteAnimation(
     Vector2 size,
     this.animation, {
     this.removeOnFinish = false,
-  }) : assert(animation != null) {
-    super.size.setFrom(size);
-  }
+  })  : assert(animation != null),
```
erickzanardo, line 33: Since size and position were added as optional to the default constructor, I think this factory should work the same way, so `size` should be optional, and a `position` should be added

lib/src/components/sprite_component.dart
```diff
@@ -24,11 +24,18 @@ class SpriteComponent extends PositionComponent {
...
-
+  SpriteComponent({
+    Vector2 position,
+    Vector2 size,
+    this.sprite,
+    this.overridePaint,
+  }) : super(position: position, size: size);
+
+  @Deprecated('Use SpriteComponent instead')
+  factory SpriteComponent.fromImage(Vector2 size, Image image) =>
```
erickzanardo, line 35: This factory still makes sense, remove the deprecation and add `size` and `position` as optional

lib/src/components/sprite_component.dart
```diff
@@ -24,11 +24,18 @@ class SpriteComponent extends PositionComponent {
...
+      SpriteComponent(size: size, sprite: Sprite(image));
+
+  @Deprecated('Use SpriteComponent instead')
```
erickzanardo, line 38: Same comment from the Animation, we are ok with breaking changes during the release candidates, just remove this factory

lib/src/components/text_box_component.dart
```diff
@@ -53,9 +53,12 @@ class TextBoxComponent extends PositionComponent {
     String text, {
     TextConfig config,
     TextBoxConfig boxConfig,
+    Vector2 position,
+    Vector2 size,
   })  : _text = text,
         _boxConfig = boxConfig ?? TextBoxConfig(),
-        _config = config ?? TextConfig() {
+        _config = config ?? TextConfig(),
+        super(position: position) {
```
erickzanardo, line 61: Size needs to be passed to super as well.

lib/src/components/text_component.dart
```diff
@@ -28,8 +29,14 @@ class TextComponent extends PositionComponent {
...
-  TextComponent(this._text, {TextConfig config})
-      : _config = config ?? TextConfig() {
+  TextComponent(
+    this._text, {
+    TextConfig config,
+    Vector2 position,
+    Vector2 size,
+  })  : assert(_text != null),
+        _config = config ?? TextConfig(),
+        super(position: position) {
```
erickzanardo, line 39: `size` needs to be passed to super as well.

erickzanardo (changes requested):
Alright, since the team has decided that it is worth to add the position and size parameters as optional on the default constructor I am onboard as well. Thanks for sharing that benchmark @pedia was really interesting.

I have added some comments though, to make the new API change more in our standards.

EDIT: We need to check if the docs needs changes because of these changes.

lib/src/components/position_component.dart
pedia: Without constructor of PositionComponent, XXComponment cannot initialize PositionComponent.{position|size}.
It's not possible add constructor to abstract class of Dart.

lib/src/components/sprite_component.dart
pedia: added

lib/src/components/text_box_component.dart
pedia: fixed

lib/src/components/text_component.dart
pedia: fixed

lib/src/components/sprite_component.dart
pedia: fixed

lib/src/components/position_component.dart
luanpotter: @erickzanardo that violates our styleguide though

lib/src/components/sprite_animation_component.dart
luanpotter: rip it off!

lib/src/components/sprite_animation_component.dart
luanpotter: consistency!

lib/src/components/sprite_component.dart
```diff
@@ -24,14 +24,19 @@ class SpriteComponent extends PositionComponent {
...
+    this.sprite,
+    this.overridePaint,
+  }) : super(position: position, size: size);
+
+  factory SpriteComponent.fromImage(
+    Vector2 size,
+    Image image, {
+    Vector2 position,
+  }) =>
+      SpriteComponent(size: size, sprite: Sprite(image), position: position);
```
luanpotter, line 39: nit: I prefer block bodies if the thing won't fit on the same line as the function dec

luanpotter (commented):
Judging by the example changes, I think this was a good call. The code looks really nice. I have some small nits but looking good already!

lib/src/components/position_component.dart
spydon: Everywhere else in the codebase we don't do spacing for this, so let's keep it consistent.:)

erickzanardo: Sure, I probably confused the styleguides, we use like I said at work

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@
  - Fix `SpriteAnimationComponent.shouldRemove` use `Component.shouldRemove`
  - Add assertion to make sure Draggables are safe to add
  - Add utility methods to the Anchor class to make it more "enum like"
+ - Change `PositionComponent` type
```
wolfenrain, line 20: I would personally make this entry a bit more explicit:

```suggestion
 - `PositionComponent` is no longer an abstract class
```

wolfenrain (approved):
One nit, but apart from that LGTM.

lib/src/components/position_component.dart
erickzanardo: I don't think this is true, you totally can add constructors to abstract classes in Dart, I have just tried this small example on dartpad to be sure:

```dart
void main() {
  final b = BlaImplementation(1);
  print(b.x);
}

abstract class Bla {
  int x;
  
  Bla(this.x);
}

class BlaImplementation extends Bla {
  BlaImplementation(int x): super(x);
}
```
I still think that this change should be reverted, I don't see any reason for this class to be concrete.

erickzanardo (changes requested):

luanpotter:
@pedia we need to rebase/merge master into this so it gets the new build script. Lmk if you want I can do that for you :)

pedia:
@luanpotter Thanks, Please.

pedia:
@erickzanardo Rollback `PositionComponent` abstract now.

lib/src/components/sprite_animation_component.dart
erickzanardo: This is not resolved.

lib/src/components/sprite_animation_component.dart
erickzanardo: This is not resolved but this method can be removed like we mentioned before, but don't forget to change the next constructor on this file as well, `fromFrameData` needs to be changed to have both size and position on its optional parameter list.

erickzanardo (changes requested):
Great! But there are still some comments that needs to be addressed.

luanpotter:
@erickzanardo can I merge and fix these two details on a followup?

erickzanardo:
> @erickzanardo can I merge and fix these two details on a followup?

Ok, I will remove my change request

erickzanardo (approved):

PR #669, opened 2021-02-17
Make the anchor class look more like an enum

luanpotter:
Anchor should really be an enum, but sadly enums in dart are not as sophisticated as java, so we need a class. However dealing with it was not easy because there was no standard way to treat it like a bounded collection of items, so I am adding some common enum helper methods to Anchor to make that integration easier.

### DIFF
--- a/lib/src/anchor.dart
+++ b/lib/src/anchor.dart
@@ -1,29 +1,61 @@
 import 'extensions/vector2.dart';
 
+/// Represents a relative position inside some 2D object with a rectangular
+/// size or bounding box.
+///
+/// Think of it as the place where you "grab" or "hold" the object.
+/// In Components, the Anchor is where the component position is measured from.
+/// For example, if a component position is (100, 100) the anchor reflects what
+/// exact point of the component that is positioned at (100, 100), as a relative
+/// fraction of the size of the object.
+///
+/// The "default" anchor in most cases is topLeft.
+///
+/// The Anchor is represented by a fraction of the size (in each axis),
+///  where 0 means top/left and 1 means right/bottom.
 class Anchor {
-  static const Anchor topLeft = Anchor(0.0, 0.0);
-  static const Anchor topCenter = Anchor(0.5, 0.0);
-  static const Anchor topRight = Anchor(1.0, 0.0);
-  static const Anchor centerLeft = Anchor(0.0, 0.5);
-  static const Anchor center = Anchor(0.5, 0.5);
-  static const Anchor centerRight = Anchor(1.0, 0.5);
-  static const Anchor bottomLeft = Anchor(0.0, 1.0);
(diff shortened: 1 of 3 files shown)


lib/src/anchor.dart
```diff
@@ -21,4 +21,27 @@ class Anchor {
   Vector2 translate(Vector2 p, Vector2 size) {
     return p - (toVector2..multiply(size));
   }
+
+  static final Map<Anchor, String> _valueNames = {
```
spydon, line 25: Isn't the current way more like an enum than this way of parsing strings?
What is the thought behind adding this second way of getting an Anchor?

lib/src/anchor.dart
```diff
@@ -21,4 +21,27 @@ class Anchor {
...
+    centerRight: 'centerRight',
+    bottomLeft: 'bottomLeft',
+    bottomCenter: 'bottomCenter',
+    bottomRight: 'bottomRight',
+  };
+
+  static final List<Anchor> values = _valueNames.keys.toList();
+
+  @override
+  String toString() {
```
erickzanardo, line 40: This could be a one liner

erickzanardo (approved):
Changes LGTM, but I am wondering about @spydon comment too

wolfenrain (approved):
Missing a CHANGELOG entry.

lib/src/anchor.dart
```diff
@@ -21,4 +21,27 @@ class Anchor {
...
+    return _valueNames[this];
+  }
+
+  static Anchor parse(String name) {
```
spydon, line 44: I think we should remove this method and the `_valueNames`, but keep the `values` one but just have a list of the predefined anchors in there, instead of taking the keys from the map.
If you don't have any good argument for why it is needed that is? Because otherwise I think it would lead to an anti-pattern where Anchors are created from strings, and that is not the intention here right?

lib/src/anchor.dart
luanpotter: we discussed offline, the key reason here is serialization

> re about the anchor class: let me explain
you are questioning why I added methods to convert to and from string and whether that is an antipattern. those methods are to make serialization easy. enums not only have a values list but they are intrinsically serializable. objects are not. if you want to store something that has an anchor as json, for example, you would need to write something like I did yourself. in fact to use them with dashbook erick had to do that by hand. so I believe having these methods makes it easier to do so if you ever wanted, and the code is in one centralized place. enums in java also offer .name and .valueOf for that purpose so I think it's quite expected that enums have to/from string conversion (maybe I can use the java names to make it clear). I feel being Serializable in a sense (even though the concept is not as precisely defined as in java/kt) is important.

erickzanardo: Is telegram offline? 👀

erickzanardo (approved):

PR #671, opened 2021-02-19
Add assert to make Draggables are used properly

luanpotter:
Analogous with Tapable

### DIFF
--- a/lib/src/game/base_game.dart
+++ b/lib/src/game/base_game.dart
@@ -2,15 +2,11 @@ import 'dart:ui';
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/rendering.dart';
-import 'package:flutter/widgets.dart' hide WidgetBuilder;
 import 'package:ordered_set/comparing.dart';
 import 'package:ordered_set/ordered_set.dart';
 
-import '../components/component.dart';
-import '../components/mixins/has_game_ref.dart';
-import '../components/mixins/tapable.dart';
-import '../components/position_component.dart';
-import '../extensions/vector2.dart';
+import '../../components.dart';
+import '../../extensions.dart';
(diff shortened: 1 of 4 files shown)


lib/src/game/base_game.dart
```diff
@@ -47,6 +43,12 @@ class BaseGame extends Game with FPSCounter {
         'Tapable Components can only be added to a BaseGame with HasTapableComponents',
       );
     }
+    if (c is Draggable) {
```
erickzanardo, line 46: I think that we could add a test for this, and for the others as well, what do you think?

erickzanardo: 🧠 🧙‍♂️

lib/src/game/base_game.dart
```diff
@@ -2,15 +2,11 @@ import 'dart:ui';
...
 import 'package:ordered_set/comparing.dart';
 import 'package:ordered_set/ordered_set.dart';
 
-import '../components/component.dart';
-import '../components/mixins/has_game_ref.dart';
-import '../components/mixins/tapable.dart';
-import '../components/position_component.dart';
-import '../extensions/vector2.dart';
+import '../../components.dart';
+import '../../extensions.dart';
```
erickzanardo, line 9: Do we know if that affects tree shaking?

erickzanardo (approved):
One small concern, but LGTM already

wolfenrain (approved):

lib/src/game/base_game.dart
luanpotter: I will revert for now until we have renan's answer

luanpotter: cc @renancaraujo

PR #673, opened 2021-02-19
Lights, Camera, Action!

luanpotter:
This adds a way more sophisticated camera (than a simple vector2) and viewports into Flame.

The code was extracted and generalized from Tales, but I've been copying it on most games. I think this is really useful.

# Viewport

The viewport converts the coordinate space of the screen into a more sane one. There can be several types of viewports; this PR adds two (Default that does nothing and FixedRatio that guarantees maximum expanded keeping the ratio).

# Camera

The camera translates the canvas following a set of rules. You can move it at will, follow a PositionComponent, among other useful features.

# Draft

This is still a draft. The basics are implemented and tested. What is missing:

* documentation
* example
* gestures

I open the PR for review because I want to discuss how to deal with gestures. Gestures come from Flutter and we just pass flutter events along, meaning the coordinate system used is the Flutter one (clearly unaware of viewport and camera transformations). In fact Flutter already have two systems, as we know, global and local. And now we have 3, so it can be a bit confusing but all 3 might be needed on different scenarios. Note that we already have a similar issue in which our drag events only have global coordinates (because of the raw detector we use for drags).

So here are the options:

1. do nothing. I provide to and from methods so we can just let the user transform as they need. this might be the most clear solution because there is no unexpected. Note that this will make tapables and draggables not respect the viewport/camera by default, but could be fixed by the user with some very simple overrides.
2. Only fix tapables and draggables engine handling: this leaves the events alone, but adds code to the base game engine so that it correctly identifies tapables and draggables. this way your tapable will only receive events such that the transformed coordinate system complies (and of course isHud components can bypass that). However if in the handler you inspect the event received it will still only have global and local coordinates. Also if you are overriding the handlers on BaseGame yourself, w/o using Tapable and Draggable, you will still need to act yourself. This is relatively simple to do and it fixes the broken expectation with T/D but it creates another inconsistency where the events being passed along have to be understood for what they are.

The problem with going further with 3 or 4 is that all our handlers use the native flutter events that just don't have the capability of holding all the information in a clean way.
3. Modify flutter events: this option we still use the same classes, so there is no api changes, but we can modify or re-create all events at the engine level to change: localCoordinates to be the transformed viewport/camera coordinates (and also fix the missing localCoordinates from drags at the same time) and the globalCoordinates can be set to be either (a) what they are now or (b) what local coordinates were now. There is no way to provide the 3 sets of coordinates so if we are going this route we basically have to choose which 2 to provide to the user. Of course they can easily get the 3rd if needed using one of the transform methods already provided on each layer (game.convertGlobal To/From Local and camera.world To/From Screen). This has the drawbacks that it might be confusing to users because the flutter docs on those events will not mention about what each field represent, so users will have to refer to our docs, and also we are monkey patching flutter fields which can be confusing and/or dangerous. But the advantage is that there is no api breaking change and is 100% backwards compatible as the default viewport/camera transformation is the identity.
4. Create our own event details classes for every single action: this will require us to create our own classes where we can provide accessors for the 3 different coordinate systems with very clear documentation on what each is and where it comes from. The user would then likely rarely need to do coordinate transformations themselves as long as they use the proper field. We can also add other details that are not the location (like if it was left middle or right click) to our class and we can hold a reference to the original event in case the user needs it. This however is an api breaking change and it's by far the hardest as it requires redefining all events. If Flutter changes some event I imagine they would only likely add new info, that as long as it doesn't contradict our transformation logic could still be accessed via the rawEvent reference until we pull it up to our own Details class.

I thought a little bit and I didn't reach a conclusion for what is the best option, so I pose this question for the community.
We can also merge this PR as is (I added a note about gestures not being transformed) and deal with it in a followup (specially if we choose the harder options).


erickzanardo:
Mark this as draft then :P

luanpotter:
AH! didn't know about that

lib/src/game/base_game.dart
```diff
@@ -31,8 +33,35 @@ class BaseGame extends Game with FPSCounter {
...
-  /// Camera position; every non-HUD component is translated so that the camera position is the top-left corner of the screen.
-  Vector2 camera = Vector2.zero();
+  /// The camera translates the coordinate space after the viewport is applied.
+  final Camera camera = Camera();
+
+  /// The viewport transforms the coordinate space depending on your chosen
+  /// implementation.
+  /// The default implementation no-ops, but you can use this to have a fixed
+  /// screen ratio for example.
+  Viewport viewport = DefaultViewport();
```
erickzanardo, line 43: Why not implement this on Game instead of BaseGame? This seems something that even games that don't use components would benefit from this

analysis_options.yaml
```diff
@@ -81,6 +81,6 @@ dart_code_metrics:
   metrics:
     number-of-arguments: 8
     number-of-methods: 28
-    lines-of-executable-code: 200
+    lines-of-executable-code: 400
```
erickzanardo, line 84: That is too many lines IMO, I think we should think on better ways of splitting the code instead of raising this so much

lib/src/game/camera.dart
```diff
@@ -0,0 +1,289 @@
...
+import 'dart:ui' show Rect;
+
+import '../../components.dart';
+import '../../game.dart';
+
+/// Utility method to smoothly transition vectors.
+void _moveToTarget(
+  Vector2 current,
+  Vector2 target,
+  double ds,
```
spydon, line 11: Maybe use a non-abbreviated name and it'll be easier to read

lib/src/game/camera.dart
```diff
@@ -0,0 +1,289 @@
...
+/// Components marked as `isHud = true` are always rendered in screen
+/// coordinates, bypassing the camera altogether.
+///
+/// Note: right now this only applies to rendering. No transformation is
+/// done on event handling (for gestures). You have to transform it yourself
+/// using [screenToWorld].
+class Camera {
+  static const DEFAULT_CAMERA_SPEED = 50.0; // in pixels/s
+  static const DEFAULT_SHAKE_INTENSITY = 75.0; // in pixels
+  static const DEFAULT_SHAKE_DURATION = 0.3; // in seconds
```
spydon, line 65: Consts in dart shouldn't be java style
https://dart.dev/guides/language/effective-dart/style#prefer-using-lowercamelcase-for-constant-names

lib/src/game/camera.dart
```diff
@@ -0,0 +1,289 @@
...
+  // Shake
+
+  /// Applies a shaking effect to the camera for [amount] seconds.
+  ///
+  /// The intensity can be controlled via the [shakeIntensity] property.
+  void shake({double amount = DEFAULT_SHAKE_DURATION}) {
+    _shakeTimer += amount;
+  }
+
+  /// Wether the camera is currently shaking or not.
```
spydon, line 278: Whether

lib/src/game/camera.dart
```diff
@@ -0,0 +1,289 @@
...
+      }
+    }
+
+    return attemptedTarget;
+  }
+
+  // Movement
+
+  /// Applies an ad-hoc movement to the camera towards the target, bypassing
+  /// follow. Once it arrives the camera will not move until resetMovement
```
spydon, line 253: [resetMovement]

lib/src/game/viewport.dart
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  @override
+  Vector2 getEffectiveSize() => _size;
+
+  @override
+  Vector2 getRawSize() => _size;
+}
+
+class FixedRatioViewport extends Viewport {
```
spydon, line 37: Maybe some short explanations of each viewport

lib/src/game/camera.dart
```diff
@@ -0,0 +1,289 @@
...
+///
+/// A camera always has a current [position], however you cannot set it
+/// directly. You must use some methods to ensure that the camera moves smoothly
+/// as the game runs. Smoothly here means that sudden snaps should be avoided,
+/// as they feel jarring to the player.
+///
+/// There are three major factors that determine the camera position:
+///
+/// * Follow
+/// If you want, you can call [followObject] at the beginning of your
```
spydon, line 35: In the forge2d viewport the follow method can take some parameters to not follow the body strictly, so that your body/component can move around a bit in the middle without the camera having to move. I think that would be awesome to have, but that could be a follow-up too.
https://github.com/flame-engine/flame_forge2d/blob/master/lib/viewport.dart#L53

spydon (commented):
This is going to be so cool!

lib/src/game/base_game.dart
luanpotter: I'm only reluctant because this basically changes the render method, and on Game the render method was supposed to be totally empty for the user to implement. if they want they can add a viewport to they Game and use it on render. it's 3 extra lines. if they don't want they don't get the burden of it

analysis_options.yaml
luanpotter: this is just because of the test file. I joined everything together because I thought it would be clearer but I can split into camera test, viewport test, etc.

lib/src/game/base_game.dart
erickzanardo: I don't see that as a problem, must callSuper is our friend for that

lib/src/game/camera.dart
luanpotter: like delta space? I think ds is a pretty standard abbreviation, like dt, we use it all over

lib/src/game/camera.dart
luanpotter: WHAAAAT? I did **not** know that at all hahahaha

Shall we add this linter rule then? and amend the styleguide?

analysis_options.yaml
erickzanardo: On spydon pr, the tests will be excluded from metrics, so that will solve you problem here too

lib/src/game/viewport.dart
luanpotter: totally forgot!

lib/src/game/camera.dart
luanpotter: I think what those parameters do is exactly the same as relativeOffset (the percentage of the screen the component occupies). we can definitely add more stuff as a followup though!

analysis_options.yaml
luanpotter: I added a comment, we should just exclude everything

erickzanardo: I disagree, I added a comment too.

lib/src/game/base_game.dart
luanpotter: that is exactly the problem, I don't want the user of Game to be forced to call super. Game should be almost like an interface (that dart doesn't have) with empty render and update and you can override and do what you want. it only hooks the bare minimum necessary. does not add anything on top. that is BaseGame, preconfigured for your convenience. you can still use viewport with game, it just won't be included automatically

erickzanardo: I still think that this should be in the Game class, if you don't want to use mustCallSuper, we could have this code hooked on an internal render method, and that internal render method would call the abstract render method that the user needs to implement. We have done something similar like that in the past

luanpotter: we do that if we need to, because it's something coming from the engine that we have to expose somehow. I don't think this is the case here, this is just something that you put on top and get some benefits, it's super easy to wire and on Game you get to wire it however you want (for example you might want to render some UI outside the viewport, etc). The BaseGame has our opinionated suggestion of how to use it. I don't see much advantage in adding it since it's so simple to add yourself, and I would rather keep Game as slim as possible. we should only have things that are strictly necessary there

spydon: Hmm, on second thought here I agree with luan.
A viewport on there is making it opinionated.

lib/src/game/camera.dart
spydon: Or make a comment about it?
I didn't know it was standing for delta space at least, even though I understood what it did when reading the code.

lib/src/game/base_game.dart
erickzanardo: If we separate the methods and leave the abstract render without anything, Game will still be slim, and since you said yourself that this is only a few lines to wire this into game, this will not add much to the Game class.

Camera is a basic function IMO, it makes a lot of sense on being on our basic Game class

spydon:
If anyone is willing to take upon the number 4 beast I think that would be the nicest for the end user, otherwise I think we should go with number 1.

lib/src/game/base_game.dart
luanpotter: I don't see what exactly do you want in these methods, the methods are already pretty separated on the viewport class, you can just hook it up. there is nothing to add on the Game class.

also we are talking about viewport right? camera is tied to base game as it uses position component

analysis_options.yaml
luanpotter: we can open a separated issue for this

luanpotter:
@spydon I am willing but I wanted to get consent before I even started such a behemoth. also I think 4 can be a followup to this PR.

lib/src/game/camera.dart
luanpotter: I will add one. maybe it can be called displacement

analysis_options.yaml
erickzanardo: You probably don't even need this anymore, just rebase yourself PR with master, Lukas did excluded the tests from metrics, so your test will be fine

lib/src/game/base_game.dart
erickzanardo: I am not talking about viewport exactly, I am talking about this feature as a whole. I think this should be on the Game class, not the BaseGame class.

The method I am talking about is not about the Camera or Viewport classes, but about that with a small change on Game methods, this could easily be wired directly on the Game class without making it "Fat"

lib/src/game/base_game.dart
```diff
@@ -32,8 +34,35 @@ class BaseGame extends Game with FPSCounter {
...
+    return viewport.getEffectiveSize();
+  }
+
+  /// This is the original Flutter widget size, without any transformation.
+  Vector2 get rawSize {
+    assertHasLayout();
+    return viewport.getRawSize();
+  }
+
+  BaseGame() {
```
erickzanardo, line 63: Can't we pass this on the Camera declaration?

lib/src/game/base_game.dart
```diff
@@ -32,8 +34,35 @@ class BaseGame extends Game with FPSCounter {
...
+  /// Which means that this is the logical size of the game screen area as
+  /// exposed to the canvas after viewport transformations.
+  /// This does not match the Flutter widget size; for that see [rawSize].
+  @override
+  Vector2 get size {
+    assertHasLayout();
+    return viewport.getEffectiveSize();
+  }
+
+  /// This is the original Flutter widget size, without any transformation.
```
erickzanardo, line 57: Not sure about this name, what do you think about canvasSize?

lib/src/game/camera.dart
```diff
@@ -0,0 +1,292 @@
...
+  /// up the camera for the new parameters).
+  Vector2 get relativeOffset => _currentRelativeOffset;
+
+  final Vector2 _currentRelativeOffset = Vector2.zero();
+  final Vector2 _targetRelativeOffset = Vector2.zero();
+
+  /// If set, this determines boundaries for the camera movement.
+  ///
+  /// The camera will never move such that a region outside the world boundaries
+  /// is shown, meaning it will stop following when the object gets close to the
```
erickzanardo, line 130: A wrong

erickzanardo:
About the gesture handling, the one I prefer, is 4, cause that would be most friendly API to the user. But for that, then I am more certain that the camera and viewport must be wired directly into the game class.

But I would settle with 1, a method that would project a coordinate into the "camera/viewport" world would be useful. I would actually do that even if we go with other solutions.

lib/src/game/camera.dart
```diff
@@ -0,0 +1,292 @@
...
+  /// position, but other than that it's just prevent movement so should not
+  /// add any non-smooth movement.
+  Rect worldBounds;
+
+  Camera();
+
+  /// This smoothly updates the camera for an amount of time [dt].
+  ///
+  /// This should be called by the Game class during the update cycle.
+  void handle(double dt) {
```
erickzanardo, line 143: I think that `update` is a better name, it fits better with our other classes and methods, what do you think?

lib/src/game/viewport.dart
```diff
@@ -0,0 +1,185 @@
...
+  ///
+  /// You probably don't need to care about this if you are using a viewport.
+  Vector2 getRawSize();
+}
+
+/// This is the default viewport if you want no transformation.
+/// The rawSize is just propagated to the effective size and no
+/// translation is applied.
+/// This basically no-ops the viewport.
+class DefaultViewport extends Viewport {
```
erickzanardo, line 75: What about `FullCanvasViewport` or `FullsizeViewport`?

lib/src/game/viewport.dart
```diff
@@ -0,0 +1,185 @@
...
+class FixedRatioViewport extends Viewport {
+  Paint borderColor;
+
+  Vector2 rawSize;
+  Vector2 scaledSize;
+  Vector2 resizeOffset;
+  Vector2 fixedSize;
+  double scale;
+
+  FixedRatioViewport(
```
erickzanardo, line 125: Suggestion: `FixedResolutionViewport`. Then how the ratio is handled could be defined by optional parameters, but this would be for the future, since we only supports fixed ratio

lib/src/game/viewport.dart
```diff
@@ -0,0 +1,185 @@
...
+  @override
+  void render(Canvas c, void Function(Canvas) renderGame) {
+    c.save();
+    c.translate(resizeOffset.x, resizeOffset.y);
+    c.scale(scale, scale);
+
+    renderGame(c);
+
+    c.restore();
+    c.drawRect(
```
erickzanardo, line 158: Suggestion, instead of drawing rects to create the borders, you could use `clip` here, less code and I think it would have better perfomance. We would not even need a border color then, because the background color of the game would already be that.

erickzanardo (commented):
Added some more comments! This is looking quite nice will be a great addition when this get merged.

analysis_options.yaml
luanpotter: yes, but I think we should treat our test code as well as treat src

lib/src/game/base_game.dart
luanpotter: No, can't access this on the initializer context

lib/src/game/camera.dart
luanpotter: ?

analysis_options.yaml
erickzanardo: I don't think so, it is really common to have big test files, and it is not such a bad thing. But it bad to have big files on src, other metrics also falls on the same thing

lib/src/game/base_game.dart
luanpotter: sure

lib/src/game/viewport.dart
luanpotter: ah good idea, let me learn about clip

lib/src/game/viewport.dart
luanpotter: I'm not sure if I understand the difference, ratio and resolution is the same thing in this context?

lib/src/game/camera.dart
erickzanardo: There is a small typo on this comment, you wrote `an wrong` but it should be `a wrong`, right? Or am I wrong? haha

lib/src/game/viewport.dart
luanpotter: I like it being called default because it's super clear it doesn't do anything. FullSize or something could lead someone to believe it will expand or something.

lib/src/game/base_game.dart
luanpotter: yes, it could. but it's so easy I don't see the point, any user can wire it up if they want, the way they want. I think we should only have on Game what is crucial and would be able to be added otherwise. like if you want to use Game and use Components you can, just add a list of Components, and handle it the way you like.

lib/src/game/viewport.dart
erickzanardo: I think that the strong thing on this viewport, is the fixed resolution, not the ratio, like when I am making a game for gameboy I need a `160×144` resolution. If the viewport will respect the aspect ratio or not is another matter

analysis_options.yaml
erickzanardo: If it is so easy to wire in the game, I don't see the point of not doing, but anyway, I think we will never come to a common sense about this by that points.

I have made another comment on the PR regarding the gesture handling that I think related to this, take a look on it please.

lib/src/game/base_game.dart
luanpotter: will tackle this on the followup

analysis_options.yaml
luanpotter: will just drop this change but we need a better way to handle it

lib/src/game/viewport.dart
luanpotter: if it respects the resolution it necessarily respects the ratio. but I will change to resolution, just bear in mind it will scale up as much as possible.

luanpotter:
I will add the example tomorrow to make sure the clipRect is working.

I will tackle (4) and docs in a followup.

spydon (approved):

lib/src/game/viewport.dart
```diff
@@ -0,0 +1,155 @@
...
+/// different ratios, you can also embed games as widgets within a Flutter app.
+/// In fact, the size of the game can even change dynamically (if the layout
+/// changes or in desktop, for example).
+///
+/// For some simple games, that is not an issue. The game will just adapt
+/// to fit the screen, so if the game world is 1:1 with screen it will just
+/// be bigger or smaller. But if you want a consistent experience across
+/// platforms and players, you should use a viewport.
+///
+/// When using a viewport, [resize] should be called by the engine with
```
erickzanardo, line 23: This doc there are some references to raw size like `getRawSize` and we changed it to `canvasSize`,  I think a replace all would be good! There some variable names as well that could be renamed

erickzanardo (approved):
Some more comments, but LGTM already

lib/src/game/base_game.dart
```diff
@@ -39,8 +41,46 @@ class BaseGame extends Game with FPSCounter {
...
+  /// screen ratio for example.
+  Viewport get viewport => _viewport;
+
+  Viewport _viewport = DefaultViewport();
+  set viewport(Viewport value) {
+    if (hasLayout) {
+      final previousSize = canvasSize;
+      _viewport = value;
+      onResize(previousSize);
+    } else {
```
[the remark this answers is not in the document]
luanpotter: the point of this setter is to call onResize after changing the viewport. your code will not work

lib/src/game/viewport.dart
```diff
@@ -0,0 +1,156 @@
...
+  /// size changes.
+  void resize(Vector2 newCanvasSize);
+
+  /// This transforms the canvas so that the coordinate system is viewport-
+  /// -aware. All your rendering logic should be put inside the lambda.
+  void render(Canvas c, void Function(Canvas c) renderGame);
+
+  /// This returns the effective size, after viewport transformation.
+  /// This is not the game widget size but for all intents and purposes,
+  /// inside your game, this size should be used as the real one.
```
spydon, line 60: https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

spydon (approved):
Tiny comment, but otherwise looks good!

lib/src/game/base_game.dart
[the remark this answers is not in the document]
erickzanardo: Just change the order then

```dart
  set viewport(Viewport value) {
     _viewport = value;
    if (hasLayout) {
      final previousSize = canvasSize;
      onResize(previousSize);
    }
   }
   ```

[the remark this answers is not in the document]
erickzanardo: Nervermind, just noticed that canvasSize uses viewport getter

spydon (approved):

PR #681, opened 2021-02-23
More followups from #667 : Unifying constructor logic + small improvements and docs

luanpotter:
@pedia 's PR also has benefits down the line in our other packages, like flame_svg, flame_flare need to be updated to support optional position argument as well. Basically every package that adds children to PosComp.

This also fix all the green-green issues from the lint PR.

### DIFF
--- a/doc/components.md
+++ b/doc/components.md
@@ -17,7 +17,7 @@ The `shouldRemove` variable can be overridden or set to true and `BaseGame` will
 
 The `isHUD` variable can be overridden or set to true (default false) to make the `BaseGame` ignore the `camera` for this element, make it static in relation to the screen that is.
 
-The `onMount` method can be overridden to run initialization code for the component. When this method is called, BaseGame ensures that all the mixins which would change this component's behaviour are already resolved.
+The `onMount` method can be overridden to run initialization code for the component. When this method is called, BaseGame ensures that all the mixins which would change this component's behavior are already resolved.
 
 The `onRemove` method can be overridden to run code before the component is removed from the game, it is only run once even if the component is removed both by using the `BaseGame` remove method and the ´Component´ remove method.
 
@@ -70,7 +70,7 @@ When implementing the `render` method for your component that extends `PositionC
 Your render method should not handle where on the screen your component should be rendered. To handle where and how your component should be rendered
 use the `position`. `angle` and `anchor` properties and flame will automatically handle the rest for you.
 
-If you really want to handle the canvas translations yourself you can just omit the `super.render(canvas)` line and surpress the warning, but for most usecases this is not recommended.
+If you really want to handle the canvas translations yourself you can just omit the `super.render(canvas)` line and suppress the warning, but for most use cases this is not recommended.
 
 If you want to know where on the screen the bounding box of the component is you can use the `toRect` method.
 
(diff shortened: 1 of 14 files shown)


doc/components.md
```diff
@@ -17,7 +17,7 @@ The `shouldRemove` variable can be overridden or set to true and `BaseGame` will
 
 The `isHUD` variable can be overridden or set to true (default false) to make the `BaseGame` ignore the `camera` for this element, make it static in relation to the screen that is.
 
-The `onMount` method can be overridden to run initialization code for the component. When this method is called, BaseGame ensures that all the mixins which would change this component's behaviour are already resolved.
+The `onMount` method can be overridden to run initialization code for the component. When this method is called, BaseGame ensures that all the mixins which would change this component's behavior are already resolved.
```
luanpotter, line 20: We should standardize american (simplified) English everywhere in code.

lib/src/components/isometric_tile_map_component.dart
```diff
@@ -33,7 +33,14 @@ class IsometricTileMapComponent extends PositionComponent {
   /// Optionally provide a new tile size to render it scaled.
   Vector2 destTileSize;
 
-  IsometricTileMapComponent(this.tileset, this.matrix, {this.destTileSize});
+  IsometricTileMapComponent(
+    this.tileset,
+    this.matrix, {
+    this.destTileSize,
```
luanpotter, line 39: size doesn't quite make sense here because destTileSize + map size is used. I should probably override the size so it returns the real size of the resulting render but that is complicated because of the transformed coordinate system. might try that later.

lib/src/sprite_animation.dart
```diff
@@ -78,7 +78,7 @@ class SpriteAnimationData {
           texturePosition: texturePosition,
           textureSize: textureSize,
           loop: loop,
-          stepTimes: stepTime != null ? List.filled(amount, stepTime) : null,
+          stepTimes: List.filled(amount, stepTime),
```
luanpotter, line 81: stepTime is @required

luanpotter: also there is an assert not null on the other constructor

doc/components.md
erickzanardo: both are correct, I don't see a reason to standardize, especially because I never know when something is British spelling or not.

No need to revert this change, I just think we don't need this standard.

lib/src/sprite_animation.dart
erickzanardo: if you leave like this, the assert will not work because the list will be filled with null values, and a List of null is not a null list.

I explicit made this ternary to make the code safer.

This should be reverted. And probably have a test added to check that.

erickzanardo (commented):
Overall looks good, just one thing that I think needs to be reverted

lib/src/sprite_animation.dart
luanpotter: but then if we want that we should add an assertion on this constructor making sure stepTime is not null. it's quite weird to explicitly add a null check if null is forbidden.

erickzanardo: If we do that assert, checking if all the positions on the list have a value, that indeed would be better. Its up to you, or revert the change, or add that new assertion, I am fine with both

doc/components.md
luanpotter: it's just a nit, but I do think standards are important, making things simpler, predictable and easier to read without having to switch mental contexts. for example you can lose precious minutes (or worse) figuring out why there is no method with the word "center" on your intellisense just to later realize it was spelled differently. however your spelling is already almost 100% american afaict, so you don't need to worry about it :)

lib/src/sprite_animation.dart
luanpotter: you don't need to check every position because they are all the same (stepTime). let me add the assert and you will see

luanpotter: had to convert to a factory because of a silly `assert_in_redirecting_constructor` restriction of dart. no idea why that is the case but factory works just as well.

doc/components.md
```diff
@@ -101,32 +101,46 @@ player.render(canvas); // it will render only if the image is loaded and the pos
...
+
 This component uses an instance of `Svg` class to represent a Component that has a svg that is rendered on the game:
 
 ```dart
 Svg svg = Svg('android.svg');
-SvgComponent android = SvgComponent.fromSvg(100, 100, svg);
-android.x = 100;
-android.y = 100;
+SvgComponent android = SvgComponent.fromSvg(Vector2.all(100.0), svg);
+android.position = Vector2.all(100.0);
```
luanpotter, line 143: this was just a cursory glance, the docs are still wildly outdated

lib/src/sprite_animation.dart
```diff
@@ -66,21 +66,24 @@ class SpriteAnimationData {
...
-  }) : this.variable(
-          amount: amount,
-          amountPerRow: amountPerRow,
-          texturePosition: texturePosition,
-          textureSize: textureSize,
-          loop: loop,
-          stepTimes: stepTime != null ? List.filled(amount, stepTime) : null,
-        );
+  }) {
+    assert(stepTime != null);
```
erickzanardo, line 77: Even better!

erickzanardo (approved):

lib/src/sprite_animation.dart
luanpotter: technically we could check every element of the list in the other constructor (but that is unrelated to this change as if you come through this it will be impossible some are null and some are not).

lmk if you like that or if it's overkill.

spydon (approved):

PR #686, opened 2021-02-25
Small fixes and nits to make null-safe simpler

luanpotter:
This is one of the many benefits of having the extra effort of the non-null branch!


lib/components.dart
```diff
@@ -28,4 +22,3 @@ export 'src/components/text_component.dart';
 export 'src/extensions/vector2.dart';
 export 'src/text_config.dart';
 export 'src/timer.dart';
```
luanpotter, line 24: probably a terrible merge/rebase accident. sadly our linter doesn't seem to care for repeated export (imports work)

lib/src/anchor.dart
```diff
@@ -86,9 +86,11 @@ class Anchor {
       return _valueNames.entries.singleWhere((e) => e.value == name).key;
     } else {
       final regexp = RegExp(r'^\Anchor\(([^,]+), ([^\)]+)\)');
-      final matches = regexp.allMatches(name).first.group;
```
luanpotter, line 89: matches is never null here

lib/src/anchor.dart
```diff
@@ -86,9 +86,11 @@ class Anchor {
...
-      assert(matches != null, 'Bad Anchor format');
-      return Anchor(double.parse(matches(1)), double.parse(matches(2)));
+      final matches = regexp.allMatches(name).first.groups([1, 2]);
+      final x = matches[0];
+      final y = matches[1];
+      assert(x != null && y != null, 'bad anchor format');
```
luanpotter, line 92: I don't find this very pretty but it's the way I found to work on the notnull world

lib/src/components/joystick/joystick_component.dart
```diff
@@ -55,8 +55,6 @@ class JoystickComponent extends JoystickController {
   void removeAction(int actionId) {
     final action = children
         .firstWhere((e) => e is JoystickAction && e.actionId == actionId);
-    if (action != null) {
-      removeChild(action);
-    }
+    removeChild(action);
```
luanpotter, line 58: firstWhere never returns null

lib/src/widgets/sprite_button.dart
```diff
@@ -42,7 +42,7 @@ class _ButtonState extends State<SpriteButton> {
       onTapUp: (_) {
         setState(() => _pressed = false);
 
-        widget.onPressed?.call();
+        widget.onPressed.call();
```
luanpotter, line 45: it's @required

analysis_options.yaml
```diff
@@ -143,7 +143,7 @@ dart_code_metrics:
     - test/**
   metrics:
     number-of-arguments: 8
-    number-of-methods: 28
+    number-of-methods: 32
```
luanpotter, line 146:
```
doc/examples/particles/lib/main.dart:
WARNING MyGame - number of methods: 30
```

again, I don't see any particular value in this metric :shrug: 
I could break down the particles example but I don't think that is necessarily better.

lib/src/components/joystick/joystick_component.dart
erickzanardo: No? What happens with no item return true for the predicate?

luanpotter: throws exception

erickzanardo: We should probably write this down somewhere and handle this later

luanpotter: wdym? I think if you try to remove something that doesn't exist it should throw an error, not fail silently

erickzanardo-dx: Yeah, but iirc, this will throw a Bad element exception, which is not that obvious on what happened, maybe we could handle this and throw a more meaningful exception with a better message

luanpotter: it will throw a no element exception witch I think it's pretty clear if you called remove and provided a non-existent id

luanpotter: `IterableElementError.noElement()`

README.md
```diff
@@ -25,13 +25,13 @@ A minimalistic Flutter game engine.
 
 ## About 1.0.0
 
-This is a release candidate for version 1.0.0 of Flame. This is not ready for production use and the documentation found here is not completely updated according to the code, also everything here can be changed before the full release.
+Our goal is to release v1 soon. We are periodically launching RCs (release candidates) as we evolve the code, and we are already quite happy with where we are (but there might be still further changes). Our documentation is still not completely updated to v1, and things are still open to change.
```
spydon, line 28: might still be*

README.md
```diff
@@ -25,13 +25,13 @@ A minimalistic Flutter game engine.
...
 
-Use this version to get a preview of how the new Flame version will be and also to give the team feedback about the new structure and/or features.
+Please use this version to get a preview of how the new Flame version will be and also to give the team feedback about the new structure and/or features.
```
spydon, line 30: I would write:
Please use this version to get a preview of the new Flame version and also ...

README.md
```diff
@@ -25,13 +25,13 @@ A minimalistic Flutter game engine.
...
 
-Checkout the branches `develop-v0.x` and `master-v0.x` for the current last 0.x version. However no new updates will be added to v0 except for bugfixes.
+The `master` branch is the bleeding edge of the v1 migration. The `master-v0.x` branch is the latest v0 release (where we are still merging some patches and crucial fixes).
 
-The current v1 release is `1.0.0-rc6` on pub. The last stable version so far is `0.29.0`. Feel free to choose the one that better suits your need.
+The current v1 release is `1.0.0-rc7` on pub. The last stable version so far is `0.29.3`. Feel free to choose the one that better suits your need.
```
spydon, line 34: Could we have that badge inline here instead so that we don't have to keep it up to date?

last -> latest

README.md
```diff
@@ -99,13 +99,13 @@ You can also show on your repository that your game is made with Flame by using
 
 ## Contributing
 
-__Warning__: We are working on bringing Flame to its first stable version, updates on `0.x` versions are frozen, except for bugs. If you want to contribute on that version, by sure that it is a bugfix. For contributions for the stable version, your PR must point to the `v1.0.0` branch and by sure to talk about your contribution to the team, which is accessible on [Discord](https://discord.gg/pxrBmy4).
+__Warning__: We are working on bringing Flame to its first stable version, updates on `0.x` versions are frozen, except for crucial bug fixes. If you want to contribute on that version, please be mindful of that, and use the `master-v0.x` branch. For contributions for v1, your PR must point to the `master` branch. If in doubt, make sure to talk about your contribution to the team, either via an issue or [Discord](https://discord.gg/pxrBmy4).
```
spydon, line 102: contribute *to* that version

lib/src/anchor.dart
spydon: Does it throw an exception if there is no match when calling first?

spydon (approved):
Lgtm! Just some grammar nits.

lib/src/anchor.dart
luanpotter: let me add a test for that case

README.md
luanpotter: not sure we can inline the tag, let me try

luanpotter: good idea, looks good, fixed

PR #687, opened 2021-02-26
Refactoring the joystick code

luanpotter:
Upon further inspection of the joystick code, I finally started to understand it. This is just a first pass of improvement, introducing the JoystickElement class, removing that nasty Utils that had to create before. After playing with it now, my main question is why have both JoystickAction and JoystickDirectional. I don't see why they both exist. But I will investigate in a followup. This is just to clear my vision a bit.


lib/src/components/joystick/joystick_action.dart
```diff
@@ -11,65 +11,58 @@ import '../../../game.dart';
...
   Vector2 _dragPosition;
-  final Paint _paintBackground;
-  final Paint _paintAction;
-  final Paint _paintActionPressed;
-  final double _sizeBackgroundDirection;
   double _tileSize;
 
+  JoystickElement backgroundDirection;
+  JoystickElement action;
+  JoystickElement actionPressed;
```
luanpotter, line 35: basically every set of (rect, paint, sprite) becomes an Element

lib/src/components/joystick/joystick_action.dart
```diff
@@ -11,65 +11,58 @@ import '../../../game.dart';
...
-  final Sprite spriteBackgroundDirection;
   final double size;
-  final double sizeFactorBackgroundDirection;
+  final double _sizeBackgroundDirection;
   final EdgeInsets margin;
   final JoystickActionAlign align;
   final bool enableDirection;
-  final Color color;
-  final double opacityBackground;
-  final double opacityKnob;
```
luanpotter, line 33: these were just never used outside the constructor

lib/src/components/joystick/joystick_action.dart
```diff
@@ -11,65 +11,58 @@ import '../../../game.dart';
...
-        _paintAction = Paint()
-          ..color = color.withOpacity(opacityKnob)
-          ..style = PaintingStyle.fill,
-        _paintActionPressed = Paint()
-          ..color = color.withOpacity(opacityBackground)
-          ..style = PaintingStyle.fill {
+    double sizeFactorBackgroundDirection = 1.5,
+    Color color = Colors.blueGrey,
+    double opacityBackground = 0.5,
+    double opacityKnob = 0.8,
```
luanpotter, line 51: in a followup we could even remove this as you can provide your own Element.paint

lib/src/components/joystick/joystick_events.dart
```diff
@@ -45,28 +54,25 @@ class JoystickDirectionalEvent {
       return JoystickMoveDirectional.idle;
     }
   }
-
-  @override
-  String toString() {
```
luanpotter, line 50: I don't see much use for this

lib/src/components/joystick/joystick_action.dart
```diff
@@ -11,65 +11,58 @@ import '../../../game.dart';
...
-    this.spriteBackgroundDirection,
+    JoystickElement backgroundDirection,
+    JoystickElement action,
+    JoystickElement actionPressed,
     this.enableDirection = false,
     this.size = 50,
-    this.sizeFactorBackgroundDirection = 1.5,
     this.margin = EdgeInsets.zero,
-    this.color = Colors.blueGrey,
     this.align = JoystickActionAlign.bottomRight,
```
luanpotter, line 47: we can probably replace this with an Anchor, allowing full customization

lib/src/components/joystick/joystick_action.dart
luanpotter: (in fact I prefer that as it simplifies things)

lib/src/components/joystick/joystick_events.dart
```diff
@@ -45,28 +54,28 @@ class JoystickDirectionalEvent {
...
+
+  /// How much the knob was moved, from 0 (center) to 1 (edge).
   final double intensity;
-  final double radAngle;
+
+  /// The direction the knob was moved towards (in radians).
+  ///
+  /// It uses the trigonometric circle convention (i.e. start on the
+  /// positive x-axis and rotates counter-clockwise).
+  final double angle;
```
luanpotter, line 70: every angle we have is in radians, also radAngle sounds a bit weird

spydon (approved):
Lgtm! How would this interact with draggables on the games?

erickzanardo:
About action and directional, I think they are different things, one represents an action button, like A, B, Y, X on the switch controller. And directional is the dpad

erickzanardo (approved):
LGTM, just missing the changelog

luanpotter:
@erickzanardo I don't think so. There is nothing specific for a dpad, you can just create 4 arrow action buttons. The Directional is used to create an analog stick and the action can be used to create either an analog stick or a button. So I don't see why have both.

luanpotter:
@spydon wdym, this has nothing to do with draggables

spydon:
> @spydon wdym, this has nothing to do with draggables

It will also make use of the drag detector right?

luanpotter:
ah yes, it uses that drag detector. but I have not changed anything related to that part.

Issue #689, opened 2021-02-28
On the optimization of collidable and hascollidables aubot collide passively


au-top:
On the optimization of collidable and hascollidables,

In a game, there may be a large number of collidable instances that will not move actively but only collide passively. Can you add attr ``` activeCollision ``` of bool type to the collidable abstract class to judge whether they are active or not

,

And change the logic of collisiondetection

Decompose the collindables parameter into

```

  List<Collidable> activeCollisionList = [];
  List<Collidable> noActiveCollisionList = [];


  collidables.forEach((element) {
    if (element.activeCollision) {
      activeCollisionList.add(element);
    }else{
      noActiveCollisionList.add(element);
    }
  });

  List<Collidable> allCollisionList = [...activeCollisionList,...noActiveCollisionList];

```

And use

```
for (var x = 0; x < activeCollisionList.length ; x++) {

final collidableX = activeCollisionList[x];

for (var y = x+1; y < allCollisionList.length ; y++) {

final collidableY = allCollisionList[y];


```

Traverse

In some passive collidable scenes, the number of frames can be significantly increased

I have a crude code

spydon:
Awesome to see a contribution to the collision detection code already, since it has only been out a few days!

spydon:
This is now merged and will be released together with rc8, thanks for your contribution @au-top!

PR #690, opened 2021-02-28
Adding CollidableType to optimize collision detection with non-active collidables

au-top:
issues
#689 

[Contributor Guide]: https://github.com/flame-engine/flame/blob/master/CONTRIBUTING.md
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[pub versioning philosophy]: https://www.dartlang.org/tools/pub/versioning


luanpotter:
I like the idea, I will defer the review to @spydon . My only thought would be adding a test just making sure the logic is not skipping any collisions or doing any additional checks.

Also since the concern here is exclusively performance, did you make any tests to see the difference in execution time/fps? By the code it seems to be much faster for larger lists by reducing the O(2) loop, but it's good to make sure.

au-top:
I used this modification on one of my games, which can significantly increase the number of frames,

It's a maze game. It's a simple prototype. The walls inside are rectangular or square collision. Before, it had poor performance. In the win end, it had only 2fps. After using the modified code and marking the correct active, its performance returned to normal (60fps)

This is the source code of the maze game

[data.dart.txt](https://github.com/flame-engine/flame/files/6056326/data.dart.txt)
[main.dart.txt](https://github.com/flame-engine/flame/files/6056327/main.dart.txt)

lib/src/components/mixins/collidable.dart
```diff
@@ -4,6 +4,7 @@ import '../../geometry/rectangle.dart';
 import 'hitbox.dart';
 
 mixin Collidable on Hitbox {
+  bool activeCollision = true;
```
luanpotter, line 7: we should probably add a comment here explaining what this entails.

another thing that crossed my mind is that you split the list into two every tick. this allows you to dynamically change this boolean without issues. but could be even better for performance to keep the lists separated from the beginning. but just an idea.

au-top: I think you are right. I have added some comments, but my English is not very good. I try my best to make the meaning clear and welcome to make corrections to them. As for the behavior of allocating and creating [list] in [collisiondetection], I think that although it may reduce some performance, its complexity can make the implementation as concise as possible



But I don't know how much the... Operator will affect performance

``` final allCollisionList = <Collidable>[

...activeCollisionList,

...noActiveCollisionList,

]; ```

File: lib/src/components/mixins/collidable.dart [hunk 2]
```diff
@@ -4,6 +4,11 @@ import '../../geometry/rectangle.dart';
 import 'hitbox.dart';
 
 mixin Collidable on Hitbox {
+  /// Set whether [collidable] is active
+  /// If [active collision] is set to true, collision with other [collisionable] will be actively checked in [collisiondetection]
```
spydon, line 8: The words inside of [ ] should correspond to what they are called in the code.

lib/src/geometry/collision_detection.dart
```diff
@@ -3,11 +3,32 @@ import '../components/mixins/collidable.dart';
...
 /// call their onCollision methods accordingly.
-void collisionDetection(List<Collidable> collidables) {
-  for (var x = 0; x < collidables.length; x++) {
-    final collidableX = collidables[x];
-    for (var y = x + 1; y < collidables.length; y++) {
-      final collidableY = collidables[y];
+void collisionDetection(
+  List<Collidable> collidables,
+) {
+  final activeCollisionList = <Collidable>[];
```
spydon, line 9: I think all the variables could drop the "List" part and made to plural, so this would be `activeCollidables` for example.

spydon: And they are collidables, not collisions

spydon (commented):
Will have another look later, but it looks like a good idea!

lib/src/components/mixins/collidable.dart
au-top: OK I'll make the changes right away

lib/src/geometry/collision_detection.dart
au-top: OK I'll make the changes right away

au-top:
Attempting to modify behavior to 
> https://github.com/flame-engine/flame/pull/690#issuecomment-787399432
To get better performance

test/components/collision_detection_make_activity_test.dart
```diff
@@ -0,0 +1,225 @@
...
+import 'package:flame/components.dart';
+import 'package:flame/extensions.dart';
+import 'package:flame/geometry.dart';
+import 'package:flame/src/geometry/collision_detection.dart'
+    as CollisionDetectionFunction;
+import 'package:test/test.dart';
+
+class _TestBlock extends PositionComponent with Hitbox, Collidable {
+  Set<Collidable> oncollidables = {};
+  Set<Collidable> cachecollidables = {};
```
luanpotter, line 11: use camelCase so it's easier to read

test/components/collision_detection_make_activity_test.dart
```diff
@@ -0,0 +1,225 @@
...
+
+class _TestBlockC extends _TestBlock {
+  _TestBlockC(Vector2 pos, Vector2 size) {
+    activeCollidable = true;
+    position = pos;
+    this.size = size;
+  }
+}
+
+class _TestBlockD extends _TestBlock {
```
luanpotter, line 46: you dont need to create multiple classes if they are all the same, just create multiple instances of the same class

test/components/collision_detection_make_activity_test.dart
```diff
@@ -0,0 +1,225 @@
...
+import 'package:flame/src/geometry/collision_detection.dart'
+    as CollisionDetectionFunction;
+import 'package:test/test.dart';
+
+class _TestBlock extends PositionComponent with Hitbox, Collidable {
+  Set<Collidable> oncollidables = {};
+  Set<Collidable> cachecollidables = {};
+  _TestBlock() {
+    addShape(HitboxCircle());
+  }
```
luanpotter, line 14: add an empty line between methods

test/components/collision_detection_make_activity_test.dart
```diff
@@ -0,0 +1,225 @@
...
+
+class _TestBlockD extends _TestBlock {
+  _TestBlockD(Vector2 pos, Vector2 size) {
+    activeCollidable = true;
+    position = pos;
+    this.size = size;
+  }
+}
+
+void _testCollisionDetection(List<Collidable> collidables) {
```
luanpotter, line 54: I'm not sure why this is needed? in order to test the function we should call the real function (CollisionDetectionFunction.collisionDetection) and not replicate it

au-top: > 我不确定为什么需要这样做？为了测试该函数，我们应该调用真实函数（CollisionDetectionFunction.collisionDetection）而不是复制它

 \_Testcollisiondetection, the function is implemented from
 > https://github.com/flame-engine/flame/blob/master/lib/src/geometry/collision_ detection.dart 
 
---

\_ The testcollisiondetection function has simple and correct logic

\_Testcollisiondetection and collis ionDetectionFunction.collisionDetection

For comparison,

To keep Collis safe ionDetectionFunction.collisionDetection There will be no conflict_ Logic of different effects of test collisiondetection

au-top: Here's (CollisionDetectionFunction.collisionDetection ）And (_ Test collisiondetection) is a different implementation

au-top:
Changes have been made and the behavior is similar
> https://github.com/flame-engine/flame/pull/690#discussion_ R584244648 

the performance is improved


This is the API now

```
class _TestBlock extends PositionComponent with Hitbox, Collidable {}
class CollidablesTestBox extends BaseGame with HasCollidables {
  CollidablesTestBox (){
      activeCollidable = false;
       //  assert error 
      // activeCollidable = true;
  }
}

final testBox = CollidablesTestBox();
final tba = _TestBlockA(Vector2(0, 0), Vector2(10, 10));

//mock add
testBox.handleCollidables({}, [tba]);

// set new activeState
testBox.setActiveCollidable(tbc, true);

//mock game running
testBox.handleCollidables({}, [tba]);

print(tba.activeCollidable);

```

au-top:
> Changes have been made and the behavior is similar
> 
> > [#690 (comment)](https://github.com/flame-engine/flame/pull/690#discussion)_ R584244648
> 
> the performance is improved
> 
> This is the API now
> 
> ```
> 
> class _TestBlock extends PositionComponent with Hitbox, Collidable {
> 
> _TestBlock(){
>       activeCollidable = false;
>        //  assert error 
>       // activeCollidable = true;
> }
> 
> }
> class CollidablesTestBox extends BaseGame with HasCollidables {
>   CollidablesTestBox (){
> 
>   }
> }
> 
> final testBox = CollidablesTestBox();
> final tba = _TestBlock();
> 
> //mock add
> testBox.handleCollidables({}, [tba]);
> 
> // set new activeState
> testBox.setActiveCollidable(tba, true);
> 
> //mock game running
> testBox.handleCollidables({}, [tba]);
> 
> print(tba.activeCollidable);
> ```

Hascollidables related test code has been completed

lib/src/components/mixins/collidable.dart
```diff
@@ -4,6 +4,26 @@ import '../../geometry/rectangle.dart';
 import 'hitbox.dart';
 
 mixin Collidable on Hitbox {
+  /// Set whether [Collidable] is active
+  /// If [activeCollidable] is set to true, collision with other [Collidable] will be actively checked in [collisionDetection]
+  /// [Collidable] whose [activeCollidable] is set to false will not actively check collisions with other [Collidable],
+  /// so as to prevent collision detection between stationary [Collidable] and other stationary [Collidable], so as to improve the performance of [collisionDetection]
+  bool _activeCollidable = null;
```
wolfenrain, line 11: Maybe it is cleaner to use a Enum value for this instead of null and check against that?

This will make it easier to maintain for null-safety but also wont create unexpected behaviour as you can just compare against enum values to get the bool value for the getter.

```dart
enum IsCollidable {
  isNotDefined,
  isActive,
  isNotActive,
}
```

That way we can keep the checks in that compare with `null`(but instead of null the appropriate enum value).

Just an idea, I will refer/follow whatever @spydon will suggest :D

lib/src/components/mixins/has_collidables.dart
```diff
@@ -1,14 +1,77 @@
...
 import '../../geometry/collision_detection.dart';
 import '../component.dart';
 
 mixin HasCollidables on BaseGame {
-  final List<Collidable> _collidables = [];
+  // final List<Collidable> _collidables = [];
+  final List<Collidable> _activeCollidables = [];
+  final List<Collidable> _noActiveCollidables = [];
+
+  List<Collidable> showActiveCollidables() => [
```
wolfenrain, line 13: This should use a `UnmodifiableListView` instead, it performance better as we don't have to copy the pointers.

lib/src/components/mixins/has_collidables.dart
```diff
@@ -1,14 +1,77 @@
...
+        ..._activeCollidables,
+      ];
+
+  List<Collidable> showNoActiveCollidables() => [
```
wolfenrain, line 17: This should use a `UnmodifiableListView` instead, it performance better as we don't have to copy the pointers.

lib/src/components/mixins/has_collidables.dart
```diff
@@ -1,14 +1,77 @@
...
+    _noActiveCollidables.removeWhere((element) {
+      if (element.activeCollidable) {
+        _activeCollidables.add(element);
+      }
+      return element.activeCollidable;
+    });
+    _noActiveCollidables.addAll(_waitAssignConllidables);
+  }
+
+  /// update a [_activeCollidable] of [Collidable]
```
wolfenrain, line 70:
```suggestion
  /// Update a [_activeCollidable] of [Collidable]
```

wolfenrain (commented):
Left a small review, few nits/suggestiosn

au-top:
I have a question, why do I need to work in [base]_ [update] in [game] exists


```

if (this is HasCollidables) {

(this as HasCollidables).handleCollidables(_ removeLater, _ addLater);

}

```

lib/src/components/mixins/has_collidables.dart
au-top: > This should use a `UnmodifiableListView` instead, it performance better as we don't have to copy the pointers.

Thank you. He will be corrected in the near future

spydon (commented):
On a second thought about this `active` field, is it really needed?
Doesn't the code both get cleaner and clearer if the user handles which `Collidable`s that are active by adding and removing them from the `_collidables` list in `HasCollidables` mixin? Either by making that list (which should be a set) into public or create access methods for it.

au-top:
@spydon 
I agree with you, but I think there are some problems in [update] of [gamebase]

```

if (this is HasCollidables) {

(this as HasCollidables).handleCollidables(_ removeLater, _ addLater);

}

```Yes https://github.com/flame-engine/flame/pull/690 I think it will be better for users to manage themselves after deleting this code, and open part of the API

**Comment by au-top** [comment]
I think calling collision detection in [gamebase]'s [update] is not a good decision.

Is it possible to delete those / or parameter control collision detection codes, remove the code about [active], and then modify the implementation of hascollidables public part 

To let users manage who to detect and with whom to detect

File: lib/src/components/mixins/collidable.dart (hunk 8 above)
Comment (au-top) [reply lib/src/components/mixins/collidable.dart:11]: I agree that it would be better if spydon is true, and I will update these codes if [Active] is really needed

> https://github.com/flame-engine/flame/pull/690#pullrequestreview-600680436

>https://github.com/flame-engine/flame/pull/690#issuecomment-787997490

File: lib/src/components/mixins/has_collidables.dart [hunk 12]
```diff
@@ -1,14 +1,77 @@
...
+    /// auto [reAssignConllidablesActive]
+    if (needReAssignConllidablesActive) {
+      reAssignConllidablesActive();
+      needReAssignConllidablesActive = false;
+    }
+
+    collisionsDetection(_activeCollidables, _noActiveCollidables);
+  }
+
+  void reAssignConllidablesActive() {
```
spydon, line 49: Spelling `reassignActiveCollidables` maybe?

spydon:
> I think calling collision detection in [gamebase]'s [update] is not a good decision.
> 
> Is it possible to delete those / or parameter control collision detection codes, remove the code about [active], and then modify the implementation of hascollidables public part
> 
> To let users manage who to detect and with whom to detect And when to call

That will make it much more work for the user, if it is a more advanced user they can skip using the `HasCollidables` mixin and do the call to the collision detection system themselves.

test/components/collision_detection_make_activity_test.dart
au-top: I apologize for the code that I am confused about, I call the real function Q_Q here, the code pointed out by luanpotter comes from https://github.com/flame-engine/flame/blob/master/lib/src/geometry/collision
This is the code from flame-master that is available and does not need to be tested to ensure the correct logic

,

[CollisionDetectionFunction.collisionDetection] is the code that needs to be tested,

I use set to merge the results of the above two functions and compare the length of the set with the previous results to see if [CollisionDetectionFunction.collisionDetection] has skipped any correct collision detection

lib/src/geometry/collision_detection.dart
```diff
@@ -1,18 +1,32 @@
...
+    /// The property [activecollidables] of [collidablex] is  true
+    final collidableX = activeCollidables[x];
+
+    /// Detection [activeCollidables]
+    for (var y = x + 1; y < activeCollidables.length; y++) {
+      aloneCollisionDetection(collidableX, activeCollidables[y]);
+    }
+
+    /// Detection [noActiveCollidables]
+    for (var y = 0; y < noActiveCollidables.length; y++) {
```
spydon, line 28: This can use the `for in` loop notation.

au-top: @spydon 
OK, I'll try to make the [active] API more usable. I agree with you, but I still have doubts about how to use [oncollision] simply. For example, when a collisionable box wants to climb over the obstacles made up of other collisionable boxes, I have doubts about this callback method. I hope to improve these interfaces so that they can be used effectively, I'd like to have a discussion. I've seen it

https://github.com/flame-engine/flame/blob/master/doc/collision_ detection.md

I think it needs some knowledge of mathematics. At the same time, I'm sorry for my lack of knowledge and experience in 2D game engine. Thank you for your attention to my news release

lib/src/geometry/collision_detection.dart
```diff
@@ -1,18 +1,32 @@
 import '../../extensions.dart';
 import '../components/mixins/collidable.dart';
 
+void aloneCollisionDetection(Collidable collidableX, Collidable collidableY) {
```
spydon, line 4: This needs a clearer name for what it does.
I'll try to come up with a suggestion.

au-top:
@spydon 
OK, I'll try to make the [active] API more usable. I agree with you, but I still have doubts about how to use [oncollision] simply. For example, when a collisionable box wants to climb over the obstacles made up of other collisionable boxes, I have doubts about this callback method. I hope to improve these interfaces so that they can be used effectively, I'd like to have a discussion. I've seen it

https://github.com/flame-engine/flame/blob/master/doc/collision_ detection.md

I think it needs some knowledge of mathematics. At the same time, I'm sorry for my lack of knowledge and experience in 2D game engine. Thank you for your attention to my news release

spydon:
> @spydon 
> OK, I'll try to make the [active] API more usable. I agree with you, but I still have doubts about how to use [oncollision] simply. For example, when a collisionable box wants to climb over the obstacles made up of other collisionable boxes, I have doubts about this callback method. I hope to improve these interfaces so that they can be used effectively, I'd like to have a discussion. I've seen it
> 
> https://github.com/flame-engine/flame/blob/master/doc/collision_ detection.md
> 
> I think it needs some knowledge of mathematics. At the same time, I'm sorry for my lack of knowledge and experience in 2D game engine. Thank you for your attention to my news release

What you are talking about sounds more like a physics engine (which you can make a simple one on top of the collision detection system), but if you want a full physics engine you should look at [Forge2D](https://github.com/flame-engine/forge2d) and flame_forge2d.

lib/src/geometry/collision_detection.dart
```diff
@@ -1,18 +1,32 @@
...
-      final points = intersections(collidableX, collidableY);
-      if (points.isNotEmpty) {
-        collidableX.onCollision(points, collidableY);
-        collidableY.onCollision(points, collidableX);
-      }
+void collisionsDetection(
+  List<Collidable> activeCollidables,
+  List<Collidable> noActiveCollidables,
+) {
+  for (var x = 0; x < activeCollidables.length; x++) {
```
spydon, line 18: I don't think we need to maintain two lists, just skipping the element in this outer loop if it is not active should be a very cheap operation.

spydon: No, because then we have to create new lists on every iteration, that is too expensive.
Just keep the original `collidables` list and check if each element is active in the loop, so just like your example, but with only one list. And then we don't have to update the other two lists in each tick to check if a collidable has changed its active state.

lib/src/components/mixins/has_collidables.dart
```diff
@@ -1,14 +1,76 @@
...
 import '../../components/mixins/collidable.dart';
 import '../../geometry/collision_detection.dart';
 import '../component.dart';
 
 mixin HasCollidables on BaseGame {
-  final List<Collidable> _collidables = [];
+  final List<Collidable> _activeCollidables = [];
+  final List<Collidable> _noActiveCollidables = [];
+
+  UnmodifiableListView<Collidable> showActiveCollidables() =>
```
wolfenrain, line 14: The return type can stay of type `List`

lib/src/geometry/collision_detection.dart
au-top: I think I know, your idea, he will be changed later, thanks

lib/src/components/mixins/has_collidables.dart
au-top: Thank you. It will be changed later

lib/src/geometry/collision_detection.dart
au-top: I may use

```
   for (var x = 0; x < allActiveCollidables.length; x++) {
    if(!allActiveCollidables[x].activeCollidable){
         continue;
    }
    for(var y=0;y<allActiveCollidables.length,y++)

```

**Note that y = 0**

Otherwise, these codes may skip some collision detection. If you have other ideas, can you use some codes to prompt me

spydon: The problem with those loops is that it will make the solution more inefficient when you have less inactive collidables than active ones, which I'm guessing will be the case for most people. I will try to think of a better solution.
Will you allow me to push to your branch? Then I can start writing the docs too.

au-top: > 这些循环的问题在于，当您的非活动可折叠对象少于活动的可折叠对象时，它将使解决方案的效率降低，我想大多数人都会遇到这种情况。我将尝试考虑一个更好的解决方案。
> 你能让我推到你的分店吗？然后，我也可以开始编写文档了。

OK, I've given you permission,In some places, I believe better improvements can be made.

spydon: Thanks!

spydon:
@au-top what do you think about the current solution? (sorry about the force-push on your branch, rebased on master)
If you are okay with this solution I'll start to write up the documentation for it.

erickzanardo (approved):
LGTM

test/components/collidable_type_test.dart
```diff
@@ -0,0 +1,262 @@
...
+    bool containsSelf = true,
+  }) {
+    return collisions
+            .toSet()
+            .containsAll(otherCollidables.toList()..remove(this)) &&
+        collisions.length == otherCollidables.length - (containsSelf ? 1 : 0);
+  }
+
+  @override
+  void onCollision(Set<Vector2> points, Collidable other) {
```
wolfenrain, line 43:
```suggestion
  void onCollision(Set<Vector2> intersectionPoints
, Collidable other) {
```

test/components/collidable_type_test.dart
```diff
@@ -0,0 +1,262 @@
...
+          true,
+        );
+        expect(
+          inactives.fold<bool>(
+            true,
+            (hasCorrectCollisions, c) => c.collisions.isEmpty,
+          ),
+          true,
+        );
+        //expect(blockB.collisions.length, 0);
```
wolfenrain, line 258: Can this be removed?

wolfenrain (approved):
Missing a CHANGELOG entry,  oh and a few nits

wolfenrain (approved):

au-top:
> @au-top what do you think about the current solution? (sorry about the force-push on your branch, rebased on master)
> If you are okay with this solution I'll start to write up the documentation for it.

It doesn't matter. I think the solution is good

PR #692, opened 2021-03-02
Second batch of improvements from the Null Safe PR

luanpotter:
These are Ericks suggestions from the NS PR

* use VoidCallback instead of void Function()
* remove param since default value is already intersect

### DIFF
--- a/lib/src/effects/combined_effect.dart
+++ b/lib/src/effects/combined_effect.dart
@@ -1,4 +1,5 @@
 import 'dart:math';
+import 'dart:ui';
 
 import 'package:meta/meta.dart';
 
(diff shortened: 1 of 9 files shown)


wolfenrain (approved):
LGTM, but missing a CHANGELOG entry.

lib/src/effects/combined_effect.dart
```diff
@@ -1,4 +1,5 @@
 import 'dart:math';
+import 'dart:ui';
```
erickzanardo, line 2: TIL this comes from `dart:ui`, I have always thought it came from widget.dart, interesting

erickzanardo (approved):

lib/src/effects/combined_effect.dart
luanpotter: it is exported in many places, I chose dart ui because its the one that fit the best with the effects

PR #712, opened 2021-03-18
Memoizing paint creation on BasicPalette to enable direct use

erickzanardo:
# Description

The BasicPalette paint field was actually a getter that always created a new Paint, that could bring perfomance issues when the BasicPalette was directly used inside a render function, which seems to me a fair use for that class, because it would create a new paint instance every time. Since the color is final on that class, it makes sense to memoize the paint value, so we don't create unnecessary instances.

I have implemented a very naive memoization function and specific for the paint class as I wasn't sure if it was something worth to be generic to be used elsewhere, but I am open to suggestions.

## Checklist

  - API is still the same, no changes needed
- [x] I read and followed the [Flutter Style Guide].

## Breaking Change

### DIFF
--- a/packages/flame/lib/src/palette.dart
+++ b/packages/flame/lib/src/palette.dart
@@ -1,11 +1,26 @@
 import 'dart:ui';
 
+class _MemoizedPaint {
+  final Color color;
+  Paint? _paint;
+
+  _MemoizedPaint(this.color);
+
+  Paint get value {
+    _paint ??= Paint()..color;
+    return _paint!;
+  }
+}
+
 class PaletteEntry {
   final Color color;
+  late _MemoizedPaint _paint;
 
-  Paint get paint => Paint()..color = color;
+  Paint get paint => _paint.value;
 
-  const PaletteEntry(this.color);
+  PaletteEntry(this.color) {
+    _paint = _MemoizedPaint(color);
+  }
 
   PaletteEntry withAlpha(int alpha) {
     return PaletteEntry(color.withAlpha(alpha));
@@ -25,6 +40,6 @@ class PaletteEntry {
 }
 
 class BasicPalette {
-  static const PaletteEntry white = PaletteEntry(Color(0xFFFFFFFF));
-  static const PaletteEntry black = PaletteEntry(Color(0xFF000000));
+  static PaletteEntry white = PaletteEntry(Color(0xFFFFFFFF));
+  static PaletteEntry black = PaletteEntry(Color(0xFF000000));
 }
--- a/packages/flame/CHANGELOG.md
+++ b/packages/flame/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
  - Fix input bug with other anchors than center
+ - Memoizing the paint creation on BasicPalette to enable direct use
 
 ## 1.0.0-rc8
  - Migrate to null safety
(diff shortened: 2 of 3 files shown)


packages/flame/lib/src/palette.dart
```diff
@@ -25,6 +40,6 @@ class PaletteEntry {
 }
 
 class BasicPalette {
-  static const PaletteEntry white = PaletteEntry(Color(0xFFFFFFFF));
-  static const PaletteEntry black = PaletteEntry(Color(0xFF000000));
+  static PaletteEntry white = PaletteEntry(Color(0xFFFFFFFF));
+  static PaletteEntry black = PaletteEntry(Color(0xFF000000));
```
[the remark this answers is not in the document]
erickzanardo: That is another option too, I went with memoization to somehow try tomara closer to the previous implementation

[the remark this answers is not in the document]
spydon: We're still in RC, break eeeeverything! :D

packages/flame/lib/src/palette.dart
```diff
@@ -1,11 +1,26 @@
...
+  final Color color;
+  Paint? _paint;
+
+  _MemoizedPaint(this.color);
+
+  Paint get value {
+    _paint ??= Paint()..color;
+    return _paint!;
+  }
+}
```
renancaraujo, line 13: This does the same thing:

```suggestion
class _MemoizedPaint {
  final Color color;
  late final Paint value = Paint()..color = color;

  _MemoizedPaint(this.color);
}
```

packages/flame/lib/src/palette.dart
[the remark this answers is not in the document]
renancaraujo: Paint has several mutable fields (color, mask, style...), keeping a static reusable instance of it to be shared possibly shared between components can lead to bugs.

renancaraujo (commented):
I am not sure about caching a paint instance, since a consumer (component) can locally change values such as blend mode in a cached instance and unintentionally make this change global.

packages/flame/lib/src/palette.dart
erickzanardo: No it doesn't, your code already create an instance of the Paint, when the original code, only create the instance when it is accessed

erickzanardo:
> I am not sure about caching a paint instance, since a consumer (component) can locally change values such as blend mode in a cached instance and unintentionally make this change global.

That is a fair point, my biggest issue with the current implementation, is that is not clear that calling `BasicaPalette.white.paint` will create a new instance every time. Maybe if we just get transform that getter into a method called `toPaint()`, or something like that, it would be clear enough that people should cache this and not use directly? wdyt?

packages/flame/lib/src/palette.dart
renancaraujo: Yes, it does, `late` makes the field be lazy, everything after `=` is executed only when accessed for the first time.

erickzanardo: Oh, that is interesting, I thought late was just a way for the compiler to "trust" the programmer.

I will not apply this though, because we will probably don't even go with this solution

renancaraujo:
Totally agree with the transform to method idea.

erickzanardo:
Closing this in favor of the new approach that arose from the discussion here.

PR #738, opened 2021-04-02
Fix camera when not following component

luanpotter:
Found this bug on camera where if it's not following a component it doesn't respect relative offset, and therefore move doesn't work.

### DIFF
--- a/packages/flame/lib/src/game/camera.dart
+++ b/packages/flame/lib/src/game/camera.dart
@@ -222,10 +222,10 @@ class Camera {
   }
 
   Vector2 _target() {
-    if (follow == null) {
-      return Vector2.zero();
-    }
     final screenDelta = gameRef.size.clone()..multiply(_currentRelativeOffset);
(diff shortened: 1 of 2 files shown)


packages/flame/lib/src/game/camera.dart
```diff
@@ -221,12 +220,13 @@ class Camera {
...
+    return gameRef.size.clone()..multiply(_currentRelativeOffset);
+  }
+
   Vector2 _target() {
-    if (follow == null) {
-      return Vector2.zero();
-    }
-    final screenDelta = gameRef.size.clone()..multiply(_currentRelativeOffset);
-    final attemptedTarget = follow! - screenDelta;
+    final target = _currentCameraDelta ?? follow ?? Vector2.zero();
```
luanpotter, line 228: this way move bypasses follow but still respect screen delta and bounds

spydon (approved):

Issue #715, opened 2021-03-19
Possibility to set a package name for asset loading


GwaedGlas:
# Problem to solve
I have implemented my first game with flame in the last days. Because I want to divide this from the app where this should be integrated, I decided to handle the flame part in a separate package.
This guides to the problem, that the assets are always loaded from assets/images/ or assets/audios, which is leaking of the package name in the file path. So it is needed to have the images in the app, instead of the package, where the path is packages/myPackageName/assets/images.

# Proposal
Having the possibility to set a prefix for the assets. In this case it would be enough to set a packagename. So the prefix "packages/myPackageName" could be generated. 
This shouldn't be a static value, because of the possibility of implementing different games for a game collection app, where you have different packages.

luanpotter:
Hey @GwaedGlas that is a good idea! In fact in flame audio (and audioplayers) I already made it configurable after a similar request. For flame itself, we have asset and images loading, and indeed those were still hardcoded (my bad).

Here is my proposed solution: https://github.com/flame-engine/flame/pull/741
What do you think?

luanpotter:
Closing since the proposed PR was merged and relased.

PR #741, opened 2021-04-03
Make prefix configurable for image and asset loading

luanpotter:
Fixing https://github.com/flame-engine/flame/issues/715

### DIFF
--- a/doc/structure.md
+++ b/doc/structure.md
@@ -1,6 +1,6 @@
 # Structure
 
-Flame has a proposed structure for your project that includes the standard Flutter `assets` directory in addition to two children: audio and images.
+Flame has a proposed structure for your project that includes the standard Flutter `assets` directory in addition to two children: `audio` and `images`.
 
 If using the following example code:
 
--- a/doc/structure.md
+++ b/doc/structure.md
@@ -1,7 +1,6 @@
 # Structure
 
-The only structure you are required to comply to when making a Flame game is having an assets folder
-with two sub folders: audio and images.
+Flame has a proposed structure for your project that includes the standard Flutter `assets` directory in addition to two children: audio and images.
 
 If using the following example code:
 
@@ -12,7 +11,7 @@ If using the following example code:
   Flame.images.load('enemy.png');
 ```
 
-The file structure would have to be:
+The file structure Flame would expect to find the files in would be:
 
 ```
 .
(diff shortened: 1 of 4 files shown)


erickzanardo (commented):
Changes LGTM, but we should mention this new feature on the docs and add a CHANGELOG entry

luanpotter:
I can add to the changelog, but I am not sure if such a small edge case is worth changing the documentation for...

erickzanardo:
> I can add to the changelog, but I am not sure if such a small edge case is worth changing the documentation for...

It sure is, otherwise how can people know that they have a different prefix? I think just a small paragraph on the structure section of our docs would be enough

luanpotter:
They would look at the code to see what the prefix is and clearly see how to change it. But if you want I will add to the docs.

erickzanardo:
> They would look at the code to see what the prefix is and clearly see how to change it. But if you want I will add to the docs.

Looking at the code is never enough, look at the amount of questions we get on AP from people asking how to stop music played by AudioCache...

luanpotter:
> Looking at the code is never enough, look at the amount of questions we get on AP from people asking how to stop music played by AudioCache...

I don't think those people would look at docs either, in fact we say how it works on the docs and readme on AP as well :p

erickzanardo (approved):

ponderMuse:
Hi,

I have started placing most of a game's code into a separate package which is then used by two different applications (one for mob and one for browsers) so the capability of having Flame load assets from package dependencies is certainly handy.

Are there any estimates as to when this updated code is likely to make it into master?

Cheers,
PM

spydon:
@ponderMuse this was released with `rc9` and `rc10` is the newest version.

ponderMuse:
@spydon I was using rc8. Have now updated to rc9 after your comment and can now load assets through package as needed. Cheers!

spydon:
> @spydon I was using rc8. Have now updated to rc9 after your comment and can now load assets through package as needed. Cheers! 

Why not use rc10? :)

ponderMuse:
@spydon can do if it is available. It's just that [here](https://pub.dev/packages/flame/install) still shows rc9 as being the latest pre-release.

spydon:
> @spydon can do if it is available. It's just that [here](https://pub.dev/packages/flame/install) still shows rc9 as being the latest pre-release.

Yeah, pub is ordering them in a weird way. rc10 is the newest.

ponderMuse:
@spydon I have tried rc10 but my project errors with:

The function 'TextConfig' isn't defined.

rc9 has the class in package:flame/src/text_config.dart).

Where can I find TextConfig in rc10 or similar?

spydon:
@ponderMuse these discusssions don't really fit on this PR, please join our discord to ask more questions https://discord.com/invite/pxrBmy4 (Our docs are updated for how to use text in rc10 too)

ponderMuse:
I will take a look at the docs. Thanks for all the good help today @spydon

PR #766, opened 2021-04-20
Adding SpriteAnimationGroupComponent

erickzanardo:
# Description

This PR adds a new component: `SpriteAnimationGroupComponent`.

This component is an extension of the `SpriteAnimationComponent` which allows a single component to hold several animations and change the current in runtime.

https://user-images.githubusercontent.com/835641/115400896-2b77db00-a1c0-11eb-8a21-1dfcf7f55e76.mov

## Checklist

- [x] I read and followed the [Flutter Style Guide].

## Breaking Change


.github/workflows/cicd.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           java-version: '12.x'
       - uses: subosito/flutter-action@v1
         with:
-          channel: 'beta'
+          channel: 'stable'
```
erickzanardo, line 21: Changed our CI to stable because apparently on beta, something changed on formatter tool, which was causa to the analyze tool work fine on local, but break on CI.

Since web and even desktop is on stable now, I think it is safe for us to use stable on CI.

examples/pubspec.yaml
```diff
@@ -24,25 +24,8 @@ flutter:
...
-    - assets/images/parallax/bg.png
-    - assets/images/parallax/foreground-trees.png
-    - assets/images/parallax/mountain-far.png
-    - assets/images/parallax/mountains.png
-    - assets/images/parallax/trees.png
+    - assets/images/
+    - assets/images/animations/
+    - assets/images/tile_maps/
+    - assets/images/layers/
+    - assets/images/parallax/
```
luanpotter, line 31: I thought I had already done this 🤔

packages/flame/lib/src/components/sprite_animation_group_component.dart
```diff
@@ -0,0 +1,67 @@
+import 'dart:ui';
+
+import '../extensions/vector2.dart';
+import '../sprite_animation.dart';
+import 'sprite_animation_component.dart';
+
+export '../sprite_animation.dart';
+
+class SpriteAnimationGroupComponent<T> extends SpriteAnimationComponent {
```
luanpotter, line 9: we could have a brief doc here explaining what this class is?

luanpotter (approved):
LGTM, great idea!

examples/pubspec.yaml
erickzanardo: Tbh, me too hahhahahah

packages/flame/lib/src/components/sprite_animation_group_component.dart
```diff
@@ -0,0 +1,67 @@
...
+  late Map<T, SpriteAnimation> animations;
+
+  /// Creates a component with an empty animation which can be set later
+  SpriteAnimationGroupComponent({
+    required this.animations,
+    Vector2? position,
+    Vector2? size,
+    this.current,
+    Paint? overridePaint,
+    bool removeOnFinish = false,
```
spydon, line 23: What does it mean for this to remove when finishing? Can it ever finish, since it can change states?

packages/flame/lib/src/components/sprite_animation_group_component.dart
```diff
@@ -0,0 +1,67 @@
...
+  }) : super(
+          position: position,
+          size: size,
+          overridePaint: overridePaint,
+          removeOnFinish: removeOnFinish,
+        );
+
+  /// Creates a SpriteAnimationGroupComponent from a [size], an [image] and [data]. Check [SpriteAnimationData] for more info on the available options.
```
spydon, line 31: Line break would be good here

packages/flame/lib/src/components/sprite_animation_group_component.dart
spydon: Ah, it means that once a "child" finishes it is removed, but if the state is chosen again it will start it again?

packages/flame/test/components/sprite_animation_group_component_test.dart
```diff
@@ -0,0 +1,50 @@
...
+enum AnimationState {
+  idle,
+  running,
+}
+
+void main() async {
+  // Generate a image
+  final image = await generateImage();
+
+  group('SpriteAnimationGroupComponent test', () {
```
spydon, line 15: Maybe a test when `removeOnFinish` is used too?

spydon (approved):
Some tiny comments. This will be sooo handy to have, can't believe we didn't have it before.

packages/flame/lib/src/components/sprite_animation_group_component.dart
erickzanardo: Yeah, whenever a state that don't loop finishes and this Flag is true, the component will be removed.

At first I thought this would be a very edge case, but I think this is quite useful. Imagine a spaceship component and one of its state is it exploding. With this Flag true and the explosion state not being a loop would make life very easy.

packages/flame/test/components/sprite_animation_group_component_test.dart
erickzanardo: This method is already tested on the parent class, I am not sure if it worth re creating those tests for this class which don't adds any new behavior to that method.

Or do you see any cases that I may be missing?

packages/flame/lib/src/components/sprite_animation_group_component.dart
spydon: I don't think I quite get it still. The whole group here is `removeOnFinish`, isn't the explosion just one of the states within the group?

erickzanardo: Yeah, I mean the way it is implemented right now, this flag is kind of a global flag for all states, any state that doesn't loop will make the component be removed.

Do you think we should have a flag per state?

spydon: Not sure what is more clear, we should explain it in the docs at least.
Your are thinking that the most used implementation will be that the initial state is a looping one and that any non-looping state should remove the component if it swaps to that one?
I think to me it would be more clear to have the flag per state.

erickzanardo: Yes that was what I was thinking.

But to be sure and cover all the cases I will create a flag per state then.

To do that I think it will not be worth that this class extends SpriteAnimatiomComponent anymore, as they will differ too much, but I don't think this is a problem.

Tomorrow I will do that follow up.

spydon: Cool cool!

.github/workflows/cicd.yml
wolfenrain: And as our doc states:
> Flame keeps it support on the stable channel. The dev, beta and master channel should work, but we don't support them. This means that issues happening outside the stable channel are not a priority.

We don't have to test for it anymore :-)

wolfenrain: But it might be interesting to have a `cicd-beta.yml` with jobs that are allowed to fail to see if we have to prepare thing for the next release.

spydon (approved):
Great, much clearer I think

PR #769, opened 2021-04-21
Making velocities and deltas have the game and global values

erickzanardo:
# Description

Deltas and velocities on the gestures events didn't had their global and scaled values. This PR changes that and also adds a Zoom example which uses those properties.

The zoom using the scroll is acting very weird, but I think this is a probably a problem with the scale rather than the example itself. But I am not sure so I could use some help here (check the video to see how it is acting)

https://user-images.githubusercontent.com/835641/115578791-f640bb00-a29b-11eb-99f0-eaa2a659014b.mov

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/examples/analysis_options.yaml
+++ b/examples/analysis_options.yaml
@@ -7,7 +7,7 @@ analyzer:
     implicit-dynamic: false
 
   plugins:
-  #  - dart_code_metrics
+    - dart_code_metrics
 
 linter:
   rules:
--- a/examples/analysis_options.yaml
+++ b/examples/analysis_options.yaml
@@ -7,7 +7,7 @@ analyzer:
     implicit-dynamic: false
 
   plugins:
-    - dart_code_metrics
+  #  - dart_code_metrics
 
 linter:
   rules:
--- a/examples/lib/stories/camera_and_viewport/zoom.dart
+++ b/examples/lib/stories/camera_and_viewport/zoom.dart
@@ -1,0 +1,50 @@
+import 'package:flame/components.dart';
+import 'package:flame/game.dart';
+import 'package:flame/gestures.dart';
+
+class ZoomGame extends BaseGame with ScrollDetector, ScaleDetector {
+  final Vector2 viewportResolution;
+  late SpriteComponent flame;
+
+  Vector2? lastDelta;
+
+  ZoomGame({
+    required this.viewportResolution,
+  });
+
+  @override
+  Future<void> onLoad() async {
+    final flameSprite = await loadSprite('flame.png');
+
+    viewport = FixedResolutionViewport(viewportResolution);
+
+    final flameSize = Vector2(149, 211);
+    add(
+      flame = SpriteComponent(
+        sprite: flameSprite,
+        size: flameSize,
+      )..anchor = Anchor.center,
+    );
+    camera.followComponent(
+      flame,
+      relativeOffset: Anchor.center.toVector2(),
+    );
+  }
+
+  static const zoomPerScrollUnit = 0.001;
+  @override
+  void onScroll(PointerScrollInfo event) {
+    camera.zoom += event.scrollDelta.game.y * zoomPerScrollUnit;
+  }
+
+  @override
+  void onScaleUpdate(ScaleUpdateInfo info) {
+    final scale = lastDelta;
+    if (scale != null) {
+      final delta = info.scale.game - scale;
+      camera.zoom += delta.y;
+    }
+
+    lastDelta = info.scale.game;
+  }
+}
--- a/examples/lib/stories/camera_and_viewport/zoom.dart
+++ b/examples/lib/stories/camera_and_viewport/zoom.dart
@@ -27,7 +27,7 @@ class ZoomGame extends BaseGame with ScrollDetector, ScaleDetector {
     );
     camera.followComponent(
       flame,
-      relativeOffset: Anchor.center.toVector2(),
+      relativeOffset: Anchor.center,
     );
   }
 
--- a/examples/lib/stories/camera_and_viewport/zoom.dart
+++ b/examples/lib/stories/camera_and_viewport/zoom.dart
@@ -6,7 +6,7 @@ class ZoomGame extends BaseGame with ScrollDetector, ScaleDetector {
   final Vector2 viewportResolution;
   late SpriteComponent flame;
 
-  Vector2? lastDelta;
+  Vector2? lastScale;
 
   ZoomGame({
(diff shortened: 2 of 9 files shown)


examples/analysis_options.yaml
```diff
@@ -7,7 +7,7 @@ analyzer:
     implicit-dynamic: false
 
   plugins:
-    - dart_code_metrics
+  #  - dart_code_metrics
```
luanpotter, line 10: :'(

packages/flame/analysis_options.yaml
```diff
@@ -7,7 +7,7 @@ analyzer:
     implicit-dynamic: false
 
   plugins:
-    - dart_code_metrics
+  #  - dart_code_metrics
```
luanpotter, line 10: :`(

examples/analysis_options.yaml
erickzanardo: I always forget to put this plugin back.

This plugin is really causing headaches for me :(

packages/flame/lib/src/gestures/events.dart
```diff
@@ -26,6 +26,23 @@ class EventPosition {
   EventPosition(this._game, this._localPosition, this._globalPosition);
 }
 
+/// [EventDelta] converts deltas based events to two different values (game and global).
+///
+/// global: this is the raw value received by the event without any scale applied to it
+/// game: the scalled value applied all the game transformations
```
luanpotter, line 32:
```suggestion
/// [global]: this is the raw value received by the event without any scale applied to it; this is always the same as local because Flutter doesn't apply any scaling.
/// [game]: the scalled value applied all the game transformations.
```

examples/lib/stories/camera_and_viewport/zoom.dart
```diff
@@ -0,0 +1,50 @@
...
+  @override
+  void onScroll(PointerScrollInfo event) {
+    camera.zoom += event.scrollDelta.game.y * zoomPerScrollUnit;
+  }
+
+  @override
+  void onScaleUpdate(ScaleUpdateInfo info) {
+    final scale = lastDelta;
+    if (scale != null) {
+      final delta = info.scale.game - scale;
```
luanpotter, line 44: why are you subtracting the lastDelta here? can't you just inc/dec the zoom by the amount scaled?

examples/lib/stories/camera_and_viewport/zoom.dart
```diff
@@ -0,0 +1,50 @@
...
+      camera.zoom += delta.y;
```
luanpotter, line 45: you should multiply/divide the current zoom by the delta (abs), not add, right?

examples/lib/stories/camera_and_viewport/zoom.dart
```diff
@@ -0,0 +1,50 @@
...
+    camera.followComponent(
+      flame,
+      relativeOffset: Anchor.center.toVector2(),
+    );
+  }
+
+  static const zoomPerScrollUnit = 0.001;
+  @override
+  void onScroll(PointerScrollInfo event) {
+    camera.zoom += event.scrollDelta.game.y * zoomPerScrollUnit;
```
luanpotter, line 37: you should multiple/divide the zoom by the delta abs right?

luanpotter (approved):
LGTM with some comments. I am not sure what is the issue from your video, is it the thing turning upside down? I think if you set the zoom to negative that would happen, and since you are adding the delta to the zoom it would happen. I don't know what is the rationale for doing that? Doesn't make sense to me to treat zoom additively, but if you allow it to get negative, I think that is expected.

examples/lib/stories/camera_and_viewport/zoom.dart
erickzanardo: Not sure I understood that, could you give me a small snippet example?

examples/lib/stories/camera_and_viewport/zoom.dart
erickzanardo: I have actually choose a very bad name for this variable.

This isn't the lastDelta but rather the lastScale.

This event don't have a delta, the scale is just the current average distance from each focal touch point.

So to have the delta we need to keep track of the last scale.

It took me a while to understand that and I forgot to rename the variable, I will rename it so the code makes more sense.

examples/lib/stories/camera_and_viewport/zoom.dart
erickzanardo: I don't think so, this particularly operation is working as expected.

This code is responsible for the effect you see on the mobile example on the video on the PR description.

luanpotter: I find it weird that the change in scale doesn't consider the current scale, hence my proposal. Also your implementation doesn't guard against negative values, which is inherent with an additive solution vs a multiplicative one (but can be solved in other ways)

examples/lib/stories/camera_and_viewport/zoom.dart
luanpotter: I am updating the code of isomash to use this and I can share that (working through another camera issue I found there rn, maybe we have a bug somewhere)

luanpotter: but what I meant was to multiply instead of add

zoom = event.scrollDelta.game.y * zoomPerScrollUnit
camera.zoom = if (zoom.sign < 1) camera.zoom / zoom.abs else camera.zoom * zoom.abs

so the effect scales with the current scales (accelerates in the global reference frame)

examples/lib/stories/camera_and_viewport/zoom.dart
erickzanardo: I may be missing something, but I am considering the current scale to calculate the new one.

I don't really think that this can get negative, because the nature of the gesture, and in my tests, this worked fine. I could try changing to your idea, but I am having a hard time picturing it without seeing some code.

examples/lib/stories/camera_and_viewport/zoom.dart
erickzanardo: Hmm, not really sure I understood the rational, but I will test and I may understand better when coding.

erickzanardo: I tried that idea but the scroll zoom got even weirder, on the first zoom, the sprite simply disappear :(

spydon (approved):
Lgtm

PR #772, opened 2021-04-22
Abstracting text API to enable custom renderers

erickzanardo:
# Description

This PR does two things:
 - Create an abstraction for text rendering on the framework, with this abstraction we can now easily introduce more ways of rendering text on Flame and we can still use the same components with different methods of text rendering.
 - Rename the `TextConfig` class to `TextPaint`: I always found it a little confusing that our class for rendering text was named `TextConfig` since it is not only a text configuration, but also a text renderer. I choose `TextPaint` for the new name because that implementation is quite tied to the Flutter's `TextPainter` class, that name was also inspired by the Flutter's duo classes `CustomPainter` and `CustomPaint`.
 
The most importante part of this PR is the abstraction, the renaming is a proposal, let me know what you think of the idea.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/doc/text.md
+++ b/doc/text.md
@@ -5,19 +5,25 @@ Flame has some dedicated classes to help you render text.
 ## TextRenderer
 
 `TextRenderer` is the abstract class used by Flame to render text. Flame provides one
-implementation for this called `TextPaint` but anyone can implement this and create
-a custom way to render text.
+implementation for this called `TextPaint` but anyone can implement this abstraction
+and create a custom way to render text.
 
 ## TextPaint
 
 A Text Paint is the built in implementation of text rendering on Flame, it is based on top of
-Flutter's `TextPainter` class (hence the name) and contains all typographical information
-required to render text; i.e., font size and color, family, etc.
+Flutter's `TextPainter` class (hence the name), it can be configured by its config class
+`TextPaingConfig` which contains all typographical information required to render text; i.e., font
+size and color, family, etc.
 
 Example usage:
 
 ```dart
-const TextPaint textPaint = TextPaint(fontSize: 48.0, fontFamily: 'Awesome Font');
+const TextPaint textPaint = TextPaint(
+  config: TextPaintConfig(
+    fontSize: 48.0,
+    fontFamily: 'Awesome Font',
+  ),
+);
 ```
 
  - `fontFamily`: a commonly available font, like Arial (default), or a custom font added in your
@@ -32,7 +38,7 @@ For more information regarding colors and how to create then, see the
 After the creation of the text paint you can use its `render` method to draw some string on a canvas:
 
 ```dart
-textPaint.render(canvas, "Flame is awesome", Position(10, 10));
+textPaint.render(canvas, "Flame is awesome", Vector2(10, 10));
 ```
 
 If you want to set the anchor of the text you can also do that in the render call, with the optional
--- a/doc/text.md
+++ b/doc/text.md
@@ -2,15 +2,22 @@
 
 Flame has some dedicated classes to help you render text.
 
-## TextConfig
+## TextRenderer
 
-A Text Config contains all typographical information required to render text; i.e., font size and
-color, family, etc.
+`TextRenderer` is the abstract class used by Flame to render text. Flame provides one
+implementation for this called `TextPaint` but anyone can implement this and create
+a custom way to render text.
+
+## TextPaint
+
+A Text Paint is the built in implementation of text rendering on Flame, it is based on top of
+Flutter's `TextPainter` class (hence the name) and contains all typographical information
+required to render text; i.e., font size and color, family, etc.
 
 Example usage:
 
 ```dart
-const TextConfig config = TextConfig(fontSize: 48.0, fontFamily: 'Awesome Font');
+const TextPaint textPaint = TextPaint(fontSize: 48.0, fontFamily: 'Awesome Font');
 ```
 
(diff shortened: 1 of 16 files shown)


spydon (approved):
Nice, lgtm!

luanpotter (commented):
>  I always found it a little confusing that our class for rendering text was named TextConfig since it is not only a text configuration, but also a text renderer

The main goal of the class was to be a text config. I just slapped the render method there because there was only one type of render and I found it to be a good place. I think there is still value in keeping the TextConfig class, because it would allow you to easily switch or use several renderers with the same config. The way you changed it, if I want to use a different Renderer I have to duplicate all the configuration for color, size, etc.

What if we keep TextConfig as a plain data class, and then add the TextRenderer interface that can be configured via TextConfig and create a TextPaintRenderer or something?

Or do you think other renderers would be so different compared to our current renderer that they would not share common configs like size, color, etc (whatever TextConfig has)?

erickzanardo:
@luanpotter I don't think all renderers will use text config, a bitmap font renderer for example I don't think would use because since the font is an image, many attributes that exists there would not be relevant.
 
I am ok with keeping the TextConfig class, but I don't think it should be an attribute of the abstract TextRenderer, but instead could be used by the renderers (TextPaint only for now) that makes sense. What do you think?

luanpotter:
> because since the font is an image, many attributes that exists there would not be relevant

By looking at TextConfig, it seems to me most if not all attributes should be used by anything rendering text. Maybe color is something you can't control on bitmap (not sure), but size, alignment, line height, font name, etc all should be configurations available right?

erickzanardo:
> By looking at TextConfig, it seems to me most if not all attributes should be used by anything rendering text. Maybe color is something you can't control on bitmap (not sure), but size, alignment, line height, font name, etc all should be configurations available right?

Color, and fontFamily don't makes sense for bitmap fonts indeed. But there is the problem, if one, or this case two attributes don't make sense for a renderer, should that renderer use the text config class for defining its style? I think it can lead to confusion having unused attributes. That is why I think I prefer that each renderer has its own configuration class.

Not sure if we could end up with an abstraction that would fit fine for text paint and for bitmap fonts, what do you think?

luanpotter:
`fontFamily` on the sprite would be the sprite name right? it's a hard problem if only one thing doesn't apply, like Color, but my idea is: TextConfig has only what is generic to all renderers, and each renderer has it's own extra attributes. now the issue is, we only have two renderers (normal and bitmap), so it's extremely hard to predict what is "common" or not moving forward. I leave it up to you to determine how "common" these things are across other types of potential renderers in the future, and if there are enough common things I think it makes sense to have a class; otherwise, each renderer can have it's own config

erickzanardo:
I wouldn't consider `fontFamily` to be the name of the sprite, because this seems to be a whole other concept.

But I think we can create a base text config, and then the text paint can extend that, that way we have a good abstraction and re use of code where we can, sounds good?

packages/flame/lib/src/text.dart
```diff
@@ -44,154 +94,149 @@ class TextConfig {
...
-  /// Creates a constant [TextConfig] with sensible defaults.
+  /// Creates a constant [TextPaint] with sensible defaults.
   ///
   /// Every parameter can be specified.
-  TextConfig({
-    this.fontSize = 24.0,
+  const TextPaintConfig({
     this.color = const Color(0xFF000000),
     this.fontFamily = 'Arial',
-    this.textAlign = TextAlign.left,
```
erickzanardo, line 72: Note to reviewers: You may notice that I removed the `textAlign` property from the configs.

I did this because I realized this property don't make much sense on a text renderer as it only has a position when rendering, there isn't any dimension, so there isn't anything to align really.

Also we already recommended to use the `Anchor` instead of the property, so I took a leap here and removed it entirely to avoid confusion.

doc/text.md
```diff
@@ -2,15 +2,28 @@
...
-color, family, etc.
+`TextRenderer` is the abstract class used by Flame to render text. Flame provides one
+implementation for this called `TextPaint` but anyone can implement this abstraction
+and create a custom way to render text.
+
+## TextPaint
+
+A Text Paint is the built in implementation of text rendering on Flame, it is based on top of
+Flutter's `TextPainter` class (hence the name), it can be configured by its config class
+`TextPaingConfig` which contains all typographical information required to render text; i.e., font
```
spydon, line 15:
```suggestion
`TextPaintConfig` which contains all typographical information required to render text; i.e., font
```

packages/flame/lib/src/text.dart
```diff
@@ -8,22 +8,72 @@ import 'extensions/size.dart';
...
 import 'memory_cache.dart';
 
-/// A Text Config contains all typographical information required to render texts; i.e., font size and color, family, etc.
-///
-/// It does not hold information regarding the position of the text to be render neither the text itself (the string).
-/// To hold all those information, use the Text component.
-///
-/// It is used by [TextComponent].
-class TextConfig {
+abstract class TextRenderer<T extends BaseTextConfig> {
```
spydon, line 11: Maybe add a short explanation regarding what the class is for?

spydon (approved):

packages/flame/lib/src/text.dart
erickzanardo: @spydon I have added a dartdoc here, could you take a look if it is ok?

spydon: Looks good!

spydon (approved):

PR #778, opened 2021-04-30
Same debugMode for children as for parent

spydon:
# Description

The children should have the same debugMode as their parent, at least initially.
This also fixes the composability example.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/test/components/composed_component_test.dart
+++ b/packages/flame/test/components/composed_component_test.dart
@@ -138,5 +138,21 @@ void main() {
       expect(child.rendered, true);
       expect(child.updated, true);
     });
+
+    test('initially same debugMode as parent', () {
+      final game = MyGame();
+      game.onResize(Vector2.all(100));
+      final child = MyTap();
+      final wrapper = MyComposed();
+      wrapper.debugMode = true;
+
+      wrapper.addChild(child);
+      game.add(wrapper);
+      game.update(0.0);
+
+      expect(child.debugMode, true);
+      wrapper.debugMode = false;
+      expect(child.debugMode, true);
+    });
   });
 }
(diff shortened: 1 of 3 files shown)


erickzanardo (approved):
LGTM

Maybe add a small test for this?

spydon:
> LGTM
> 
> Maybe add a small test for this?

Good idea, added one now

packages/flame/test/components/composed_component_test.dart
```diff
@@ -138,5 +138,21 @@ void main() {
       expect(child.rendered, true);
       expect(child.updated, true);
     });
+
+    test('initially same debugMode as parent', () {
+      final game = MyGame();
+      game.onResize(Vector2.all(100));
```
luanpotter, line 144: just out of curiosity, do you need to call onResize here since your test doesn't care about it?

luanpotter (approved):
LGTM.

one thing that I was thinking though, should we get rid of a per-component copy of debugMode and instead whoever wants to use it adds a HasGameRef?

packages/flame/test/components/composed_component_test.dart
spydon: It was just a copy of the previous test, this can be removed

spydon: Ah, it can't be removed, because the game needs to have `hasLayout = true` before it can do `add`

spydon:
> LGTM.
> 
> one thing that I was thinking though, should we get rid of a per-component copy of debugMode and instead whoever wants to use it adds a HasGameRef?

I think it is pretty handy to have per-component `debugMode` so that you can debug only the parts that you are interested in, but anyhow that should be different from what it looks today anyways. But we can have that as a follow-up I think, to decide how we want to have it.

PR #793, opened 2021-05-16
Possibility to change priority and reorder the component list

spydon:
# Description

This adds the possibility to change priority of components through `BaseGame`.
It is a breaking change since you are not supposed to override the priority field anymore.

## Checklist

## Breaking Change

## Related Issues


doc/game.md
```diff
@@ -62,6 +62,18 @@ want to remove a list of components.
 Any component on which the `remove()` method has been called will also be removed. You can do this
 simply by doing `yourComponent.remove();`.
 
+## Changing component priorities (render/update order)
+To update a component with a new priority you have to call either `BaseGame.changePriority`, or
+`BaseGame.changePriorities` if you want to change the priorities of many components at once.
+This design is due to that the components doesn't always have access to the component list and
```
erickzanardo, line 68:
```suggestion
This design is due to the fact that the components doesn't always have access to the component list and
```

examples/lib/stories/components/components.dart
```diff
@@ -4,6 +4,12 @@ import 'package:flame/game.dart';
 import '../../commons/commons.dart';
 import 'composability.dart';
 import 'debug.dart';
+import 'priority.dart';
+
+const priorityInfo = '''
+On this example, click on the squared to bring them to the front by changing the
```
erickzanardo, line 10:
```suggestion
On this example, click on the square to bring them to the front by changing the
```

examples/lib/stories/components/priority.dart
```diff
@@ -0,0 +1,57 @@
...
+    canvas.drawRect(size.toRect(), paint);
+  }
+
+  static Paint _randomPaint() {
+    final rng = Random();
+    final color = Color.fromRGBO(
+      rng.nextInt(256),
+      rng.nextInt(256),
+      rng.nextInt(256),
+      0.9,
```
erickzanardo, line 40: This seems like a great addition for a Color extension

packages/flame/lib/src/components/base_component.dart
```diff
@@ -171,6 +173,13 @@ abstract class BaseComponent extends Component {
     }
   }
 
+  Future<void> addChildren(
```
erickzanardo, line 176: This method needs to be async? You don't use any async operations on the method body

packages/flame/lib/src/components/component.dart
```diff
@@ -27,14 +27,15 @@ abstract class Component {
...
+  int get priority => _priority;
+  int _priority;
 
   /// Whether this component should be removed or not.
   ///
   /// It will be checked once per component per tick, and if it is true, BaseGame will remove it.
   bool shouldRemove = false;
 
-  Component({this.priority = 0});
+  Component({int? priority = 0}) : _priority = priority ?? 0;
```
erickzanardo, line 38: I am almost sure that this will have the same effect and will prevent the user to pass null to the parameter

```suggestion
  Component({int priority = 0}) : _priority = priority;
```

packages/flame/lib/src/components/base_component.dart
spydon: Ah, forgot that the user should be able to wait for the children to be loaded. Does it look correct now?

erickzanardo: Did you changed anything? Can't spot any difference, there is still nothing being awaited on the  function body

spydon: Maybe I commented before I had pushed, it is there now at least.
```dart
    await Future.wait(
      children.map(
        (child) => addChild(child, gameRef: gameRef),
      ),
    );
```

erickzanardo: 👍

erickzanardo (approved):

packages/flame/lib/src/game/base_game.dart
```diff
@@ -238,6 +238,69 @@ class BaseGame extends Game with FPSCounter {
...
+          if (child == component.parent) {
+            foundParent = child;
+            return false;
+          }
+          return true;
+        });
+        if (foundParent != null) {
+          break;
+        }
+      }
```
luanpotter, line 280: I think this could be extracted to a useful method on BaseComponent, what do you think? (findParent)

packages/flame/lib/src/game/base_game.dart
```diff
@@ -238,6 +238,69 @@ class BaseGame extends Game with FPSCounter {
...
+      foundParent?.reorderChildren();
```
luanpotter, line 281: technically this will rebalance the same component multiple times when using `changePriorities` with siblings, not sure if that is a huge issue.

packages/flame/test/components/priority_test.dart
```diff
@@ -0,0 +1,119 @@
+import 'package:flame/components.dart';
+import 'package:flame/game.dart';
+import 'package:test/test.dart';
+
+class PriorityComponent extends BaseComponent {
+  PriorityComponent(int priority) : super(priority: priority);
+}
+
+bool isSorted<T>(Iterable<T> list, [int Function(T, T)? compare]) {
```
luanpotter, line 9: do we need a generic isSorted? we could just do
```
p = components.map(e => e.priority)
expect(p, orderedEquals(p.sorted()))
```

luanpotter: (assuming we only ever care about List<Component>, otherwise the func is much more generic

luanpotter (approved):
LGTM with comments

packages/flame/lib/src/game/base_game.dart
spydon: Wait... Why am I doing this, `BaseComponent` already knows its parent, so for all of those we don't need to check and I'm missing to check for normal `Component` here. Maybe we should move the parent field to `Component` and this could be done a lot more efficiently. I'll re-do this part later today or tomorrow.

packages/flame/test/components/priority_test.dart
spydon: Much better idea, fixed that now

packages/flame/lib/src/game/base_game.dart
spydon: Fixed this issue by putting all the parent in a set first and then reordering their children

packages/flame/lib/src/game/base_game.dart
spydon: Should be fixed now, please re-review :)

examples/lib/stories/components/priority.dart
spydon: I'll do this in a follow-up

luanpotter (approved):
LGTM.

I wonder why component.parent is not already a BaseComp?? could it ever not be? only Base can have children right?

PR #794, opened 2021-05-16
Adding coverage checking

erickzanardo:
# Description

Adds a coverage checking on our CI to improve our tests

The idea is that as we add more coverage, we increase the value on the min_coverage file and PRs will only fail when a PR has a coverage less than the min.

## Checklist

## Breaking Change

## Related Issues


scripts/check_coverage.dart
```diff
@@ -0,0 +1,12 @@
+import 'dart:io';
+
+void main(args) {
```
erickzanardo, line 3: Note to reviewer: I was planning on make this all in bash, but it seems that bash can't work with floating point numbers, so I chose dart because dart is already available

scripts/test.sh
```diff
@@ -2,16 +2,29 @@
...
   cd $dir
   echo "Testing $dir"
-  flutter test
+  flutter test --coverage
   test_result=$?
   if [ $test_result -ne 0 ]; then
     exit $test_result
   fi
+
+  if [ -f ".min_coverage" ]; then
```
erickzanardo, line 16: Note to reviewers: Run coverage check only if the project has a minimum value set already, decided to do this way because we may have projects on the mono repo which don't make sense to check coverage

scripts/test.sh
```diff
@@ -2,16 +2,33 @@
 
 echo "Starting Flame Tester"
 echo "---------------------"
+root_dir=$(pwd)
```
spydon, line 5:
```suggestion
root_dir="$(dirname "$0")/.."
```

scripts/test.sh
```diff
@@ -2,16 +2,33 @@
...
+
+  if [ -f ".min_coverage" ]; then
+    min_coverage=$(cat .min_coverage)
+
+    coverage_summary=$(lcov --summary coverage/lcov.info)
+    echo "$coverage_summary"
+
+    current_cov=$(echo "$coverage_summary" | grep lines | grep -Eow '\d+\.\d+')
+
+    dart "$root_dir/scripts/check_coverage.dart" "$current_cov" "$min_coverage"
```
spydon, line 24:
```suggestion
    dart "$(dirname $0)/check_coverage.dart" "$current_cov" "$min_coverage"
```

scripts/test.sh
spydon: I think `root_dir` is only used in one place, so you can just remove this line and do the other suggestion if you want.

spydon (approved):
Lgtm, just one suggestion

scripts/test.sh
```diff
@@ -2,16 +2,33 @@
...
     exit $test_result
   fi
+
+  if [ -f ".min_coverage" ]; then
+    min_coverage=$(cat .min_coverage)
+
+    coverage_summary=$(lcov --summary coverage/lcov.info)
+    echo "$coverage_summary"
+
+    current_cov=$(echo "$coverage_summary" | grep lines | grep -Eow '\d+\.\d+')
```
spydon, line 22:
```suggestion
    current_cov=$(echo "$coverage_summary" | grep lines | grep -Pow '\d+\.\d+')
```

scripts/test.sh
erickzanardo: What is the difference?

About the place, I need it before because the current dir is changed by `cd` commands, so if I use directly on the place I use root_dir, it will not be the value I expect.

scripts/test.sh
erickzanardo: damn, with this change, the script works on linux, but not on Mac :(, I think I will not use grep and come with more cross platform solution.

scripts/test.sh
spydon: The difference is that with just `pwd` you have to be in the correct directory for it to work, with `dirname` you can be in any directory. It doesn't matter that you have changed directory since it takes the path of the script.

renancaraujo (approved):

luanpotter (commented):
The script LGTM, I like the idea of having code coverage but not sure I agree it should block the PR, as there are legitimate cases where a change would reduce coverage as percentage (removing dead code for example). I feel coverage is much more of an indicator than a hard rule.

erickzanardo:
> The script LGTM, I like the idea of having code coverage but not sure I agree it should block the PR, as there are legitimate cases where a change would reduce coverage as percentage (removing dead code for example). I feel coverage is much more of an indicator than a hard rule.

Well, if it makes sense to lower the percentage limit for some reason, then the PR can just justify why and change limit file and life continues

erickzanardo:
@luanpotter we can add a new entry to the pull request template, something like this:

 - [ ] My PR don't decrease the minimum test coverage, or I have a special case and explained why on the PR description

scripts/test.sh
erickzanardo: @spydon I couldn't make the `dirname $0` work. The command behaves like you explained to me but when I use it to create the path to the dart file I want to execute, dart can't find the file because the current session is on a different folder. With pwd I have an absolute path and it seems to be more secure.

But I am open to suggestion here.

.github/pull_request_template.md
```diff
@@ -8,6 +8,7 @@ Before you create this PR confirm that it meets all requirements listed below by
 
 - [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
 - [ ] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors and are passing (See [Contributor Guide]).
+- [ ] My PR don't decreases the code coverage, or I have __a very special case__ and explained on the PR description why this PR decreases the coverage.
```
wolfenrain, line 11:
```suggestion
- [ ] My PR does not decrease the code coverage, or I have __a very special case__ and explained on the PR description why this PR decreases the coverage.
```

wolfenrain (approved):

scripts/test.sh
```diff
@@ -2,16 +2,32 @@
 
 echo "Starting Flame Tester"
 echo "---------------------"
+root_dir=$(pwd)
```
spydon, line 5:
```suggestion
root_dir="$(realpath $(dirname "$0")/..)"
```

PR #797, opened 2021-05-17
Add hoverables

luanpotter:
# Description

Analogous to Tapables and Draggables.

## Checklist

## Breaking Change

### DIFF
--- a/doc/input.md
+++ b/doc/input.md
@@ -138,10 +138,10 @@ class MyGame extends Game with TapDetector {
 You can also check more complete examples
 [here](https://github.com/flame-engine/flame/tree/main/examples/lib/stories/controls/).
 
-## Tapable and Draggable components
+## Tapable, Draggable and Hoverable components
 
-Any component derived from `BaseComponent` (most components) can add the `Tapable` and/or the
-`Draggable` mixins to handle taps and drags on the component.
+Any component derived from `BaseComponent` (most components) can add the `Tapable`, the
+`Draggable`, and/or the `Hoverable` mixins to handle taps, drags and hovers on the component.
 
 All overridden methods return a boolean to control if the event should be passed down further along
 to components underneath it. So say that you only want your top visible component to receive a tap
@@ -229,8 +229,7 @@ Minimal component example (this example ignores pointerId so it wont work well i
 multi-drag):
 
 ```dart
-import 'package:flame/components/component.dart';
-import 'package:flame/components/mixins/draggable.dart';
+import 'package:flame/components.dart';
 
 class DraggableComponent extends PositionComponent with Draggable {
 
@@ -274,6 +273,25 @@ class MyGame extends BaseGame with HasDraggableComponents {
 
 **Note**: `HasDraggableComponents` uses an advanced gesture detector under the hood and as explained
 further up on this page, shouldn't be used alongside basic detectors.
+
+### Hoverable components
+
+Just like the others, this mixin allows for easy wiring of your component to listen to hover states
+and events.
+
+By adding the `HasHoverableComponents` mixin to your base game, and by using the mixin `Hoverable` on
+your components, they get a `isHovered` field and a couple methods (`onHoverStart`, `onHoverEnd`) that
+you can override if you want to listen to the events.
+
+```dart
+  bool isHovered = false; // note: you should not modify this value yourself
+  void onHoverEnter(PointerHoverInfo event) {}
+  void onHoverLeave(PointerHoverInfo event) {}
+```
+
+The provided event info is from the mouse move that triggered the action (entering or leaving).
+While the mouse move keeping inside or outside, no events are fired and those mouse move events are
+not propagated. Only when the state is changed that the handlers are triggered.
 
 ## Hitbox
 The `Hitbox` mixin is used to make detection of gestures on top of your `PositionComponent`s more
--- a/doc/input.md
+++ b/doc/input.md
@@ -283,7 +283,7 @@ your components, they get an `isHovered` field and a couple of methods (`onHover
 you can override if you want to listen to the events.
 
 ```dart
-  bool isHovered = false; // note: you should not modify this value yourself
+  bool isHovered = false;
   void onHoverEnter(PointerHoverInfo event) {}
   void onHoverLeave(PointerHoverInfo event) {}
 ```
--- a/doc/input.md
+++ b/doc/input.md
@@ -166,8 +166,7 @@ void onTapUp(TapUpInfo event) {}
 Minimal component example:
 
 ```dart
-import 'package:flame/components/component.dart';
-import 'package:flame/components/mixins/tapable.dart';
+import 'package:flame/components.dart';
 
 class TapableComponent extends PositionComponent with Tapable {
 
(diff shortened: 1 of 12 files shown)


doc/input.md
```diff
@@ -275,6 +273,25 @@ class MyGame extends BaseGame with HasDraggableComponents {
 **Note**: `HasDraggableComponents` uses an advanced gesture detector under the hood and as explained
 further up on this page, shouldn't be used alongside basic detectors.
 
+### Hoverable components
+
+Just like the others, this mixin allows for easy wiring of your component to listen to hover states
+and events.
+
+By adding the `HasHoverableComponents` mixin to your base game, and by using the mixin `Hoverable` on
+your components, they get a `isHovered` field and a couple methods (`onHoverStart`, `onHoverEnd`) that
```
spydon, line 282:
```suggestion
your components, they get an `isHovered` field and a couple of methods (`onHoverStart`, `onHoverEnd`) that
```

doc/input.md
```diff
@@ -275,6 +273,25 @@ class MyGame extends BaseGame with HasDraggableComponents {
...
+
+```dart
+  bool isHovered = false; // note: you should not modify this value yourself
+  void onHoverEnter(PointerHoverInfo event) {}
+  void onHoverLeave(PointerHoverInfo event) {}
+```
+
+The provided event info is from the mouse move that triggered the action (entering or leaving).
+While the mouse move keeping inside or outside, no events are fired and those mouse move events are
+not propagated. Only when the state is changed that the handlers are triggered.
```
spydon, line 293:
```suggestion
not propagated. Only when the state is changed the handlers are triggered.
```

packages/flame/lib/src/components/mixins/hoverable.dart
```diff
@@ -0,0 +1,42 @@
+import 'package:meta/meta.dart';
+
+import '../../../game.dart';
+import '../../game/base_game.dart';
+import '../../gestures/events.dart';
+import '../base_component.dart';
+
+mixin Hoverable on BaseComponent {
+  bool isHovered = false;
```
spydon, line 9: Nice! We don't have anything like this on the other mixins, I think this would be pretty good to have on `Draggable` too.

spydon (approved):
LGTM! Maybe we can do a follow-up to add `isDragged` to `Draggable` so that we are consistent with the structure of these mixins.

doc/input.md
```diff
@@ -275,6 +273,25 @@ class MyGame extends BaseGame with HasDraggableComponents {
...
+
+Just like the others, this mixin allows for easy wiring of your component to listen to hover states
+and events.
+
+By adding the `HasHoverableComponents` mixin to your base game, and by using the mixin `Hoverable` on
+your components, they get an `isHovered` field and a couple of methods (`onHoverStart`, `onHoverEnd`) that
+you can override if you want to listen to the events.
+
+```dart
+  bool isHovered = false; // note: you should not modify this value yourself
```
erickzanardo, line 286: Can't we transform this into a getter, so the user can't change it?

doc/input.md
```diff
@@ -275,6 +273,25 @@ class MyGame extends BaseGame with HasDraggableComponents {
...
+  void onHoverEnter(PointerHoverInfo event) {}
+  void onHoverLeave(PointerHoverInfo event) {}
+```
+
+The provided event info is from the mouse move that triggered the action (entering or leaving).
+While the mouse move keeping inside or outside, no events are fired and those mouse move events are
```
erickzanardo, line 292: Found this sentence a little confusing, what do you think of this suggestion?

```suggestion
While the mouse movement is kept inside or outside, no events are fired and those mouse move events are
```

examples/lib/stories/controls/controls.dart
```diff
@@ -18,61 +19,62 @@ void addControlsStories(Dashbook dashbook) {
...
           ),
         );
       },
-      codeLink: baseLink('gestures/draggables.dart'),
+      codeLink: baseLink('controls/draggables.dart'),
+    )
+    ..add(
+      'Hoverables',
+      (_) => GameWidget(game: HoverablesGame()),
+      codeLink: baseLink('controls/hoverables.dart'),
```
erickzanardo, line 68: Add some info about this example, could be just some instructions on how the example works.

erickzanardo (commented):
A couple of comments, but overall looks pretty good addition!

doc/input.md
luanpotter: it is a getter, they can always override it right?

luanpotter: on the mixin level I can't control what the impl will do with it

erickzanardo: I thought it was a field, so da good then

erickzanardo: Actually, looking on the component, it is a field, I am confused

luanpotter: not sure I understand your suggestions, we need a way to set it and a way to get it, it's a field on the mixin but can be overwritten by the user

erickzanardo: My suggestion is that this is a getter to the user of the component and internally, we set this using a setter that isn't exposed to the user. That way the user can't change the field at all.

luanpotter: I thought more about it and there might be a way to hide it from the user! (slightly hide it)

packages/flame/lib/src/components/mixins/hoverable.dart
```diff
@@ -0,0 +1,48 @@
+import 'package:meta/meta.dart';
+
+import '../../../game.dart';
+import '../../game/base_game.dart';
+import '../../gestures/events.dart';
+import '../base_component.dart';
+
+mixin Hoverable on BaseComponent {
+  bool _isHovered = false;
+  bool get isHovered => _isHovered;
```
erickzanardo, line 10: Excellent! That was pretty much what I had in mind.

packages/flame/test/components/draggable_test.dart
```diff
@@ -28,15 +28,11 @@ void main() {
...
-      try {
-        await game2.add(DraggableComponent());
-      } catch (e) {
-        hasError = true;
-      }
-
-      expect(hasError, true);
+      expect(
+        () => game2.add(DraggableComponent()),
+        throwsA(const TypeMatcher<AssertionError>()),
```
erickzanardo, line 34: I believe you can use isA here instead, give it a test, I don't fully remember is this is the correct syntax.
```suggestion
        throwsA(isA<AssertionError>()),
```

packages/flame/test/components/hoverable_test.dart
```diff
@@ -0,0 +1,162 @@
...
+
+      _triggerMouseMove(game, 11, 6);
+      expect(a.isHovered, true);
+      expect(b.isHovered, false);
+      expect(c.isHovered, false);
+    });
+  });
+}
+
+// TODO(luan) we can probably provide some helpers to facilitate testing events
```
erickzanardo, line 152: `flame_test` 😎

erickzanardo (approved):
One nit but LGTM already

packages/flame/lib/src/components/mixins/hoverable.dart
luanpotter: yeah what I didn't know is that we had "final" method in dart for the mutator. it's not compiler safe but at least is annotated so good enough for me

PR #814, opened 2021-05-22
Parse color from hex

luanpotter:
I was looking at the src for Tales to find useful bits of code to port back into Flame. I have some good ideas to tackle in the future. This one was a simple one that I was not sure about. Not sure how overall useful it is, since it assumes the `#RRGGBB` format, which is common but not unique. What are your thoughts? Worth it or not?


spydon:
How did you use it in tales? Since color accepts an ARGB directly in the constructor it would be pretty much as easy to write those colors in that format or did you use it when parsing files or something?

luanpotter:
Yes it is only used for parsing external files. Of course for hardcoded colours you should use Dart's builtin stuff. That is why I am unsure if this is worth adding or not.

spydon:
I guess it depends on what philosophy we have around the extensions; whether they should contain everything or more clean with a bar just above the bare necessities. I don't have any preference on what route we take, as long as we are consistent.

erickzanardo:
I think this is worth, it is always a hassle to parse colors from a string in files.

I would just mention on the docs that this is focused for that use and for colors declared directly in the code, it is not recommended to use it.

I would even suggest to create more methods to parse other formats like the rgb, rgba, etc

spydon (approved):

luanpotter:
> parse other formats like the rgb, rgba, etc

good idea, I will rename this one to make it clear it's RGB and add one for RGBA

I can make it support other things too like `#ccc`

doc/util.md
```diff
@@ -170,6 +170,10 @@ Methods:
  - `darken`: Darken the shade of the color by an amount between 0 to 1.
  - `brighten`: Brighten the shade of the color by an amount between 0 to 1.
 
+Factories:
+- `fromRGBHexString`: Parses an RGB color from a valid hex string (e.g. #1C1C1C).
```
erickzanardo, line 174:
```suggestion
- `fromRGBHexString`: Parses a RGB color from a valid hex string (e.g. #1C1C1C).
```

packages/flame/lib/src/extensions/color.dart
```diff
@@ -35,4 +35,70 @@ extension ColorExtension on Color {
       blue + ((255 - blue) * amount).round(),
     );
   }
+
+  /// Parses an RGB color from a valid hex string (e.g. #1C1C1C).
```
erickzanardo, line 39:
```suggestion
  /// Parses a RGB color from a valid hex string (e.g. #1C1C1C).
```

erickzanardo (approved):
Couple of possible nits (I am unsure if you should use a or an there, still comment to bring that to your attention)

But LGTM arleady

doc/util.md
luanpotter: Since RGB is an initialism, not an acronym, I believe that is correct, because R is pronounced with a vowel.

packages/flame/lib/src/extensions/color.dart
luanpotter: image

PR #825, opened 2021-05-27
Make it possible to work with the new Text inheritance chain

luanpotter:
# Description

Assuming you have an arbitrary TextComponent, that got it's painter (which stands in as a config) via a parameter, and you want to modify some property, you cannot do that in a generic way. This adds a couple functions to facilitate such operations; also, it makes text component type safe by defining what implementation of text renderer it uses as a generic parameter.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/lib/src/text.dart
+++ b/packages/flame/lib/src/text.dart
@@ -13,6 +13,11 @@ import 'memory_cache.dart';
 ///
 /// See [TextPaint] for the default implementation offered by Flame
 abstract class TextRenderer<T extends BaseTextConfig> {
+  /// A registry containing default providers for every [TextRenderer] subclass;
+  /// used by [createDefault] to create default parameter values.
+  ///
+  /// If you add a new [TextRenderer] child, you can register it by adding it,
+  /// alongisde a provider lambda, to this map.
   static Map<Type, TextRenderer Function()> defaultCreatorsRegistry = {
     TextRenderer: () => TextPaint(),
     TextPaint: () => TextPaint(),
--- a/packages/flame/lib/src/text.dart
+++ b/packages/flame/lib/src/text.dart
@@ -46,6 +46,10 @@ abstract class TextRenderer<T extends BaseTextConfig> {
       measureTextHeight(text),
     );
   }
+
+  /// Creates a new instance of this painter but transforming the [config]
+  /// object via the provided lambda.
+  TextRenderer<T> copyWith(T Function(T) transform);
 }
 
 /// A Text Config contains all typographical information required to render texts; i.e., font size, text direction, etc.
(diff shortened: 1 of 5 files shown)


packages/flame/lib/src/text.dart
```diff
@@ -46,6 +46,19 @@ abstract class TextRenderer<T extends BaseTextConfig> {
       measureTextHeight(text),
     );
   }
+
+  /// Creates a new instance of this painter but transforming the [config]
+  /// object via the provided lambda.
+  TextRenderer<T> copyWith(T Function(T) transform);
+
+  /// Given a generic type [T], creates a default renderer of that type.
+  static T createDefault<T extends TextRenderer>() {
```
erickzanardo, line 55: I am not sure about this, this method will make it impossible for people to create their own implementation right? All implementation will need to be provided by Flame

erickzanardo (commented):
One concern

packages/flame/lib/src/text.dart
luanpotter: yeah that makes sense

packages/flame/lib/src/text.dart
```diff
@@ -13,6 +13,11 @@ import 'memory_cache.dart';
 ///
 /// See [TextPaint] for the default implementation offered by Flame
 abstract class TextRenderer<T extends BaseTextConfig> {
+  static Map<Type, TextRenderer Function()> defaultCreatorsRegistry = {
+    TextRenderer: () => TextPaint(),
+    TextPaint: () => TextPaint(),
+  };
+
```
erickzanardo, line 20: I wonder if we should just expose the map, or make it private and have a `registerDefaultCreator` or something like that.

I think I prefer the method, but no strong opinions, I leave it up to you.

packages/flame/test/text_test.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:flame/components.dart';
+import 'package:test/test.dart';
+
```
erickzanardo, line 3: We should add a small text to check if the default registration is working properly.

packages/flame/lib/src/text.dart
erickzanardo: Also, whichever solution we follow, we need to quote this on the docs.

erickzanardo (commented):
A few comments, but that was pretty much the idea.

packages/flame/lib/src/text.dart
luanpotter: I didn't see why make it private, the user can have full control if they want to, for instance, change one of the existing defaults. Just give the power. Also, modifying this at all is a very, very niche thing, so I didn't want to include this since it's already documented on the code.

erickzanardo: I don't see how this documented on the code, there is not dartdoc on it, or am I missing something?

luanpotter: I meant this

image

But I can add a dartdoc as well

erickzanardo (approved):

PR #838, opened 2021-06-06
Added Matrix4 extension

wolfenrain:
# Description

*Replace this paragraph with a description of what this PR is doing. If you're modifying existing behavior, describe the existing behavior, how this PR is changing it, and what motivated the change. If this is a breaking change, specify explicitly which APIs have been changed*

## Checklist

## Breaking Change

## Related Issues

None


doc/util.md
```diff
@@ -241,3 +242,34 @@ Operators:
  - `&`: Combines two `Vector2`s to form a Rect, the origin should be on the left and the size on the
   right.
  - `%`: Modulo/Remainder of x and y separately of two `Vector2`s.
+
+### Matrix4
+
+This class comes from the `vector_math` pacakage. We have created a few extension methods on top
```
spydon, line 248:
```suggestion
This class comes from the `vector_math` package. We have created a few extension methods on top
```

packages/flame/lib/src/extensions/matrix4.dart
```diff
@@ -0,0 +1,82 @@
...
+    } else {
+      out.setFrom(position);
+    }
+    return transform2(out);
+  }
+
+  /// Create a scaled matrix.
+  ///
+  /// Scale by a [Vector3], [Vector4], or x,y,z
+  static Matrix4 scale(dynamic x, [double? y, double? z]) => Matrix4.zero()
```
spydon, line 79: `Matrix4.identity()` can be used here instead of `zero()`

packages/flame/lib/src/extensions/matrix4.dart
```diff
@@ -0,0 +1,82 @@
+import 'package:vector_math/vector_math_64.dart';
+
+export 'package:vector_math/vector_math_64.dart' hide Colors;
+
+extension Matrix4Extension on Matrix4 {
+  /// A first row and first column value.
+  double get m11 => storage[0];
```
spydon, line 7: Isn't it better to teach the users to use the `entry(int row, int column)` method?
I think that gives more clear code.

wolfenrain: This extension was based on a conversation me and @luanpotter had a few months ago about adding more unity/godot like getters to the Matrix4 class. Because whenever you google matrix4 based code you get examples that are based on `mYX` properties. So to make it easier for people to work with the Matrix4 I came with the idea to add them as getters.

spydon: Okay, might make sense... I didn't get up much of that when I googled, seems to be different ways of getting entries in different engines and languages. Can you send some examples? 
I stumbled upon one, and that one started on `m00` at least. I think if later there are examples and tutorials specifically for Flutter and Dart it will be confusing since there it would use `entry` and our users would have a higher chance of getting up such docs, on the other hand that would work too. Also imho it feels a bit old school/C++ to have entries like this.

wolfenrain: http://rbwhitaker.wikidot.com/monogame-basic-matrices This weirdly enough came up as first when I googled for unity.. But it does showcase it at the `Custom Matrices` part.

doc/util.md
```diff
@@ -241,3 +242,34 @@ Operators:
...
+  - `m23`: Returns the value at the second row and third column.
+  - `m24`: Returns the value at the second row and fourth column.
+  - `m31`: Returns the value at the third row and first column.
+  - `m32`: Returns the value at the third row and second column.
+  - `m33`: Returns the value at the third row and third column.
+  - `m34`: Returns the value at the third row and fourth column.
+  - `m41`: Returns the value at the fourth row and first column.
+  - `m42`: Returns the value at the fourth row and second column.
+  - `m43`: Returns the value at the fourth row and third column.
+  - `m44`: Returns the value at the fourth row and fourth column.
```
erickzanardo, line 271: Maybe we could omit that `Returns the value at the` as it repeats on all items ?

erickzanardo (approved):
one optional nit, LGTM already

doc/util.md
wolfenrain: Documentation wise it makes sense to put it on each? As it explains what the getter does. Just saying first row second column is imo not saying anything

erickzanardo: idk, imo just saying `The first row...` is enough for, especially since this is aligned under the getters section.

Either is up to you, totally optional

wolfenrain: Oh wait I thought this was in the dart code... I have updated this as I do agree on that part

packages/flame/lib/src/extensions/matrix4.dart
```diff
@@ -0,0 +1,81 @@
...
+      out.setFrom(position);
+    }
+    return transform2(out);
+  }
+
+  /// Create a scaled matrix.
+  ///
+  /// Scale by a [Vector3], [Vector4], or x,y,z
+  static Matrix4 scale(dynamic x, [double? y, double? z]) =>
+      Matrix4.identity()..scale(x, y, z);
```
spydon, line 80: Methods that can't be written on one line should be blocks according to the effective dart style guide, if I remember correctly.

spydon (approved):
Just one syntax nit, otherwise it looks good to me! I think I can get used to the `m11-m44` style.
Would be great with a `translateVector2` method in the future too.

packages/flame/lib/src/extensions/matrix4.dart
wolfenrain: Fixed

wolfenrain:
> Would be great with a `translateVector2` method in the future too.

@spydon what did u mean with this exactly?

spydon:
> > Would be great with a `translateVector2` method in the future too.
> 
> @spydon what did u mean with this exactly?

Right now we have to do `Matrix4.translate(x, y)`, would be nice with a `Matrix4.translateVector2(Vector2(x, y))` too.
The normal one only takes `Vector3` and `Vector4`.

packages/flame/lib/src/extensions/matrix4.dart
```diff
@@ -0,0 +1,87 @@
...
+    } else {
+      out.setFrom(position);
+    }
+    return transform2(out);
+  }
+
+  /// Create a scaled matrix.
+  ///
+  /// Scale by a [Vector3], [Vector4], or x,y,z
+  static Matrix4 scale(dynamic x, [double? y, double? z]) {
```
spydon, line 84: Hmm, hard with this name, it might be confusing since it is clashing with the already existing method.

doc/util.md
```diff
@@ -241,3 +242,34 @@ Operators:
...
  - `%`: Modulo/Remainder of x and y separately of two `Vector2`s.
+
+### Matrix4
+
+This class comes from the `vector_math` package. We have created a few extension methods on top
+of what is already offered by `vector_math`.
+
+Methods:
+  - `transform2`: Create a new `Vector2` by transforming the given `Vector2` using the `Matrix4`.
+  - `transformed2`: Transform the input `Vector2` into the output `Vector2`.
```
spydon, line 253: Missing `translateVector2`

doc/util.md
```diff
@@ -241,3 +242,34 @@ Operators:
...
+  - `m23`: The second row and third column.
+  - `m24`: The second row and fourth column.
+  - `m31`: The third row and first column.
+  - `m32`: The third row and second column.
+  - `m33`: The third row and third column.
+  - `m34`: The third row and fourth column.
+  - `m41`: The fourth row and first column.
+  - `m42`: The fourth row and second column.
+  - `m43`: The fourth row and third column.
+  - `m44`: The fourth row and fourth column.
```
luanpotter, line 271: can't we just say mXY returns row and col y

packages/flame/CHANGELOG.md
```diff
@@ -2,27 +2,28 @@
...
+ - The `size` property of `TextComponent` was not correctly set
+ - The `anchor` property was being incorrectly passed along to `TextRenderer`
  - All components take priority as an argument on their constructors
- - Fix renderRotated
- - Use QueryableOrderedSet for Collidables
- - Refactor TextBoxComponent
- - Fix bugs with TextBoxComponent
+ - Fix `renderRotated`
+ - Use `QueryableOrderedSet` for `Collidable`s
+ - Refactored `TextBoxComponent`
```
luanpotter, line 18: Refactor

packages/flame/CHANGELOG.md
```diff
@@ -2,27 +2,28 @@
...
+ - Fixed bugs with `TextBoxComponent`
```
luanpotter, line 19: Fix

packages/flame/CHANGELOG.md
```diff
@@ -2,27 +2,28 @@
...
  - Improve error message for composed components
  - Add `anchor` for `ShapeComponent` constructor
  - Fix rendering of polygons in `ShapeComponent`
  - Add `SpriteAnimation` support to parallax
  - Fix `Parallax` alignment for images with different width and height
  - Fix `ImageComposition` image bounds validation
+ - Added `Matrix4` extensions
```
luanpotter, line 26: Add

packages/flame/lib/src/extensions/matrix4.dart
luanpotter: if storage is a 1d array I think an `entry(i, j)` method alongside the m-getters is also useful

packages/flame/lib/src/extensions/matrix4.dart
```diff
@@ -0,0 +1,87 @@
...
+    return Vector2(
+      (position.x * m11) + (position.y * m21) + m41,
+      (position.x * m12) + (position.y * m22) + m42,
+    );
+  }
+
+  /// Transform a copy of [position] of type [Vector2] using the transformation
+  /// defined by this. If a [out] parameter is supplied, the copy is stored in
+  /// [out].
+  Vector2 transformed2(Vector2 position, [Vector2? out]) {
```
luanpotter, line 71: not a super fan of out parameters, is this coming from a specific need, or copying from other places?

packages/flame/lib/src/extensions/matrix4.dart
luanpotter: yeah I would rather have scale2, scale3 and not take dynamic

packages/flame/lib/src/extensions/matrix4.dart
```diff
@@ -0,0 +1,87 @@
...
+  double get m42 => storage[13];
+
+  /// A fourth row and third column value.
+  double get m43 => storage[14];
+
+  /// A fourth row and fourth column value.
+  double get m44 => storage[15];
+
+  /// Translate this matrix by a [Vector2].
+  void translateVector2(Vector2 vector) {
```
luanpotter, line 55: should this be translate2 to be similar to the others?

luanpotter (commented):
Some comments

packages/flame/lib/src/extensions/matrix4.dart
spydon: I really dislike this syntax too, but 8t was very useful in Forge2D for example, because then you could create one object less.

packages/flame/lib/src/extensions/matrix4.dart
wolfenrain: I decided to let it follow the already existing method (as the only way to use it is on the extension), and follow the same standards that the `vector_math` library already had. Because otherwise you have difference ways of doing the same kind of thing on one class.

packages/flame/CHANGELOG.md
wolfenrain: The `Added` was based on the other entries in our CHANGELOG when something got added.

packages/flame/lib/src/extensions/matrix4.dart
wolfenrain: This is following the standards of the `vector_math` class. The `transformed3` and `transformed4` methods have it as well.

packages/flame/CHANGELOG.md
wolfenrain: The `Refactored` was based on the other entries in our CHANGELOG where it started the sentence with it.

doc/util.md
wolfenrain: When writing these docs I was wondering... why are we adding all of this when we can just link to our generated API's? Especially for these extensions classes there is no extra value (like explaining usecases) to define each method and getter explicitly as the API docs will already do that for us.

packages/flame/lib/src/extensions/matrix4.dart
wolfenrain: That method already exists.

packages/flame/CHANGELOG.md
spydon: I can see two `Add` just from the preview here :smile:

packages/flame/CHANGELOG.md
spydon: We shouldn't change the lines in the changelog after they have been merged, then you lose the ability to directly see in which PR which entry is related to. I think that is more important than having the backticks there.

packages/flame/test/extensions/matrix4_test.dart
```diff
@@ -0,0 +1,63 @@
...
+    test('test m31', () => expect(matrix4.m31, matrix4.storage[8]));
+
+    test('test m32', () => expect(matrix4.m32, matrix4.storage[9]));
+
+    test('test m33', () => expect(matrix4.m33, matrix4.storage[10]));
+
+    test('test m34', () => expect(matrix4.m34, matrix4.storage[11]));
+
+    test('test m41', () => expect(matrix4.m41, matrix4.storage[12]));
+    test('test m42', () => expect(matrix4.m42, matrix4.storage[13]));
```
spydon, line 40: New line before this one. (Or even better imho, remove all newlines between the `test mXY` tests)

renancaraujo (approved):

packages/flame/CHANGELOG.md
wolfenrain: Indeed, how much it will be annoying for me to see, I should not pollute the reference to the original PR. Will revert these changes.

packages/flame/CHANGELOG.md
wolfenrain: Both are sorta equally used the same at the beginning of the sentence (Added=31, Add=37). I'll change it to Add so we can get it merged :D

packages/flame/test/extensions/matrix4_test.dart
wolfenrain: Removed the lines inbetween.

PR #859, opened 2021-06-26
Extract shared children logic when handling components to ComponentSet

luanpotter:
# Description

This unifies how we handle "children" for both *BaseGame* and *BaseComponent* by creating a wrapper over OrderedSet.

This wrapper does all the handling of addLater and removeLater, making sure that BaseComponent does not have any concurrency modifications issues.

It also unifies the APIs, keeping most component list related methods inside the `components` (or `children` namespaces). I do expose a few most important methods, like `add`, specially since add does some extra work as well. But more specific things, like `remove`, `length`, `clear`, I left inside the namespace for clarity. I consider removing even `add` but thought that was too idiomatic `game.add`. But open to suggestions on that area.

I also didn't touch `propagateToChildren` here, nor rebalancing. But I believe those two can be simplified now that both Base* use the same structure. That is just implementation details though, which I might look into next. Also wanna draw a state flowchart to better explain the difference between mount and load (that I had forgotten).

## Checklist

## Breaking Change

### DIFF
--- a/examples/lib/stories/components/composability.dart
+++ b/examples/lib/stories/components/composability.dart
@@ -9,7 +9,7 @@ class Square extends PositionComponent {
   }
 }
 
-class ParentSquare extends Square {
+class ParentSquare extends Square with HasGameRef {
   ParentSquare(Vector2 position, Vector2 size) : super(position, size);
 
   @override
@@ -27,7 +27,7 @@ class ParentSquare extends Square {
       Square(Vector2(70, 200), Vector2(50, 50), angle: 5),
     ];
 
-    children.forEach(addChild);
+    children.forEach((c) => addChild(c, gameRef: gameRef));
   }
 }
 
--- a/examples/lib/stories/components/priority.dart
+++ b/examples/lib/stories/components/priority.dart
@@ -52,6 +52,6 @@ class Priority extends BaseGame with HasTapableComponents {
       Square(Vector2(170, 150)),
       Square(Vector2(110, 150)),
     ];
-    addAll(squares);
+    components.addAll(squares);
   }
 }
--- a/examples/lib/stories/components/priority.dart
+++ b/examples/lib/stories/components/priority.dart
@@ -52,6 +52,6 @@ class Priority extends BaseGame with HasTapableComponents {
       Square(Vector2(170, 150)),
       Square(Vector2(110, 150)),
     ];
-    components.addAll(squares);
+    addAll(squares);
   }
 }
--- a/packages/flame/example/lib/main.dart
+++ b/packages/flame/example/lib/main.dart
@@ -70,7 +70,7 @@ class MyGame extends BaseGame with DoubleTapDetector, TapDetector {
 
     final handled = components.any((c) {
       if (c is PositionComponent && c.toRect().overlaps(touchArea)) {
-        remove(c);
+        components.remove(c);
         return true;
       }
       return false;
--- a/packages/flame/lib/src/components/base_component.dart
+++ b/packages/flame/lib/src/components/base_component.dart
@@ -134,13 +134,7 @@ abstract class BaseComponent extends Component {
   /// Get a list of non removed effects
   List<ComponentEffect> get effects => _effectsHandler.effects;
 
-  /// Uses the game passed in, or uses the game from [HasGameRef] otherwise,
-  /// to prepare the child component before it is added to the list of children.
-  /// Note that this component needs to be added to the game first if
-  /// [this.gameRef] should be used to prepare the child.
-  /// For children that don't need preparation from the game instance can
-  /// disregard both the options given above.
-  Future<void> addChild(Component child, {Game? gameRef}) async {
+  void prepare(Component child, {Game? gameRef}) {
     if (this is HasGameRef) {
       final c = this as HasGameRef;
       gameRef ??= c.hasGameRef ? c.gameRef : null;
@@ -158,26 +152,16 @@ abstract class BaseComponent extends Component {
       child._parent = this;
       child.debugMode = debugMode;
     }
-
-    final childOnLoadFuture = child.onLoad();
-    if (childOnLoadFuture != null) {
-      await childOnLoadFuture;
-    }
-    children.add(child);
-    if (isMounted) {
-      child.onMount();
-    }
   }
 
-  Future<void> addChildren(
-    Iterable<Component> children, {
-    Game? gameRef,
-  }) async {
-    await Future.wait(
-      children.map(
-        (child) => addChild(child, gameRef: gameRef),
-      ),
-    );
+  /// Uses the game passed in, or uses the game from [HasGameRef] otherwise,
+  /// to prepare the child component before it is added to the list of children.
+  /// Note that this component needs to be added to the game first if
+  /// [this.gameRef] should be used to prepare the child.
+  /// For children that don't need preparation from the game instance can
+  /// disregard both the options given above.
+  Future<void> addChild(Component child, {BaseGame? gameRef}) {
+    return children.addChild(child, gameRef: gameRef);
   }
 
   bool containsChild(Component c) => children.contains(c);
@@ -219,7 +203,7 @@ abstract class BaseComponent extends Component {
   }
 
   ComponentSet createComponentSet() {
-    final components = ComponentSet.createDefault();
+    final components = ComponentSet.createDefault(prepare);
     if (this is HasGameRef) {
       components.register<HasGameRef>();
     }
--- a/packages/flame/lib/src/components/base_component.dart
+++ b/packages/flame/lib/src/components/base_component.dart
@@ -164,8 +164,22 @@ abstract class BaseComponent extends Component {
     return children.addChild(child, gameRef: gameRef);
   }
 
+  /// Adds mutiple children.
+  ///
+  /// See [addChild] for details (or `children.addChildren()`).
+  Future<void> addChildren(List<Component> cs, {BaseGame? gameRef}) {
+    return children.addChildren(cs, gameRef: gameRef);
+  }
+
+  /// Whether the children list contains the given component.
+  ///
+  /// This method uses reference equality.
   bool containsChild(Component c) => children.contains(c);
 
+  /// Call this if any of this component's children priorities have changed
+  /// at runtime.
+  ///
+  /// This will call `rebalanceAll` on the [children] ordered set.
   void reorderChildren() => children.rebalanceAll();
 
   /// This method first calls the passed handler on the leaves in the tree,
--- a/packages/flame/lib/src/components/base_component.dart
+++ b/packages/flame/lib/src/components/base_component.dart
@@ -1,9 +1,6 @@
-import 'dart:collection';
 import 'dart:ui';
 
 import 'package:meta/meta.dart';
-import 'package:ordered_set/comparing.dart';
-import 'package:ordered_set/ordered_set.dart';
 
 import '../../game.dart';
 import '../../gestures.dart';
@@ -12,6 +9,7 @@ import '../effects/effects_handler.dart';
 import '../extensions/vector2.dart';
 import '../text.dart';
 import 'component.dart';
+import 'component_set.dart';
 import 'mixins/has_game_ref.dart';
 
 /// This can be extended to represent a basic Component for your game.
@@ -22,22 +20,13 @@ import 'mixins/has_game_ref.dart';
 abstract class BaseComponent extends Component {
   final EffectsHandler _effectsHandler = EffectsHandler();
 
-  final OrderedSet<Component> _children =
-      OrderedSet(Comparing.on((c) => c.priority));
+  late final ComponentSet children = createComponentSet();
 
   /// If the component has a parent it will be set here
   BaseComponent? _parent;
 
   @override
   BaseComponent? get parent => _parent;
-
-  /// The children list shouldn't be modified directly, that is why an
-  /// [UnmodifiableListView] is used. If you want to add children use the
-  /// [addChild] method, and if you want to propagate something to the children
-  /// use the [propagateToChildren] method.
-  UnmodifiableListView<Component> get children {
-    return UnmodifiableListView<Component>(_children);
-  }
 
   /// This is set by the BaseGame to tell this component to render additional debug information,
   /// like borders, coordinates, etc.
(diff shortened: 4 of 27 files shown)


examples/lib/stories/components/priority.dart
```diff
@@ -52,6 +52,6 @@ class Priority extends BaseGame with HasTapableComponents {
       Square(Vector2(170, 150)),
       Square(Vector2(110, 150)),
     ];
-    addAll(squares);
+    components.addAll(squares);
```
luanpotter, line 55: addAll is one that was considering if could be added proxied

examples/lib/stories/components/composability.dart
```diff
@@ -27,7 +27,7 @@ class ParentSquare extends Square {
       Square(Vector2(70, 200), Vector2(50, 50), angle: 5),
     ];
 
-    children.forEach(addChild);
+    children.forEach((c) => addChild(c, gameRef: gameRef));
```
luanpotter, line 30: this was actually a bug. if any child had a HasGameRef it would not be populated. my new assert catches that.

packages/flame/example/lib/main.dart
```diff
@@ -70,7 +70,7 @@ class MyGame extends BaseGame with DoubleTapDetector, TapDetector {
 
     final handled = components.any((c) {
       if (c is PositionComponent && c.toRect().overlaps(touchArea)) {
-        remove(c);
+        components.remove(c);
```
luanpotter, line 73: again, another method that I decided not to proxy. open to suggestions

packages/flame/lib/src/components/base_component.dart
```diff
@@ -236,4 +201,12 @@ abstract class BaseComponent extends Component {
   Vector2 eventPosition(PositionInfo info) {
     return isHud ? info.eventPosition.widget : info.eventPosition.game;
   }
+
+  ComponentSet createComponentSet() {
+    final components = ComponentSet.createDefault(prepare);
+    if (this is HasGameRef) {
+      components.register<HasGameRef>();
```
luanpotter, line 208: this allows HasGameRef to do `query<HasGameRef>()`, which should improve performance.
also I make all ordered sets QOS by default. so no more complications into whether we can use QOS's features or not

packages/flame/lib/src/components/mixins/has_collidables.dart
```diff
@@ -1,12 +1,9 @@
...
-
 import '../../../game.dart';
 import '../../components/mixins/collidable.dart';
 import '../../geometry/collision_detection.dart';
 
 mixin HasCollidables on BaseGame {
   void handleCollidables() {
-    final qos = components as QueryableOrderedSet;
-    collisionDetection(qos.query<Collidable>());
+    collisionDetection(components.query<Collidable>());
```
luanpotter, line 7: everything is QOS now

packages/flame/lib/src/components/mixins/has_game_ref.dart
```diff
@@ -17,9 +17,10 @@ mixin HasGameRef<T extends Game> {
   set gameRef(T gameRef) {
     _gameRef = gameRef;
     if (this is BaseComponent) {
+      // TODO(luan) this is wrong, should be done using propagateToChildren
```
luanpotter, line 20: on a HasGameRef > no game ref > HasGameRef chain this will not work.

I will revisit this when trying to generify propageToChildren later.

packages/flame/lib/src/game/base_game.dart
```diff
@@ -97,14 +83,14 @@ class BaseGame extends Game with FPSCounter {
   ///
   /// You can return a specific sub-class of OrderedSet, like
   /// [QueryableOrderedSet] for example, that we use for Collidables.
-  OrderedSet<Component> createOrderedSet() {
-    final comparator = Comparing.on<Component>((c) => c.priority);
+  ComponentSet createComponentSet() {
+    final components = ComponentSet.createDefault(
+      (c, {BaseGame? gameRef}) => prepare(c),
```
luanpotter, line 88: gameRef is always `this` on this case, no need to wire that up

packages/flame/lib/src/game/base_game.dart
```diff
@@ -156,38 +147,8 @@ class BaseGame extends Game with FPSCounter {
   /// This methods is an async operation since it await the `onLoad` method of the component. Nevertheless, this method only need to be waited to finish if by some reason, your logic needs to be sure that the component has finished loading, otherwise, this method can be called without waiting for it to finish as the BaseGame already handle the loading of the component.
   ///
   /// *Note:* Do not add components on the game constructor. This method can only be called after the game already has its layout set, this can be verified by the [hasLayout] property, to add components upon a game initialization, the [onLoad] method can be used instead.
-  Future<void> add(Component c) async {
-    assert(
-      hasLayout,
-      '"add" called before the game is ready. Did you try to access it on the Game constructor? Use the "onLoad" method instead.',
-    );
```
luanpotter, line 163: moved to prepare

packages/flame/lib/src/game/base_game.dart
```diff
@@ -156,38 +147,8 @@ class BaseGame extends Game with FPSCounter {
...
-    prepare(c);
-    final loadFuture = c.onLoad();
-
-    if (loadFuture != null) {
-      await loadFuture;
-    }
```
luanpotter, line 169: this is done by ComponentSet now

packages/flame/lib/src/game/base_game.dart
```diff
@@ -156,38 +147,8 @@ class BaseGame extends Game with FPSCounter {
...
-  }
-
-  /// Marks a list of components to be removed from the components list on the next game tick
-  void removeAll(Iterable<Component> components) {
-    _removeLater.addAll(components);
-  }
-
-  /// Marks all existing components to be removed from the components list on the next game tick
-  void clear() {
-    _removeLater.addAll(components);
```
luanpotter, line 190: proxy cleanup

packages/flame_forge2d/lib/forge2d_game.dart
```diff
@@ -54,15 +53,6 @@ class Forge2DGame extends BaseGame {
     }
   }
 
-  @override
-  void remove(Component component) {
```
luanpotter, line 58: game has no "remove" no more :) 
also this would not work for removeAll before.

packages/flame/lib/src/components/mixins/draggable.dart
```diff
@@ -84,7 +84,7 @@ mixin HasDraggableComponents on BaseGame {
   }
 
   void _onGenericEventReceived(bool Function(Draggable) handler) {
-    for (final c in components.toList().reversed) {
+    for (final c in components.reversed()) {
```
spydon, line 87: I think I might have written this initially, but we should probably just iterate over the list in reverse instead of creating a new reversed list to iterate over.

packages/flame/lib/src/game/base_game.dart
```diff
@@ -156,38 +147,8 @@ class BaseGame extends Game with FPSCounter {
...
+  Future<void> add(Component c) {
+    return components.addChild(c);
```
spydon, line 151: I think that if we have `add` here, we should also have the other methods proxied.

spydon (commented):
Nicely streamlined with QOS! I do think we should keep the proxies though (or alternatively remove `game.add` too.

examples/lib/stories/components/composability.dart
luanpotter: no, but they could

packages/flame/lib/src/components/mixins/draggable.dart
luanpotter: that does exactly the same thing as before

examples/lib/stories/components/composability.dart
spydon: But this is an example that the writer of the example is in total control over writing that game? I don't see the need to give the children a `gameRef` when they don't need it?

packages/flame/lib/src/game/base_game.dart
luanpotter: I thought about it, but the problem is there are so many collection methods, so where do we draw the line?

packages/flame/lib/src/components/mixins/draggable.dart
spydon: That's what I'm saying, that it was inefficient before too and that we should consider not creating a new list every time an event is fired. So either maintain another list in the reverse order inside of QOS, or iterate it from the end.

packages/flame/lib/src/game/base_game.dart
spydon: I would say that the ones we had was a pretty good place to draw the line, but I do see your point.

spydon: Now when I have thought about it a bit more I think I actually prefer the proxies removed, like you have done (possibly adding `addAll` too like you said). Since removal can just as well be done on a component level for single components, and the rest might be clearer if they are on the collection. I think `add` is good to keep though since it is so heavily used.

luanpotter: add, addAll, remove, clear? I thought it could be confusing, specially game.clear vs game.components.clear. but no strong opinion, I can definitely add them again as I said. also the baseComponent didn't expose all those, so we should unify that as well. but let's here what everyone has to say

packages/flame/lib/src/components/mixins/draggable.dart
luanpotter: ah gotcha, let me see how to iterate properly. I think that was the reason I "extracted" to a method but then I forgot to look into it

examples/lib/stories/components/composability.dart
luanpotter: that is true, but I thought it would lead to confusing behavior to allow this, might cause not so clever users to get into weird states and think it's a bug

spydon: If that is the case, maybe we should consider having the `gameRef` on all `BaseComponent` as default instead?

spydon: Ah right, that wouldn't solve the case when children are added to the component before the component is added to the game though...

packages/flame/lib/src/components/mixins/draggable.dart
luanpotter: I don't think it's possible, sadly. unless we keep a secondary copy of the entire set in sync with negative priorities. you can't reverse an interator just by the nature of the interface that it exposes. you have to first materialize the whole list

spydon: Okay, let's handle this later if it becomes a bottle neck.

spydon (approved):
I think we should add `addAll` and after that I think it looks good!

examples/lib/stories/components/composability.dart
luanpotter: I think we can improve that on followups, it's not directly related to this PR

packages/flame/lib/src/game/base_game.dart
luanpotter: I will add addAll back, and only that. we can reconsider the api later of course

examples/lib/stories/components/priority.dart
spydon: Now when `addAll` is back that one should be used

packages/flame/example/lib/main.dart
spydon: Maybe we should start doing `c.remove()` in our examples, not sure which one is clearer.

packages/flame_forge2d/example/lib/blob_sample.dart
```diff
@@ -94,7 +94,7 @@ class BlobSample extends Forge2DGame with TapDetector {
     final worldCenter = screenToWorld(size * camera.zoom / 2);
     final blobCenter = worldCenter + Vector2(0, 10);
     final blobRadius = Vector2.all(6.0);
-    addAll(createBoundaries(this));
+    components.addAll(createBoundaries(this));
```
erickzanardo, line 97: Can we keep addAll here?

packages/flame_forge2d/example/lib/sprite_body_sample.dart
```diff
@@ -46,7 +46,7 @@ class SpriteBodySample extends Forge2DGame with TapDetector {
   @override
   Future<void> onLoad() async {
     _pizzaImage = await images.load('pizza.png');
-    addAll(createBoundaries(this));
+    components.addAll(createBoundaries(this));
```
erickzanardo, line 49: Can we keep addAll here?

packages/flame_forge2d/lib/body_component.dart
```diff
@@ -133,4 +133,10 @@ abstract class BodyComponent<T extends Forge2DGame> extends BaseComponent
   bool containsPoint(Vector2 point) {
     return body.fixtures.any((fixture) => fixture.testPoint(point));
   }
+
+  @override
+  void onRemove() {
```
erickzanardo, line 138: This seems to be a different fix right? It deserves its own PR, or its own entry on the changelog

examples/lib/stories/components/priority.dart
luanpotter: ah I added only to BaseGame. let me add to BaseComp

examples/lib/stories/components/composability.dart
luanpotter: we only need to pass when the parent doesn't have it

packages/flame/lib/src/components/base_component.dart
```diff
@@ -236,4 +201,12 @@ abstract class BaseComponent extends Component {
   Vector2 eventPosition(PositionInfo info) {
     return isHud ? info.eventPosition.widget : info.eventPosition.game;
   }
+
+  ComponentSet createComponentSet() {
```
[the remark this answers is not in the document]
luanpotter: yes, you might want to extend it

packages/flame_forge2d/lib/body_component.dart
luanpotter: it's not a fix, just changing the way it's done to work well with this PR

erickzanardo: Still, at least a changelog entry is needed right, this is on another package.

erickzanardo (approved):
LGTM once the changelog entry for forged2d is added

packages/flame_forge2d/lib/body_component.dart
luanpotter: I don't see why, there is no functional difference right? just impl detail

luanpotter: but I can add one

PR #880, opened 2021-07-14
Initial implementation of paint effects

erickzanardo:
# Description

This PR adds an initial implementation for the paint effects (like OpacityEffect for example)

https://user-images.githubusercontent.com/835641/125707623-dd2bb61b-b20b-4a59-9de3-7fdced99f058.mov

https://user-images.githubusercontent.com/835641/126021233-ab073c2e-26d4-41e2-bc5d-a62765125f04.mov

## Checklist

- [ ] I read and followed the [Flame Style Guide].

## Breaking Change

## Related Issues


packages/flame/lib/src/effects/opacity_effect.dart
```diff
@@ -0,0 +1,82 @@
...
+    component?.paint.color = _original;
+  }
+
+  @override
+  void update(double dt) {
+    super.update(dt);
+    final color = component?.paint.color;
+    if (color != null) {
+      component?.paint.color = _original.withOpacity(
+          _original.opacity - _difference * curveProgress,
```
renancaraujo, line 78:
```suggestion
         lerpDouble(_original.opacity, _final.opacity, curveProgress);
```

Didnt test burt it should work.

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -0,0 +1,11 @@
+import 'dart:ui';
+
+import '../../../components.dart';
+import '../../palette.dart';
+
+mixin HasPaint on BaseComponent {
```
renancaraujo, line 6: I would document this justifying its existence and what uses it [the effect].

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -0,0 +1,11 @@
...
+  /// Use this to change the paint used (to apply tint or opacity).
+  ///
+  /// If not provided the default is full white (no tint).
+  Paint paint = BasicPalette.white.paint();
```
renancaraujo, line 10: The very name of the property is not clear on what it is about. Paint of what? the whole component? Who uses it? etc

packages/flame/lib/src/components/mixins/has_paint.dart
renancaraujo: Also, there is no reference on how to use it on custom components.

packages/flame/lib/src/components/mixins/has_paint.dart
spydon: Isn't that the point? So that it can be used for anything.

packages/flame/lib/src/components/mixins/has_paint.dart
spydon, line 6: Is it possible for this to have a list of paints (or a map) maybe? I'm thinking that many components might have more than one paint.
And then we could specify with index / map key which of them that should be changed in the effect?

renancaraujo:
What if instead of another mixin, we make the effect work out of the box with any component?

Only if we could provide a way to make the components receive a callback to be invoked on components prerender.

Something like
```dart

void preRenderHook(Canvas c) {
 // prepare layer or something like that
}

compoenen.addPreRenderCallback(preRenderHook);
```

packages/flame/lib/src/effects/opacity_effect.dart
```diff
@@ -0,0 +1,82 @@
+import 'dart:ui';
+
+import 'package:flutter/animation.dart';
+
+import '../components/mixins/has_paint.dart';
+import 'effects.dart';
+
+class OpacityEffect extends ComponentEffect<HasPaint> {
```
spydon, line 8: I would create a `ColorEffect` that this one extends and then the color effect just lerps each color field separately (RGBA) from one `Color` to another `Color`, since `opacity` is a field in `Color`.

spydon (commented):
Great job! I put in some suggestions, and cool that you managed to hook in to the `ComponentEffect` without any breaking change. :smile:

spydon:
> What if instead of another mixin, we make the effect work out of the box with any component?
> 
> Only if we could provide a way to make the components receive a callback to be invoked on components prerender.
> 
> Something like
> 
> ```dart
> void preRenderHook(Canvas c) {
>  // prepare layer or something like that
> }
> 
> compoenen.addPreRenderCallback(preRenderHook);
> ```

I think that would be more work for the user, the effects can be implemented for any `BaseComponent` currently, do we really need it for `Component`? Since that is supposed to barely have anything implemented on it.

renancaraujo:
> I think that would be more work for the user, the effects can be implemented for any `BaseComponent` currently, do we really need it for `Component`? Since that is supposed to barely have anything implemented on it.

That example was internal code for the effect to use. The user code would be nothing but the addEffect call. No mixin. 

It would be less work for the user since it wouldn't have to use a mixin in its component and figure out what that paint is for. 

With the proposed implementation, the opacity effect will only work if the said component is `HasPaint`.  To reproduce it on a custom component that is not a sprite one, it would require consulting the `paint`'s alpha and interpret it into its internal logic. It is counterintuitive. 

> Since that is supposed to barely have anything implemented on it.
The implementations on `Component` are all related to lifecycle, which is the case.

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: Unsure what you mean about justifying its existence

erickzanardo: Hmmm, it is possible sure, I can't think on components that may need more than one paint, do you have an example?

My fear of needing to pass a name or index to the effects API is that the effects api may end up a little confusing

spydon:
> It would be less work for the user since it wouldn't have to use a mixin in its component and figure out what that paint is for.
Yeah, I agree that could be confusing, that is why I suggested to have a map that can have named keys in the `HasPaint` mixin.

> With the proposed implementation, the opacity effect will only work if the said component is `HasPaint`. To reproduce it on a custom component that is not a sprite one, it would require consulting the `paint`'s alpha and interpret it into its internal logic. It is counterintuitive.
I'm not sure I understand what you mean here, what is it that would happen in the `preRenderHook`?

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: Yeah, I am with Lukas on this one, this mixin just says that a component has a paint and add a paint property like the names says

packages/flame/lib/src/effects/opacity_effect.dart
erickzanardo: I will governor a try, but I think that we need to have separate effects because iirc to change the color we need to use color blending

packages/flame/lib/src/components/mixins/has_paint.dart
spydon: Anything with a background and a foreground, but I agree it might become confusing.
But the constructor could default to the first if there is no map-key specified? If it most often would be just one.

packages/flame/lib/src/effects/opacity_effect.dart
spydon: You could use something like this:
```dart
Color.fromARGB(
      lerp(original.alpha, end.alpha, curveProgress),
      lerp(original.red, end.green, curveProgress),
      lerp(original.green, end.green, curveProgress),
      lerp(original.blue, end.blue,  );`
```

erickzanardo:
I don't know how this could be counterintuitive, we use a lot of mixins in Flame to add behaviors or characteristics to our components.

And I am unsure if we can prepare the canvas somehow to make this work, I could test

spydon:
> I don't know how this could be counterintuitive, we use a lot of mixins in Flame to add behaviors or characteristics to our components.
> 
> And I am unsure if we can prepare the canvas somehow to make this work, I could test

I think it will be more confusing if we fiddle with the canvas for this, especially if there would be several `Opacity/Color` effects affecting different parts of a component (like a child for example).

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: @spydon I still need to improve docs and make tests for this but I have applied the changes to enable this suggestion, lmk what do you think.

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -0,0 +1,69 @@
+import 'dart:ui';
+
+import '../../../components.dart';
+import '../../palette.dart';
+
+const _kMain = 'main';
```
spydon, line 6: I think the k thing isn't recommended by the style guide anymore

spydon (approved):
Lgtm!! So let's try a colorblending effect in a later PR? :)

erickzanardo:
> Lgtm!! So let's try a colorblending effect in a later PR? :)

@spydon indeed we need color blend to apply a color to the paint, since I tested and I already had it working, I included it on the PR

packages/flame/lib/src/effects/color_effect.dart
```diff
@@ -0,0 +1,61 @@
...
+  final double duration;
+  final String? paintId;
+
+  late Paint _original;
+
+  late ColorTween _tween;
+
+  ColorEffect({
+    required this.color,
+    required this.duration,
```
spydon, line 19: Should we have `speed` here too, like the other effects? Maybe it doesn't make sense to express it as `speed`.

spydon (approved):

packages/flame/lib/src/effects/color_effect.dart
erickzanardo: Yeah, I thought on using speed here, but I can't visualize how to use it on these kind o effects, I think speed is just the result of the difference of color by the duration.

But if you have a good idea on how to use speed here, let me know and I can change

packages/flame/lib/src/effects/opacity_effect.dart
erickzanardo: Indeed can't use like that to tint the paint, I have added a `tint` method on the mixin and a color effect though

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: I believe it still is: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#naming

doc/effects.md
```diff
@@ -180,14 +180,67 @@ Usage example:
...
-    isInfinite: true, 
+    isInfinite: true,
     isAlternating: true);
 myComponent.addEffect(combination);
 ```
 
 An example of how to use the `CombinedEffect` can be found
 [here](https://github.com/flame-engine/flame/tree/main/examples/lib/stories/effects/combined_effect.dart).
 
+## Common for paint effects
```
luanpotter, line 191: I think instead of having `## Common for MoveEffect, ScaleEffect and RotateEffect (SimplePositionComponentEffects)` and this, we could have second-level `##` with the type of effect, eg `## Position Effects`, `## Paint Effects`, followed by a brief desc and second level `###` chapters for each type (here and above). Wdyt?

packages/flame/lib/src/components/mixins/has_paint.dart
luanpotter: can't this be a static variable on the mixin?

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -0,0 +1,81 @@
...
+
+/// Adds a collection of paints to a component
+///
+/// Component will always have a main Paint that can be accessed
+/// by the [paint] getter/setter and other paints can be manipulated/accessed
+/// using [getPaint], [setPaint] and [deletePaint]
+mixin HasPaint on BaseComponent {
+  final Map<String, Paint> _paints = {
+    'main': BasicPalette.white.paint(),
```
luanpotter, line 15: constant?

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -0,0 +1,81 @@
...
+
+const _kMain = 'main';
+
+/// Adds a collection of paints to a component
+///
+/// Component will always have a main Paint that can be accessed
+/// by the [paint] getter/setter and other paints can be manipulated/accessed
+/// using [getPaint], [setPaint] and [deletePaint]
+mixin HasPaint on BaseComponent {
+  final Map<String, Paint> _paints = {
```
luanpotter, line 14: do we have any comp that has more than one paint? I found this map quite confusing to understand

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: Could be, but what's the benefit?

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: Not yet, but with this we could allow users to have custom components with more than one paints. Initially I had a hard time understanding why to have a collection of paint but @spydon made some fair points about this, there were a discussion about this point on this PR, take a look on it for more info

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: Can't have constant in a method call like this

doc/effects.md
erickzanardo: Yeah, I think it could look better indeed, but I just followed the current structure. Changing to your proposed structure would make me need to change more things on the doc unrelated to this PR, so I think we could do it, but in a another PR. wdyt?

luanpotter: Sure, it can be a followup

packages/flame/lib/src/components/mixins/has_paint.dart
luanpotter: I think you can, just tested on dartpad. it's not a method call, just a map literal initializer

image

but you can use constants on method calls too, in fact, anywhere that you can use a variable, you can use a constant, afaik

packages/flame/lib/src/components/mixins/has_paint.dart
luanpotter: he means explaining what it is supposed to be used for

packages/flame/lib/src/components/mixins/has_paint.dart
luanpotter: yeah I agree, HasPaint makes it clear that it has a paint, what it is used for has to be specified by each component that uses the mixin.

packages/flame/lib/src/components/mixins/has_paint.dart
luanpotter: global variables should be avoided at all costs, they have more scope than necessary. the lesser the scope, the better. would declare a local variable as a field if it doesn't have to be?

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: `BasicPalette.white.paint()` is a method call and used to initialize the value of the map, so we can't use const here

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: I don't think this will change anything, this variable is private to this file, so nobody else will have access to it. If I declare it as static const, the only real thing that will change is that I will have to write more to access it because I will need to write `HasPaint._kMain` instead of only `_kMain`.

If this variable wasn't private, I would agree with you, but since it isn't I don't see the point and see no problem on it be "global".

But if you feel too strong about it, I can change.

packages/flame/lib/src/components/mixins/has_paint.dart
luanpotter: hum, I see. I thought it was adding a lot of complexity, but I can see the use case for multiple paints. I have a different proposal though, because I dislike having arbitrary "strings" as keys, as that is not type safe, and subject to errors. we could use enums as keys. for example, see very bare-bones implementation (just to explain what I mean):

```dart
class Component {}

// generic HasPaints mixin has all the functionality and works with one or many paints
mixin HasPaints<T> on Component {
  Map<T, String> values = {};
}

// has paint always forced to have one paint
enum DefaultKey {  Main }
typedef HasPaint<T> = HasPaints<DefaultKey>;

class SimpleComponent extends Component with HasPaint {
  SimpleComponent() {
    values[DefaultKey.Main] = "bar";
    // this will not compile
    // values["Mani"] = "";
  }
}

// for complex components, you are forced to be type safe with your keys
enum EnumA { 
  Foo, Bar
}


class ComplexComponent extends Component with HasPaints<EnumA> {
  ComplexComponent() {
    values[EnumA.Foo] = "bar";
    // this will not compile
    // values[DefaultKey.Main] = "";
  }
}
```

That way whoever uses `HasPaint` is forced to have a single paint and ignores all the complexity, and however wats to use `HasPaints` has nothing to do with "main" and can create an enum to represent their paints the way they like, without magic strings.

erickzanardo: Interesting idea, I didn't liked the need for the user to need to initialize the main paint on the component, user will need to know the existence of the defaultkey enum, etc.

I will think if I can think on a middle ground solution for this

luanpotter: ah that was just an example to prove it's type safe, you can have easy to use methods defined on the mixins to faciltiate access

packages/flame/lib/src/components/mixins/has_paint.dart
luanpotter: I don't think being a method call makes any difference, in fact:

```dart
const constant = 'key';

String methodCall() {
  return 'dynamic content';
}

class A {
  final Map<String, String> foo = {
    constant: methodCall(),
  };
}

void main() {
  print(A().foo[constant]);
}
```

works just fine. and I don't think there is any difference between `methodCall()` and `BasicPalette.white.paint()`...

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: true, but I think I can think on a good way for implementing that, will implement it tonight :D

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: lol, now I get it, you want the key to be the constant, I thought you wanted to make the whole map const. Either way, I think I will change this as a whole because of our discussion on the type safe thing.

packages/flame/lib/src/components/mixins/has_paint.dart
luanpotter: Even being private, it has more scope than it needs, as there might be more than one scope/class per file (though we normally don't do that). Also, you wouldn't need to do `HasPaint._kMain`; you won't need the weird `k`, because then it will be a normal static constant, like `HasPaint._main`, like we have on many other classes, so we don't need to even consider this weird `k` thing until the need actually arises of having a global variable (which I think we should avoid at all costs as a matter of principle).

luanpotter: Also I think you won't need to call it with the full identifier since it will have the right scope, eg:

```dart
class Foo {
  static const Bar = "foobar";
  
  void run() {
    print(Bar);
  }
}

void main() {
  Foo().run();
}
```

packages/flame/lib/src/components/mixins/has_paint.dart
luanpotter: yes, agree, that will totally bypass the magic string thing (if it works), but just wanted to comment this too in case that doesn't work too well

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: I disagree with you on that, I don't find the `k` weird, it is heavily used alongside Flutter for describing keys (something this variable is). It don't have more scoped than it should, it is tied to the scope of this file so I don't see how it has more scope than it should.

That said, I will probably change this whole thing because our other discussion because of the type safe thing, so this doesn't matter anymore hahahah

luanpotter: (it works if `Bar` is private too)

luanpotter: Hum, I don't think it has anything to do with being a key, rather, it's a prefix for global variables, as explained on the styleguide you linked. if you want to make clear it's a key you can add `key` to the variable name, not `k`. in fact the examples on the styleguide there are not keys at all. and the styleguide also explicitly says: `However, where possible avoid global constants`, and I don't see any reason for it to be global here when class-level would work just as well. the scope file > scope class, so it's bigger than it needs to be rn.

but yeah, hopefully we can get rid of it entirely with some more type-safe approach

spydon: > I disagree with you on that, I don't find the `k` weird, it is heavily used alongside Flutter for describing keys (something this variable is). It don't have more scoped than it should, it is tied to the scope of this file so I don't see how it has more scope than it should.
> 
> That said, I will probably change this whole thing because our other discussion because of the type safe thing, so this doesn't matter anymore hahahah

I think there is a very good argument against it [https://dart.dev/guides/language/effective-dart/style#dont-use-prefix-letters](here).
And there is a very interesting discussion about it [https://github.com/flutter/flutter/issues/20858](here).
Otherwise, why not put types on all variables too? :thinking:

erickzanardo: it seems to me that flutter style guide and dart effective have different opinions them.

The `k` isn't the type, but rather to simbolize that it is a key, so I think they are different things, example of flutter using the k prefix: https://api.flutter.dev/flutter/foundation/kIsWeb-constant.html

Either way, I will get rid of this part of the code as result of the discussion that me an Luan about the type safe stuff, I thought on a different way of storing this which I think will look way better and will not need this key anymore.

spydon: It is not a type, but it shows something that the compiler (and your IDE) already knows, that's what I meant. 
But like you said, pointless discussion. :smile:

examples/lib/stories/effects/opacity_effect.dart
```diff
@@ -0,0 +1,44 @@
...
+            initialIsAlternating: true,
+          ),
+        ),
+    );
+  }
+
+  @override
+  void onTap() {
+    if (sprite.paint.color.opacity == 1) {
+      sprite.addEffect(OpacityEffect.fadeOff());
```
spydon, line 39: fade out?

spydon (approved):

examples/lib/stories/effects/opacity_effect.dart
erickzanardo: lol, what the hell I was thinking...

Changed!

spydon (approved):
Good job! Lgtm

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -0,0 +1,74 @@
...
+
+  /// Shortcut for changing the color of the paint
+  void setColor(Color color, {T? paintId}) {
+    getPaint(paintId).color = color;
+  }
+
+  /// Applies a color filter to the paint which will make
+  /// things rendered with the paint looking like it was
+  // tinted with the given color
+  void tint(Color color, {T? paintId}) {
```
luanpotter, line 71: this is sometimes called `key` and sometimes `paintId`, but they are the same thing right?

doc/effects.md
```diff
@@ -180,14 +180,67 @@ Usage example:
...
+
+Flame provides an useful mixin that adds paint variables to your components, that is used by some
+of Flame's own components like `SpriteComponent`, `SpriteAnimationComponent`, but you can also use
+it on you any custom component that your game has, and any component that uses that mixin can have
+paint affects applied to it.
```
luanpotter, line 196:
```suggestion
Flame provides an useful mixin that adds paint variables to your components. This mixin is used by some
of Flame's own components like `SpriteComponent`, `SpriteAnimationComponent`, but you can also use
it on you any of your custom components. Any component that uses this mixin can have
paint effects applied to it.
```

doc/effects.md
```diff
@@ -180,14 +180,67 @@ Usage example:
...
+
+By default all effects will animate the main paint of the component, that can be changed by passing
+a `paintId` to the effect contructor. This can be useful for when you have a component that have
+multiple paints, for example a component which has a foreground and background layer.
```
luanpotter, line 200:
```suggestion
By default, all effects will animate the main paint of the component. That can be changed by passing
a `paintId` [note: or key?] to the effect contructor. This can be useful for when you have a component that have
multiple paints; for example, a component which has a foreground and background layer.
```

I think before this paragraph explaining how effects behaves wrt to multiple paints, we should have a paragraph dedicated to explaining how the `HasPaint` mixing allows for one main paint and other, optional sub-paints that you can key by providing a generic type parameter (because that is not super easy to grasp so it's good to be precise before even talking about effects).

doc/effects.md
```diff
@@ -180,14 +180,67 @@ Usage example:
...
+
+Check below for a list of the available paint effects
```
luanpotter, line 202:
```suggestion
Below is a list of the available paint effects.
```

doc/effects.md
```diff
@@ -180,14 +180,67 @@ Usage example:
...
+
+## OpacityEffect
+
+This effect allows you animate the opacity of your paint, it receives a double for the opacity,
+which must be between 0 and 1 (including) and a duration in seconds, represented by a double as
+well.
```
luanpotter, line 208:
```suggestion
This effect allows you animate the opacity of your paint. It receives a double for the opacity,
which must be between 0 and 1 (including), and a duration in seconds, represented by a double as
well.
```

doc/effects.md
```diff
@@ -180,14 +180,67 @@ Usage example:
...
+);
+```
+
+An example of how to use the `OpacityEffect` can be found
+[here](https://github.com/flame-engine/flame/tree/main/examples/lib/stories/effects/opacity_effect.dart).
+
+## ColorEffect
+
+This effect will change the base color of the paint, making things rendered with it, looks like
+it was tinted with the passed color.
```
luanpotter, line 227:
```suggestion
This effect will change the base color of the paint, causing the rendered component to be tinted by the provided color.
```

doc/effects.md
```diff
@@ -180,14 +180,67 @@ Usage example:
...
+
+
+Usage Example:
```
luanpotter, line 230:
```suggestion
Usage example:
```

consistency is key :p

doc/effects.md
```diff
@@ -180,14 +180,67 @@ Usage example:
...
+```dart
+myComponent.addEffect(
+  ColorEffect(
+    color: const Color(0xFF00FF00),
+    duration: 0.5,
+  ),
+);
+```
+
+An example of how to use the `ColorEffect` can be found
```
luanpotter, line 241:
```suggestion
A more in-depth example can be found
```

examples/lib/stories/effects/opacity_effect.dart
```diff
@@ -0,0 +1,44 @@
...
+    add(
+      SpriteComponent(
+        sprite: flameSprite,
+        position: Vector2(300, 100),
+        size: Vector2(149, 211),
+      )..addEffect(
+          OpacityEffect(
+            opacity: 0,
+            duration: 0.5,
+            initialIsInfinite: true,
```
luanpotter, line 29: this is not related to this PR at all, and just an opinion:
I think this parameter should be called just `isInfinite`. I see on the impl why it's "initial" and that it can change, but the parameter name doesn't have to have the same name of the field. from the perspective of the constructor callee it feels a bit weird to me to call it "initial". thoughts?

examples/lib/stories/effects/opacity_effect.dart
```diff
@@ -0,0 +1,44 @@
...
+            initialIsAlternating: true,
+          ),
+        ),
+    );
+  }
+
+  @override
+  void onTap() {
+    if (sprite.paint.color.opacity == 1) {
```
luanpotter, line 38: nit: save `sprite.paint.color.opacity` to a variable

packages/flame/CHANGELOG.md
```diff
@@ -14,6 +14,9 @@
  - Add `BaseComponent.remove` and `BaseComponent.removeAll` helpers for removing children
  - Rename `Camera.cameraSpeed` to `Camera.speed`
  - Rename `addShape` to `addHitbox` in `Hitbox` mixin
+ - Generalize `paint` usage on fields
```
luanpotter, line 17: on components?

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -0,0 +1,74 @@
+import 'dart:ui';
+
+import '../../../components.dart';
+import '../../palette.dart';
+
+/// Adds a collection of paints to a component
+///
+/// Component will always have a main Paint that can be accessed
+/// by the [paint] attribute and other paints can be manipulated/accessed
+/// using [getPaint], [setPaint] and [deletePaint]
```
luanpotter, line 10:
```suggestion
/// using [getPaint], [setPaint] and [deletePaint] by a key of generic type [T], that can be omited if the component only have one paint.
```

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -0,0 +1,74 @@
...
+mixin HasPaint<T extends Object> on BaseComponent {
+  final Map<T, Paint> _paints = {};
+
+  Paint paint = BasicPalette.white.paint();
+
+  /// Gets a paint from the collection, returns the main paint if no [key] is provided
```
luanpotter, line 16:
```suggestion
  /// Gets a paint from the collection.
  ///
  /// Returns the main paint if no [key] is provided.
```

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -0,0 +1,74 @@
...
+
+    if (_paint == null) {
+      throw ArgumentError('No Paint found for $key');
+    }
+
+    return _paint;
+  }
+
+  /// Sets a paint on the collection
+  void setPaint(T key, Paint paint) => _paints[key] = paint;
```
luanpotter, line 32: why no make key nullable here too (and set `paint`)?

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -0,0 +1,207 @@
+import 'dart:ui';
+
+import 'package:flame/components.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+class MyComponent extends PositionComponent with HasPaint {}
```
luanpotter, line 6: I think it would we could make the tests clearer by having two components, one with HasPaint and a single paint and one with `HasPaint<String>` and type safe (or even an enum).

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -0,0 +1,74 @@
...
+
+import '../../../components.dart';
+import '../../palette.dart';
+
+/// Adds a collection of paints to a component
+///
+/// Component will always have a main Paint that can be accessed
+/// by the [paint] attribute and other paints can be manipulated/accessed
+/// using [getPaint], [setPaint] and [deletePaint]
+mixin HasPaint<T extends Object> on BaseComponent {
```
luanpotter, line 11: `T extends Object` === `T`? everything extends Object right?

luanpotter: ahh sorry I see the difference in dart. so I was thinking we could have a check:

```
if (T == Object)
```

and forbid using the `_paints` map in this case. that way people are forced to either use HasPaint w/o any key type, which will fallback to Obj and not allow using the map, or provide a specific key type and "enable" the map.

luanpotter: this assert method could be called on every method that provides a non-null `key/paintId`

luanpotter: eg (dartpad):
```dart
void main() {
  A().foo();
  A<String>().foo();
}

class A<T extends Object> {
  void foo() {
    if (T != Object) {
      print('faaa $T');
    }
  }
}
```

examples/lib/stories/effects/opacity_effect.dart
erickzanardo: Yeah, I confess that this puzzled me as well, I am gonna put a trello card about this so we can discuss and do this change on a followup maybe, since this isn't directly related to this PR.

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: Sorry, I didn't understood if this is just your train of thought, or if there is a suggestion here.

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: I could, but I don't see much point since `paint` is an attribute now. Also to make it nullable, we would need to make it named, so we would need to write more.

Unsure, no hard opinions here, I think we don't need to do that, but if you really think it will be good, I can implement.

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: yeah, I will keep them all paintId

packages/flame/lib/src/effects/color_effect.dart
erickzanardo: Can we resolve this @spydon ?

spydon: For sure!

examples/lib/stories/effects/opacity_effect.dart
erickzanardo: Unsure what you mean @spydon , `initialIsInfinite` and the other property are properties from the base effect class right?

spydon: Yeah, but they are not named arguments there. All the effects that use that call the named arguments without `initial`.

erickzanardo: ahhhh, got it, that slipped my attention, will change here then

packages/flame/lib/src/components/mixins/has_paint.dart
luanpotter: yes, I think we should have a check

packages/flame/lib/src/components/mixins/has_paint.dart
luanpotter: yeah I think you are right, not much point in having

luanpotter (approved):
LGTM but I think we should have a check to prevent the abusing of T=Object as an "anything" key.

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: can you explain more what this check would be? I am lost here

spydon (approved):

packages/flame/lib/src/components/mixins/has_paint.dart
luanpotter: I think we should add a check so that we enforce one of two cases:
* use specifies the generic type T and thus can use the map
* user does not and thus T fallsback to Object, in which case we block using the map
Otherwise, everyone that doesn't specific T will be able to add any keys they want to components that should not have extra paints.

PR #886, opened 2021-07-28
Improve STYLEGUIDE

luanpotter:
# Description

This is a proposal to better organize the styleguide's sections and subsections, moving some stuff around, removing the use of bullet items when not needed, etc.

There is no significant content change proposed (yet).

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -109,6 +109,7 @@ void something(int smaller, int bigger) {
   // ...
 }
 ````
+
 ## Comments
 
 * For any `//` comments, always add a space after the second slash and before the next character.
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -1,15 +1,17 @@
 # Style Guide
 
-This is a general style guide that shall govern over the Flame Engine repository. The aim is to keep
-a common stable general environment. This includes high level guidance to help with simple decisions
+This is a general style guide that shall govern over all Flame repositories. The aim is to keep a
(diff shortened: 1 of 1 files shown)


STYLEGUIDE.md
```diff
@@ -1,15 +1,17 @@
 # Style Guide
 
-This is a general style guide that shall govern over the Flame Engine repository. The aim is to keep
-a common stable general environment. This includes high level guidance to help with simple decisions
+This is a general style guide that shall govern over all Flame repositories. The aim is to keep a
+common stable general environment. This includes high level guidance to help with simple decisions
```
wolfenrain, line 4: "common stable general" sounds strange to me :/ what are we trying to convey here?

STYLEGUIDE.md
```diff
@@ -90,24 +92,39 @@ import 'game.dart';
...
+
+Example:
+
+````dart
+void something(int smaller, int bigger) {
+  assert(small < bigger, 'smaller is not smaller than bigger');
+  // ...
+}
+````
+## Comments
```
wolfenrain, line 112:
```suggestion

## Comments
```

wolfenrain (approved):

spydon (approved):

erickzanardo (approved):

STYLEGUIDE.md
luanpotter: i have no idea what I was thinking hahaha must have been a lapse.

luanpotter: AH that was already there haha. that explains why I don't know :P 

improved wording!

PR #887, opened 2021-07-29
Fix parallax fullscreen bug when game is resized

spydon:
# Description

Fix parallax fullscreen bug when game is resized. Before if the game was zoomed out the parallax would be minimized when the game was resized.

## Checklist

- [xj] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

### DIFF
--- a/packages/flame/lib/src/components/parallax_component.dart
+++ b/packages/flame/lib/src/components/parallax_component.dart
@@ -78,6 +78,7 @@ class ParallaxComponent<T extends BaseGame> extends PositionComponent
     )..parallax = parallax;
   }
 
+  @mustCallSuper
   @override
   void onGameResize(Vector2 size) {
     super.onGameResize(size);
--- a/packages/flame/lib/src/components/parallax_component.dart
+++ b/packages/flame/lib/src/components/parallax_component.dart
@@ -1,9 +1,11 @@
 import 'dart:async';
 import 'dart:ui';
 
+import 'package:flame/components.dart';
 import 'package:flutter/painting.dart';
 import 'package:meta/meta.dart';
 
+import '../../game.dart';
 import '../assets/images.dart';
 import '../extensions/vector2.dart';
 import '../game/game.dart';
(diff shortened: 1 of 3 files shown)


packages/flame/lib/src/components/parallax_component.dart
```diff
@@ -73,14 +78,15 @@ class ParallaxComponent extends PositionComponent {
     )..parallax = parallax;
   }
 
-  @mustCallSuper
```
erickzanardo, line 76: Why remove this line?

spydon: Because when I used this on my own `ParallaxComponent` when I found this bug I realized one might want to do their own logic here and then who are we to stop them? :man_shrugging:

erickzanardo: Yeah, but isn't the super logic important on this case? Or is this so simple that would make sense for the user to skip super?

spydon: It is important if you have children, otherwise it will just get in your way if you want to handle this yourself.
But I'm up for any, there are cons and pros with having and not having this line.
Maybe it is better to keep it and let the advanced users just ignore the warning?

erickzanardo: If there are use cases for ignoring it, I am ok on removing the mustCallSuper, and just add a note on the docs for when you have children, you may face issues if super isn't called? Wdyt?

spydon: I re-added it already :sweat_smile: 
I think the users that want to implement their own logic there will be advanced enough to know what they are doing, there isn't as much of a use-case anymore now when I changed the logic of that method in this PR.

erickzanardo: Hahhahah, alright, up to you!

erickzanardo (approved):

PR #892, opened 2021-07-31
Scale for PositionComponent

spydon:
# Description

If we want to do this we have to take it into consideration in:
 - Collision detection
 - The positioning in `preRender` has to change
 - If `width` and `height` should be with or without scaling
 - ...

Pros:
The scaling factor will affect all children and some component implementations won't have to take `size` into consideration when they need to scale.

Cons:
Makes parts of the code base more messy.

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors and are passing (See [Contributor Guide]).
- [x] My PR does not decrease the code coverage, or I have __a very special case__ and explained on the PR description why this PR decreases the coverage.
- [x] I updated/added relevant documentation (doc comments with `///`) and updated/added examples in `doc/examples`.
- [x] I have formatted my code with `./scripts/format.sh` and the Flame analyzer (`./scripts/analyze.sh`) does not report any problems.
- [x] I read and followed the [Flame Style Guide].
- [x] I have added a description of the change under `[next]` in `CHANGELOG.md`.
- [x] I removed the `Draft` status, by clicking on the `Ready for review` button in this PR.
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Does your PR require Flame users to manually update their apps to accommodate your change?

- [x] Yes, this is a breaking change (please indicate a breaking change in `CHANGELOG.md`).

## Related Issues
#891

[issue database]: https://github.com/flame-engine/flame/issues
[Contributor Guide]: https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md
[Flame Style Guide]: https://github.com/flame-engine/flame/blob/main/STYLEGUIDE.md

PR #904, opened 2021-08-05
Use Transform2D in PositionComponent

### LINKED ISSUE #891: PositionComponent can be scale()-able
# What could be improved
Currently, a `PositionComponent` can be moved, rotated, and flipped horizontally or vertically. It would be great if it could also be scaled by a certain % factor.

Note that `PositionComponent` already applies scale factors when flipping, but they are limited to ±1. It would be a simple extension to allow the component to replace `renderFlipX` and `renderFlipY` flags with variables `scaleX` and `scaleY`, which could change their signs when "flip X" or "flip Y" operation is requested, but also allowing the user to apply any other scale transform.

# Why should this be improved
A scale transform can be useful in multiple situations:
  - short animation to make a button larger when clicked;
  - a buff on a hero making him stronger and visually bigger;
  - simulate entities that are moving closer to the screen, or further into the distance;
  - etc.

# Any risks?
Variables `renderFlipX` and `renderFlipY` will be replaced with methods for flipping the component horizontally/vertically. The `containsPoint()` method will need to be adjusted to take the scaling factors into account.

st-pasha:
# Description

This is a performance optimization change. Since `PositionComponent` is one of the most basic building blocks of the engine, it is probably worthwhile making it more optimal.

This change adds `PositionComponent._transform` field, which is a common `Transform2D` class.

Added helper functions `flipHorizontally()`, `flipVertically()`, `flipHorizontallyAroundCenter()` and `flipVerticallyAroundCenter()`. The latter 2 methods can be used as replacements for fields `renderFlipX` and `renderFlipY`.

Added methods to transform between local, parent, and global (absolute) coordinate space. These methods are intended for hit-testing, collision detections, and similar needs. In particular, current methods such as `absoluteTopLeftPosition()` can now be expressed as `localToAbsolute(Anchor.topLeft * size)`. Having a clear distinction between different coordinate spaces should allow easier construction of nested component trees.

## Checklist

- [x] My PR does not decrease the code coverage, 
- [x] I have formatted my code with `flutter format .` and the Flame analyzer (`flutter analyze .`) does not report any problems.
- [x] I am willing to follow up on review comments in a timely manner.

## Breaking Change

- [x] This is a breaking change:
  - properties `renderFlipX` and `renderFlipY` removed. Instead you can use methods `flipHorizontally()`, `flipHorizontallyAroundCenter()`, `flipVertically()`, or `flipVerticallyAroundCenter()`.
  - removed `.angleTo()` -- the method was unused and was calculating a meaningless result.

## Related Issues

Closes #891


spydon:
It is a good idea, but this will introduce some bugs in current code that use the `component.position.setFrom` since there is no way of setting `recalculateTransform` when that is done.
We could also store values that the vectors had on the last generation of the matrix and compare towards those ones whether it should be regenerated or not, it is not as efficient but it will avoid these bugs. I'm thinking something in the lines of `ShapeCache` if you have looked at that part of the code.

st-pasha:
There is already a setter for `position`, so instead of `component.position.setFrom(x)` one could simply write `component.position = x`, and it even looks better.

Of course,  it could be difficult to find all places where `component.position.setFrom` is used. In such a case we could replace the getter for `position` with the one that returns an instance of `MutableVector2`,  which stores a reference to its owner and overrides the `setFrom` setter to mark the owner as dirty and issue a warning to the user that this API is deprecated and will be removed in the future. Eventually, the `position` getter would return `_position.clone()` in order to prevent accidental modification of the internal component's state.

Though I admit, this PR is not as non-breaking as I thought it was.

st-pasha:
I changed the code to use `_NotifyingVector2` which keeps a reference to the parent class and marks it dirty upon all modifications. Somehow, this solution doesn't feel quite right...

The few remaining tests that are failing are all in `camera_and_viewport_test.dart` and related to the changes in how the transform is applied to the canvas. These tests should probably be updated.

spydon:
Have you had a look at #892 too? These PRs have quite some overlap.

st-pasha:
Indeed they do, for example, they both affect `camera_and_viewport_test.dart` unit test in the same way.
However, the focus is different. This PR prepares the `PositionComponent` making it easy to add the `scale` parameter; and also doing some performance optimizations to avoid unnecessary calculations. The #892 takes it to the next step and enables the `scale` parameter to be settable, which affects many other places in the engine, and is much more breaking. Thus, this PR serves as a prequel for #892, and should make implementing it easier.

spydon:
> Indeed they do, for example, they both affect `camera_and_viewport_test.dart` unit test in the same way.
> However, the focus is different. This PR prepares the `PositionComponent` making it easy to add the `scale` parameter; and also doing some performance optimizations to avoid unnecessary calculations. The #892 takes it to the next step and enables the `scale` parameter to be settable, which affects many other places in the engine, and is much more breaking. Thus, this PR serves as a prequel for #892, and should make implementing it easier.

I don't think this will serve as a prequel, since it requires more discussion around whether we want to go this way or not. It does complicate things a fair bit. But it is up to the team too, maybe I'll get some comments I haven't thought about in #892, but other than that I think #892 is more straight forward for a merge than this one.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -192,4 +345,46 @@ abstract class PositionComponent extends BaseComponent {
       canvas.translate(-width / 2, -height / 2);
     }
   }
+} // ignore: number-of-methods
+
+class _NotifyingVector2 extends Vector2 {
```
spydon, line 350: This is pretty smart, maybe we can have this one even more general to be used in more places and just have a callback that is called whenever the `x` and `y` is changed.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -135,20 +206,106 @@ abstract class PositionComponent extends BaseComponent {
...
 
+  /// Test whether the `point` (given in global coordinates) lies within this
+  /// component. The top and the left borders of the component are inclusive,
+  /// while the bottom and the right borders are exclusive.
   @override
   bool containsPoint(Vector2 point) {
-    final rectangle = Rectangle.fromRect(toAbsoluteRect(), angle: angle)
-      ..position = absoluteCenter;
-    return rectangle.containsPoint(point);
+    final local = absoluteToLocal(point);
```
spydon, line 304: This does not take angle into consideration

packages/flame/lib/src/components/position_component.dart
```diff
@@ -135,20 +206,106 @@ abstract class PositionComponent extends BaseComponent {
   PositionComponent({
     Vector2? position,
     Vector2? size,
-    this.angle = 0.0,
+    double angle = 0.0,
     this.anchor = Anchor.topLeft,
```
spydon, line 210: If the anchor is changed we also need to recalculate the matrix right?
(Couldn't comment on the correct place since nothing was changed around there)

packages/flame/lib/src/components/position_component.dart
```diff
@@ -135,20 +206,106 @@ abstract class PositionComponent extends BaseComponent {
...
+  }
+
+  /// The total transformation matrix for the component. This matrix combines
+  /// translation, rotation and scale transforms into a single entity. The
+  /// matrix is cached and gets recalculated only as necessary.
+  Matrix4 get transformMatrix {
+    if (_recalculateTransform) {
+      final m = _transformMatrix.storage;
+      final cosA = math.cos(_angle);
+      final sinA = math.sin(_angle);
```
spydon, line 231: Can't you use `_transformMatrix.rotate/translate/scale`?
I think readability here will give more than the small performance gain in not calling the `Matrix4` methods.

packages/flame/lib/src/components/position_component.dart
st-pasha: Perhaps it is not that obvious, but `absoluteToLocal()` chains `parentToLocal()` along the path to the root of the component tree. At the same time, `parentToLocal()` applies an inverse of the `_transformMatrix` to the given `point`. Thus, this method takes into account all the transforms that were composed within the transform matrix: translations, rotations, scaling, and flips.

packages/flame/lib/src/components/position_component.dart
st-pasha: A more general version could simply take a callback function that should be invoked when any change occurs.

spydon: Yeah, that's what I meant. :)

packages/flame/lib/src/components/position_component.dart
spydon: Aah makes sense! Could you add that to the dartdoc comments? (if it is not already there)
Someone asked for these types of methods on discord the other day.

packages/flame/lib/src/components/position_component.dart
st-pasha: Yes, you're right, since the transform depends on the anchor point. I'll add a setter for the anchor.

packages/flame/lib/src/components/position_component.dart
st-pasha: I've added a comment to clarify what's going here. This way there are both the human-readable and CPU-optimized versions side by side. In addition, I added a test `'transform matrix'` to verify that these two forms are equivalent (except for the scale, the scale is currently not settable).

Hopefully, this would solve the readability concerns while retaining efficiency.

spydon: That is a good middle ground. 

Were you able to join discord btw?
Then we could discuss how to tackle the merge conflicts that we will get between the two PRs.
https://discord.gg/qTexXyqR

packages/flame/lib/src/game/transform2d.dart
```diff
@@ -0,0 +1,172 @@
+import 'dart:math' as math;
+import 'package:flutter/foundation.dart';
+import 'package:vector_math/vector_math_64.dart';
+
+class Transform2D extends ChangeNotifier {
```
spydon, line 5: This will be soo useful, good idea!

packages/flame/lib/src/game/transform2d.dart
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  void _notify() {
+    _recalculate = true;
+    notifyListeners();
+  }
+}
+
+// ignore: prefer_mixin
+class NotifyingVector2 extends Vector2 with ChangeNotifier {
```
st-pasha, line 134: Note: I created a separate PR for this class.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -121,14 +138,27 @@ abstract class PositionComponent extends BaseComponent {
...
 
-  /// Anchor point for this component. This is where flame "grabs it".
-  /// The [position] is relative to this point inside the component.
-  /// The [angle] is rotated around this point.
-  Anchor anchor;
+  /// Flip the component vertically around its anchor point.
+  void flipVertically() => _transform.flipVertically();
+
+  /// Flip the component horizontally around its center line.
+  void flipHorizontallyAroundCenter() {
```
spydon, line 148: Will these work if they are called two times in a row? To first flip it, and then flip it back?

st-pasha: I hope so. Probably can add a test for this.

spydon:
@st-pasha can you rebase this on `main` again?
And then if you feel that it is ready for review you can move it from `Draft` to `Ready for review`.

st-pasha:
@spydon Not ready yet. I want to move out `Transform2D` into a separate PR with its own set of tests. Then this PR will be about integrating Transform2D into the `PositionComponent`; and I can start working in parallel on integrating Transform2D into `Camera`, and maybe even into the effects system.

st-pasha:
@spydon -- I was thinking that perhaps we should change `.width` and `.height` back so that it refers to the original (local) size of the component so that `size == (width, height)`.

That way we could restore the setters for those fields, and avoid possible misunderstandings. If someone needs scaled width/height, they could use `scaledSize.x` and `scaledSize.y`.

spydon:
> @spydon -- I was thinking that perhaps we should change `.width` and `.height` back so that it refers to the original (local) size of the component so that `size == (width, height)`.
> 
> That way we could restore the setters for those fields, and avoid possible misunderstandings. If someone needs scaled width/height, they could use `scaledSize.x` and `scaledSize.y`.

Sounds good to me

st-pasha:
@spydon One other API nitpick: we have properties
- `.topLeftPosition`
- `.absoluteTopLeftPosition`
- `.position`
- `.absolutePosition`
- `.center`
- `.absoluteCenter`
and these names appear to be somewhat inconsistent. Should we do something about this? For example, perhaps rename `topLeftPosition` into simply `topLeft`? Alternatively, we could rename `center` -> `centerPosition`, and add `anchorPosition` as the synonym for `position`. WDYT?

When reviewing, please check the names of the methods that I added and see if they are reasonable enough.

examples/lib/stories/input/joystick_advanced.dart
```diff
@@ -75,7 +75,7 @@ class JoystickAdvancedGame extends BaseGame
         right: 160,
         bottom: 60,
       ),
-      onPressed: () => player.renderFlipY = !player.renderFlipY,
+      onPressed: () => player.flipVertically,
```
erickzanardo, line 78:
```suggestion
      onPressed: player.flipVertically,
```

spydon:
> @spydon One other API nitpick: we have properties
> 
>     * `.topLeftPosition`
> 
>     * `.absoluteTopLeftPosition`
> 
>     * `.position`
> 
>     * `.absolutePosition`
> 
>     * `.center`
> 
>     * `.absoluteCenter`
>       and these names appear to be somewhat inconsistent. Should we do something about this? For example, perhaps rename `topLeftPosition` into simply `topLeft`? Alternatively, we could rename `center` -> `centerPosition`, and add `anchorPosition` as the synonym for `position`. WDYT?
> 
> 
> When reviewing, please check the names of the methods that I added and see if they are reasonable enough.

Good idea, these names have grown organically into that inconsistency.
I'm not sure which version I like the most, with the `position` postfix it is maybe easier to understand what the field is, but without it it is less to write.
What do you think @flame-engine/flame-colab?

examples/lib/stories/components/composability.dart
```diff
@@ -41,20 +41,20 @@ class Composability extends BaseGame with TapDetector {
 
   @override
   Future<void> onLoad() async {
-    _parent = ParentSquare(Vector2.all(200), Vector2.all(300))
+    _parent = ParentSquare(Vector2.all(400), Vector2.all(250))
```
spydon, line 44: What is the reason to change things in this class?

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,326 @@
...
+  /// rotated around its anchor point in the clockwise direction if the
+  /// angle is positive, or counterclockwise if the angle is negative.
+  double get angle => _transform.angle;
+  set angle(double a) => _transform.angle = a;
+
+  /// Rotation angle (in degrees) of the component. The component will be
+  /// rotated around its anchor point in the clockwise direction if the
+  /// angle is positive, or counterclockwise if the angle is negative.
+  double get angleDegrees => _transform.angleDegrees;
+  set angleDegrees(double a) => _transform.angleDegrees = a;
```
spydon, line 98: I think we intentionally left this out to not have a mix of radians and degrees throughout the games that are made.
Do you rememember @flame-engine/flame-colabc

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,326 @@
...
-  Vector2 get absoluteTopLeftPosition {
-    final p = parent;
-    if (p is PositionComponent) {
-      return p.absoluteTopLeftPosition + topLeftPosition;
-    } else {
-      return topLeftPosition;
+  /// Convert local coordinates of a point [p] inside the component into
+  /// the parent's coordinate space. Point [p] can be either a [Vector2],
+  /// an [Anchor] (relative to component's [size]), or an [Offset].
+  Vector2 positionOf(dynamic p) {
```
spydon, line 174: I wonder if we'd maybe rather have this as two methods, so that the type system can catch errors before runtime.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,326 @@
...
+
+  /// Flip the component vertically around its center line.
+  void flipVerticallyAroundCenter() {
+    final delta = (1 - 2 * _anchor.y) * height;
+    _transform.x += -delta * math.sin(_transform.angle);
+    _transform.y += delta * math.cos(_transform.angle);
+    _transform.flipVertically();
   }
 
-  double angleTo(PositionComponent c) => position.angleTo(c.position);
```
spydon, line 160: Why was this removed?

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,326 @@
...
+/// then the picture is rotated by [angle] around the anchor point, and
+/// scaled by [scale] also around the anchor point.
+///
+/// The [size] property of the [PositionComponent] is used primarily for
+/// tap and collision detection. Thus, the [size] should be set equal to
+/// the approximate bounding rectangle of the rendered picture. If you
+/// do not specify the size of a PositionComponent, then it will be
+/// equal to zero and the component won't be able to respond to taps.
+///
+class PositionComponent extends BaseComponent {
```
spydon, line 40: Good that you removed `abstract`, I did that in my PR too, but this will probably be merged first.

examples/lib/stories/components/composability.dart
st-pasha: It was close to the edge of the screen, so part of the picture was cut off when it rotated. And I wanted to be able to see  it nice and clear, because this example is the only real visual test for nest positioned components, and for the debug paint mode.

packages/flame/lib/src/components/position_component.dart
st-pasha: You think there's potential for confusion? It's always easy to remove this property if need be.

packages/flame/lib/src/components/position_component.dart
st-pasha: Because this thing never actually worked. For example, if you have a component at (100, 100), and another component at (100, 110) then the angle from first to second is ~0.047 (or about 3 degrees). This has no geometrical meaning.

Now, I'm not against introducing some notion of an angle, but an angle requires two lines. The first line would be through the anchors of two components, but what about the other line? We can either pick some arbitrary fixed direction (like (0,1)), or make it a parameter to `angleTo`, or introduce a notion of "principal direction of a sprite" as an extra property of the PositionComponent.

But whatever we do, this needs a serious discussion, and perhaps can be considered as a new feature to be added. Right now, however, removing `angleTo` is justified because the current implementation doesn't work at all...

packages/flame/lib/src/components/position_component.dart
st-pasha: Good point. What should they be named? And should we care about `Offset`s at all?

packages/flame/lib/src/components/position_component.dart
st-pasha: Yeah, I was following your lead here :) Someone would have to update [this picture](https://flame-engine.org/docs/#/components) though

packages/flame/lib/src/components/position_component.dart
spydon: Ah right, since the lines both now go from (0, 0).
What we should do is something like this then:
`(c.position - position).angleTo(Vector2(0, 1));`

packages/flame/lib/src/components/position_component.dart
spydon: Indeed, I'll update that with the other PR since the game classes will be put in there too.

packages/flame/lib/src/components/position_component.dart
spydon: No, `Offset` can be skipped. We're trying to keep everything possible as `Vector2`.
A name suggestion could be to keep this as `positionOf` and introduce `positionOfAnchor` (also check out the `Anchor` method that does conversion to a position).

st-pasha: done

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,333 @@
...
-  bool containsPoint(Vector2 point) {
-    final rectangle = Rectangle.fromRect(toAbsoluteRect(), angle: angle)
-      ..position = absoluteCenter;
-    return rectangle.containsPoint(point);
+  //----------------------------------------------------------------------------
+  // Mutators
+  //----------------------------------------------------------------------------
+
+  /// Flip the component horizontally around its anchor point.
+  void flipHorizontally() => _transform.flipHorizontally();
```
spydon, line 259: I think these two could cause some confusion, because gestures will not be handled where the component is rendered with this right? Since it could be flipped outside of its bounds when the anchor is something else than `center`?
If that is the case, I think we should remove these two and let the `flip*AroundCenter` take their names.

spydon (commented):
One last comment!

packages/flame/lib/src/components/position_component.dart
st-pasha: Yes, so I was actually thinking about this, and I believe it would likely depend on a specific use case.

For example, one situation when image flipping may be used is when you have a character in a side-view game, and that character is normally going left-to-right, but then you want to go back (right-to-left) and you flip the sprite then. Now, we need to pick a character sprite whose anchor is off-center.  For example, this (the small [+] indicates the anchor):
mouse

Now your mouse is sitting on a floor, and you want it to go the other direction. How you would prefer to flip it:
mice
To me, the second option is much better, since the body of the mouse remains visually in place. 

Same goes for vertical flipping. Right now the mouse is floor-bound, so her anchor is on the floor. Now imagine she can climb the ceilings too: I would want to flip the image vertically for that. However, when she's climbing the ceiling, I would want her anchor to be one ceiling, and the body of the mouse below the anchor -- thus, the flip should occur around the anchor, not around the center line.

What is your thinking here, when would flipping around the center be preferable?

As for your concerns that this would interfere with tap detection -- it shouldn't, as long as gesture detectors use `PositionComponent`'s own `containsPoint()`, or when implementing their own, use the provided `toLocal()` method. The `Transform2D` object handles all these complexities in one centralized location, which was the main point of that class.

spydon: Aah, that's great! Didn't get that `toLocal` took care of that case, but it's obvious now when you've explained it.

spydon (approved):
Very good job! Such a nice clean-up.

packages/flame/CHANGELOG.md
```diff
@@ -31,6 +31,12 @@
  - Add class `NotifyingVector2`
  - Add class `Transform2D`
  - Remove `FPSCounter` from `BaseGame`
+ - Refactor `PositionComponent` to work with `Transform2D`. Several methods were updated to work
```
wolfenrain, line 34: I think it would be cleaner to split up the other sentence into multiple entries.

- renderFlipX replace by..
- angleTo removed...
- positioncomponent debug...

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,360 @@
...
+/// [position] on the screen, then the picture is rotated for [angle]
+/// radians around the anchor point, and finally scaled by [scale] also
+/// around the anchor point.
+///
+/// The [size] property of the [PositionComponent] is used primarily for
+/// tap and collision detection. Thus, the [size] should be set equal to
+/// the approximate bounding rectangle of the rendered picture. If you
+/// do not specify the size of a PositionComponent, then it will be
+/// equal to zero and the component won't be able to respond to taps.
+///
```
wolfenrain, line 61:
```suggestion
```

packages/flame/lib/src/components/position_component.dart
wolfenrain: We did indeed decide to not introduce two types of angles. And we are using only radians throughout the whole engine.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,360 @@
...
+  Vector2 get scaledSize => _scaledSize;
+
+  /// Measure the distance (in parent's coordinate space) between this
+  /// component's anchor and the [other] component's anchor.
+  double distance(PositionComponent other) =>
+      position.distanceTo(other.position);
+
+  //----------------------------------------------------------------------------
+  // Coordinate transformations
+  //----------------------------------------------------------------------------
```
wolfenrain, line 179: I am not sure if these kind of comments is something we want to add, and if we do why not use a code folding structure like the commonly used `//#region YourRegion` and `//#endregion` syntax

wolfenrain (commented):
Awesome stuff! Left a few comments

packages/flame/lib/src/components/position_component.dart
st-pasha: ok, I'll remove these methods then

packages/flame/CHANGELOG.md
st-pasha: sure, will do

packages/flame/lib/src/components/position_component.dart
st-pasha: Thanks, I didn't know about this syntax!

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,349 @@
...
+  /// modified by the user.
+  final Vector2 _scaledSize;
+  Vector2 get scaledSize => _scaledSize;
+
+  /// Measure the distance (in parent's coordinate space) between this
+  /// component's anchor and the [other] component's anchor.
+  double distance(PositionComponent other) =>
+      position.distanceTo(other.position);
+
+  //#region Coordinate transformations
```
erickzanardo, line 170: Just curious on why the `#` here?

erickzanardo (approved):
Awesome stuff, thanks for the contribution

packages/flame/lib/src/components/position_component.dart
st-pasha: That's what @wolfenrain [suggested](https://github.com/flame-engine/flame/pull/904#discussion_r690249010). For some reason, I can't find any mention of code regions on [effective Dart](https://dart.dev/guides/language/effective-dart/documentation), but the syntax does work: Android Studio recognizes these comments and offers an opportunity to fold each section.

erickzanardo: niice, didn't knew about that too! I wonder if Vim will recognise that 🤔

wolfenrain (approved):
LGTM

packages/flame/lib/src/components/position_component.dart
```diff
@@ -230,4 +361,4 @@ abstract class PositionComponent extends BaseComponent {
     size.setValues(rect.width, rect.height);
     topLeftPosition = rect.topLeft.toVector2();
   }
-}
+} // ignore: number-of-methods
```
luanpotter, line 364: I think we can just bump the codemetrics baseline instead, I don't think it's necessarily a good metric anyway and I prefer to avoid "ignoring" rules

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,349 @@
...
+  /// may visually appear larger or smaller due to application of [scale].
+  double get height => _size.y;
+  set height(double h) => _size.y = h;
+
+  /// The "physical" size of the component. This is the size of the
+  /// component as seen from the parent's perspective, and it is equal to
+  /// [size] * [scale]. This is a computed property and cannot be
+  /// modified by the user.
+  final Vector2 _scaledSize;
+  Vector2 get scaledSize => _scaledSize;
```
luanpotter, line 163: by exposing this someone could do `component.scaledSize.x = 10` and that would alter the private field `_scaledSize` reference right?

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,349 @@
...
+/// radians around the anchor point, and finally scaled by [scale] also
+/// around the anchor point.
+///
+/// The [size] property of the [PositionComponent] is used primarily for
+/// tap and collision detection. Thus, the [size] should be set equal to
+/// the approximate bounding rectangle of the rendered picture. If you
+/// do not specify the size of a PositionComponent, then it will be
+/// equal to zero and the component won't be able to respond to taps.
+class PositionComponent extends BaseComponent {
+  final Transform2D _transform;
```
luanpotter, line 62: I have a feeling this Transform2D class ends up being an implementation of the Projector interface I created. it could certainly be used for camera, viewport stuff as well!

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,349 @@
...
+  /// or the [anchor] change.
+  void _onModifiedSizeOrAnchor() {
+    _transform.offset = Vector2(-_anchor.x * _size.x, -_anchor.y * _size.y);
+  }
 
   @override
   void renderDebugMode(Canvas canvas) {
-    if (this is Hitbox) {
-      (this as Hitbox).renderHitboxes(canvas);
+    super.renderDebugMode(canvas);
```
luanpotter, line 315: just a comment for a potential followup, but we should find a way to extract all this debug related code into a separated class, to improve readability

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,349 @@
...
-      _preRenderMatrix.translate(size.x / 2, size.y / 2);
-      _preRenderMatrix.scale(
-        renderFlipX ? -1.0 : 1.0,
-        renderFlipY ? -1.0 : 1.0,
-      );
-      canvas.transform(_preRenderMatrix.storage);
-      canvas.translate(-size.x / 2, -size.y / 2);
-      _preRenderMatrix.setIdentity();
-    }
+    canvas.transform(transformMatrix.storage);
```
luanpotter, line 346: NICE!

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,349 @@
...
-      return p.absoluteTopLeftPosition + topLeftPosition;
-    } else {
-      return topLeftPosition;
+  /// Similar to [positionOf()], but applies to any anchor point within
+  /// the component.
+  Vector2 positionOfAnchor(Anchor anchor) {
+    if (anchor == _anchor) {
+      return position;
     }
+    return positionOf(Vector2(anchor.x * size.x, anchor.y * size.y));
```
luanpotter, line 196: we should think of a simpler way of doing `a * b`  (with `*` being a naive per coordinate product)

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,349 @@
...
+  Vector2 absolutePositionOfAnchor(Anchor anchor) {
+    var point = positionOfAnchor(anchor);
+    var ancestor = parent;
+    while (ancestor != null) {
+      if (ancestor is PositionComponent) {
+        point = ancestor.positionOf(point);
+      }
+      ancestor = ancestor.parent;
+    }
+    return point;
```
luanpotter, line 224: could extract an `absolutePositon` method here to be used twice?

packages/flame/test/components/position_component_test.dart
```diff
@@ -285,4 +295,301 @@ void main() {
...
+
+        final transform = Matrix4.identity()
+          ..translate(x, y)
+          ..rotateZ(angle)
+          ..translate(
+            -component.anchor.x * component.width,
+            -component.anchor.y * component.height,
+          );
+        for (var j = 0; j < 16; j++) {
+          expect(component.transformMatrix[j], closeTo(transform[j], 1e-10));
```
luanpotter, line 472: I think we have an `expectDouble` util that has a default tolerance

packages/flame/test/components/position_component_test.dart
```diff
@@ -285,4 +295,301 @@ void main() {
...
+        ..anchor = Anchor.center;
+
+      // Rotate the component in small increments counterclockwise
+      // and track the coordinate of its top-right corner
+      for (var i = 0; i < 30; i++) {
+        component.angle = -i / 10;
+        final cosA = math.cos(i / 10);
+        final sinA = math.sin(i / 10);
+        final expectedX = 50 + 5 * (0.8 * cosA - 0.6 * sinA);
+        final expectedY = 20 - 5 * (0.6 * cosA + 0.8 * sinA);
```
luanpotter, line 408: I am a bit unclear where the 0.8/0.6 come from, is that the width/height divided by 10? and then times 5 to do half? if so can we simplify? if not can we have a comment explaining the reason

packages/flame/lib/src/components/position_component.dart
```diff
@@ -1,218 +1,349 @@
...
+  /// as the point where Flame "grabs" the component. All transforms
+  /// occur around this point: the [position] is where the anchor point
+  /// will end up after the component is translated; the rotation and
+  /// scaling also happen around this anchor point.
+  Anchor _anchor;
+  Anchor get anchor => _anchor;
+  set anchor(Anchor anchor) {
+    _anchor = anchor;
+    _onModifiedSizeOrAnchor();
+  }
```
luanpotter, line 134: a good thing to explain on the docs is what happens when you change the anchor. either
1. the position of the anchor on world space is constant and the component moves to fit the new anchor local position
2. the position of the component on world space is constant and the anchor moves to fit the new local relative position

I don't see clearly which of those is the case by reading the comments on anchor, should probably specify for posterity (as both would make sense to me from different point of views)

packages/flame/lib/src/components/position_component.dart
luanpotter: probably need a clone() here

luanpotter (approved):
LGTM!

GREAT refactor 🎉 
I have some comments, among followups and some more practical docs or impl details.
But I love the new Transform2D based approach!

packages/flame/lib/src/components/position_component.dart
st-pasha: Good point. I'll add some clarifying note here.

packages/flame/lib/src/components/position_component.dart
st-pasha: You're right, I didn't think about that. Might as well remove the `_scaledSize` variable altogether, and let the getter compute the value on the fly.

packages/flame/lib/src/components/position_component.dart
st-pasha: I think the problem here is that `Vector2` already defines `*` operator as
```dart
Vector2 operator *(double scale) => clone()..scale(scale);
```
which is not what we want. Then there is the `multiply()` method, but it works in-place:
```dart
void multiply(Vector2 arg);
```

packages/flame/lib/src/components/position_component.dart
st-pasha: I've upped the `number-of-methods` parameter to 50

packages/flame/lib/src/components/position_component.dart
st-pasha: You're right, `Transform2D` should definitely implement `Projector`. However, if you don't mind, let's defer that to another PR, so that this one wouldn't have to grow even bigger.

packages/flame/lib/src/components/position_component.dart
st-pasha: thank you

packages/flame/test/components/position_component_test.dart
st-pasha: sure, i've added some explanation here

packages/flame/test/components/position_component_test.dart
st-pasha: I'm sorry, but I don't like that util very much :(

For once, the default tolerance is too high. These computations should easily hold up to 1e-12 precision if not lower. (The machine precision for doubles is 1e-16 near 1, but given compounding effects from multiple calculations and taking into account scale factors, 1e-10 is more than enough precision).

The second reason is that when the test fails, `closeTo` provides a much better description of what happened. For example, this is the error message I get from `expectDouble()`:
```
  Expected: <true>
    Actual: <false>
```
And here's the same error message with `closeTo()`:
```
  Expected: a numeric value within <1e-13> of <99.64973343376249>
    Actual: <99.6497334337626>
     Which:  differs by <1.1368683772161603e-13>
```

packages/flame/lib/src/components/position_component.dart
st-pasha: done

st-pasha:
So, is there anything else that needs fixing in this PR, or is it good to be merged?

erickzanardo:
> So, is there anything else that needs fixing in this PR, or is it good to be merged?

Not really, I am going to merge it, sorry for the delay to merge, busy days lately.

Issue #900, opened 2021-08-05
Components don't compose properly without HasGameRef


st-pasha:
Here I'm creating a simple component, and then adding another component to the first:
```
class AComponent extends BaseComponent {}

void main() {
  group('game ref tests', () {
    test('nested components without gameRef', () {
      final game = MyGame() .. onResize(Vector2.zero());
      final a = AComponent();
      game.add(a);
      game.update(0);
      final aa = AComponent();
      a.addChild(aa);
      game.update(0);
    });
  });
}
```
However, this produces the following error:
```
Testing started at 6:47 PM ...
/Users/pasha/flutter/bin/flutter --no-color test --machine --start-paused --plain-name "nested components without gameRef" test/components/has_game_ref_test.dart

dart:core                                                _AssertionError._throwNew
package:flame/src/components/base_component.dart 155:9   BaseComponent.prepare
package:flame/src/components/component_set.dart 75:12    ComponentSet.addChild
package:flame/src/components/base_component.dart 176:21  BaseComponent.addChild
test/components/has_game_ref_test.dart 42:9              main.<fn>.<fn>
===== asynchronous gap ===========================
dart:async                                               _completeOnAsyncError
package:flame/src/components/component_set.dart          ComponentSet.addChild
package:flame/src/components/base_component.dart 176:21  BaseComponent.addChild
test/components/has_game_ref_test.dart 42:9              main.<fn>.<fn>

'package:flame/src/components/base_component.dart': Failed assertion: line 155 pos 9: '!isMounted': Parent was already added to Game and has no HasGameRef; in this case, gameRef is mandatory.
```

I  believe this is a bug since mixins shouldn't be mandatory.

st-pasha:
I couldn't find any clear documentation on when in the component's lifecycle the `gameRef` is guaranteed to become available.

For example, if it is ok for the mixin to work starting from the `onMount()` event, then I can fix this bug easily. On the other hand, if you want `gameRef` to be accessible in `onLoad()`, then it will be more complicated.

spydon:
Yeah, `gameRef` has to be accessible from `onLoad` since it can be used to load sprites etc.

Right now you have to either add the children to the parent component before it is being added to the game, or your parent has to have the `HasGameRef` mixin. This is due to that the `gameRef` is used to prepare the component. This could potentially be solved by the restructuring of things so that `Game` is just a component, because then the parent could be responsible for preparing its children instead of the `Game`. If you have any other ideas we are all ears.

[...]

PR #906, opened 2021-08-06
Game as a Component

spydon:
# Description

 - Turns `Game` into a `Component`
 - Turns `ComponentEffect`s into `Component`s
 - Adds `Component.changeParent`
 - Removes `onMount` and replaces it with `onParentChange`
 - `onLoad` now only runs once for each component, so that you can initialize `late` variables in there

Things that this will make simpler:
- No code duplication between `BaseComponent` and `BaseGame`
- Multi-stage games
- `TransformEffect` that @st-pasha suggested
- Changing the parent of a component
- ...

In this example a square is move from one parent game to another and back again every 5 seconds:
https://user-images.githubusercontent.com/744771/132994425-ddeff3fd-9ff8-4093-be80-b1dcdc61b688.mp4

Component (and Game) Lifecycle:

component_lifecycle

Component inheritance tree:

diagram

This also restructures how to work with the more low-level game API:

game_lowlevel

## Checklist

## Breaking Change

## Related Issues

#900


packages/flame/lib/src/components/base_component.dart
```diff
@@ -20,14 +18,6 @@ import 'mixins/has_game_ref.dart';
 abstract class BaseComponent extends Component {
```
st-pasha, line 18: Now that the children were moved into `Component`, this description needs to be updated.

spydon: Yeah, I'm thinking of refactoring away `BaseComponent` completely since it is mostly the effects system left, and that can also be a component. Lots of docs and descriptions and re-ordering left to to be done. :)

packages/flame/lib/src/components/component.dart
```diff
@@ -93,10 +190,45 @@ abstract class Component {
   ///   myImage = await gameRef.load('my_image.png');
   /// }
   /// ```
-  Future<void>? onLoad() => null;
+  Future<void> onLoad() async {}
```
luanpotter, line 193: why not `Future<void>?`?

st-pasha:
In #899 I was envisioning that `BaseGame` would have member `_root` of type `BaseComponent`. One of the benefits was the ability to quickly swap out the root for another component. Would this still be possible with this approach?

spydon:
> In #899 I was envisioning that `BaseGame` would have member `_root` of type `BaseComponent`. One of the benefits was the ability to quickly swap out the root for another component. Would this still be possible with this approach?

Yes, it will. I'm going to add a method that is named `changeParent(Component c)` or something like that on `Component` which will remove the component from its current parent and then set the new parent and re-initialize the tree under the new parent..

packages/flame/lib/src/components/component.dart
spydon: The `GameWidget` was not happy about that, I didn't look any further into why that was.
Is it a lot more expensive to do a function call than to do a null check?

st-pasha: When the `prepare()` function calls `component.onLoad()`, it checks if the return value is `null` and in that case avoids `await`-ing it.  Thus, any component that doesn't override `onLoad` can be added slightly faster.

packages/flame/lib/src/components/mixins/has_game_ref.dart
```diff
@@ -1,7 +1,7 @@
 import '../../../components.dart';
 import '../../../game.dart';
 
-mixin HasGameRef<T extends BaseGame> on Component {
+mixin HasGameRef<T extends Game> on Component {
```
st-pasha, line 4: This can be simplified even more:
```dart
mixin HasGameRef<T> on Component {
  T? _gameRef;

  T get gameRef {
    if (_gameRef == null) {
      Component? c = parent;
      while (c != null) {
        if (c is HasGameRef<T> && c._gameRef != null) {
          _gameRef = c._gameRef;
          return _gameRef!;
        } else if (c is T) {
          _gameRef = c;
          return c;
        } else {
          c = c.parent;
        }
      }
      throw StateError("Cannot find reference $T in the component tree");
    }
    return _gameRef!;
  }

  @override
  void onRemove() {
    super.onRemove();
    _gameRef = null;
  }
}
```
Then neither `Game` nor `Component` will have to deal with the `HasGameRef`, it works on its own.
I suspect something like this can be done for `Tappable` and `Draggable` as well.

packages/flame/lib/src/game/game.dart
```diff
@@ -20,7 +21,7 @@ import 'projector.dart';
 ///
 /// Subclass this to implement the [update] and [render] methods.
 /// Flame will deal with calling these methods properly when the game's widget is rendered.
-abstract class Game extends Projector {
+abstract class Game extends Component implements Projector {
```
st-pasha, line 24: Since `Game` is now a component, it should probably declare itself to be `_isMounted`

packages/flame/lib/src/components/mixins/has_game_ref.dart
spydon: Good idea!

packages/flame/lib/src/components/component.dart
spydon: Reverted it.

packages/flame/lib/src/components/mixins/has_game_ref.dart
```diff
@@ -1,27 +1,31 @@
...
 
   T get gameRef {
-    final ref = _gameRef;
-    if (ref == null) {
-      throw 'Accessing gameRef before the component was added to the game!';
+    if (_gameRef == null) {
+      var c = parent;
+      while (c != null) {
+        if (c is HasGameRef<T> && c._gameRef != null) {
+          _gameRef = c._gameRef;
```
st-pasha, line 12: Even easier:
```suggestion
        if (c is HasGameRef<T>) {
          _gameRef = c.gameRef;
```
That way `_gameRef` will be cached in the closest ancestor of the current component.

packages/flame/lib/src/components/mixins/has_game_ref.dart
st-pasha: Glad to see that it actually worked! :)

packages/flame/lib/src/components/component_set.dart
```diff
@@ -90,10 +98,17 @@ class ComponentSet extends QueryableOrderedSet<Component> {
     Iterable<Component> components, {
     BaseGame? gameRef,
```
st-pasha, line 99: `gameRef` no longer needed here

spydon: Ah, haven't pushed the latest commits.
Its still quite some work left, especially with the effects.

spydon: This one I had apparently forgotten about, good catch!

packages/flame/lib/src/effects/effects.dart
```diff
@@ -28,34 +35,91 @@ abstract class ComponentEffect<T extends BaseComponent> {
...
+
+  /// The percentage of the effect that has passed including [preOffset] and
+  /// [postOffset].
+  double percentage = 0.0;
+
+  /// The outcome the curve function, only updates after [preOffset] and before
+  /// [postOffset]
   double curveProgress = 0.0;
+
+  /// How long time it takes for the effect to peak, which means right before it
```
rundun, line 50: "How much time it takes..." ?

packages/flame/lib/src/effects/effects.dart
```diff
@@ -28,34 +35,91 @@ abstract class ComponentEffect<T extends BaseComponent> {
...
+  /// starts any potential reversal or reset. Including offsets.
```
rundun, line 51: Including both offsets

packages/flame/lib/src/effects/effects.dart
```diff
@@ -28,34 +35,91 @@ abstract class ComponentEffect<T extends BaseComponent> {
...
+  /// after it has reached [peakTime] if [isAlternating] is true, and reset to
+  /// zero if it is not.
   double currentTime = 0.0;
+
+  /// When an effect reaches the end, and the beginning if it is alternating,
+  /// it will overshoot 0.0 and [peakTime], this time is added to the next time
+  /// step.
   double driftTime = 0.0;
+
+  /// Whether the effect is going forward or is reversing.
```
rundun, line 64: Can we have an example of forward and backwards in the comment?

spydon: Is it clearer now?

rundun: Ahh right its clearer now thanks

packages/flame/lib/src/components/component_set.dart
```diff
@@ -71,29 +72,45 @@ class ComponentSet extends QueryableOrderedSet<Component> {
...
   /// This method can be considered sync for all intents and purposes if no
   /// onLoad is provided by the component.
-  Future<void> addChild(Component c, {BaseGame? gameRef}) async {
-    prepare(c, gameRef: gameRef);
-
-    final loadFuture = c.onLoad();
-    if (loadFuture != null) {
-      await loadFuture;
+  Future<void> addChild(Component component) async {
+    print('Trying to addChild $component to $this');
```
luanpotter, line 76: print

examples/lib/stories/input/draggables.dart
```diff
@@ -58,13 +62,15 @@ class DraggableSquare extends PositionComponent
...
+  late final DraggableSquare square;
 
   DraggablesGame({required this.zoom});
 
   @override
   Future<void> onLoad() async {
     camera.zoom = zoom;
-    add(DraggableSquare());
+    square = DraggableSquare();
+    add(square);
```
luanpotter, line 73: nit: `add(square = DraggableSquare());` :P (just bc we do that most of the times)

packages/flame/lib/src/effects/effects.dart
```diff
@@ -28,34 +35,95 @@ abstract class ComponentEffect<T extends BaseComponent> {
...
+  /// has [isAlternating] set to true, it will do the effect backwards back to
+  /// its original state.
   int curveDirection = 1;
+
+  /// Which curve that the effect is following.
   Curve curve;
 
+  /// The time (in seconds) that should pass before the component starts each
+  /// iteration.
+  double preOffset;
```
luanpotter, line 76: this is a totally unrelated addition right?

examples/lib/stories/components/game_in_game.dart
```diff
@@ -0,0 +1,40 @@
...
+const gameInGameInfo = '''
+This example shows two games having another game as a parent.
+One game contains draggable components and the other is a rotating square with
+other square children.
+After 5 seconds, one of the components from the game with draggable squares
+changes its parent from its original game to the component that is rotating.
+After another 5 seconds it changes back to its original parent, and so on.
+''';
+
+class GameInGame extends BaseGame with HasDraggableComponents {
```
luanpotter, line 16: wow!

examples/lib/stories/components/game_in_game.dart
```diff
@@ -0,0 +1,40 @@
...
+    await add(composedGame);
+    await add(draggablesGame);
+    Component? swapChild;
+    final t = Timer(5, callback: () {
+      swapChild ??= draggablesGame.square;
+      swapChild!.changeParent(
+        swapChild!.parent == draggablesGame
+            ? composedGame.parentSquare
+            : draggablesGame,
+      );
```
luanpotter, line 35: i think if you save to variable you can avoid the `!`:

```suggestion
      final child = swapChild ??= draggablesGame.square;
      child.changeParent(
        child.parent == draggablesGame
            ? composedGame.parentSquare
            : draggablesGame,
      );
```

examples/lib/stories/effects/scale_effect.dart
```diff
@@ -23,16 +23,17 @@ class ScaleEffectGame extends BaseGame with TapDetector {
       ..size = Vector2.all(20)
       ..anchor = Anchor.center;
 
-    square.addChild(childSquare);
     add(square);
+    // TODO(spydon): fix so that children added before add also works
```
luanpotter, line 27: this seems like an important thing to be included!

packages/flame/lib/src/components/base_component.dart
```diff
@@ -4,30 +4,16 @@ import 'package:meta/meta.dart';
...
-  /// If the component has a parent it will be set here
-  BaseComponent? _parent;
-
-  @override
-  BaseComponent? get parent => _parent;
-
+/// The difference between this and [Component] is that the [BaseComponent]
+/// implements the [update] and [render] methods, meanwhile you have to do that
+/// yourself in [Component].
+class BaseComponent extends Component {
```
luanpotter, line 16: if we are moving the children up and this is just a shin for propagating update and render, maybe we can just get rid of it?

packages/flame/lib/src/components/component.dart
```diff
@@ -56,30 +74,135 @@ abstract class Component {
...
+      } else if (shouldContinue && child is BaseGame) {
+        shouldContinue = child.propagateToChildren<T>(handler);
+      }
+      if (!shouldContinue) {
+        break;
+      }
+    }
+    return shouldContinue;
+  }
+
```
luanpotter, line 198: /// Finds the closest parent up the hierarchy that satisfies type=T, or null if none is found

packages/flame/lib/src/components/base_component.dart
```diff
@@ -4,30 +4,16 @@ import 'package:meta/meta.dart';
...
-  late final ComponentSet children = createComponentSet();
-
-  /// If the component has a parent it will be set here
-  BaseComponent? _parent;
-
-  @override
-  BaseComponent? get parent => _parent;
-
+/// The difference between this and [Component] is that the [BaseComponent]
+/// implements the [update] and [render] methods, meanwhile you have to do that
```
luanpotter, line 14: implements the [...] methods => adds a default implement for the [...] methods that passes along the calls to its children

packages/flame/lib/src/components/component_set.dart
```diff
@@ -30,7 +30,8 @@ class ComponentSet extends QueryableOrderedSet<Component> {
 
   /// This is the "prepare" function that will be called *before* the
   /// component is added to the component list by the add/addAll methods.
-  final void Function(Component child, {BaseGame? gameRef}) prepare;
+  /// It is also called when the component changes parent.
+  final void Function(Component child) prepare;
```
luanpotter, line 34: nice! never liked the need of that ref

packages/flame/lib/src/components/component_set.dart
st-pasha: There's a linter rule https://dart-lang.github.io/linter/lints/avoid_print.html to help catch these

st-pasha:
This is great work, @spydon -- can't wait to see it merged!

packages/flame/lib/src/components/component_set.dart
spydon: Since this is a draft I'm having these to debug things, I'll remove them once I mark it as "Ready for review". :)

packages/flame/lib/src/effects/effects.dart
spydon: I thought that it would be needed to make the `CombinedEffect` work like it used to work, we'll see if it is actually possible to make the `CombinedEffect` alternate with the current structure.

packages/flame/lib/src/components/base_component.dart
spydon: I thought about that too, the only extra thing that it has now is the debug stuff and that might be worth having on the `Component` anyways? What do you think @flame-engine/flame-colab?

luanpotter: that should be extracted to a separate file imo, and then it can be "hooked up" with a couple lines at any level we desire. I think it's worth it

packages/flame/lib/src/game/base_game.dart
```diff
@@ -20,13 +19,10 @@ import 'viewport.dart';
 /// This is a more complete and opinionated implementation of Game.
 ///
 /// BaseGame should be extended to add your game logic.
-/// [update], [render] and [onResize] methods have default implementations.
+/// [update], [render] and [onGameResize] methods have default implementations.
 /// This is the recommended structure to use for most games.
 /// It is based on the Component system.
 class BaseGame extends Game {
```
st-pasha, line 25: Would it make sense to call this `FCSGame`, or `ComponentBasedGame`, or something like that, in order to clearly distinguish it from `OxygenGame` and `Forge2DGame`?

packages/flame/lib/src/components/component_set.dart
st-pasha: There is https://dart-lang.github.io/linter/lints/avoid_print.html linter rule to catch these kinds of things

packages/flame/lib/src/game/base_game.dart
spydon: I like `BaseGame` as a name, since it is the base game in Flame and since `OxygenGame` and `Forge2DGame` lives outside of Flame. `Forge2DGame` even extends `BaseGame`. 
I think adding `FCS` there instead will confuse users more than helping them.  
(I don't like acronyms in names either)

spydon: Could be that I'm just very used to it though, we could have a poll for this in discord among the users if any more people of the team think that it could be a good idea to change the name. :)

st-pasha: Yeah, i don't like 'FCS' either. How about `FlameGame`? (I don't know how to do polls on discord)

spydon: What is wrong with `BaseGame` though? I'm open for a discussion to change it, but I don't see any problems with the current name.

st-pasha: Well, I'm just trying to think which name would best capture the intent and purpose of that class.
For example, the `Game` class can be described as "The base game class", because it's at the root of the hierarchy of various game classes. The `BaseGame` class is currently described as a "more opinionated implementation of a Game", but that's not very precise. We could say that `BaseGame` is a "component-based implementation of a Game", so it's a base class for all Flame-engine games, but not **the** base game class. So, a `FlameGame` could be appropriate, but only if my understanding of the scope of the name "Flame" is correct.

packages/flame/CHANGELOG.md
```diff
@@ -46,6 +46,11 @@
  - Added `FocusNode` on the game widget and improved keyboard handling in the game.
  - Added ability to have custom mouse cursor on the `GameWidget` region
  - Change sprite component to default to the Sprite size if not provided
+ - `Game` is now a `Component`
+ - `ComponentEffect` is now a `Component`
+ - `HasGameRef` can now operate independently from `Game`
+ - `preOffset` and `postOffset` for effects to handle time before and after an effect
+ - Removes `component.onMount` since `onLoad` can be used for all its purposes
```
st-pasha, line 53: The current implementation allows putting expensive initialization code into the `onLoad` method, which serves as a late-stage constructor, and is guaranteed to run only once. The `onMount`, on the other hand, runs every time the component is mounted into the game tree. For example, if the component is removed from the game, and then attached at a new place in the tree, we'd run `onRemove` and then `onMount` again. Thus, `onMount` can be used to perform initialization that requires awareness of the surrounding game tree.

So, how are these tasks going to be separated now?

examples/lib/stories/components/composability.dart
```diff
@@ -29,32 +29,26 @@ class ParentSquare extends Square with HasGameRef {
...
     ];
 
-    children.forEach((c) => addChild(c, gameRef: gameRef));
+    addAll(children);
   }
 }
 
-class Composability extends BaseGame with TapDetector {
-  late ParentSquare _parent;
+class Composability extends BaseGame with HasDraggableComponents {
```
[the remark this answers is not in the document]
spydon: Still fiddling around with how I want to have the `GameInGame` example, since that moves a `Draggable` component from one game to this one. But I will probably remove the mixin in the end anyways since it will be confusing with the spinning.

packages/flame/CHANGELOG.md
spydon: `onLoad` would also run again if it changed parent, since that is part of `ComponentSet.addChild`.
Maybe it shouldn't though.

packages/flame/lib/src/effects/effects.dart
```diff
@@ -11,51 +11,134 @@ export './rotate_effect.dart';
...
+  /// Which curve that the effect is following.
   Curve curve;
 
+  /// The time (in seconds) that should pass before the effect starts each
+  /// iteration.
+  double preOffset;
+
+  /// The time (in seconds) that should pass before the effect ends each
+  /// peak.
+  double postOffset;
```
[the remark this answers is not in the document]
spydon: Good idea, fixed!

packages/flame/CHANGELOG.md
st-pasha: This could potentially be problematic if `onLoad` initializes `late final` variables (which it usually does).

spydon: Good point, maybe there should be a new lifecycle method called `onParentChange` or something like that which runs after `onLoad` (`onMount` ran before).

st-pasha: I just tested this with a couple of print statements, and in the current `main` branch the sequence of events is `onGameResize -> onLoad -> onMount -> (update -> render)*`.

spydon: Alright, then it was just wrong in the docs. What do you think of the new name?

st-pasha: Frankly, when comparing `onParentChange` vs `onMount`, I like the latter slightly more. Maybe because I'm already used to it, but also because it pairs better with `onRemove`.

So, am I correct in understanding that `onLoad` will now run only when the component is already mounted? I mean, this is obviously nice to have if you need to access `gameRef` within the `onLoad` handler, but adds a bit more confusion with respect to the fact that now as the component is mounted we run `onLoad` first, before the `onMount`. I guess it should be explained really clearly in the docs then.

spydon: I think it sounds better with `onMount`, but I think it is really vague what it actually means. Especially when it is changing parents, is it mounted again on that tree? `onMount` sounds more like a synonym to `onAttach` to me.

I have updated the section in the docs about the lifecycle of the `Component` and the `Game` a bit.

This is how the lifecycle looks now:
https://github.com/flame-engine/flame/blob/ad385c16713ffc44e2d64510bac7a13ea5be74db/doc/images/component_lifecycle.png

st-pasha: But `onAttach` is for the `Game` class only, not for a generic component, right? It could have a more verbose name in that case, something like `onFlutterTreeAttach`.

As for the vagueness of `onMount`, I guess it would help if we started by clarifying what the `isMounted` property actually means. We could start with the notion of the component tree, and how it can be traversed up and down. That component is called "mounted" when it is placed inside that tree. That there are mechanisms in place to ensure that all components in the tree will get updated/rendered on every game tick, and that they will properly receive input events such as taps or mouse clicks.

Then we would say that a component can be removed (detached) from the component tree, in which case it will be considered "unmounted", and then added back into the tree, maybe at another place, in which case call the `onMount` handler again.

As for `onParentChange` -- it immediately raises several questions: (1) is `onRemove` called at all when you change a component's parent? (2) if you remove the component from a tree and then later add it back for the same parent -- does this qualify as a "parent change"?

spydon: True, that could be a good name for `onAttach`/`onDetach`.

Doesn't `onMount` raise the same questions as the ones that you spell out in the last paragraph?

st-pasha: The second question is avoided because "onMount" doesn't imply a change of parent -- only that the component is being put into the component tree. So I know that every time a component is being placed into a tree, the game calls `onMount`. The first question becomes trivial if we state that for every `onMount` there will be a call to `onRemove` later on. So that even though a parent change can be as simple as `this.parent = newParent`, we would still do it in two steps.

spydon: I don't know, I still feel that `onMount` needs more explanation than `onParentChange` by just reading their names (calling `changeParent` with the same parent really is an edge case, which will be explained in the docs).

st-pasha: I would say, a thorough explanation of when the method is called is needed in any case. Then it is the matter of having a name that best matches that explanation.

There is a flag `isMounted` to describe whether the component is currently in the game tree or not. This flag is more harmonious with the `onMount` name. Or are you planning to change the flag's name as well?

To me, `onParentChange` sounds like it is called whenever the component's `parent` property changes. Which leaves one wondering whether it'll be called (a) only when the `parent` property changes, or also when it is initialized? (b) when `parent` property becomes `null` because the component is removed from the tree.

spydon: I talked to the team about what they thought and they got quite confused about `onParentChange` so I will rename it to `onMount` and re-add `isMounted` (since that was removed, but I don't remember what my reasoning for removing it was).
I'll try to make it clear how it works in the docs too. :)

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -135,10 +138,15 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
...
-
-  Future<void> get _gameLoaderFutureCache =>
-      _gameLoaderFuture ?? (_gameLoaderFuture = widget.game.onLoad());
+  Future<void> get _loaderFuture {
+    final onLoad = widget.game.onLoadCache;
+    final onMount = widget.game.onMount;
+    if (onLoad != null) {
+      return onLoad.then((_) => onMount());
+    } else {
+      return Future<void>.value().then((_) => onMount());
```
spydon, line 147: There must be a better way than this to encapsulate a void function call in a Future?

st-pasha: I guess you could avoid `if` using the `??` operator:
```
(onLoad ?? Future<void>.value()).then((_) => onMount());
```

spydon: Then `onMount` wouldn't be called when `onLoad` is defined. Or maybe I misunderstood your suggestion?

I want to chain them when onLoad is defined and only call `onMount` if `onLoad` is not defined.

What my comment was about was removing the empty future with the `then` and encapsulating `onMount` differently in a future, if it is possible.

spydon: Ooh I missed the parenthesis on the phone, that is a good suggestion! I still want to remove the weird empty Future though.

doc/game.md
```diff
@@ -62,7 +62,20 @@ want to remove a list of components.
...
+## Lifecycle
+
+![Game Lifecycle Diagram](images/game_lifecycle.png)
+
+When a game first is added to a Flutter widget tree the following lifecycle methods will be called
+in order: `onGameResize`, `onLoad`, `onMount` and `onAttach`. After that it goes on to call
+`update` and `render` back and forth every tick, until the widget is removed from the tree.
+Once the `GameWidget` is removed from the tree, `onRemove` and `onDetach` is called.
+
+If the game is added to another `Component` instead of a `GameWidget`, `onAttach` and `onDetach`
```
erickzanardo, line 74: Hmm, after reading this, I guess it makes sense to differentiate `onAttach` and `onDetach` from the other methods.

Though I still don't like much `onFlutterAttach` and `onFlutterDetach`... Maybe `onGameAttach`, or `onRenderBoxAttach`?

packages/flame_oxygen/lib/src/oxygen_game.dart
```diff
@@ -66,5 +66,6 @@ abstract class OxygenGame extends Game {
 
   @override
   @mustCallSuper
+  //ignore: must_call_super
```
erickzanardo, line 69: Is that right? 🤔

packages/flame/pubspec.yaml
```diff
@@ -7,13 +7,13 @@ dependencies:
...
   meta: ^1.3.0
-  ordered_set: ^3.1.0
+  ordered_set: ^3.2.0
   vector_math: '>=2.1.0 <3.0.0'
 
 dev_dependencies:
   flutter_test:
     sdk: flutter
-  test: ^1.16.0
+  test: ^1.17.10
```
erickzanardo, line 16: I wonder if we need to keep depending on this, we already get test methods on the `flutter_test` dependency.

Either way this is unrelated to your PR, so no need to remove it on this PR.

packages/flame/lib/src/timer.dart
```diff
@@ -63,6 +63,10 @@ class Timer {
   void resume() {
     _running = true;
   }
+
+  TimerComponent asComponent() {
```
erickzanardo, line 67: I am unsure about this method, as this kind of thing don't exist in any other component.

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -217,7 +218,8 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
     });
   }
 
-  // widget overlay stuff
```
erickzanardo, line 220: Thank you

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -190,9 +194,6 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
       // Reset mouse cursor
       _initMouseCursor();
       addMouseCursorListener();
-
-      // Reset the loader future
-      _gameLoaderFuture = null;
```
erickzanardo, line 195: I think we need to keep this right? If the game on this widget changed, onLoad needs to run again

packages/flame/lib/src/components/component.dart
```diff
@@ -95,8 +291,67 @@ abstract class Component {
   /// ```
   Future<void>? onLoad() => null;
 
+  Future<void>? _onLoadCache;
+
+  /// Since [onLoad] only should run once throughout a components lifetime it is
+  /// cached so that it can be reused when the parent component/game/widget
+  /// changes.
+  Future<void>? get onLoadCache => _onLoadCache ?? (_onLoadCache = onLoad());
```
erickzanardo, line 299: Can't we keep this on the game as it was before? Seems a little bit weird to have a `onLoadCache` getter and if the user overrides this, bad things can happen.

packages/flame/lib/src/device.dart
```diff
@@ -12,7 +13,7 @@ class Device {
       // TODO(erickzanardo): We probably could use dart:html and implement this for web as well
       return Future.value();
     }
-    return SystemChrome.setEnabledSystemUIOverlays([]);
+    return SystemChrome.setEnabledSystemUIMode(SystemUiMode.immersiveSticky);
```
erickzanardo, line 16: `immersiveSticky` 🤔  hahahahaha

erickzanardo: Btw, isn't this supported on web now? Maybe we could check and remove the if above (can be a follow up too)

packages/flame/lib/src/effects/combined_effect.dart
```diff
@@ -1,109 +1,70 @@
...
+import '../../components.dart';
 import 'effects.dart';
 
-class CombinedEffect extends PositionComponentEffect {
-  final List<PositionComponentEffect> effects;
-  final double offset;
+/// The [CombinedEffect] is just a container for multiple effects, all settings
+/// are handled on each individual effect. [CombinedEffect] is not an effect
+/// itself.
+class CombinedEffect extends ComponentEffect with EffectsHelper {
```
erickzanardo, line 10: What is this `EffectsHelper`? I am not finding it on the PR and seems that it is a new thing right?

packages/flame/lib/src/effects/opacity_effect.dart
```diff
@@ -55,36 +61,43 @@ class OpacityEffect extends ComponentEffect<HasPaint> {
...
   }
 
   @override
   void update(double dt) {
+    if (isPaused) {
+      return;
+    }
     super.update(dt);
-    component?.setOpacity(
+    if (hasCompleted()) {
```
erickzanardo, line 97: Comparing the changes on this Effect and the changes on the Color effect, seems that we are repeating quite an amount of code on the effects, we should consider thinking on a way to share more logic on our effects library (Also non blocker for this PR)

packages/flame/lib/src/effects/effects.dart
```diff
@@ -243,40 +381,19 @@ abstract class PositionComponentEffect
...
-    _setComponentState(endPosition, endAngle, endSize, endScale);
+  void setComponentToPeakState() {
+    _setComponentState(peakPosition, peakAngle, peakSize, peakScale);
   }
 }
 
-abstract class SimplePositionComponentEffect extends PositionComponentEffect {
-  double? duration;
-  double? speed;
+mixin EffectsHelper on Component {
```
erickzanardo, line 389: Ah, found it 😅 , this file had too many changes, GitHub collapsed it.

What if we call this `EffectsHandlers` instead of `EffectsHelper`?

examples/lib/stories/components/components.dart
```diff
@@ -28,5 +24,11 @@ void addComponentsStories(Dashbook dashbook) {
       'Debug',
       (_) => GameWidget(game: DebugGame()),
       codeLink: baseLink('components/debug.dart'),
+    )
+    ..add(
+      'Game-in-game',
```
erickzanardo, line 29: Oh, you should record this example running and showcase it on the PR description.

examples/lib/stories/input/draggables.dart
```diff
@@ -34,6 +34,9 @@ class DraggableSquare extends PositionComponent
 
   @override
   bool onDragUpdate(int pointerId, DragUpdateInfo info) {
+    if (parent is! DraggablesGame) {
```
erickzanardo, line 37: Why is this necessary now?

examples/lib/stories/parallax/no_fcs.dart
```diff
@@ -29,6 +29,7 @@ class NoFCSParallaxGame extends Game {
   }
 
   @override
+  // ignore: must_call_super
```
erickzanardo, line 32: Why not just add super.update?

doc/game.md
erickzanardo: SGTM too

packages/flame/lib/src/game/game_widget/game_widget.dart
st-pasha: How about `Future(onMount)`?

packages/flame/test/components/composed_component_test.dart
```diff
@@ -149,67 +166,46 @@ void main() {
...
 
-      wrapper.addChild(child);
-      game.add(wrapper);
+      wrapper.add(child);
+      await game.add(wrapper);
       game.update(0.0);
 
       expect(child.debugMode, true);
     });
-    test('fail to add child if no gameRef can be acquired', () {
```
luanpotter, line 195: NICE!

packages/flame/test/components/component_lifecycle_test.dart
```diff
@@ -0,0 +1,114 @@
...
+          'prepared: false',
+          'onGameResize',
+          'prepared: true',
+          'onLoad',
+          'onMount',
+        ],
+      );
+    });
+
+    test('Correct lifecycle on parent change', () async {
```
luanpotter, line 75: we should define a standard for casing on test names (for followup)

packages/flame/pubspec.yaml
luanpotter: good q, I imagine it might be bc of this import `package:test/fake.dart`? def worth testing out though

packages/flame/lib/src/test_helpers/expect_vector2.dart
```diff
@@ -2,10 +2,11 @@ import 'package:test/test.dart';
 
 import '../../extensions.dart';
 
+// Default is that one pixel away from the expected value is accepted
```
luanpotter, line 5: this is a big change, is it to solve some existing flakyness or newly introduced one?

packages/flame/lib/src/game/base_game.dart
```diff
@@ -273,27 +234,33 @@ class BaseGame extends Game {
...
       final wasUpdated = changePriority(
         component,
         priority,
         reorderRoot: false,
       );
       if (wasUpdated) {
-        if (component.parent != null && component.parent is BaseComponent) {
-          parents.add(component.parent! as BaseComponent);
+        if (component.parent != null) {
+          parents.add(component.parent!);
```
luanpotter, line 246: nit: save `component.parent` to a variable to avoid the bang. here and L223

packages/flame/lib/src/game/base_game.dart
luanpotter: originally `more opinionated implementation` meant not only adding components, but a lot of other "default" things. I am not sure if that is the case anymore, but that was the original idea and hence the name

packages/flame/lib/src/effects/opacity_effect.dart
luanpotter: should we add an issue just to keep track of the followups? I fear some good comments might get lost when this is merged

packages/flame/lib/src/effects/effects.dart
luanpotter: or HasEffects ?

packages/flame/lib/src/components/sprite_animation_component.dart
```diff
@@ -10,8 +10,15 @@ import 'position_component.dart';
 export '../sprite_animation.dart';
 
 class SpriteAnimationComponent extends PositionComponent with HasPaint {
+  /// The animation used by the component.
   SpriteAnimation? animation;
+
+  /// If the component should be removed once the animation has finished.
+  /// (Needs the animation to have `loop = false`, since it will never finish
+  /// otherwise.
```
luanpotter, line 18:
```suggestion
  /// otherwise).
```

packages/flame/lib/src/components/particle_component.dart
```diff
@@ -5,7 +5,7 @@ import 'component.dart';
 
 /// Base container for [Particle] instances to be attach
 /// to a [Component] tree. Could be added either to BaseGame
-/// or an implementation of BaseComponent.
+/// or an implementation of Component.
```
luanpotter, line 8:
```suggestion
/// or an implementation of [Component].
```

packages/flame/lib/src/components/mixins/hoverable.dart
```diff
@@ -35,12 +35,13 @@ mixin HasHoverableComponents on BaseGame {
       return true; // always continue
     }
 
-    for (final c in components.reversed()) {
-      if (c is BaseComponent) {
-        c.propagateToChildren<Hoverable>(_mouseMoveHandler);
+    for (final c in children.reversed()) {
```
luanpotter, line 38: this is another piece of code that seems to be repeated a lot. I think there must be a better abstraction than `propagateToChildren` that requires each place to have less code...

packages/flame_oxygen/lib/src/oxygen_game.dart
spydon: It is since Oxygen isn't using the children so it would be an unnecessary check through an empty component tree

packages/flame/lib/src/timer.dart
spydon: I'll remove it and we can discuss whether we want it or not and possibly put up a future PR.
I found the method a bit handy, had it on the particles too.

packages/flame/lib/src/game/game_widget/game_widget.dart
spydon: It will, the cache is just in the game now instead

packages/flame/lib/src/components/component.dart
spydon: If we have it like it was before `onLoad` might be called several times, and then you can't to late final initialization in `onLoad`.

packages/flame_oxygen/lib/src/oxygen_game.dart
erickzanardo: Hm, I think that is there cases where it makes sense not to call super, we should remove the annotation and explain on the docs, otherwise I believe we should respect it everytime

packages/flame/lib/src/effects/combined_effect.dart
spydon:
```dart
mixin EffectsHelper on Component {
  void clearEffects() {
    children.removeAll(effects);
  }

  List<ComponentEffect> get effects {
    return children.isRegistered<ComponentEffect>()
        ? children.query<ComponentEffect>()
        : [];
  }
}
```
For now it is just those two helper methods (it lives in `effects.dart`).

packages/flame/lib/src/effects/opacity_effect.dart
spydon: Opened #945

packages/flame/lib/src/effects/effects.dart
spydon: `EffectsHandlers` sounds like it would do something, but it is just two helper methods, you don't need this mixin to use effects if you don't want to. `HasEffects` sounds like it would be similar to `HasDraggables/Tappables` etc, but it is not.

erickzanardo: On a second thought, since it is only two methods couldn't we just add them to components directly?

packages/flame/lib/src/game/game_widget/game_widget.dart
spydon: Lol, I thought I tried that! Thanks

packages/flame/lib/src/effects/effects.dart
spydon: Hmm, I thought we didn't want a lot of extra methods on there?
But it would probably make it easier for the user in the end, so I'd be for it.

examples/lib/stories/components/components.dart
spydon: Done!

examples/lib/stories/parallax/no_fcs.dart
spydon: Done!

packages/flame/test/components/component_lifecycle_test.dart
spydon: Opened #946

packages/flame/lib/src/test_helpers/expect_vector2.dart
spydon: It is for the existing one when the `SequenceEffect` is `isInfinite` and `isAlternating`.

st-pasha: If this is for a small number of tests only, then wouldn't it be better to pass `epsilon=1` in those specific tests? That way all other tests can remain more strict.

spydon: True, I'll do that

packages/flame/lib/src/game/base_game.dart
spydon: Opened discussion #948, since this can be done as a follow-up after this PR is merged.

packages/flame/lib/src/components/component.dart
```diff
@@ -3,88 +3,284 @@ import 'dart:ui';
...
+
+  /// Returns a [TextPaint] object with the [debugColor] set as color for the
+  /// text.
+  TextPaint get debugTextPaint => TextPaint(
+        config: TextPaintConfig(
+          color: debugColor,
+          fontSize: 12,
+        ),
+      );
+
```
erickzanardo, line 99: Both these should be memoized to avoid creating instances on every access

packages/flame/lib/src/components/component.dart
erickzanardo: We should try to make this an internal thing somehow though. I feel that this method could confuse the user when overriding the `onLoad` method for their own components.

spydon: If only dart had `protected`... Any ideas for how to make it internal? Since it needs to be reached by the `GameWidget` and from `ComponentSet`.

erickzanardo:
Yeah, protected is truly missed. Can't think of anything right now, but
maybe if we just rename this, the confusion would be avoided? What if we
call this `loadedCache` or something? I believe that the `on` prefix is
what will cause the confusion.

Em seg., 13 de set. de 2021 às 08:33, Lukas Klingsbo <
***@***.***> escreveu:

packages/flame/lib/src/components/component.dart
spydon: Forgot that `@protected` exists, so I used that now

packages/flame/lib/src/components/component.dart
spydon: Good idea, fixed with `ValueCache`

packages/flame_oxygen/lib/src/oxygen_game.dart
spydon: Done

examples/lib/stories/input/draggables.dart
spydon: Because I use this game in the game-in-game example, and it becomes very weird when the square is moved to the other game if it can still be dragged around there since the parent is spinning.

packages/flame/lib/src/components/mixins/hoverable.dart
spydon: Opened #949 for that, since it isn't really related to this PR.

examples/lib/stories/input/draggables.dart
erickzanardo: Oh, this is the example code! Alright

packages/flame/.min_coverage
```diff
@@ -1 +1 @@
-57.0
+59.7
```
erickzanardo, line 1: ❤️

packages/flame/lib/src/components/component.dart
```diff
@@ -3,88 +3,300 @@ import 'dart:ui';
...
+    isPrepared = false;
+    isMounted = false;
+    _parent = null;
+    nextParent?.add(this);
+    nextParent = null;
+  }
+
+  /// Prepares and registers a component to be added on the next game tick
+  ///
+  /// This methods is an async operation since it await the [onLoad] method of
```
erickzanardo, line 200:
```suggestion
  /// This method is an async operation since it await the [onLoad] method of
```

packages/flame/lib/src/components/component.dart
```diff
@@ -3,88 +3,300 @@ import 'dart:ui';
...
+  /// at runtime.
+  ///
+  /// This will call [ComponentSet.rebalanceAll] on the [children] ordered set.
+  void reorderChildren() => children.rebalanceAll();
+
+  /// This method first calls the passed handler on the leaves in the tree,
+  /// the children without any children of their own.
+  /// Then it continues through all other children. The propagation continues
+  /// until the handler returns false, which means "do not continue", or when
+  /// the handler has been called with all children
```
erickzanardo, line 257:
```suggestion
  /// the handler has been called with all children.
```

packages/flame/lib/src/components/component.dart
```diff
@@ -93,10 +305,71 @@ abstract class Component {
...
+  Future<void>? _onLoadCache;
+
+  /// Since [onLoad] only should run once throughout a components lifetime it is
+  /// cached so that it can be reused when the parent component/game/widget
+  /// changes.
+  @internal
+  Future<void>? get onLoadCache => _onLoadCache ?? (_onLoadCache = onLoad());
+
+  /// Called after the component has successfully run [prepare] and [onLoad] and
+  /// called before the component is added to its new parent.
```
erickzanardo, line 320: too many called words :P
```suggestion
  /// before the component is added to its new parent.
```

packages/flame/lib/src/components/component.dart
```diff
@@ -93,10 +305,71 @@ abstract class Component {
...
+  ///
+  /// Whenever [onMount] returns something, the parent will wait for the
+  /// [Future] to be resolved before adding the component on the list.
+  /// If `null` is returned, the component is added right away.
+  ///
+  /// The default implementation just returns null.
```
erickzanardo, line 326: this wasn't true
```suggestion
```

packages/flame/lib/src/components/component.dart
```diff
@@ -93,10 +305,71 @@ abstract class Component {
...
+  ///
```
erickzanardo, line 327: Remove the blank line that will result on the above comment
```suggestion
```

packages/flame/lib/src/effects/effects.dart
erickzanardo: Usually I wouldn't want too, but idk, Helper extensions seems weird to me, but I have no logical good reason for this, is just a personal thing. So I leave it up to you.

packages/flame/pubspec.yaml
erickzanardo: Makes sense, we probably still needs it then.

packages/flame/lib/src/components/component.dart
```diff
@@ -3,88 +3,300 @@ import 'dart:ui';
...
 
+  /// Returns whether this [Component] is in debug mode or not.
+  /// When a child is added to the [Component] it gets the same [debugMode] as
+  /// its parent has when it is prepared.
+  ///
+  /// Returns `false` by default. Override it, or set it to true, to use debug
+  /// mode.
+  /// You can use this value to enable debug behaviors for your game and many
+  /// components will
+  /// show extra information on the screen when debug mode is activated
```
erickzanardo, line 75:
```suggestion
  /// show extra information on the screen when debug mode is activated.
```

erickzanardo (approved):

examples/lib/stories/input/keyboard.dart
```diff
@@ -1,13 +1,12 @@
...
 
 import 'package:flame/game.dart';
-
 import 'package:flame/input.dart';
 import 'package:flame/palette.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 
-class KeyboardGame extends Game with KeyboardEvents {
+class KeyboardGame extends FlameGame with KeyboardEvents {
```
luanpotter, line 9: would this be a good candiate for a Game example? since it was before & it uses no components

packages/flame/lib/src/components/position_component.dart
```diff
@@ -83,6 +85,18 @@ class PositionComponent extends BaseComponent {
...
+  @mustCallSuper
+  void update(double dt) {
+    super.update(dt);
+  }
+
+  @override
+  @mustCallSuper
+  void render(Canvas canvas) {
+    super.render(canvas);
+  }
```
luanpotter, line 98: do you need to override just to call super?

packages/flame/lib/src/fps_counter.dart
```diff
@@ -1,6 +1,6 @@
 import 'package:flutter/scheduler.dart';
 
-import 'game/game.dart';
+import '../game.dart';
```
luanpotter, line 3: would this file fit well inside the `game` directory?

packages/flame/lib/src/game/base_game.dart
```diff
@@ -1,285 +0,0 @@
-import 'dart:ui';
```
luanpotter, line 1: farewell, old friend!

packages/flame/lib/src/game/camera/camera_wrapper.dart
```diff
@@ -1,10 +1,11 @@
...
 import 'camera.dart';
 
-/// This class encapsulates BaseGame's rendering functionality. It will be
+/// This class encapsulates FlameGame's rendering functionality. It will be
 /// converted into a proper Component in a future release, but until then
-/// using it in any code other than the BaseGame class is unsafe and
+/// using it in any code other than the FlameGame class is unsafe and
 /// not recommended.
 class CameraWrapper {
   // TODO(st-pasha): extend from BaseComponent
```
luanpotter, line 11: this comment probably needs updating as BasComp is no more

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -23,17 +23,18 @@ typedef OverlayWidgetBuilder<T extends Game> = Widget Function(
   T game,
 );
 
-/// A [StatefulWidget] that is in charge of attaching a [Game] instance into the flutter tree
-///
+/// A [StatefulWidget] that is in charge of attaching a [Game] instance into the
+/// flutter tree.
```
luanpotter, line 27:
```suggestion
/// Flutter tree.
```

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -135,10 +138,11 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
 
   MouseCursor? _mouseCursor;
 
-  Future<void>? _gameLoaderFuture;
-
-  Future<void> get _gameLoaderFutureCache =>
-      _gameLoaderFuture ?? (_gameLoaderFuture = widget.game.onLoad());
+  Future<void> get _loaderFuture {
```
luanpotter, line 141: I think you can have async getters no problem:
```
class A {
  Future<void> get foo async {
    await bar();
  }
  
  Future<void> bar() async {}
}
```
would that make this slightly simpler?

packages/flame/lib/src/game/mixins/loadable.dart
```diff
@@ -0,0 +1,51 @@
+import 'package:meta/meta.dart';
+
+mixin Loadable {
```
luanpotter, line 3: nice abstraction! maybe a brief dartdoc on the mixin level for when it's used?

packages/flame/lib/src/game/mixins/loadable.dart
```diff
@@ -0,0 +1,51 @@
...
+  /// Example:
+  /// ```dart
+  /// @override
+  /// void onMount() {
+  ///   position = parent!.size / 2;
+  /// }
+  /// ```
+  void onMount() {}
+
+  void onRemove() {}
```
luanpotter, line 50: comments?

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -0,0 +1,228 @@
...
+import 'camera/camera.dart';
+import 'camera/camera_wrapper.dart';
+import 'mixins/game.dart';
+
+/// This is a more complete and opinionated implementation of Game.
+///
+/// FlameGame should be extended to add your game logic.
+/// [update], [render] and [onGameResize] methods have default implementations.
+/// This is the recommended structure to use for most games.
+/// It is based on the Component system.
```
luanpotter, line 22: Flame Component system ?

luanpotter (approved):
GET THIS BABY IN!

packages/flame/lib/src/components/position_component.dart
spydon: This was needed before when `Component` didn't have `@mustCallSuper`, since OxygenGame shouldn't be force to call super, but now since it is using the low-level API I can re-add `@mustCallSuper` on `Component` instead. :)

packages/flame/lib/src/game/camera/camera_wrapper.dart
spydon: Fixed

packages/flame/lib/src/fps_counter.dart
spydon: Indeed it would, in `game/mixins`.

examples/lib/stories/input/keyboard.dart
spydon: Done

packages/flame/lib/src/game/mixins/loadable.dart
spydon: Added!

packages/flame/lib/src/game/game_widget/game_widget.dart
luanpotter: I was thinking:
```
if (foo != null)
  await foo()
if (bar != null)
  await bar()
```
but maybe it's not better, not sure

PR #907, opened 2021-08-08
NotifyingVector2 class

st-pasha:
# Description

Added class `NotifyingVector2`, which will be used for `Transform2D` in PR #904.
Moving this class out into a separate PR to make it easier to review.

## Checklist

- [x] My PR does not decrease the code coverage.
- [x] I have formatted my code with `flutter format .` and the Flame analyzer (`flutter analyze`) does not report any problems.

## Breaking Change

## Related Issues

WIP for #904


spydon (approved):
Lgtm, good job!

packages/flame/lib/src/game/notifying_vector2.dart
```diff
@@ -0,0 +1,190 @@
...
+  set y(double y) {
+    super.y = y;
+    notifyListeners();
+  }
+
+  @override
+  Float64List get storage {
+    // If we allow this, then the user would be able to modify the internal
+    // state of the vector without triggering notifications to the listeners.
+    throw UnsupportedError(
```
luanpotter, line 186: one option would be to return a copy. but that could lead to misunderstandings

packages/flame/lib/src/game/notifying_vector2.dart
```diff
@@ -0,0 +1,190 @@
...
+/// value of this vector changes.
+///
+/// This class can be used as a regular [Vector2] class. However, if you do
+/// subscribe to notifications, don't forget to eventually unsubscribe in
+/// order to avoid resource leaks.
+///
+/// Direct access to this vector's [storage] is not allowed.
+///
+// ignore: prefer_mixin
+class NotifyingVector2 extends Vector2 with ChangeNotifier {
```
luanpotter, line 16: some places we don't allow for you to set your own instance, eg PositionComponent.position IIRC has some custom setters. I wonder what scenarios this would not work with

packages/flame/lib/src/game/notifying_vector2.dart
```diff
@@ -0,0 +1,190 @@
...
+/// functionality. This allows any interested party to be notified when the
+/// value of this vector changes.
+///
+/// This class can be used as a regular [Vector2] class. However, if you do
+/// subscribe to notifications, don't forget to eventually unsubscribe in
+/// order to avoid resource leaks.
+///
+/// Direct access to this vector's [storage] is not allowed.
+///
+// ignore: prefer_mixin
```
luanpotter, line 15: we should consider disable this rule haha

packages/flame/lib/src/game/notifying_vector2.dart
spydon: Isn't that exactly the case that this will be used for? So that #904 can recalculate the matrix for example

st-pasha: Yes, this is how it is being used by `Transform2D`: it exposes `.position` as a `NotifyingVector2`, and subscribes to any changes in that vector. Thus, when the user says, for example, `transform.position.setFrom(v)` the `transform` gets notified about the fact that its position has changed.

packages/flame/lib/src/game/notifying_vector2.dart
st-pasha: I guess the primary reason for requesting `vector2d.storage` is to be able to manipulate that data. If you just need to read the data, then there are `.x`, `.y` and `.xy` getters.

packages/flame/lib/src/game/notifying_vector2.dart
luanpotter: ah, I see. I thought this was more for users of Flame to use on their components, but rather the idea is to use it ourselves on PositionComponent. gotcha.

luanpotter (approved):
LGTM

packages/flame/lib/src/game/notifying_vector2.dart
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  @override
+  set y(double y) {
+    super.y = y;
+    notifyListeners();
+  }
+
+  @override
+  Float64List get storage {
```
wolfenrain, line 183: We could return a [UnmodifiableFloat64ListView](https://api.flutter.dev/flutter/dart-typed_data/UnmodifiableFloat64ListView-class.html), that one throw errors on data modifying and would still allow people to read the storage data for copying and such

wolfenrain (changes requested):
LGTM, but missing a changelog entry ;-)

packages/flame/lib/src/game/notifying_vector2.dart
st-pasha: Nice, didn't know about that one.

packages/flame/lib/src/game/notifying_vector2.dart
st-pasha: will do

packages/flame/analysis_options.yaml
```diff
@@ -94,7 +94,7 @@ linter:
     - prefer_is_not_empty
     - prefer_is_not_operator
     - prefer_iterable_whereType
-    - prefer_mixin
+#    - prefer_mixin
```
wolfenrain, line 97: Should we just remove that line in total? Or have a follow up PR where we clean this up(as we reference it in code as well I believe).

/cc @flame-engine/flame-colab

wolfenrain (approved):
LGTM, left one review but that can be a follow up on our side I think

packages/flame/analysis_options.yaml
spydon: What does it do? Prefer a mixin over what?

spydon: Found it [here](https://dart-lang.github.io/linter/lints/prefer_mixin.html), I don't really get how the two cases that they show there can be comparable though.

st-pasha: @wolfenrain In my projects, I prefer to comment-out linter rules (perhaps even with a comment why), as opposed to straight-out deleting them, because then it's clear which rules, were deliberately deleted, and which were never there to begin with. For example, if the linter gets updated with new rules, I can then compare the new list vs what I have in the project and add the difference. However, if you have different conventions in `flame`, then it's fine too.

@spydon In my case, the analyzer complained that I use `... with ChangeNotifier`, even though `ChangeNotifier` is declared as a `class` and not a `mixin`. However, the docs for `ChangeNotifier` state that it can be used both as a parent and as a mixin, so the rule was giving a false positive.

wolfenrain: I personally don't have any preference for it, makes sense to document certain rules that are turned off as well.

But I know we have removed rules in the past, that's why I added it. I think we should choose a convention and stick to it, and your explanation has gotten my vote haha

spydon: Let's keep it like this for now and we can discuss it in the meeting later and possibly do a follow-up. :)

PR #914, opened 2021-08-10
Transform2D class

st-pasha:
# Description

New class `Transform2D` whose purpose is to describe a compound 2d transform: a combination of translation, rotation, reflection, and scaling.

This class is intended to be used in `PositionComponent`, and also possibly in `Camera` and `PositionComponentEffect`.

## Checklist

- [x] My PR does not decrease the code coverage, 

## Breaking Change

## Related Issues

WIP for #904

### DIFF
--- a/packages/flame/lib/src/game/notifying_vector2.dart
+++ b/packages/flame/lib/src/game/notifying_vector2.dart
@@ -13,7 +13,13 @@ import 'package:vector_math/vector_math_64.dart';
 /// Direct modification of this vector's [storage] is not allowed.
 ///
 class NotifyingVector2 extends Vector2 with ChangeNotifier {
-  NotifyingVector2() : super.zero();
+  factory NotifyingVector2(double x, double y) => NotifyingVector2.zero() ..setValues(x, y);
+
+  NotifyingVector2.zero() : super.zero();
+
+  factory NotifyingVector2.all(double v) => NotifyingVector2.zero() ..splat(v);
+
+  factory NotifyingVector2.copy(Vector2 v) => NotifyingVector2.zero() ..setFrom(v);
 
   @override
   void setValues(double x, double y) {
--- a/packages/flame/lib/src/game/notifying_vector2.dart
+++ b/packages/flame/lib/src/game/notifying_vector2.dart
@@ -13,13 +13,15 @@ import 'package:vector_math/vector_math_64.dart';
 /// Direct modification of this vector's [storage] is not allowed.
 ///
 class NotifyingVector2 extends Vector2 with ChangeNotifier {
-  factory NotifyingVector2(double x, double y) => NotifyingVector2.zero() ..setValues(x, y);
+  factory NotifyingVector2(double x, double y) =>
+      NotifyingVector2.zero()..setValues(x, y);
 
   NotifyingVector2.zero() : super.zero();
 
-  factory NotifyingVector2.all(double v) => NotifyingVector2.zero() ..splat(v);
+  factory NotifyingVector2.all(double v) => NotifyingVector2.zero()..splat(v);
 
-  factory NotifyingVector2.copy(Vector2 v) => NotifyingVector2.zero() ..setFrom(v);
+  factory NotifyingVector2.copy(Vector2 v) =>
+      NotifyingVector2.zero()..setFrom(v);
 
   @override
   void setValues(double x, double y) {
--- a/packages/flame/lib/src/game/transform2d.dart
+++ b/packages/flame/lib/src/game/transform2d.dart
@@ -42,8 +42,7 @@ class Transform2D extends ChangeNotifier {
         _angle = 0,
         _position = NotifyingVector2(),
         _scale = NotifyingVector2()..setValues(1, 1),
-        _offset = NotifyingVector2()
-  {
+        _offset = NotifyingVector2() {
     _position.addListener(_markAsModified);
     _scale.addListener(_markAsModified);
     _offset.addListener(_markAsModified);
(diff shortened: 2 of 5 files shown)


packages/flame/lib/src/game/transform2d.dart
```diff
@@ -0,0 +1,203 @@
...
+  final NotifyingVector2 _scale;
+  final NotifyingVector2 _offset;
+  static const tau = 2 * math.pi;
+
+  Transform2D()
+      : _transformMatrix = Matrix4.identity(),
+        _recalculate = true,
+        _angle = 0,
+        _position = NotifyingVector2(),
+        _scale = NotifyingVector2()..setValues(1, 1),
```
spydon, line 44:
```suggestion
        _scale = NotifyingVector2.all(1.0),
```
It should be possible to do it like this to skip the initial notification that would otherwise incur right?
Not that there are any listeners, but still.

packages/flame/lib/src/game/transform2d.dart
```diff
@@ -0,0 +1,203 @@
...
+    _position.addListener(_markAsModified);
+    _scale.addListener(_markAsModified);
+    _offset.addListener(_markAsModified);
+  }
+
+  factory Transform2D.copy(Transform2D other) => Transform2D()
+    ..angle = other.angle
+    ..position = other.position
+    ..scale = other.scale
+    ..offset = other.offset;
```
spydon, line 55: I think you would want clones of these values and not the same references?

packages/flame/lib/src/game/transform2d.dart
```diff
@@ -0,0 +1,203 @@
...
+
+  /// Check whether this transform is equal to other, up to the given
```
spydon, line 57:
```suggestion
  /// Check whether this transform is equal to [other], up to the given
```

packages/flame/lib/src/game/transform2d.dart
```diff
@@ -0,0 +1,203 @@
...
+  /// [tolerance]. Setting tolerance to zero will check for exact equality.
+  /// Transforms are considered equal if their rotation angles are the same
+  /// or differ by a multiple of 2π, and if all other transform parameters:
+  /// translation, scale, and offset are the same.
+  ///
+  /// The [tolerance] parameter is in absolute units, not relative.
+  bool closeTo(Transform2D other, {double tolerance = 1e-10}) {
```
spydon, line 64: I don't think it matters, but the tolerance value here would be a much larger change for for example `scale` than for `position`, in most cases.

spydon (commented):
Looks good, just some minor comments

packages/flame/lib/src/game/transform2d.dart
st-pasha: A line `..position = other.position` invokes the setter
```
set position(Vector2 position) => _position.setFrom(position);
```
where `.setFrom` ultimately copies the values:
```
  void setFrom(Vector2 other) {
    final otherStorage = other._v2storage;
    _v2storage[1] = otherStorage[1];
    _v2storage[0] = otherStorage[0];
  }
```
So, in this case the code that I wrote should work just fine.

Also, I have a test `test('.copy', ...)` that verifies that a clone does not get modified when the original transform is modified.

packages/flame/lib/src/game/transform2d.dart
st-pasha: Yes, that's the downside of absolute error... I mostly used it because it's slightly less code to write, but perhaps it's better to switch to relative tolerance then?

packages/flame/lib/src/game/transform2d.dart
spydon: Aah, of course! Apparently I hadn't woken up yet when I did the review.

packages/flame/lib/src/game/transform2d.dart
spydon: I don't think it will make much difference in reality since epsilon probably will be quite small in most cases anyways, I think we can just re-visit this if we experience problems with it in the future.

packages/flame/lib/src/game/transform2d.dart
st-pasha: done

packages/flame/lib/src/game/transform2d.dart
st-pasha: done

spydon (approved):

packages/flame/lib/src/game/notifying_vector2.dart
```diff
@@ -13,7 +13,15 @@ import 'package:vector_math/vector_math_64.dart';
 /// Direct modification of this vector's [storage] is not allowed.
 ///
 class NotifyingVector2 extends Vector2 with ChangeNotifier {
-  NotifyingVector2() : super.zero();
+  factory NotifyingVector2(double x, double y) =>
+      NotifyingVector2.zero()..setValues(x, y);
```
spydon, line 17: This doesn't need to be a factory right? It could just be a constructor that invokes super?

st-pasha: I'm not particularly sure what's going on here, but if I remove `factory` keyword it gives an error
```
The constructor 'Vector2 Vector2(double x, double y)' is a factory constructor, but must be a generative constructor to be a valid superinitializer.
```
Also, this is how they've set up constructors in the base `Vector2` class, so I just followed the lead.

spydon: I see, let's go with that then.

erickzanardo (approved):

wolfenrain (approved):
LGTM! Appreciate the detailed in-code documentation about the matrix behavior, that will  help in the long run when people have to look at the code again!

Issue #915, opened 2021-08-12
Occasional test failure


st-pasha:
I'm occasionally seeing test failures from running `flutter test` that look like this:
```
00:08 +310: /Users/pasha/github/st-pasha/flame/packages/flame/test/effects/size_effect_test.dart: SizeEffect can be infinite                                                                                                                 
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following TestFailure object was thrown running a test:
  Expected: a numeric value within <0.01> of <0.0>
  Actual: <82.29216244576264>
   Which:  differs by <82.29216244576264>
Size is not correct

When the exception was thrown, this was the stack:
#0      fail (package:test_api/src/frontend/expect.dart:153:31)
#1      _expect (package:test_api/src/frontend/expect.dart:148:3)
#2      expect (package:test_api/src/frontend/expect.dart:57:3)
#3      expectVector2 (package:flame/src/test_helpers/expect_vector2.dart:11:3)
#4      effectTest (file:///Users/pasha/github/st-pasha/flame/packages/flame/test/effects/effect_test_utils.dart:61:5)
<asynchronous suspension>

The test description was:
  SizeEffect can be infinite
════════════════════════════════════════════════════════════════════════════════════════════════════
00:08 +311 -1: /Users/pasha/github/st-pasha/flame/packages/flame/test/effects/size_effect_test.dart: SizeEffect can be infinite [E]                                                                                                          
  Test failed. See exception logs above.
  The test description was: SizeEffect can be infinite
  
00:08 +327 -1: /Users/pasha/github/st-pasha/flame/packages/flame/test/effects/sequence_effect_test.dart: SequenceEffect can alternate and be infinite                                                                                        
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following TestFailure object was thrown running a test:
  Expected: <true>
  Actual: <false>
Angle is not correct

When the exception was thrown, this was the stack:
#0      fail (package:test_api/src/frontend/expect.dart:153:31)
#1      _expect (package:test_api/src/frontend/expect.dart:148:3)
#2      expect (package:test_api/src/frontend/expect.dart:57:3)
#3      expectDouble (package:flame/src/test_helpers/expect_double.dart:9:3)
[...]

PR #916, opened 2021-08-13
Reproducible random tests

st-pasha:
# Description

- Added two functions to help with the randomized unit testing: `testRandom()` and `testWidgetsRandom()`. The purpose of these functions is to make the randomized tests reproducible.
- Refactored all effects tests to use these functions. 

## Checklist

- [x] My PR does not decrease the code coverage.

## Breaking Change

## Related Issues

WIP for #915

### DIFF
--- a/packages/flame/lib/src/test_helpers/random_test.dart
+++ b/packages/flame/lib/src/test_helpers/random_test.dart
@@ -50,7 +50,9 @@ void testRandom(
 }
 
 typedef TestWidgetsCallback = Future<void> Function(
-    Random random, WidgetTester widgetTester);
+  Random random,
+  WidgetTester widgetTester,
+);
 
 /// This function is equivalent to `testWidgets(name, body)`, except that
 /// it is better suited for randomized testing: it will create a Random
--- a/packages/flame/lib/src/test_helpers/random_test.dart
+++ b/packages/flame/lib/src/test_helpers/random_test.dart
@@ -1,0 +1,99 @@
+import 'dart:math';
+import 'package:meta/meta.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+final _rnd = Random(); // for generating seeds
+const _maxSeed = 4294967296;
+
+/// This function is equivalent to `test(name, body)`, except that it is
+/// better suited for randomized testing: it will create a Random
+/// generator and pass it to the test body, but also record the seed
+/// that was used for creating the random generator. Thus, if a test
+/// fails for a specific rare seed, it would be easy to reproduce this
+/// failure.
+///
+/// In order for this to work properly, all random number generation
+/// within `testRandom()` must be performed through the provided random
+/// generator.
+///
+/// Example of use:
+///
+///     testRandom('description', (Random random) {
+///       expect(random.nextDouble() == random.nextDouble(), false);
+///     });
+///
+/// Then if the test output shows that the test failed with seed `S`,
+/// simply adding parameter `seed=S` into the function will force it
+/// to run for that specific seed.
+///
+@isTest
+void testRandom(
+  String name,
+  void Function(Random random) body, {
+  String? testOn,
+  Timeout? timeout,
+  dynamic skip,
+  dynamic tags,
+  Map<String, dynamic>? onPlatform,
+  int? retry,
+  int? seed,
+}) {
+  seed ??= _rnd.nextInt(_maxSeed);
+  test(
+    '$name [seed=$seed]',
+    () => body(Random(seed)),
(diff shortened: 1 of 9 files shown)


spydon (approved):
Nice! Lgtm

erickzanardo (approved):

packages/flame/lib/src/test_helpers/random_test.dart
```diff
@@ -0,0 +1,99 @@
+import 'dart:math';
+import 'package:meta/meta.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+final _rnd = Random(); // for generating seeds
```
wolfenrain, line 5: If it is for seed generating I would personally give it a name that would reflect that usage.

packages/flame/lib/src/test_helpers/random_test.dart
```diff
@@ -0,0 +1,99 @@
...
+///
+/// In order for this to work properly, all random number generation
+/// within `testRandom()` must be performed through the provided random
+/// generator.
+///
+/// Example of use:
+///
+///     testRandom('description', (Random random) {
+///       expect(random.nextDouble() == random.nextDouble(), false);
+///     });
```
wolfenrain, line 23: This should be in a code block using backticks instead. That is the standard we been handling in our documentation

packages/flame/lib/src/test_helpers/random_test.dart
```diff
@@ -0,0 +1,99 @@
...
+///
+/// Then if the test output shows that the test failed with seed `S`,
+/// simply adding parameter `seed=S` into the function will force it
+/// to run for that specific seed.
+///
```
wolfenrain, line 28: Noticed u always put an empty doc line in your supplied docs.

Havent had a "nit" about it yet but it isn't according to our (unwritten) dartdoc standards :-). Any reason why u do this?

packages/flame/test/effects/move_effect_test.dart
```diff
@@ -2,14 +2,19 @@ import 'dart:math';
...
 import 'package:flame/components.dart';
 import 'package:flame/effects.dart';
+import 'package:flame/src/test_helpers/random_test.dart';
 import 'package:flutter_test/flutter_test.dart';
 
 import 'effect_test_utils.dart';
 
-void main() {
-  final random = Random();
+class Elements {
```
wolfenrain, line 10: Is this `Elements` the same as the one in `combined_effect_test`? Wondering if we can somehow prevent duplicate code

wolfenrain (changes requested):
Few nits

packages/flame/test/effects/move_effect_test.dart
wolfenrain: Did a quick look at it again and looks like we can have some sort of base class with the duplicate code like the `randomVector2`

packages/flame/lib/src/test_helpers/random_test.dart
st-pasha: Sure, I'll do it.

packages/flame/lib/src/test_helpers/random_test.dart
st-pasha: ty, good to know

packages/flame/lib/src/test_helpers/random_test.dart
st-pasha: No problem, I'll remove the extra line.

As for the reason, I normally code in C++ / Python, and in C++ the doc-comments look like this:
```
/**
 * doc...
 */
```
while  in Python they are
```
"""
doc...
"""
```
so in both cases, there is a visual empty line after the doc-comment

packages/flame/test/effects/move_effect_test.dart
st-pasha: Indeed, there seems to be a fair amount of code duplication there. However, I don't think it's within the scope of this PR to fix that, is it?

spydon: I think the code duplication for the `Elements` class can be fixed in this PR since it is introduced in this PR, it should be fairly quick to fix I think?

st-pasha: It wasn't introduced in this PR -- I merely took all the initialization code at the top of each `main()` and moved it into a class. But you're right about that it should be fairly quick to fix.

st-pasha: ok done

wolfenrain (approved):
LGTM

PR #932, opened 2021-08-30
Function to load all images in prefixed path

spydon:
# Description

To simplify loading all image assets in the specified (or default) path into the cache.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/CHANGELOG.md
+++ b/packages/flame/CHANGELOG.md
@@ -41,6 +41,7 @@
  - Update `Camera` docs to showcase usage with `Game` class
  - Fixed a bug with `worldBounds` being set to `null` in `Camera`
  - `MockCanvas` is now strongly typed and matches numeric coordinates up to a tolerance
+ - Add `loadAllImages` to `Images`, which loads all images from the prefixed path
 
 ## [1.0.0-releasecandidate.13]
  - Fix camera not ending up in the correct position on long jumps
--- a/packages/flame/lib/src/assets/images.dart
+++ b/packages/flame/lib/src/assets/images.dart
@@ -1,10 +1,11 @@
 import 'dart:async';
-import 'dart:convert' show base64;
+import 'dart:convert' show base64, json;
 import 'dart:typed_data';
 import 'dart:ui';
 import 'dart:ui' as ui show decodeImageFromPixels;
 
 import 'package:flutter/foundation.dart';
+import 'package:flutter/services.dart';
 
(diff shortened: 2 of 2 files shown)


st-pasha (commented):
From the API standpoint, there are now the following methods:
```
Images.load(String filename)
Images.loadAll(List<String> filenames)
Images.loadAllImages()
Images.loadAllFromPattern(Pattern pattern)
```
This is somewhat confusing, in particular, the 2nd and the 3rd methods sound too similar, and it's not immediately clear from their names what the difference is. So, perhaps we could go with something like
```
Images.load(String filename)
Images.loadList(List<String> filenames)
Images.loadAllAssets(Pattern? pattern)
```

erickzanardo (approved):

spydon:
> From the API standpoint, there are now the following methods:
> 
> ```
> Images.load(String filename)
> Images.loadAll(List<String> filenames)
> Images.loadAllImages()
> Images.loadAllFromPattern(Pattern pattern)
> ```
> 
> This is somewhat confusing, in particular, the 2nd and the 3rd methods sound too similar, and it's not immediately clear from their names what the difference is. So, perhaps we could go with something like
> 
> ```
> Images.load(String filename)
> Images.loadList(List<String> filenames)
> Images.loadAllAssets(Pattern? pattern)
> ```

These are only for images though, not for all assets (like audio for example).

We do have the `loadAll` pattern for passing in a list in different places in the API too.
What about:
```
> Images.load(String filename)
> Images.loadAll()
> Images.loadList(List<String> filenames)
> Images.loadAllAssets(Pattern? pattern)

spydon:
@st-pasha I can't come up with any better names, do you have any other suggestions by chance?
I'd like to keep the `loadAll` for the list, since that pattern is used both in dart and in other places in Flame.
And I think it is clearer for the user to have the method that loads from a pattern and the one that loads all images separate, instead of relying on a nullable argument. Any ideas from @flame-engine/flame-colab maybe?

erickzanardo:
> @st-pasha I can't come up with any better names, do you have any other suggestions by chance?
> I'd like to keep the `loadAll` for the list, since that pattern is used both in dart and in other places in Flame.
> And I think it is clearer for the user to have the method that loads from a pattern and the one that loads all images separate, instead of relying on a nullable argument. Any ideas from @flame-engine/flame-colab maybe?

I am ok the current choice of names, can't think on better options too.

Issue #931, opened 2021-08-26
Camera thoughts 


st-pasha:
So, I've been pondering about the camera and how to integrate it with `Transform2D` and had some thoughts about the general direction in which this class can be developed:

1. **`Camera` and `Viewport` can be merged into a single entity.** This is both from the efficiency standpoint (currently both the camera and the viewport will translate/zoom the canvas), and from the practical standpoint: currently both classes try to map between the screen coordinates and world coordinates, and also in the real-world a viewport is simply a property of a camera.

2. **`Camera` shouldn't be a property of the `Game`**, instead, it could be a simple `Component`. There are several reasons for this: first, not all games (or game views) need a camera. For example, a simple puzzle game where everything fits on a screen, or an inventory screen, or character view, or trade UI, etc. -- all don't need a camera. At the same time, there could be situations where multiple cameras are needed. For example: a split-screen multiplayer game; a game which allows magnification of a small part of a screen, such as magnifying glass functionality, or a sniper scope; minimap functionality (minimap is a camera into a separate zoomed-out map of the world); a game which shows the views from multiple security cameras at the same time; and so on. Even now, in order to distinguish components that shouldn't be affected by the camera, we introduced the `.isHud` flag. But that flag is unnecessary if we simply keep the "Hud" components in a separate component tree.

3. **`Camera` can be made more modular.** Currently, the camera comes with many built-in features that make it really easy to implement the most common camera behaviors: move, follow, shake. However, there is no easy way to augment or add to this functionality. For example, if I wanted the camera that follows the character horizontally but not vertically, there is no way to do that. Or if I wanted a camera that zooms out when the character gains speed (like in GTA). Or a camera that can smoothly zoom in and center on a certain object. Or a Perlin-noise-based shake. Or a world boundary that is circular instead of rectangular (someone asked for that), or an even weirder shape.
One possibility here is to split the "camera behavior" logic into a separate class or several classes, and allow those to be overridden and swapped out at will. I am not sure whether the effects (such as shake) should be kept separate from the "normal" behaviors such as follow or apply-world-boundary.

st-pasha:
Closing, see the final proposal in #1161.

PR #933, opened 2021-08-31
CameraComponent

### LINKED ISSUE #934: Components with `isHud = true` can be rendered behind normal components
# Current bug behaviour
Components with `isHud = true` can be rendered behind normal components.

# Expected behaviour
Components with `isHud = true` should be rendered on top of all components that have `isHud = false`

# Steps to reproduce
Run the advanced joystick example and steer the sprite down over the buttons or the joystick.
https://examples.flame-engine.org/#/Input_Joystick%20Advanced

# Flutter doctor output

```
Output of: flutter doctor -v
```

* Flame version: 1.0.0 main branch
* Platform affected: all

st-pasha:
# Description

This PR performs certain preliminary refactorings in order to enable further work on implementing multi-camera support. Currently, the main obstacle for having multiple cameras is that the camera is merged into the `BaseGame` class and dissociated from its viewport.

In this PR, we
- Put `Viewport` inside the `Camera`
  
  Not only it is difficult to imagine a camera without a viewport, or a viewport without a camera -- the `Camera` actually needs to know about its viewport in order to function correctly. Before, the viewport was kept as a property of the BaseGame, and the camera had to look up that property using a `gameRef`. Obviously, such an approach wouldn't work if we want to have multiple cameras. Now, the `Viewport` is tightly coupled to its `Camera` (though it remains a separate class). As an additional benefit, the camera no longer needs to keep the `gameRef` reference.

- Create a `CameraComponent` component that wraps the `Camera` functionality for an FCS game.

  In order for a game to have multiple cameras, the camera must become a component. However, since we want to be able to reuse the `Camera` class across many different types of games (including OxygenGame, Forge2DGame, etc), we encapsulate the "component" functionality in a separate small class `CameraComponent`.

- `camera.dart` and `viewport.dart` moved into the `game/camera/` folder, for easier maintenance (and in anticipation of more camera-related files).

## Checklist

## Breaking Change

  Note: the only breaking change is that `viewport` now needs to be accessed as `camera.viewport`. It would be easy to add this as a simple getter into `BaseGame`, but I'm not sure we want that.

## Related Issues

WIP for #931
Closes #934


packages/flame/lib/src/game/camera/camera_component.dart
```diff
@@ -0,0 +1,43 @@
...
+      world.forEach((component) {
+        if (!component.isHud) {
+          canvas.save();
+          component.renderTree(canvas);
+          canvas.restore();
+        }
+      });
+      canvas.restore();
+      // Then render the HUD
+      world.forEach((component) {
```
erickzanardo, line 34: I wonder the perfomance impacts of running through the component set twice

erickzanardo (commented):
The viewport joining the camera change I am ok with. But I am still unsure about having camera as a component.

By keeping it as an attribute on the game, when the game as component refactor is done, we will be able to have multiple cameras and I believe that it would be way easier for the components to access the camera through the HasGameRef mixin

packages/flame/lib/src/game/camera/camera_component.dart
st-pasha: Agree, running through the list twice is suboptimal. Ideally, we would want to have them as two separate component trees: one is the **world**, and another is the **hud**, and then the performance concerns would disappear. However, I left this refactoring step for the future, since it requires #906 to be merged.

In the meanwhile, it is not clear whether this double-loop setup made things better or worse. On the one hand, as you noted, we iterate through the component tree twice. On the other hand, we no longer need to apply camera transform for every non-hud component: this is done only once outside the loop. Plus, the new approach ensures that any hud components are actually drawn on top of the world, not below it.

st-pasha:
@erickzanardo 
> I am still unsure about having camera as a component. By keeping it as an attribute on the game, when the game as component refactor is done, we will be able to have multiple cameras and I believe that it would be way easier for the components to access the camera through the HasGameRef mixin

If I understand you correctly, you're saying that we can achieve multi-camera functionality by having one or more extra games inside the main game. Thus, a `Game` is the world, and the camera is fused with the game because it is necessary to have a camera in order to see the world.

My thinking, however, is that a `Game` is more of an abstract entity that ties together multiple parts that together make up the game: the world (which is a component tree), the camera (which is a component that renders the world), the hud (which is probably tied to the camera). What I want to achieve is:
1. The ability to have multiple cameras observing the same world (e.g. think multiplayer split-screen). It is possible that this could be achieved via a special "multi-camera" which somehow ties together several cameras -- but this would still require a special render logic that is different from the current rendering logic in `BaseGame`, which means that logic must be moved out into a `CameraComponent` class?
2. The ability to have a camera into the world inside the world itself. This could be used for example for creating mirror reflections (such as in the water). I am still not sure what would be needed to accomplish this, but it seems to me that the only way to mount a camera inside the game world is to have the camera as a component.

I'm still very much unsure what the overall architecture should be. Perhaps we could have a `CameraMixin` on `Game` that sets up some default camera functionality?

packages/flame/lib/src/game/base_game.dart
```diff
@@ -24,33 +23,20 @@ import 'viewport.dart';
...
-    }
-    _combinedProjector = Projector.compose([camera, value]);
-  }
-
-  late Projector _combinedProjector;
-
-  final Vector2 _sizeBuffer = Vector2.zero();
+  // When the Game becomes a Component (#906), this could be added directly
+  // into the component tree.
+  late final CameraComponent _cameraComponent;
```
spydon, line 39: Why can it not be added directly in to the component tree already?

st-pasha: I was hitting a snag here: the `onGameResize` method passes `canvasSize` to the camera, but `size` to all other elements. I'm not exactly sure how to approach this issue, but here's what I've been thinking so far:

Currently, BaseGame passes `size = viewport.effectiveSize/camera.zoom` to every component. Essentially, this is the size of the game world visible through the viewport. There is a problem, however: if the camera's zoom level changes, or if the viewport's effective size changes for whatever reason -- the `onGameResize` wouldn't be called. I believe this would be a potential source of bugs for any component that relies on `onGameResize`.

On the other hand, the only component in Flame that currently uses `onGameResize` is `ParallaxComponent`, and even it ignores the passed `size` parameter. So, maybe instead of having `onGameResize` as a lifecycle method, a component would instead subscribe to resize events on either the Game itself or on the camera's viewport?

Honestly, I'm not sure what the correct solution should be. Obviously, the main camera needs to be informed of the canvas size in order to size and position its viewport correctly. On the other hand, components that are "within" the camera do not care about the canvas size -- for them, the viewport is the equivalent of the canvas. For example, if someone wants to have a camera within the camera - the internal camera will only care about the dimensions made available to it by the external camera. So, perhaps we could make each camera responsible for calling `onResize` on all the components that it observes, passing along its viewport size?

packages/flame/lib/src/game/camera/viewport.dart
```diff
@@ -66,29 +66,26 @@ abstract class Viewport extends Projector {
   /// any viewport.
   ///
   /// You probably don't need to care about this if you are using a viewport.
-  Vector2 get canvasSize;
+  Vector2? canvasSize;
```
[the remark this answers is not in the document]
st-pasha: The `canvasSize` is not known (`null`) before everything was laid out. I could have had `hasLayout` boolean flag instead, but it seems that having the property nullable is more straightforward.

packages/flame/lib/src/game/base_game.dart
spydon: Right, makes sense. We'll have to think about how to solve that.

> So, perhaps we could make each camera responsible for calling onResize on all the components that it observes, passing along its viewport size?

How would that work if a component is observed by multiple cameras?

st-pasha: > How would that work if a component is observed by multiple cameras?

You're absolutely right, this is quite puzzling.
First, it should probably be noted that the vast majority of self-respecting components do not care about the resize events. This is because a "normal" world continues its existence without any regard to whether it's observed through the camera or not. 

There could be interesting examples when you'd want to know whether an object is actively observed by the user or not -- either from an optimization standpoint or to implement some peculiar game mechanic. However, in this case, you'd be interested not only in viewport changes but also in any changes to the camera's target as well.

Viewport changes are obviously of interest to the HUD elements of the camera, but those elements won't be observed by another camera, so it's safe to call onResize for them.

The `ParallaxComponent` is an interesting use-case because, while being a part of the game world, it actually cares about how it is observed. But then, if the same parallax is being observed from two different cameras -- what do we do? I'm not sure what the right solution to parallax is. Maybe instead of relying on the resize event, it could always go into the camera and fetch the current world window size?

spydon: Maybe we should wait with having the camera as a component, until it can actually be used as a component?
But then just have your restructured camera directly on `Game`?

st-pasha: Ok, I made the `CameraComponent` to be a non-component. Now it can't be mounted into the component tree, but at the same time, it will be quick to restore its status as a `Component` when it has more fleshed-out functionality.

spydon: Great!

st-pasha:
Could this possibly be reviewed and/or merged?

packages/flame/lib/src/game/camera/camera_component.dart
```diff
@@ -0,0 +1,41 @@
+import 'dart:ui';
+import '../../../components.dart';
+import 'camera.dart';
+
+// TODO(st-pasha): extend from BaseComponent
+class CameraComponent {
```
spydon, line 6: This should be named something differently now when it is not a component. CameraController or something maybe?

st-pasha: As you remember, this class **is**, in fact, a `Component`, except that we temporarily orphaned it until we could figure out how it should interact with the `Game` and some other classes. Renaming it now will serve no purpose since we'd have to rename it back in the next PR.

spydon: Since we don't have a solution for how to make it interact with the component tree when it is a component itself we will most likely release `rc14` before that has been fixed (we are planning to do this soon after the "Game as a component" PR is merged), and I don't think we want classes named `*Component` released if they can't be added to the component tree.

st-pasha: Ok, fair enough. Renaming is easy anyway.

spydon (approved):
LGTM!

packages/flame/lib/src/game/camera/camera_component.dart
```diff
@@ -0,0 +1,49 @@
...
+
+/// Do not use.
+///
+/// This class encapsulates BaseGame's rendering functionality. It will be
+/// converted into a proper Component in a future release, but until then
+/// using it in any code other than from the BaseGame class is unsafe and
+/// therefore not recommended.
+///
+/// "Co" means unfinished Component.
+class CameraCo {
```
erickzanardo, line 13: I think I would prefer to have this as `CameraController` and remove the comments about not using this outside base game. In the past implementation, camera was possible to be used on the `Game` class, so this is basically the same.

Also I would name this file `camera_controller` instead of component.

st-pasha: Unfortunately, `CameraController` would not be a proper name for this class, since it doesn't actually control the camera. Currently, the logic for controlling the camera lives inside the `Camera` class itself.

The purpose of this class is that it's a Camera component (i.e. it enables the use of Camera class within the Flame Component System), except that we downgraded it from an actual Component into a not-quite-component for now, because it is not yet ready to be used just about anywhere a Component might be used.

erickzanardo: We could name it only camera then, like it was in the past. I understand that this will became a component later, but probably for the next release, it won't. So it would better have a proper name and file name and then the comment above wouldn't be needed anymore.

Then on a future change, it could easily be promoted again to a component

st-pasha: But we already have a class called `Camera`, so we can't call this one `Camera` too. I guess we could call it `Camera2`, or `CameraWrapper`, or?

wolfenrain: For sure not `Camera2`.. Getting android flashbacks here

spydon: CameraWrapper sounds good to me!

st-pasha: ok, renamed once again

erickzanardo (approved):

packages/flame/lib/src/game/base_game.dart
```diff
@@ -24,33 +23,20 @@ import 'viewport.dart';
 /// This is the recommended structure to use for most games.
 /// It is based on the Component system.
 class BaseGame extends Game {
+  BaseGame() {
+    components = createComponentSet();
+    _cameraComponent = CameraWrapper(Camera(), components);
```
wolfenrain, line 28: Should this property not reflect the class name for the time being, even if it is internal?

As far as I understood is that when #906 gets merged this property wont exist anyway?

wolfenrain (approved):

packages/flame/lib/src/game/camera/camera_wrapper.dart
```diff
@@ -0,0 +1,47 @@
...
+import '../../../components.dart';
+import 'camera.dart';
+
+/// Do not use.
+///
+/// This class encapsulates BaseGame's rendering functionality. It will be
+/// converted into a proper Component in a future release, but until then
+/// using it in any code other than the BaseGame class is unsafe and
+/// not recommended.
+class CameraWrapper {
```
luanpotter, line 11: how about CameraRenderer?

packages/flame/lib/src/game/camera/camera_wrapper.dart
```diff
@@ -0,0 +1,47 @@
+import 'dart:ui';
+import '../../../components.dart';
+import 'camera.dart';
+
+/// Do not use.
```
luanpotter, line 5: I don't think we need this comment, since we are still on RCs everything can be changed at any time

packages/flame/lib/src/game/camera/camera_wrapper.dart
spydon: Haha we have changed the name several times already, and it doesn't really matter, both since it is internal and since the new structure won't require this. :)

packages/flame/lib/src/game/camera/camera_wrapper.dart
```diff
@@ -0,0 +1,47 @@
...
+      camera.apply(canvas);
+      world.forEach((component) {
+        if (!component.isHud) {
+          canvas.save();
+          component.renderTree(canvas);
+          canvas.restore();
+        }
+      });
+      canvas.restore();
+      // Then render the HUD
```
luanpotter, line 37: this is breaking the behaviour, currently components are drawn in priority order, regardless of isHUD!

packages/flame/lib/src/game/camera/camera.dart
```diff
@@ -151,6 +161,18 @@ class Camera extends Projector {
     canvas.transform(_transformMatrix(position, zoom).storage);
   }
 
+  Vector2? _canvasSize;
+  Vector2 get canvasSize {
+    assert(_canvasSize != null,
+        'Property `canvasSize` cannot be accessed before the layout stage');
```
luanpotter, line 167:
```suggestion
    assert(
        _canvasSize != null,
        'Property `canvasSize` cannot be accessed before the layout stage',
    );
```

packages/flame/lib/src/game/camera/camera_wrapper.dart
spydon: Right, I guess this could actually break components like the Parallax.
Since the parallax would be rendered on top of all the other components.

packages/flame/lib/src/game/camera/viewport.dart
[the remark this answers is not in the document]
luanpotter: do you ever need to check for `viewport.canvasSize == null` though? otherwise it can be just a `late Vector2`, so you don't need the `!` everywhere (just assume it will be there always after layout)

packages/flame_forge2d/example/lib/widget_sample.dart
```diff
@@ -31,8 +31,9 @@ class WidgetSample extends Forge2DGame with TapDetector {
   Body createBody() {
     final bodyDef = BodyDef()
       ..angularVelocity = 3
-      ..position =
-          screenToWorld(Vector2.random()..multiply(viewport.effectiveSize))
+      ..position = screenToWorld(
+        Vector2.random()..multiply(camera.viewport.effectiveSize),
```
luanpotter, line 35: I wonder if we should expose this property, that seems to be used everywhere, move readily inside baseGame. like canvasSize or originalSize, alongside the `size` (that is transformed)

packages/flame/lib/src/game/camera/camera_wrapper.dart
spydon: Maybe HUD is actually a bad name for it? Since it isn't always really "heads-up".

luanpotter: yeah the definition would be ignoreCamera or something. note that it still applies viewport. because that is how HUD works. but I could see you wanting ignoreViewport as well... I guess isHUD is just a nice way to encapsulate it for most use cases

packages/flame/lib/src/game/camera/camera_wrapper.dart
luanpotter: sgtm!

packages/flame/lib/src/game/camera/camera_wrapper.dart
st-pasha: > currently components are drawn in priority order, regardless of isHUD!

That's true; but the current behavior is considered a bug (#934), so this PR actually fixes that bug. Should this perhaps be noted more explicitly somewhere?

spydon: I guess that #934 is invalid (my fault), we should clean that up to remove confusion.
Would we want all three, `ignoreCamera`, `ignoreViewport` and `isHud` (having `isHud` working like it is in this branch) maybe?

packages/flame_forge2d/example/lib/widget_sample.dart
st-pasha: So far this has been a mechanical rename in order not to break any existing behavior. However, going forward it might be useful to review the notion of various sizes and `viewport.effectiveSize` in particular. I am not so sure that its current ubiquitous use is necessarily warranted.

For one, there is `windowSize = viewport.effectiveSize / camera.zoom`, which is the size of the game world currently visible through the camera. This could be useful, for example, if you want to avoid rendering objects that are out of the camera's view. 

Then, there's `viewportSize`, which is the size of the viewport in its own coordinate frame. For a top-level camera, this would be in pixels. This property can be useful, for example, for laying out various HUD elements that are attached to the viewport.

I'm not sure about the scenarios where the original `effectiveSize` would be useful.

In any case, I would prefer to leave the complexities of various size definitions and where they should be used outside the scope of this PR.

packages/flame/lib/src/game/camera/camera_wrapper.dart
st-pasha: > Would we want all three, ignoreCamera, ignoreViewport and isHud maybe?

Ideally, we'd want neither. The semantics of these properties is too vague and error-prone. Consider the current `isHud` for example: at this moment this property works only for elements that are attached directly to the BaseGame class. If such elements appear at a more deep level in the components hierarchy, they would not be rendered correctly. This includes the current `ParallaxComponent` class too since apparently, it declares itself as `isHud`.

A better approach would be to model these properties via the components' place in the game tree:
```
Game
 +-- World (most of the in-game components go here)
 +-- Camera (views the World)
 |   +-- HUD (elements that are rendered on top of the camera's viewport)
 +-- Other (elements that are completely independent from the camera/viewport)
```

luanpotter: I agree with this in parts. I think having the camera as a component that can have children makes a lot of sense. That allow for multiple cameras and hud elements can be out of the camera. however even hud elements should be subject to viewport. so I think it would look like this:

```
game (has a viewport)
         + -- "hud" components (not transformed by any camera)
         + -- camera 1 comp
         |        + -- components transformed by camera1
         + -- camera 2 comp
                  + -- components transformed by camera2
```

I don't think it makes any sense for a game to have more than one viewport. I don't see how that would apply (the viewport draws constraints on the screen, doesn't just translate stuff inside it).
so having that in mind I am not sure unifying camera and viewport would work for this structure. wdyt?

luanpotter: cc @erickzanardo @spydon @wolfenrain @renancaraujo as I think this is an impactful decision

st-pasha: A viewport is a rectangle (or some other shape) on the screen where a camera renders the world. The viewport's primary function is to clip the world (which is probably too large to fit on a screen anyway). The camera's function, on the other hand, is to move the world within the viewport, i.e. to choose the point within the world that the camera is currently "looking at". As such, a camera and a viewport are 2 parts of a single entity, and one cannot operate without the other.

For example, I'm currently making a demo game where there is a world that can be viewed normally on a screen, but also the ability to use a magnifying glass to look at some parts of the world more closely. The magnifying glass here is a separate camera (with its own custom viewport), which can be dynamically rendered on top of the main world. The magnifying glass even has some "hud" elements: the frame and the handle. These are bound to the viewport (i.e. move together with the viewport) but are not actually clipped by the viewport.

spydon: How does that work when you are having the components affected by both cameras? Does it works because they are only seen through the last camera in the hierarchy?

Sounds to me like it make sense to have the viewport in the camera then. What about having the components that are affected by the camera as children to a camera component? I think we can disregard having two cameras on the same level with overlapping components for now. (The issue we talked about earlier)

st-pasha: Yes, there are some components whose rendering logic might depend on the settings of the camera, such as zoom level, or the position of the viewport. For example, GoogleMaps gradually removes smaller details of the landscape as you zoom out. The Parallax is another example, where you have a large image and want to render only that portion of it which is visible through the viewport. In cases like these, when there are two cameras, the elements will want to render differently for each camera. 

I can see two solutions here:
1. Add `Camera` argument to the `render()` method.
2. Add static variable `Camera.current` and then have each camera set/clear that variable during its render call.

The first approach is more robust, but also more disruptive to existing user code. The second approach is more conservative but is essentially a hack to pass down the camera via unorthodox means.

spydon: If the components are children of the camera, like in @luanpotter's tree, then we wouldn't need to do any of those solutions for `render` right?
We would have to make`update` only run once per component though, even though it can be a child of many `Camera`s.

st-pasha: If we have a world that is observed through multiple cameras at the same time, then we can't have the world to be mounted to all those cameras. While it is superficially possible to `addChild(world)` on multiple cameras, it would result in `world` not having a proper notion of a parent, and further unknown consequences caused by the fact that our component tree is no longer a *tree*. 

A more sane approach would be for the camera to *know* about the world that it observes (via a property, similarly how in this PR the camera knows about the `ComponentSet` of its parent), but the world wouldn't actually be *mounted* to the camera. That way we wouldn't have to worry about duplicate `update()` calls either.

Note also that if we have a component within the world that cares about being observed, and there are multiple cameras looking at the world, then ultimately the component will have to be rendered differently even within the single render cycle.

spydon:
We've decided to merge this and I opened discussion #950 where we can talk about how to follow-up on this (it will probably be for RC15 though, since we want to get out rc14 soon).

PR #938, opened 2021-09-06
Default sprite component size to sprite size

luanpotter:
I noticed that for some reason, if we don't provide an explicit `size` for a `SpriteComponent`, it fallbacks to null. I think there is no reason not to fallback to the `Sprite` original size, if present. Thoughts?


spydon (approved):
Lgtm, just missing changelog entry

erickzanardo (commented):
LGTM if we write a test for this 😉

spydon (approved):

erickzanardo (approved):

PR #941, opened 2021-09-08
More improvements for TextBoxComponent

luanpotter:
# Description

This is a 2-in-1, but I think it's succint enough for easy review.

1. Small tweaks to TextBoxComponent
    - onLoad now waits for the first cache, allowing you to `await add` this component if desired
    - customizable pixelRatio in case you will further enlarge the rendered TBC
2. Small tweaks to testing infrastructure to support testing changes on TextBoxComponent
    - support drawImage on MockCanvas
    - add a TextBoxComponent
    - add a new `ContainsAtLeastMockCanvas` to facilitate testing with `MockCanvas` when you don't care for the whole sequence of commands (I suggested this on a previous PR)

## Checklist

## Breaking Change


packages/flame/lib/src/components/text_box_component.dart
```diff
@@ -74,9 +75,11 @@ class TextBoxComponent<T extends TextRenderer> extends PositionComponent {
     TextBoxConfig? boxConfig,
     Vector2? position,
     int? priority,
+    double? pixelRatio,
   })  : _text = text,
         _boxConfig = boxConfig ?? TextBoxConfig(),
         _textRenderer = textRenderer ?? TextRenderer.createDefault<T>(),
+        pixelRatio = pixelRatio ?? window.devicePixelRatio,
```
luanpotter, line 82: this also avoids repeating `window.devicePixelRatio` twice

packages/flame/lib/src/components/text_box_component.dart
```diff
@@ -74,9 +75,11 @@ class TextBoxComponent<T extends TextRenderer> extends PositionComponent {
     TextBoxConfig? boxConfig,
     Vector2? position,
     int? priority,
+    double? pixelRatio,
```
luanpotter, line 78: sadly default values have to be const in dart :/

packages/flame/lib/src/components/text_box_component.dart
```diff
@@ -210,17 +216,18 @@ class TextBoxComponent<T extends TextRenderer> extends PositionComponent {
     _textRenderer.render(c, line, Vector2(_boxConfig.margins.left, dy));
   }
 
-  void redrawLater() async {
-    _cache = await _redrawCache();
+  Future<void> redraw() async {
+    final newSize = _recomputeSize();
```
luanpotter, line 220: I noticed before we updated the size before actually finishing redrawing, I figure this is slightly better like so though it hardly makes any practical difference

packages/flame/lib/src/components/text_box_component.dart
```diff
@@ -210,17 +216,18 @@ class TextBoxComponent<T extends TextRenderer> extends PositionComponent {
...
   }
 
   @override
   void update(double dt) {
     super.update(dt);
     _lifeTime += dt;
     if (_previousChar != currentChar) {
-      size = _recomputeSize();
-      redrawLater();
+      redraw();
```
luanpotter, line 230: I drawing was too expensive I considered that these futures could potentially race. however (1) testing several examples I never saw that happening, (2) seems very unlikely for reasonable typing speed vs rendering speeds, and (3) would be extremely convoluted to keep track of futures and have them cancelled. but if you think it's important I can address that on a followup.

packages/flame/lib/src/test_helpers/canvas_commands/image_command.dart
```diff
@@ -0,0 +1,20 @@
+import 'dart:ui';
+import 'command.dart';
+
+/// canvas.drawImage()
```
luanpotter, line 4: I don't compare the actual image, see comment on the `mockCanvas.drawImage` impl

packages/flame/lib/src/test_helpers/mock_canvas.dart
```diff
@@ -115,6 +118,34 @@ class MockCanvas extends Fake implements Canvas, Matcher {
     return mismatchDescription.add(matchState['description'] as String);
   }
 
+  bool containsAtLeast(covariant MockCanvas other, Map matchState) {
```
luanpotter, line 121: since everything was private, I figured having this here was the easiest way to implement a different Matcher using simple inheritance, rather than composition. but I am not super familiar on how to re-use and combine these Matchers, so open to suggestions for other implementations

luanpotter:
cc @st-pasha can't add you as reviewer for some reason, but you might wanna take a look at the testing part

spydon (approved):

packages/flame/lib/src/test_helpers/mock_canvas.dart
```diff
@@ -115,6 +118,34 @@ class MockCanvas extends Fake implements Canvas, Matcher {
...
+    if (_saveCount != 0) {
+      return _fail('Canvas finished with saveCount=$_saveCount', matchState);
+    }
+    final useTolerance = max(tolerance, other.tolerance);
+    final remainingExpectedCommands = other._commands.toList();
+    for (final expectedCommand in _commands) {
```
st-pasha, line 127: If I'm not mistaken, `this._commands` contains "actual" commands, not expected ones. Well, either that or the other way around, but they can't be both expected :)

erickzanardo (approved):

packages/flame/lib/src/test_helpers/contains_at_least_mock_canvas.dart
```diff
@@ -0,0 +1,12 @@
+import 'mock_canvas.dart';
+
+/// Similar to [MockCanvas], but only asserts that the provided list of expected
+/// commands at least appear in the actual canvas.
+/// Additional commands that the actual Canvas might have are ignored by this
+/// Matcher.
+class ContainsAtLeastMockCanvas extends MockCanvas {
```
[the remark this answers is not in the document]
luanpotter: that works too, I like both constructors, let me do that

packages/flame/lib/src/test_helpers/mock_canvas.dart
```diff
@@ -115,6 +118,34 @@ class MockCanvas extends Fake implements Canvas, Matcher {
...
+        return expectedCommand.equals(cmd);
+      });
+      if (idx == -1) {
+        return _fail(
+          'Expected canvas command not found: $expectedCommand. '
+          'Actual commands: $_commands',
+          matchState,
+        );
+      } else {
+        remainingExpectedCommands.removeAt(idx);
```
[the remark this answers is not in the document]
luanpotter: doing what it does was my original intent but in the case I use turns out both would work. I'm not sure which is more useful in the long run, could even add another parameter to choose if the order matters or not. (so we end up with 2x2=4 options for full flexibility)

packages/flame/lib/src/test_helpers/mock_canvas.dart
luanpotter: yes my bad! I got confused during it and forgot to rename one of the instances

packages/flame/lib/src/test_helpers/mock_canvas.dart
```diff
@@ -115,6 +118,34 @@ class MockCanvas extends Fake implements Canvas, Matcher {
...
+    final remainingExpectedCommands = other._commands.toList();
+    for (final expectedCommand in _commands) {
+      final idx = remainingExpectedCommands.indexWhere((cmd) {
+        if (expectedCommand.runtimeType != cmd.runtimeType) {
+          return false;
+        }
+
+        expectedCommand.tolerance = cmd.tolerance = useTolerance;
+        return expectedCommand.equals(cmd);
+      });
```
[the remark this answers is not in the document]
luanpotter: that might not work too well if you are expecting multiple equal commands though right? bc since they can be in any order I don't want to end up matching them by type in a way that then leads to an error where if I happen to choose a better matching by content it would actually work out

luanpotter:
I made mockCanvas to take an enum to express its configuration. we can add other options (like containsExactOrder, etc) on followups as needed

Issue #960, opened 2021-09-17
Multi-version documentation


st-pasha:
## What could be improved

Currently, the main documentation site ([flame-engine.org](flame-engine.org)) shows only documentation for the latest release candidate. However, it would greatly benefit the users if we were able to show documentation for multiple versions of the library:
  - last stable releast (0.29.4);
  - latest release candidate (1.0.0.rc14)
  - previous release candidate(s) (1.0.0.rc13, ...)
  - current version in development (`main` branch)
  - not public, but still available somewhere: documentation for every current PR

The purpose here is to allow users to read the documentation for the specific version that they are using. Too often we see complaints that the documentation does not reflect the code.

## How

There could be several potential approaches here, but IMHO the simplest one is to use the external documentation hosting platform [readthedocs.io](readthedocs.io). Its features are the following:
  - Free hosting for open source projects;
  - Support for building and hosting multiple versions of the project at the same time;
  - Building from PR branches;
  - Fully customizable build process.

Here's how it would work in practice:
  - Locally, run `pip install sphinx` and then inside the `docs` folder: `sphinx-quickstart`;
  - This will create a `Makefile` and the configuration file `conf.py`;
  - (optional) run `make html` to verify that documentation can be generated properly, add the build directory to `.gitignore`. The resulting documentation won't look very nice, but that's ok for the initial draft;
  - create an account on `readthedocs.io` and point it to our GitHub repo.

That's it for the initial version. Later we can tweak the theme/style to make the generated docs look exactly like we want. 

## More information

- Readthedocs can be configured to serve at the `flame-engine.org` address;
- Readthedocs uses python `sphinx` library in order to convert markdown doc files into html.
  - PRO: the process is highly customizable - you can quite literally modify every single step in the generation pipeline, for example, if we want to create extensions for the Markdown syntax;
  - CON: Python language is not primary for Flame, so it may not be that easy.
- Local development is very easy: `make html` and it will build the entire documentation site locally.
- This proposal is relatively risk-free: it only involves adding 2 (or 3) new files without modifying any of the existing files. The site can be kept parallel to the existing one while we evaluate its feasibility.
- In order to enable readthedocs to build documentation for the past versions of Flame, those versions must have the same configuration file `conf.py`. Thus, once we finalize the settings in conf.py, we'd want to push that file to each of the released branches that we plan to document.

spydon:
Sounds good to me, especially since @renancaraujo has given it a thumbs up.
[...]

PR #966, opened 2021-09-20
Sphinx docs

st-pasha:
# Description

Prepare documentation folder to make it build-able with Sphinx (see #960 for reasons why).

I tried to modify one of the original sphinx themes to use a dark background. The results of how it currently looks like are shown in the screenshot below:
Screen Shot 2021-09-20 at 1 54 47 AM

In order to build the docs locally, you'd need to first run `pip install -r requirements.txt`, and then `make html`.

## Checklist

## Breaking Change

## Related Issues

WIP for #960


spydon (approved):

doc/conf.py
```diff
@@ -0,0 +1,73 @@
+# Configuration file for the Sphinx documentation builder.
```
luanpotter, line 1: would it be possible to move any build related files into a specific folder? to separate them from the actual markdown

luanpotter:
Great idea! We do need to find a way to wire it up for the website deploy

doc/conf.py
st-pasha: I've put all the config files into the `doc/_conf/` folder.
It works fine locally: you just need to be in that folder when running `make html` command.

doc/README.md
```diff
@@ -53,3 +53,48 @@ You can also check out the
 [awesome flame repository](https://github.com/flame-engine/awesome-flame#articles--tutorials),
 It contains quite a lot of good tutorials and articles written by the community to get you started
 with Flame.
+
+```{toctree}
```
luanpotter, line 57: This breaks it on github view: 
image

I wonder if we can find a way to make it work both ways. or at least omit it on github

.readthedocs.yaml
```diff
@@ -0,0 +1,14 @@
+# [Read the Docs] configuration file
```
luanpotter, line 1: can this also be inside the _config folder? or does that break anything?

luanpotter (commented):
I think we can merge this and followup on make sure the deploy works later. I just have a couple final comments I left on the PR, plus:

* should we use flame icon & logo instead of Blue Fire? since this is flame-related doc?
* can the static folder be inside the config folder too? (not sure if possible)

.readthedocs.yaml
st-pasha: This file can only be in the repository root, or ReadTheDocs won't be able to find it. The only other option is to not have this file at all, and instead set up all configuration options through their web interface. But that's considered more fragile, so [not recommended](https://docs.readthedocs.io/en/stable/config-file/index.html).

doc/README.md
st-pasha: I've put the table of contents into a separate file. This way the readme can stay exactly the way it is now, but there is also a new `index.md` file, and new `toc.rst` file.

st-pasha:
> should we use flame icon & logo instead of Blue Fire? since this is flame-related doc?

I thought the change to Blue Fire / blue color was across the board. But if you're saying Flame remains orange/red as it was before, then it's my misunderstanding. I've changed the logo and favicon to use the correct colors now.

st-pasha:
> can the static folder be inside the config folder too? (not sure if possible)

The folder can be renamed/moved, no problem. However, the purpose of this folder is not to provide build configuration, but rather it's a store for static content that will end up being copied into the resulting HTML website. This typically includes things like stylesheets, images, scripts, etc. It is pretty much equivalent to the `assets` folder in Android/iOS project. Perhaps we should rename it as such?

It would probably make sense to move the current `images` folder inside that folder -- but that's better to do as a separate PR.

st-pasha:
After all the tweaks, the doc now renders like this:

Screen Shot 2021-09-29 at 11 20 37 PM

spydon:
> After all the tweaks, the doc now renders like this:
> 
> Screen Shot 2021-09-29 at 11 20 37 PM
> 

Very nice!!

luanpotter:
> the purpose of this folder is not to provide build configuration

I was thinking more of a way to separate any and all "build" files from the markdowns, so someone browsing on github doesn't get lost reading some config file or css. but maybe the name isn't the best then, how about:

```
+ /doc
|--+ /sphinx
|   |-- /config
|   |-- /static
|-- actual md files
```

woudl that work?

luanpotter (approved):
LGTM, awesome work, thanks @st-pasha

st-pasha:
@luanpotter  We could make nested folders like you're suggesting. The only question then is what do we do with `images/`? Conceptually, they're part of the static assets, so should go into that folder. But at the same time, they are very close to being the "content", and in fact, aren't really sphinx-specific. I guess we could have a rule that those images that are embedded into the page(s) will go into the root `images/` folder, while those images/assets that are needed specifically for sphinx would go into the `sphinx/images` folder? It is possible that at some point we'd want to add custom fonts (if Arial is too boring), or scripts (for example in order to embed dart samples into the page) -- but these won't be renderable on GitHub anyway.

Btw, sphinx uses `_` in front of a name to signify "special" folders, resembling the convention for private variables in a class. Do we want to stick to this convention (so it'd be `_sphinx/` instead `sphinx/`)?

erickzanardo (approved):

luanpotter:
@st-pasha totally agree with your distinction between two kinds of images folders, one for actual content and one for sphinx related stuff!

for the `_`, if it's a normal sphinx convention, I am ok with naming it `_sphinx`!

luanpotter (approved):
LGTM

PR #971, opened 2021-09-22
Upgrade and move flame_tiled to the monorepo

luanpotter:
image

The only thing not working are the flips. But I believe that has been broken for a while due to the sprite batch migration.


spydon (approved):
Lgtm!

luanpotter:
I will merge this for now and we can fix the flips problem later (cc @wolfenrain )

Issue #973, opened 2021-09-22
Perfomance gets very bad after multiple components have been added to the game


erickzanardo:
# Current bug behaviour
Flame perfomance gets bad after multiple components are added (even if they are removed)

# Expected behaviour
Flame perfomance should be good, even after multiple components have been added/removed

# Steps to reproduce
https://github.com/flame-engine/flame_rc14_perfomance_issue

clone and run the following repo, after a few seconds shooting the FPS goes very low

# Flutter doctor output

```
[✓] Flutter (Channel stable, 2.5.1, on macOS 11.5.1 20G80 darwin-arm, locale en-GB)
    • Flutter version 2.5.1 at /Users/erickzanardo/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision ffb2ecea52 (5 days ago), 2021-09-17 15:26:33 -0400
    • Engine revision b3af521a05
    • Dart version 2.14.2

[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
    • Android SDK at /Users/erickzanardo/Library/Android/sdk
    • Platform android-31, build-tools 31.0.0
    • ANDROID_HOME = /Users/erickzanardo/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.8+10-b944.6916264)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 12.5.1, Build version 12E507
    • CocoaPods version 1.10.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 4.2)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.8+10-b944.6916264)

[✓] VS Code (version 1.60.1)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.26.0

[✓] Connected device (3 available)
[...]

PR #974, opened 2021-09-23
Make onLoad mandatory

luanpotter:
This fixes https://github.com/flame-engine/flame/issues/973

The issue was actually a compound of two things, neither in itself a bug with Flame but together they led the implementation provided to have issues.

1. The first issue is the Collidables had it's own list of collidables to keep track. There is no reason to do that now with QOS. That is super brittle because if remove hooks are not setup properly (which is what happened), that will keep an ever growing list of components (memory leak). And worse, those components will be kept being checked n*n-1 for collisions. I changed it to use `query` for a simpler and less brittle implementation. The list was cleared though, so this alone was not the issue.

2. The second "issue" is that onLoad was not marked as mustCallSuper as of the refactor. That caused @erickzanardo to forget to call super.onLoad on the sample game, which in turn caused collidables not to be cleared. Again, not an issue on itself, but definitely brittle. So I added mustCallSuper to onLoad.

3. A third unrelated issue I noticed is that the has_collidables mixin, even though it is a Game mixin, was put in the components folder. So I took the opportunity to fix that.

With these fixes, I believe now it's not easy to fall into this trap again. Also the trap no longer exists to begin with (by using QOS instead of keepint track of lists).

I tested this against an updated branch of erick's example and the collisions seem to be working. @spydon lmk if there was any reason for this impl or if it was just a relic for when we didn't have QOS.

PR for the example side: https://github.com/flame-engine/flame_rc14_perfomance_issue/pull/1


packages/flame/lib/src/game/mixins/has_collidables.dart
```diff
@@ -29,6 +19,6 @@ mixin HasCollidables on FlameGame {
   }
 
   void handleCollidables() {
-    collisionDetection(collidables);
+    collisionDetection(children.query<Collidable>());
```
st-pasha, line 22: Isn't the issue here that we need to be able to handle collisions between components that are in separate parts of the component tree? `children.query` will not traverse the component tree, primarily because it doesn't know that components are arranged into a tree; but even if it did know that -- it would still be inefficient to re-create the list of collidables on every tick.

erickzanardo: I think Pasha has a good point here, I am almost sure that if we remove this internal tracking of components here, we will undo a task that @spydon recently did, so the component children are considered on the collisions detections.

Maybe for this fix, we should just keep the addition of the must call super on onLoad and revert this change?

packages/flame/lib/src/components/mixins/collidable.dart
```diff
@@ -19,15 +19,6 @@ mixin Collidable on Hitbox {
 
   void onCollision(Set<Vector2> intersectionPoints, Collidable other) {}
   void onCollisionEnd(Collidable other) {}
-
-  @override
-  void onRemove() {
-    final parentGame = findParent<FlameGame>();
-    if (parentGame is HasCollidables) {
-      parentGame.collidables.remove(this);
```
spydon, line 27: So how is the Collidable removed from the list now?

spydon (changes requested):

packages/flame/lib/src/game/mixins/has_collidables.dart
luanpotter: ah, I see, didn't think of that case. but I feel that normally we do these things using `propagateToChildren`, instead of keeping track of ad-hoc lists, right?

packages/flame/lib/src/components/mixins/collidable.dart
luanpotter: I removed this list entirely

packages/flame/lib/src/game/mixins/has_collidables.dart
spydon: > ah, I see, didn't think of that case. but I feel that normally we do these things using `propagateToChildren`, instead of keeping track of ad-hoc lists, right?

That will be less performant, and we have done it like that on the things where the parent covers the child, but apparently people don't necessarily use the Collidables like that (from the issues we have seen on discord). The best thing would be if we had a method to query a tree like we have the other flat query, then we could also use that for gestures too.

luanpotter: that would be a great addition I think! for now I think I will bring back the collidables and just make sure remove is called properly; we can iterate later on a generic recursive `query` method

st-pasha: `propagateToChildren` would not be ideal here, since we need to iterate over every pair of elements in the list.

luanpotter:
Simplified this to just make onLoad mandatory for now. We can re-think how collidables work later.

spydon (approved):

erickzanardo (approved):
LGTM

PR #975, opened 2021-09-23
Fix PositionComponent.toRect()

### LINKED ISSUE #970: flipping components - underlying rect on wrong position
# Current bug behaviour

The flipped component has its .toRect() not on the visual sprite but to the right of it. This bug manifests in many instances and causes weird behaviors of components jumping around and/or touch inputs not being passed correctly.

# Expected behaviour

component.toRect() should be at the same location as the visual sprite.

# Steps to reproduce

use a SpriteComponent and give it anchor.center
flip it with .flipHorizontally()

as expected the sprite component is flipped in the centerline but the underlying .toRect is not around the visual sprite but to the right of it. 

In one of my cases I for example used the mixin TapDetector and in the onTapDown method I used 

```
if (component
        .toRect()
        .contains(info.eventPosition.global.toOffset())) {...}
```

This condition becomes true if touched to the right of the visual sprite component on the screen. Its false if touched directly!

Flame version: 1.0.0-releasecandidate.14
Platform affected: android

st-pasha:
# Description

Method `.toRect()` in `PositionComponent` now always returns the bounding rectangle of the component. Specifically, this method will now behave correctly even if the component was flipped, scaled, or rotated. When the component is rotated, we still return the smallest axis-aligned rectangle that is needed to circumscribe the component. This bounding rectangle will usually be larger in size than the dimensions of the component.

The method `.toAbsoluteRect()` is handled similarly.

## Checklist

## Breaking Change

## Related Issues

Closes #970


packages/flame/lib/src/components/position_component.dart
```diff
@@ -332,17 +332,45 @@ class PositionComponent extends Component {
...
+      final xs = [topLeft.x, topRight.x, bottomLeft.x, bottomRight.x]..sort();
+      final ys = [topLeft.y, topRight.y, bottomLeft.y, bottomRight.y]..sort();
+      return Rect.fromLTRB(xs.first, ys.first, xs.last, ys.last);
+    }
+  }
 
-  /// Mutates position and size using the provided [rect] as basis.
-  /// This is a relative rect, same definition that [toRect] use
-  /// (therefore both methods are compatible, i.e. setByRect ∘ toRect = identity).
+  // FIXME
```
erickzanardo, line 373: What is missing here?

packages/flame/lib/src/components/position_component.dart
```diff
@@ -332,17 +332,45 @@ class PositionComponent extends Component {
...
+  }
 
-  /// Returns the absolute position/size of this component.
-  /// Absolute because it takes any possible parent position into consideration.
-  Rect toAbsoluteRect() => absoluteTopLeftPosition.toPositionedRect(scaledSize);
+  /// The bounding rectangle of the component in global coordinate space.
+  ///
+  /// This is similar to [toRect()], except the rectangle is projected into the
+  /// outermost coordinate frame.
+  Rect toAbsoluteRect() {
```
erickzanardo, line 359: This method and the other one are super similar, we should abstract it to an internal method so we don't repeat ourselves

luanpotter: +1. the "find axis aligned bounding box" bit could even be a generic method that we provide

packages/flame/lib/src/components/position_component.dart
st-pasha: 1. It is not the inverse of `toRect()`: what if the component is rotated and/or scaled?
2. Do we want to reset the angle here?
3. If the component is flipped, do we want to unflip it, or keep as-is?
4. Changing the `size` of the component is questionable: it does not affect the visible size (unless the component uses its `size` property to render itself, which is not recommended), so do we want to modify the `scale` instead?

Generally, I'm not sure what the purpose of this method is and how it is expected to be used, so don't know in which direction to take it.

packages/flame/lib/src/components/position_component.dart
st-pasha: ok

packages/flame/lib/src/components/position_component.dart
erickzanardo: @luanpotter can you answser @st-pasha here?

Either we should not replace the original docs of the method for a FIXME. We could

 - If this is indeed wrong, put a FIXME comment with why it is broken, but leave the original dartdoc on the method
 - Fix the implementation and update the dartdoc if needed

st-pasha: Ok, I restored the comment to how it was before. The concerns about its functionality could be posted as a new issue instead.

st-pasha: Filed as #978

spydon (approved):
Lgtm!

wolfenrain (approved):
LGTM

PR #979, opened 2021-09-25
No need to pre-register types in ComponentSet

st-pasha:
# Description

It is no longer necessary to call `ComponentSet.register<T>()` before being able to use `ComponentSet.query<T>()`. The list of components matching `T` will be computed on demand. Note that this does not result in performance degradation, since the amount of work needed to construct the filtered list is the same in both cases. In fact, this approach provides a minor performance benefit in case some of the elements are removed from the ComponentSet before querying, or some type gets queried only on a fraction of the ComponentSets.

## Checklist

## Breaking Change

### DIFF
--- a/packages/flame/lib/src/components/component_set.dart
+++ b/packages/flame/lib/src/components/component_set.dart
@@ -138,8 +138,7 @@ class ComponentSet extends QueryableOrderedSet<Component> {
   /// This is equivalent to `List.whereType<C>()`, except the result of this
   /// operation is memoized for each distinct type [C]. The type [C] may be
   /// "registered" in advance using `register<C>()`, or it may not, in which
-  /// case the type will be registered at the time of the first query. There
-  /// is no performance difference between these two approaches.
+  /// case the type will be registered at the time of the first query.
   @override
   List<C> query<C extends Component>() {
     if (!isRegistered<C>()) {
--- a/packages/flame/lib/src/components/component_set.dart
+++ b/packages/flame/lib/src/components/component_set.dart
@@ -133,6 +133,21 @@ class ComponentSet extends QueryableOrderedSet<Component> {
     _removeLater.addAll(this);
   }
 
+  /// Return sublist of all components of type [C].
+  ///
+  /// This is equivalent to `List.whereType<C>()`, except the result of this
+  /// operation is memoized for each distinct type [C]. The type [C] may be
+  /// "registered" in advance using `register<C>()`, or it may not, in which
+  /// case the type will be registered at the time of the first query. There
+  /// is no performance difference between these two approaches.
+  @override
+  List<C> query<C extends Component>() {
+    if (!isRegistered<C>()) {
+      super.register<C>();
+    }
+    return super.query<C>();
+  }
+
   /// Materializes the component list in reversed order.
   Iterable<Component> reversed() {
     return toList().reversed;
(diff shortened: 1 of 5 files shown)


spydon (commented):
Shouldn't this change be directly on OrderedSet instead? And it could be a slight performance degradation for the tick where you run the query if you have added a lot of components before, instead of having the query registered from the start?

luanpotter:
Yeah the idea around pre-registering is that you are forced to keep track from the start and pay no cost during the actual game execution

st-pasha:
On every game tick you're going through the list of all components, and invoke `update()` and `render()` on each -- both containing potentially complicated logic. Compared to that, simply checking the type of an object adds such a minuscule cost (within a single game frame) that it would unlikely be measurable, let alone noticeable.

spydon:
> On every game tick you're going through the list of all components, and invoke `update()` and `render()` on each -- both containing potentially complicated logic. Compared to that, simply checking the type of an object adds such a minuscule cost (within a single game frame) that it would unlikely be measurable, let alone noticeable.

But this will not be done within that loop, if I'm not mistaken. So the whole tree needs to be traversed again? I think that could be noticable in some cases.

st-pasha:
> So the whole tree needs to be traversed again? I think that could be noticable in some cases.

We're currently performing the same or greater amount of work on every tick when we traverse the tree in the reverse order (i.e. serializing the tree into a list and then reversing that list). While I do believe that reverse traversal should be optimized, because it runs every tick, this example does show that an extra tree traversal is not noticeable even when it runs on every tick, and even less so when it runs for a single tick.

From a more practical standpoint, this PR makes it so that `query<>` does not cause the game to crash. Crashing is definitely worse than extra <1% execution time in a single tick.

It also removes `register<Collidable>` because collidables keep their own flattened list of components, and do not rely on `query`.

spydon:
> > So the whole tree needs to be traversed again? I think that could be noticable in some cases.
> 
> We're currently performing the same or greater amount of work on every tick when we traverse the tree in the reverse order (i.e. serializing the tree into a list and then reversing that list). While I do believe that reverse traversal should be optimized, because it runs every tick, this example does show that an extra tree traversal is not noticeable even when it runs on every tick, and even less so when it runs for a single tick.
> 
> From a more practical standpoint, this PR makes it so that `query<>` does not cause the game to crash. Crashing is definitely worse than extra <1% execution time in a single tick.
> 
> It also removes `register<Collidable>` because collidables keep their own flattened list of components, and do not rely on `query`.

Good point, approved

spydon (approved):
Just a changelog entry missing

packages/flame/lib/src/components/component_set.dart
```diff
@@ -133,6 +133,21 @@ class ComponentSet extends QueryableOrderedSet<Component> {
     _removeLater.addAll(this);
   }
 
+  /// Return sublist of all components of type [C].
+  ///
+  /// This is equivalent to `List.whereType<C>()`, except the result of this
+  /// operation is memoized for each distinct type [C]. The type [C] may be
+  /// "registered" in advance using `register<C>()`, or it may not, in which
+  /// case the type will be registered at the time of the first query. There
+  /// is no performance difference between these two approaches.
```
spydon, line 142: The last part of this comment should be changed or removed since it isn't true.

st-pasha: What I meant is that the total amount of CPU time spent on computing the cached list will be the same in both cases. Hence, neither of the approaches can be claimed to be faster or slower than the other.

spydon: That isn't necessarily true either, since a tree with many different types of components will take longer to go through if the query isn't registered first.

st-pasha: ok, I removed that sentence then.

wolfenrain (approved):

PR #986, opened 2021-09-30
Add flame_flare

luanpotter:
Note: this has to run with `--no-sound-null-safety`, bc flare doesn't support NS.

image


erickzanardo (approved):

packages/flame_flare/lib/flame_flare.dart
```diff
@@ -0,0 +1,24 @@
...
+import 'package:flare_flutter/asset_provider.dart';
+import 'package:flare_flutter/flare_actor.dart';
+import 'package:flare_flutter/flare_controller.dart';
+import 'package:flare_flutter/provider/asset_flare.dart';
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/rendering.dart';
+
+part 'src/flare_actor_component.dart';
+part 'src/flare_animation.dart';
+part 'src/flare_particle.dart';
```
spydon, line 24: `part` is not recommended to use right?

luanpotter: that is how it was before, I will take a look how other more updated packages do

PR #992, opened 2021-10-03
Moving mixin assertions to their own prepare method

erickzanardo:
# Description

As we noticed on the previous discussions, dart can handle multiple inheritance, so we probably don't need the `prepareComponent` method on the `FlameGame` class.

This PR is a first step for the removal of that method, it moves all the assertions that were done on that method to the mixins that the assertions are related to.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/lib/src/components/mixins/collidable.dart
+++ b/packages/flame/lib/src/components/mixins/collidable.dart
@@ -1,3 +1,5 @@
+import 'package:meta/meta.dart';
+
 import '../../../components.dart';
 import '../../../game.dart';
 import '../../components/position_component.dart';
@@ -27,6 +29,21 @@ mixin Collidable on Hitbox {
       parentGame.collidables.remove(this);
     }
     super.onRemove();
+  }
+
+  @override
+  @mustCallSuper
+  void prepare(Component parent) {
+    super.prepare(parent);
+
+    if (isPrepared) {
+      final parentGame = findParent<FlameGame>();
+      assert(
+        parentGame is HasCollidables,
+        'You can only use the Hitbox/Collidable feature with games that has '
+        'the HasCollidables mixin',
+      );
+    }
   }
 }
 
--- a/packages/flame/test/components/mixins/has_collidable_test.dart
+++ b/packages/flame/test/components/mixins/has_collidable_test.dart
@@ -1,0 +1,23 @@
+import 'package:flame/components.dart';
+import 'package:flame/game.dart';
+import 'package:flame_test/flame_test.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+class NonCollidableGame extends FlameGame {}
+
+class MyCollidable extends PositionComponent with Hitbox, Collidable {}
+
+void main() {
+  group('HasCollidables', () {
+    flameTest<NonCollidableGame>(
+      "can't add collidables to a game without HasCollidables",
+      createGame: () => NonCollidableGame(),
+      verify: (game) {
+        expect(
+          () => game.add(MyCollidable()),
+          throwsAssertionError,
+        );
+      },
+    );
+  });
+}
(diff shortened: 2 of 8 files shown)


packages/flame/lib/src/components/mixins/collidable.dart
```diff
@@ -28,6 +30,21 @@ mixin Collidable on Hitbox {
     }
     super.onRemove();
   }
+
+  @override
+  @mustCallSuper
+  void prepare(Component parent) {
+    super.prepare(parent);
+
+    if (isPrepared) {
```
luanpotter, line 39: so if I am understanding this, if isPrepared is false it means this is being add as a child from a component that has not yet been added to any game.

and then later when that component is added to a game, it and its child (this) will be prepared again.

is that right? if so I think having a test for this case is important as it's non trivial but seems easy to test and make sure it's working

packages/flame/test/components/mixins/has_collidable_test.dart
```diff
@@ -0,0 +1,23 @@
...
+import 'package:flutter_test/flutter_test.dart';
+
+class NonCollidableGame extends FlameGame {}
+
+class MyCollidable extends PositionComponent with Hitbox, Collidable {}
+
+void main() {
+  group('HasCollidables', () {
+    flameTest<NonCollidableGame>(
+      "can't add collidables to a game without HasCollidables",
```
luanpotter, line 13: single quotes...

luanpotter: ah I see why

packages/flame/test/game/base_game_test.dart
```diff
@@ -223,4 +223,14 @@ void main() {
...
     expect(game.children.isEmpty, equals(true));
   });
+
+  test("can't add a component to a game that don't have layout yet", () {
+    final game = MyGame();
+    final component = MyComponent();
+
+    expect(
+      () => game.add(component),
+      throwsAssertionError,
```
luanpotter, line 233: should we assert the actual message on each test? since we have 3 different asserts, by just checking this you might not be getting the assert you think you are

luanpotter (approved):
LGTM with some comments, NICE cleanup!

wolfenrain (changes requested):
Missing CHANGELOG entry :-)

packages/flame/lib/src/components/mixins/collidable.dart
erickzanardo: Yeah, that is what I understood from my investigation too

packages/flame/lib/src/components/mixins/collidable.dart
```diff
@@ -28,6 +30,21 @@ mixin Collidable on Hitbox {
     }
     super.onRemove();
   }
+
+  @override
+  @mustCallSuper
+  void prepare(Component parent) {
```
st-pasha, line 36: All of these `prepare()` methods could instead be `onLoad()`s, which run when the component is already added to the tree.

erickzanardo: Unfortunately not, onLoad runs only once and it will be harder for the user to catch the error as it runs later in the lifecycle and is async

st-pasha: `onLoad` runs either immediately after `prepare()`, or on the next tick. How does it make any difference when the assertion is thrown?

erickzanardo: Hmm, maybe you are correct, add is already async so it wouldn't change much for that matter.

But still, we have the problem that `onLoad` is executed only once during the component existence, and we should run those checking everytime the component would be added to a game or new parent.

st-pasha: Ah, so you mean a situation like a component is added to a game, then removed, then added once again to another game -- but this time the assertion won't be valid anymore?
This is a valid use-case, but then wouldn't the `onMount()` method be the proper place for such an assert?

Issue #991, opened 2021-10-03
Decide on exact linting rule for `prefer_mixin`


luanpotter:
We have a linter rule commented out on some lint configurations files:

```yaml
 #    - prefer_mixin 
```

We should study what it does and what it's for and if we wanna use it or not, and either have it enforced or not at all. Commented out code is a recipe for rot.

st-pasha:
Here are some notes on this topic: https://github.com/dart-lang/language/blob/master/accepted/2.1/super-mixins/feature-specification.md

PR #993, opened 2021-10-04
Unify all lint rules to be consistent across packages

luanpotter:
# Description

Unify all lint rules to be consistent across packages 

## Checklist

## Breaking Change

## Related Issues
https://github.com/flame-engine/flame/issues/991

### DIFF
--- a/packages/flame/lib/src/game/game_render_box.dart
+++ b/packages/flame/lib/src/game/game_render_box.dart
@@ -9,7 +9,6 @@ import '../extensions/size.dart';
 import 'game_loop.dart';
 import 'mixins/game.dart';
 
-// ignore: prefer_mixin
 class GameRenderBox extends RenderBox with WidgetsBindingObserver {
   BuildContext buildContext;
   Game game;


luanpotter:
Found some context here: https://github.com/flutter/flutter/issues/40808

spydon:
Seems like it will take a long time for all these classes that should be mixins in flutter to become mixins so I think we can close this and open this once they are done with that work.
The title of the PR is already solved by `flame_lint`.

st-pasha:
I feel like, in general, we are substantially over-using the mixin pattern. In many cases when we declare something as a mixin, it would have better been a component instead.

spydon:
> I feel like, in general, we are substantially over-using the mixin pattern. In many cases when we declare something as a mixin, it would have better been a component instead.

Do you have any examples?

st-pasha:
> Do you have any examples?

Yes: `Tappable` is a mixin, but could just as easily be a component. Similarly for other event handlers. The advantage of components over mixins is that the mixins are "baked-in", whereas components can be easily added/removed at runtime.

spydon:
> Yes: `Tappable` is a mixin, but could just as easily be a component. Similarly for other event handlers. The advantage of components over mixins is that the mixins are "baked-in", whereas components can be easily added/removed at runtime.

I can see the benefit of that, but then you can't handle the specific gesture logic on your component, you'll have to create another component that you then add to the component where you want this logic. It feels like it becomes unnecessarily nested.

st-pasha:
What if there is a predefined component that takes a callback as an argument: `Tappable(TapCallback onTap)`.
Then, inside your component you can say `add(Tappable(this.handleTap1))`. Though this seems a bit more code than merely adding a mixin and defining `onTap()` method, but the advantage is that the callback can be changed at runtime or even removed at all.

Ok, i'm not actually sure how big of an advantage this is.

luanpotter:
@spydon actually I don't agree this is solved; the rule is still commented out on Flame Lint:

https://github.com/flame-engine/flame/blob/main/packages/flame_lint/lib/analysis_options.yaml#L94

I'm not saying we need to wait years until Flutter decides to fix itself; rather, given the current context we should decide right now if we want this rule or not (seems to me the answer is not). And then, regardless of the decision, either have the rule or not on the file, and not having it commented out.

luanpotter:
Closing in lieu of https://github.com/flame-engine/flame/pull/1117

PR #1001, opened 2021-10-06
Improved testing API

luanpotter:
# Description

<details>
<summary> ~~first proposal~~ </summary>
Proposal: make the createGame provider for the `flameTest` optional.
Advantages: cleaner, simpler for most use cases.
Disadvantages: not as type safe because if converts a compile error to a runtime error if T != Game (and dart doesn't have overloading).
Other option would be to create a new method where T===FlameGame and it calls this one (but would need to have a different name).
</details>

### Second Proposal

So I came up with a better idea. It came to me that from all the parameters required to run a test, there are two types:

* name and actually test -> these are of course always unique per test. can't have a test with duped name or code
* everything else -> these most commonly have a single value, if not for the entire application, at least for a set of tests

So the proposal is to split this into two steps. the common parameters are provided via the constructor and the test functions only take two parameters. now, when you do wanna change something, I provide a copy constructor. I considered optional parameters on the two test functions but I choose the copy constructor for two reasons: (1) promotes reuse of code between tests even when there are variations and (2, most importantly) the test functions signature now is identical to the test/widgetTest functions from the platform, so it looks more natural

Thoughts?

Note: just a proposal yet, I will (try) to add tests later (tests that test tests are hard haha)

## Checklist

- [ ] I read and followed the [Flame Style Guide].

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/test/components/hoverable_test.dart
+++ b/packages/flame/test/components/hoverable_test.dart
@@ -10,7 +10,11 @@ import 'package:test/test.dart';
 
 class _GameWithHoverables extends FlameGame with HasHoverableComponents {}
 
+final withHoverables = FlameTester(() => _GameWithHoverables());
+
 class _GameWithoutHoverables extends FlameGame {}
+
+final withoutHoverables = FlameTester(() => _GameWithoutHoverables());
 
 class HoverableComponent extends PositionComponent with Hoverable {
   int enterCount = 0;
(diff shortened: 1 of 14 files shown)


erickzanardo:
Leeeets do eeeet

st-pasha:
Is there any reason why we can't have `FlameTester(MyGame())` and instead have to do `FlameTester(() => MyGame())?

erickzanardo:
We need a fresh instance for each test case

st-pasha:
> We need a fresh instance for each test case

Makes sense, thanks.

luanpotter:
We will soon be able to do `FlameTester(MyGame)` with the dart 2.15! (constructor tear off)

PR #1031, opened 2021-10-18
[WIP] Effects

st-pasha:
# Description

This is a work-in-progress effort to make effects more powerful, customizable, and reusable. Here's the outline of what's happening in this PR:

First, I split the original `ComponentEffect` class into 2 parts. One part, which I call `FlameAnimationController`, is responsible for handling the time. That is, all the logic that handles how the effect progresses over a certain duration, then waits, then progresses backwards, then waits some more, then repeats, etc. The reasons to put this logic into a separate class are the following:
  - it makes the main `ComponentEffect` class less complicated;
  - the logic in animation controller does not dependent on the components system, which makes it possible to reuse those classes in non-FCS games;
  - `FlameAnimationController` is an abstract interface, which allows multiple concrete implementations, including user-defined ones. Currently I have only `StandardAnimationController`, but planning to add a `RandomAnimationController` too;
  - each individual effect now accepts the `controller` object instead of a dozen or so individual animation-related parameters. This makes it much easier to document, test, and extend this functionality.

The second part of the original `ComponentEffect` is called `EffectComponent` (tentative name), and it is simply the logic that allows inserting effects into the game tree as regular components. This class is the base for all other effects.

The `Transform2DEffect` extends `EffectComponent` and represents effects that are intended to apply to objects having a `Transform2D` property, for example to a `PositionComponent` or a `Camera`.

There are also several move/rotate effects that extend from the `Transform2DEffect`. My goals when redesigning the original move / rotate effects are as follows:
  - the effects should target the `Transform2D` object directly, and thus be applicable to any component having such property;
  - the effects should be *mutually composable*: it should be possible to add multiple effects (even of the same type) to a component, and they should not interfere with each other. For example, I may want to move a component to another point, but also superimpose a "shake" effect while it moves.

Currently missing: tests, examples, more documentation, some more effects.

## Checklist

## Breaking Change


st-pasha:
Closing, as the functionality implemented in PRs #1047, #1052, #1061, #1063 and #1077

chongbo2013:
Parallel Effect pls

spydon:
> Parallel Effect pls 

You can already do parallel effects, if they dont interfere with each other. Try adding two (or more) `MoveByEffect`s for example, like one that goes up and then down and one forward. Many use that to simulate a jump forward.

PR #1058, opened 2021-10-31
Fix events + isHud

luanpotter:
I still need to write up an example to showcase this, and maybe some test.

But hopefully this PR is a good starting point for understanding the issue.

Currently we combine both projectos on game (viewport & camera). And say that game `is a Projector`. but for hud elements we ignore that, while we would want to ignore the camera only, and still apply the viewport. So instead I make the game `has` 2 projectors, that way we can use the appropriated one when desired. also change the events to expose both (viewport only, and both). I didn't see any use to expose camera only.

### DIFF
--- a/packages/flame/lib/src/game/mixins/game.dart
+++ b/packages/flame/lib/src/game/mixins/game.dart
@@ -24,7 +24,7 @@ import 'loadable.dart';
 /// methods to use it in a `GameWidget`.
 /// Flame will deal with calling these methods properly when the game's widget
 /// is rendered.
-mixin Game on Loadable implements Projector {
+mixin Game on Loadable {
   final images = Images();
   final assets = AssetsCache();
 
@@ -157,17 +157,8 @@ mixin Game on Loadable implements Projector {
     return _gameRenderBox!.localToGlobal(point.toOffset()).toVector2();
   }
 
-  @override
-  Vector2 unprojectVector(Vector2 vector) => vector;
-
-  @override
-  Vector2 projectVector(Vector2 vector) => vector;
-
-  @override
-  Vector2 unscaleVector(Vector2 vector) => vector;
-
-  @override
-  Vector2 scaleVector(Vector2 vector) => vector;
+  Projector projector() => IdentityProjector();
+  Projector viewportProjector() => IdentityProjector();
 
   /// Utility method to load and cache the image for a sprite based on its
   /// options.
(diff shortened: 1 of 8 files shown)


spydon:
I think there are some huds where we want to ignore the viewport, is that possible? Imagine a joystick for example, you would want that one placed outside the viewport if you have a different aspect ratio on your phone, so that the joystick can still be reached comfortably from the edge.

st-pasha:
> I think there are some huds where we want to ignore the viewport, is that possible? 

We should replace the `isHud` flag with component's position within the game tree: if the component is attached to a viewport, then it is "HUD"; a "regular" component will be attached to the World instance. The components like Joystick can be attached to the `Game` object directly.

spydon:
> > I think there are some huds where we want to ignore the viewport, is that possible?
> 
> We should replace the `isHud` flag with component's position within the game tree: if the component is attached to a viewport, then it is "HUD"; a "regular" component will be attached to the World instance. The components like Joystick can be attached to the `Game` object directly.

What do you mean with the world instance? Isn't that the same as being attached directly to the game right now?
Or do you mean that we should create a layer which is called world where it goes through both the camera and the viewport, but when being attached directly to the game instance it doesn't go through any of them?

I think it sounds like a pretty good idea, the default `game.add` should still add to the world instance though I think since that is the most common operation, and then maybe `game.viewport.add` adds a component as a hud that respects the viewport but not the camera. But what would the syntax then be for adding something that doesn't take the viewport or the camera into consideration?

luanpotter:
> I think there are some huds where we want to ignore the viewport, is that possible

I agree we have to improve how isHud, camera & viewport work. But for now this just fixes the existing bug on how they were supposed to work. So I think this is at least a step in the right direction

packages/flame/lib/src/game/mixins/game.dart
```diff
@@ -157,17 +157,8 @@ mixin Game on Loadable implements Projector {
...
-
-  @override
-  Vector2 projectVector(Vector2 vector) => vector;
-
-  @override
-  Vector2 unscaleVector(Vector2 vector) => vector;
-
-  @override
-  Vector2 scaleVector(Vector2 vector) => vector;
+  Projector projector() => IdentityProjector();
```
erickzanardo, line 160: Should we make these attributes instead of methods, so we avoid instantiating this more than necessary?

luanpotter: Good point, forgot you could override attributes on dart. let me see how it looks

spydon (approved):

erickzanardo (approved):
LGTM

PR #1061, opened 2021-10-31
New Rotate effect

st-pasha:
# Description

This is a new implementation of `RotateEffect`, based on `Transform2DEffect`. The differences are:
  - the class is conceptually much simpler: the constructor now takes only 2 parameters: an angle and an effect controller;
  - the effect can be applied to any object with a `Transform2D` member;
  - the effect is composable: it is possible to apply several rotation effects to the same object.

A new example is added to demonstrate this effect.

## Checklist

## Breaking Change

## Related Issues

Part of #1031


packages/flame/lib/src/effects2/rotate_effect.dart
```diff
@@ -0,0 +1,60 @@
...
+    return _RotateToEffect(angle, controller);
+  }
+
+  /// The magnitude of the effect: how much the target should turn as the
+  /// progress goes from 0 to 1.
+  double _angle;
+
+  /// This variable keeps track of the effect's `progress` from the last time
+  /// that the `apply()` method was called.
+  double _lastProgress = 0;
```
spydon, line 33: Maybe this is something that `Transform2DEffect` should keep track of? Will it not be used by more effects later?

packages/flame/lib/src/effects2/rotate_effect.dart
```diff
@@ -0,0 +1,60 @@
...
+
+  @override
+  void apply(double progress) {
+    final dProgress = progress - _lastProgress;
```
spydon, line 37: Will this way not create a larger rounding error until the effect is done? Instead of looking at how much the current progress should be converted to an angle?

spydon (commented):
Looks pretty good, just some comments.

packages/flame/lib/src/effects2/rotate_effect.dart
st-pasha: I consider it highly unlikely. A double has 16 digits of precision in its mantissa. For an average 1-second effect at 100fps, in the worst-case scenario we'd still retain 14 digits of precision. That's more than enough for a typical game.

Typically, the loss-of-precision occurs with floating points when adding and then subtracting quantities of significantly different magnitudes (such as `A + a - A`, where `A` is much larger than `a`), however this is not the scenario that we have here.

packages/flame/lib/src/effects2/rotate_effect.dart
st-pasha: I guess you're right, that would save us from having to redefine `reset()` method too.

packages/flame/lib/src/effects2/rotate_effect.dart
spydon: Airtight, as long as we have tests for it, because I remember having problems with precision in the other effects system so that it was 1 pixel off when it ended.

packages/flame/lib/src/effects2/rotate_effect.dart
```diff
@@ -0,0 +1,50 @@
...
+
+  @override
+  void apply(double progress) {
+    final dProgress = progress - previousProgress;
+    target.angle += _angle * dProgress;
+    super.apply(progress);
+  }
+}
+
+class _RotateToEffect extends RotateEffect {
```
spydon, line 39: Does this really need to be done as a separate class, can't this be done in the factory?

packages/flame/test/effects2/rotate_effect_test.dart
```diff
@@ -0,0 +1,126 @@
+import 'package:flame/components.dart';
+import 'package:flame/game.dart';
+import 'package:flame/src/effects2/rotate_effect.dart';
+import 'package:flame/src/effects2/standard_effect_controller.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('RotateEffect', () {
```
spydon, line 8: It would be nice to have some test that covers a longer timespan or `isInfinite`, and `isAlternating` (is it still called that?) to see that it always hits the correct angle for min and max on the curve when it has run as a lone effect on a component for a longer while.

packages/flame/test/effects2/transform2d_effect_test.dart
```diff
@@ -8,9 +8,6 @@ import 'package:flutter_test/flutter_test.dart';
 
 class MyEffect extends Transform2DEffect {
```
spydon, line 9: Is `Transform2DEffect` abstract? Otherwise that could just be used directly.

examples/lib/stories/effects/effects.dart
```diff
@@ -67,5 +68,18 @@ void addEffectsStories(Dashbook dashbook) {
       'Color Effect',
       (_) => GameWidget(game: ColorEffectGame()),
       codeLink: baseLink('effects/color_effect.dart'),
+    )
+    ..add(
+      'Rotate Effect (v2)',
+      (_) => GameWidget(game: RotateEffect2Game()),
+      codeLink: baseLink('effects/rotate_effect2.dart'),
+      info: '''
+        The outer rim rotates at a different speed forward and reverse, and
```
erickzanardo, line 77: Add this string as a static const variable on the `RotateEffect2Game` class for improved organisation.

examples/lib/stories/effects/rotate_effect2.dart
```diff
@@ -0,0 +1,196 @@
...
+  final double _radius;
+  final double _width;
+  late final Path _northPath, _southPath;
+  late final Paint _northPaint, _southPaint;
+
+  @override
+  Future<void> onLoad() async {
+    super.onLoad();
+    _northPaint = Paint()..color = const Color(0xff387fcb);
+    _southPaint = Paint()..color = const Color(0xffa83636);
```
erickzanardo, line 125: nit, since these variables are already late final, they can be initialised together with the declaration.

examples/lib/stories/effects/rotate_effect2.dart
```diff
@@ -0,0 +1,196 @@
...
+  @override
+  Future<void> onLoad() async {
+    super.onLoad();
+    _bgPaint = Paint()
+      ..style = PaintingStyle.stroke
+      ..strokeWidth = _width
+      ..color = const Color(0xffb6a241);
+    _marksPaint = Paint()
+      ..style = PaintingStyle.stroke
+      ..color = const Color(0xff3d3b26);
```
erickzanardo, line 172: Same comment here, they can be initialised together with the declarion.

packages/flame/test/effects2/transform2d_effect_test.dart
st-pasha: Yes, it's abstract because in itself it doesn't do anything (other than keeping track of `lastProgress` variable). So I don't see how this do-nothing behavior would be useful anywhere, other than in this particular test file.

examples/lib/stories/effects/effects.dart
st-pasha: ok

examples/lib/stories/effects/rotate_effect2.dart
st-pasha: ok

examples/lib/stories/effects/rotate_effect2.dart
st-pasha: yep

packages/flame/lib/src/effects2/rotate_effect.dart
st-pasha: This is a private class, invoked from the factory constructor `RotateEffect.to()`.
Can it be merged with the "main" RotateEffect class? It can, but it'll bring an extra variable `_destinationAngle`, a boolean parameter `_isRelative`, and an override for `onStart()` which is not needed by the main class. I don't think this would improve code clarity...

Also, I've been thinking about eventually adding another flavor of the rotate effect: `RotateEffect.towards()`, which would be similar to `RotateEffect.to()`, except that it would perform the rotation along the shortest path (for example, rotating from 359º to 0º should not make a full circle, but instead choose a 1º arc). It would be easier to implement such class as another override, instead of converting `_isRelative` into an enum.

packages/flame/test/effects2/rotate_effect_test.dart
st-pasha: I've added a test for an effect running for 1000 seconds at ~100fps, and the observed numeric discrepancy at the end is on the order of 1e-12 to 1e-11.

packages/flame/lib/src/effects2/rotate_effect.dart
```diff
@@ -0,0 +1,50 @@
...
+///
+/// This effect applies incremental changes to the component's angle, and
+/// requires that any other effect or update logic applied to the same component
+/// also used incremental updates.
+class RotateEffect extends Transform2DEffect {
+  RotateEffect.by(double angle, EffectController controller)
+      : _angle = angle,
+        super(controller);
+
+  factory RotateEffect.to(double angle, EffectController controller) {
```
spydon, line 23: Commenting here instead since it has more with this line to do than the ongoibg discussing further down.

Will it not be clearer to have an argument here which specifies whether it goes the shortest path or not? `to` vs. `towards` doesn't really make it clear which one that goes the shortest way I think. Or `towards` is pretty clear, but not `to`, possibly a better name could also be used?

st-pasha: So, here was my thinking: 
I started with a `MoveEffect`, which has two flavors: `MoveEffect.to` which moves to a specific point on screen, and `MoveEffect.by` which shifts the current position by the specified offset. The first name is nice and clear, the second sounds a bit weird, but I couldn't think of anything better. Then came the `RotateEffect`. By analogy, I made `RotateEffect.by` (which is now nice and clear), and `RotateEffect.to` (which is ok).

I imagine we'd add `ScaleEffect.to()` / `ScaleEffect.by()` later on too. And it would be nice to have all these names consistent with each other.

Now, `RotateEffect.towards()` is different from the rest, but that's because rotation has this weird property that 0º==360º.

Having said that, I agree that the distinction between `towards` and `to` may not be immediately clear to a casual reader. But the mnemonics is clear here: `towards`= in general direction of, whereas `to`= go there. The use cases are also quite different: you would use `.towards` in a game like space-shooter, where you need to orient your character or projectiles in the direction of an enemy; while `.to` has more niche purposes, such as rotary phones, various knobs on a control panel, roulette, spin-wheel, etc.

spydon: Alright, I can't come up with any better name either, and i know that you'll document it well so it should be fine.

spydon (approved):

PR #1062, opened 2021-11-01
Missing lint rules

st-pasha:
This PR adds all currently missing lint rules -- as comments -- into the main `analysis_options.yaml` file.

The reason for this change is because the list of rules in `analysis_options.yaml` has been compiled a long time ago, and since then the set of lint rules has grown considerably. However, we don't know which of the rules that are currently missing in the `analysis_options.yaml` because they were deliberately deleted, versus those that were never there. This PR addresses this problem.

The next step(s) after this PR is merged should be as follows:
  - check the commented-out rules one-by-one, and decide in which of the categories they fall:
    - if the rule should obviously be added, then uncomment it and run `flutter analyze` on the entire project. If there were any changes, then submit those changes + the uncommented rule as a separate PR;
    - if the rule should obviously not be added (for example because it is incompatible with one of the existing rules), then add a comment at the bottom of the `analysis_options.yaml` file explaining the rationale for this rule's exclusion. The explanation is needed, because reasons change over time, and what makes perfect sense today may no longer apply in the future. Also, remove the "?" after the rule;
    - if it's not clear whether the rule should or should not be added, then let's open a discussion for that rule somewhere and make a decision once there is a consensus.


erickzanardo:
Not sure I understand why we would add all the lint rules commented.

If we want to include or remove a lint rule, we could just open a PR for it.

st-pasha:
> Not sure I understand why we would add all the lint rules commented.

This PR establishes the scope of which rules are missing: there are 70. I could have added most of these rules as uncommented, but that would be too big, and too controversial. Approaching each rule one-by-one is much easier.

erickzanardo:
But why add all of them commented? Then un comment them?

If there is any rule that you want to propose, I think you could just add a proposal for that rule and then we can discuss about that specific one.

st-pasha:
Well, ideally we want to have as many linter rules as possible -- because generally they are useful. Still, there are those that we'd want to exclude, because reasons.

Currently, there are 70 lint rules missing from the list, and it is impossible to tell which ones were excluded deliberately, and which ones were never even considered, probably because they are too new. This PR makes it easier to tell those apart: 
  1. rules that were considered and found useful are uncommented;
  2. rules that were considered and found not applicable are commented, and there is an explanation at the end of the file for the reasons why we don't like that specific rule;
  3. rules that were never considered are commented out and have a question mark, so that they can later be evaluated and turned either into category 1 or category 2.

> I think you could just add a proposal for that rule and then we can discuss about that specific one.

That would work nicely if there was only a few rules missing. But there are 70, so we need a system.

spydon:
> This PR adds all currently missing lint rules -- as comments -- into the main `analysis_options.yaml` file.
> 
> The reason for this change is because the list of rules in `analysis_options.yaml` has been compiled a long time ago, and since then the set of lint rules has grown considerably. However, we don't know which of the rules that are currently missing in the `analysis_options.yaml` because they were deliberately deleted, versus those that were never there. This PR addresses this problem.
> 
> The next step(s) after this PR is merged should be as follows:
>   - check the commented-out rules one-by-one, and decide in which of the categories they fall:
>     - if the rule should obviously be added, then uncomment it and run `flutter analyze` on the entire project. If there were any changes, then submit those changes + the uncommented rule as a separate PR;
>     - if the rule should obviously not be added (for example because it is incompatible with one of the existing rules), then add a comment at the bottom of the `analysis_options.yaml` file explaining the rationale for this rule's exclusion. The explanation is needed, because reasons change over time, and what makes perfect sense today may no longer apply in the future. Also, remove the "?" after the rule;
>     - if it's not clear whether the rule should or should not be added, then let's open a discussion for that rule somewhere and make a decision once there is a consensus.

We have to mention these categories somewhere then, and there is another category too, the new rules that are added that are not yet in our list.

I see a lot of rules here that we could enable so I think this could be a good way to start keeping track of what rules we have decided to use, and which we have actively decided not to use.

erickzanardo:
I don't think we need to have "as many rules as possible". We need to have only the ones that we are conformable and happy with.

Many rules are opposite sides of a coin, so it is normal to not have all rules on the project.

I still prefer to, when w want a rule to be included, a PR can be made with the proposal, no need to add a bunch of comments to the analysis option.

st-pasha:
> I don't think we need to have "as many rules as possible". We need to have only the ones that we are conformable and happy with.

These are not opposites. We can have as many rules as possible within the subset of those that we're comfortable with. However, defining the bounds of that subset is not trivial. Usually it would require one to consider all rules, then filter out the unhappy ones, and what's left is the happy subset. 

Thus, this PR makes a first step in that direction: it reveals the set of *all* rules.

> Many rules are opposite sides of a coin, so it is normal to not have all rules on the project.

Perfectly normal, I agree.

> I still prefer to, when we want a rule to be included, a PR can be made with the proposal, no need to add a bunch of comments to the analysis option.

Agree with the first part, disagree with the second. Before anyone can start to *want* to include a rule, they must know what rules are out there. Which involves comparing the current list of rules against the total list of rules, which takes work. And what we don't want is to repeat that work 70 times over. Hence, this PR: it gathers the list of things to consider.

You can think of each commented-out rule as an open GitHub issue. We can make PRs to close some of those issues one way or another; but in general it's much easier to work when there's some kind of work plan. Which is exactly what this PR creates.

erickzanardo:
My point is, we don't need to define any subset. This will only complicate something that is simple IMO.

Want a new rule? Send a proposal and we can discuss about it, no need to spend time going through all the rules available. We have many, many more important things to care about right now.

Anyhow, I leave this in the hands of the rest of the team, if majority wants to go with this, that is ok.

spydon:
@renancaraujo suggested that we would have the discussions regarding the new rules and rules that we have decided against as a GitHub discussion page instead, I think that will be much cleaner and it will be a better forum than to have some comments directly on the file and then the full discussion somewhere else.
So I vote for closing this.

spydon (approved):

PR #1063, opened 2021-11-01
RemoveEffect

st-pasha:
# Description

Added `RemoveEffect` for removing components. This could be especially useful at the end of a sequence of effects. For example, when you make the component fade out using an opacity effect, then add a `RemoveEffect` to physically remove the component from the game tree after it is no longer visible.

Also, added `SimpleEventController` -- which is a much simpler version of `StandardEventController`, and can be used in those situations where you don't need anything fancy.

## Checklist

## Breaking Change

## Related Issues

WIP for #1031


packages/flame/lib/src/effects2/destroy_effect.dart
```diff
@@ -0,0 +1,17 @@
+
+import 'effect.dart';
+import 'simple_effect_controller.dart';
+
+/// This simple effect, when attached to a component will cause that component
+/// to be removed from the game tree and destroyed after `delay` seconds.
+class DestroyEffect extends Effect {
```
erickzanardo, line 7: I would name this RemoveEffect to match the shoulRemove attrition, or the removeFromParent method

packages/flame/lib/src/effects2/simple_effect_controller.dart
```diff
@@ -0,0 +1,53 @@
...
+/// linearly over [duration] seconds.
+///
+/// The [duration] can be 0, in which case the effect will jump from 0 to 1
+/// instantaneously.
+///
+/// The [delay] parameter allows to delay the start of the effect by the
+/// specified number of seconds.
+///
+/// See also: [StandardEffectController]
+class SimpleEffectController extends EffectController {
```
erickzanardo, line 15: What about LinerEffectController ?

erickzanardo (approved):
Couple of suggestions but LGTM

packages/flame/lib/src/effects2/destroy_effect.dart
st-pasha: sounds good

packages/flame/lib/src/effects2/simple_effect_controller.dart
st-pasha: The main idea behind this controller is that it is as simple as possible (while still being useful). So it is certainly linear, but only because linear curve is the simplest. From this perspective, I think the current name is better because it conveys the limitation of this class: that we don't want to make it more complicated in the future.

spydon:
This seems like it overlaps a bit with the particle system, but I can definitely see how this is useful.

examples/lib/stories/effects/effects.dart
```diff
@@ -67,5 +68,14 @@ void addEffectsStories(Dashbook dashbook) {
       'Color Effect',
       (_) => GameWidget(game: ColorEffectGame()),
       codeLink: baseLink('effects/color_effect.dart'),
+    )
+    ..add(
+      'Remove Effect',
+      (_) => GameWidget(game: RemoveEffectGame()),
+      codeLink: baseLink('effects/remove_effect_game.dart'),
+      info: '''
```
spydon, line 76: We have started to move in these info sections to variables in the example class that is it running (it is far from done), so that the info text can be seen easiliy when looking at the code too.

spydon: Like this:  https://github.com/flame-engine/flame/blob/main/examples/lib/stories/collision_detection/multiple_shapes.dart#L12

st-pasha: Ah, that makes sense. I'll do the same here.

examples/lib/commons/circle_component.dart
```diff
@@ -18,4 +18,11 @@ class CircleComponent extends PositionComponent {
     super.render(canvas);
     canvas.drawCircle(Offset(radius, radius), radius, paint);
   }
+
+  @override
+  bool containsPoint(Vector2 point) {
+    final local = absoluteToLocal(point);
+    final center = Vector2.all(radius);
+    return local.distanceToSquared(center) <= radius * radius;
```
spydon, line 26: I think we should convert this helper class to a `ShapeComponent` with a circle shape, since this is code duplication from what is in the `Circle` that extends `Shape`.

packages/flame/lib/src/effects2/simple_effect_controller.dart
```diff
@@ -0,0 +1,54 @@
...
+
+  final double duration;
+  final double delay;
+  double _timer = 0.0;
+
+  @override
+  bool get started => _timer >= delay;
+
+  @override
+  bool get completed => _timer >= delay + duration;
```
spydon, line 29: A note for later, when creating the `SequenceEffect` we need to carry over the time that overflowed in here to the next effect.

examples/lib/commons/circle_component.dart
st-pasha: Hmm, but the problem with the `Circle` is that it is not a `Component`; and it also keeps its own position and size, so can't be cleanly converted into a `PositionComponent` either.

packages/flame/lib/src/effects2/simple_effect_controller.dart
st-pasha: noted

examples/lib/commons/circle_component.dart
spydon: It can, you just use the `ShapeComponent`.

spydon: Ah the shape keeps its initial position + the component position (and same for the other fields), how sloppy of me.
But that will be easy to fix, putting up a PR for that.

spydon: You can keep this component like it is and I'll do the refactoring later when my PR is in. :)

spydon (approved):

Issue #1075, opened 2021-11-04
`ShapeComponent.containsPoint` does not take `scale` into consideration


spydon:
# Current bug behaviour
If you scale a `ShapeComponent` the input events will miss the component if they aren't still inside of the original hitbox/size.

Since `Shape` is using a global coordinate system for the hitbox shapes it should be refactored so that only the `HitboxShape`s use the global coordinate system and that `containsPoint` that is directly on `Shape` can be passed a transformed point, like in `PositionComponent` with `absoluteToLocal`.

# Expected behaviour
That the events would be passed to the `ShapeComponent` as long as the are inside its bounds.

# Steps to reproduce
Put a scale on one of the components in the examples that take input.

# More environment information
Create a list of more environment information, like:
* Flame version: 1.0.0-releasecandidate.16
* Platform affected: all
* Platform version affected: all

# More information

st-pasha:
A more reliable approach would be to re-think the design of `ShapeComponent` so that it wouldn't need to maintain its own transform that mirrors the transform of `PositionComponent`.

spydon:
> A more reliable approach would be to re-think the design of `ShapeComponent` so that it wouldn't need to maintain its own transform that mirrors the transform of `PositionComponent`.

And good idea of how to do that without having to duplicate code from `Shape`?

spydon:
> Is it possible to make Shape transform-agnostic?

Not easily I think, one idea is to treat it as a sub-component to `ShapeComponent`. 
So instead of having the transform variables that are initially set in `Shape` being transferred to the `ShapeComponent` those variables could be relative to it and we can pass in the local point instead of the global one, the problem with that is that if we add the same `Shape` as a hitbox in the `ShapeComponent` it will not work with other hitboxes.

st-pasha:
What if a `HitboxShape` was a component that could be added to a `PositionComponent` as a child? Then the parent's transform would apply to the child automatically.

spydon:
> What if a `HitboxShape` was a component that could be added to a `PositionComponent` as a child? Then the parent's transform would apply to the child automatically.

The collision detection system needs to be able to live outside of FCS, but it could maybe `HitboxShape` could be wrapped somehow, that would most likely not end up much different from how it is now though.

st-pasha:
> The collision detection system needs to be able to live outside of FCS

[...]

Issue #980, opened 2021-09-25
Improvements on collision detection.


namzug16:
# What could be improved

The way collision detection works can be improved and can be even made 4 times faster(more or less).

# Any risks?

No that I can think of

# More information

At the moment, every time we add a component to a game that has as a mixin ```HasCollidables```, this component will be added to a list of ```Collidables``` and then all the list will be evaluated with the function ```collisionDetection(List<Collidable> collidables)```. The problem with this implementation ( if I can call it a problem) is that one ```Collidable``` will check for collisions with all the other ```Collidable```s even if we don't want it to collide with everything. 

We could change this by setting the concept of "Layers", a layer would be an ```int``` that this ```Collidable``` is occupying and then we could specify with which layers can this component collide. 

These layers then would be stored on a map like so:
```dart
Map<int, List<Collidable>> collidables = {};
```
by doing this we can improved a lot the performance of the function ```collisionDetection()```.

This is the new implementation of the function that I came out with:
```dart
void collisionDetectionMap(Map<int, List<Collidable>> collidables) {
  // l stands for layer
  for (var l = 0; l < collidables.length; l++) {
    for(var x = 0; x < collidables[l]!.length; x++){

      final collidableX = collidables[l]![x];
      if (collidableX.collidableType != CollidableType.active) {
        break;
      }

      // all the collidables that this component can collide with
      final collidablesY = <Collidable>[];

      for(final lY in collidableX.collisionLayersToCollideWith){
        if(lY == l){
          // I think this can still be improved
          // here I'm adding any other component of the same layer
          // in case the layer is inside the layers this component can collide with

          // if this part stays like so, the entire function will be only 3.8(more or less)
          // faster than the original, in the other hand it this part is changed
          // the function could be more than 4 times faster
          collidablesY.addAll(collidables[lY]!.toList().sublist(0, x) + collidables[lY]!.toList().sublist(x+1));
        }else{
          collidablesY.addAll(collidables[lY]!.toList());
        }
      }

      for(final collidableY in collidablesY){
        if (collidableY.collidableType == CollidableType.inactive) {
          break;
        }

        final intersectionPoints = intersections(collidableX, collidableY);
        if (intersectionPoints.isNotEmpty) {
          collidableX.onCollision(intersectionPoints, collidableY);
          collidableY.onCollision(intersectionPoints, collidableX);
          final collisionHash = _combinedHash(collidableX, collidableY);
          _collidableHashes.add(collisionHash);
        } else {
[...]

PR #1076, opened 2021-11-04
`ShapeComponent` and `Hitbox` to take transform of parents full ancestor tree into consideration

spydon:
# Description

The hitboxes didn't take the full ancestor tree of the parent component into consideration, now they do (except for multiple scales).

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/doc/components.md
+++ b/doc/components.md
@@ -442,7 +442,7 @@ final circle = CircleComponent(radius: 200.0, position: Vector2(100, 200), paint
 
 ### RectangleComponent
 A `RectangleComponent` can be created in two ways, depending on if it is a square or not.
-To create a `RectangleComponent` that is 300px wide and 200px in height you can do the following:
+To create a `RectangleComponent` that is 300 in width and 200 in height you can do the following:
 
 Example:
 ```dart
@@ -456,7 +456,7 @@ final rectangle = RectangleComponent(
 
 To create a square you can instead use the slightly simpler named constructor
 `RectangleComponent.square`. This is an example of how to create a red square with width and height
-200px:
+200:
 
 ```dart
 final paint = BasicPalette.red.paint()..style = PaintingStyle.stroke;
--- a/doc/components.md
+++ b/doc/components.md
@@ -431,7 +431,7 @@ the shape of the specific component, it also takes all the arguments that can be
 `PositionComponent`.
 
 ### CircleComponent
-A `CircleComponent` can be created only defining it's `radius`, but you most likely want to pass it
+A `CircleComponent` can be created only by defining its `radius`, but you most likely want to pass it
 a `position` and maybe `paint` (the default is white) too.
 
 Example:
--- a/doc/components.md
+++ b/doc/components.md
@@ -424,7 +424,7 @@ Three example implementations can be found in the
 
 ## ShapeComponents
 
-There exists three basic components which can be used if you want to draw geometrical shapes as
+There exist three basic components that can be used if you want to draw geometrical shapes as
 components on the screen. Since the `ShapeComponent`s are `PositionComponent`s you can use effects
 on them. All `ShapeComponent`s take a `Paint` as an argument and then arguments to define
 the shape of the specific component, it also takes all the arguments that can be passed to the
--- a/doc/components.md
+++ b/doc/components.md
@@ -441,7 +441,7 @@ final circle = CircleComponent(radius: 200.0, position: Vector2(100, 200), paint
 ```
 
 ### RectangleComponent
-A `RectangleComponent` can be created in two ways, depending on if it is a square or not.
+A `RectangleComponent` can be created in two ways, depending on if it's a square or not.
 To create a `RectangleComponent` that is 300 in width and 200 in height you can do the following:
 
 Example:
--- a/doc/components.md
+++ b/doc/components.md
@@ -422,6 +422,99 @@ It is also possible to create custom renderers by extending the `ParallaxRendere
 Three example implementations can be found in the
 [examples directory](https://github.com/flame-engine/flame/tree/main/examples/lib/stories/parallax).
 
+## ShapeComponents
+
+There exists three basic components which can be used if you want to draw geometrical shapes as
+components on the screen. Since the `ShapeComponent`s are `PositionComponent`s you can use effects
+on them. All `ShapeComponent`s take a `Paint` as an argument and then arguments to define
+the shape of the specific component, it also takes all the arguments that can be passed to the
+`PositionComponent`.
+
+### CircleComponent
+A `CircleComponent` can be created only defining it's `radius`, but you most likely want to pass it
+a `position` and maybe `paint` (the default is white) too.
+
+Example:
+```dart
+final paint = BasicPalette.red.paint()..style = PaintingStyle.stroke;
+final circle = CircleComponent(radius: 200.0, position: Vector2(100, 200), paint: paint);
+```
+
+### RectangleComponent
+A `RectangleComponent` can be created in two ways, depending on if it is a square or not.
+To create a `RectangleComponent` that is 300px wide and 200px in height you can do the following:
+
+Example:
+```dart
+final paint = BasicPalette.red.paint()..style = PaintingStyle.stroke;
+final rectangle = RectangleComponent(
+  size: Vector2(300.0, 200.0),
+  position: Vector2(100, 200),
+  paint: paint,
+);
+```
+
+To create a square you can instead use the slightly simpler named constructor
+`RectangleComponent.square`. This is an example of how to create a red square with width and height
+200px:
+
+```dart
+final paint = BasicPalette.red.paint()..style = PaintingStyle.stroke;
+final square = RectangleComponent.square(
+  size: 200.0,
+  position: Vector2(100, 200),
+  paint: paint,
+);
+```
+
+### PolygonComponent
+The `PolygonComponent` is the most complicated of the `ShapeComponent`s since you'll have to define
+all the "corners" of your polygon. You can create the `PolygonComponent` in two different ways,
+either you use the default constructor which takes a list of `Vector2` where each of them should be
+between -1.0 and 1.0 that describes the ration of the length from the center to the edge of the size
+of the component. So
+`[Vector2(1.0, 1.0), Vector2(1.0, -1.0), Vector2(-1.0, -1.0), Vector2(-1.0, 1.0)]`
+would describe a rectangle that fills the full size of the component. Remember to define the list in
+a counter clockwise manner (if you think in the screen coordinate system where the y-axis is
+flipped, otherwise it is clockwise).
+
+So to create a diamond shaped `PolygonComponent` which is slightly smaller than the defined size you
+would do this:
+
+```dart
+final vertices = ([
+  Vector2(0.0, 0.9),  // Middle of top wall
+  Vector2(-0.9, 0.0), // Middle of left wall
+  Vector2(0.0, -0.9), // Middle of bottom wall
+  Vector2(0.9, 0.0),  // Middle of right wall
+]);
+
+final diamond = PolygonComponent(
+  normalizedVertices: vertices,
+  size: Vector2(200, 300),
+  position: Vector2.all(500),
+)
+```
+
+If you instead want to define your polygon from absolute points you can do that too with the
+`PolygonComponent.fromPoints` factory. When using that one you don't have to define a `size` or a
+`position` either since it will be calculated for you, but if you decide to add those arguments
+anyways they will override what has been calculated from your list of vertices.
+
+Example (diamond shape again):
+
+```dart
+final vertices = ([
+  Vector2(100, 100),  // Middle of top wall
+  Vector2(50, 150), // Middle of left wall
+  Vector2(100, 200), // Middle of bottom wall
+  Vector2(200, 150),  // Middle of right wall
+]);
+
+final diamond = PolygonComponent.fromPoints(vertices);
+)
+```
+
 ## SpriteBodyComponent
 
 See [SpriteBodyComponent](forge2d.md#SpriteBodyComponent) in the Forge2D documentation.
--- a/doc/components.md
+++ b/doc/components.md
@@ -423,12 +423,13 @@ Three example implementations can be found in the
 [examples directory](https://github.com/flame-engine/flame/tree/main/examples/lib/stories/parallax).
 
 ## ShapeComponents
-
-There exist three basic components that can be used if you want to draw geometrical shapes as
-components on the screen. Since the `ShapeComponent`s are `PositionComponent`s you can use effects
-on them. All `ShapeComponent`s take a `Paint` as an argument and then arguments to define
+The `ShapeComponent` is a basic component that can be used if you want to draw geometrical shapes as
+components on the screen.  Since the `ShapeComponent` is a `PositionComponent`s you can use effects
+on it. All `ShapeComponent`s take a `Paint` as an argument and then arguments to define
 the shape of the specific component, it also takes all the arguments that can be passed to the
 `PositionComponent`.
+
+There are three implementations of `ShapeComponent`, which are the following:
 
 ### CircleComponent
 A `CircleComponent` can be created only by defining its `radius`, but you most likely want to pass it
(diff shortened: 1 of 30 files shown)


packages/flame/lib/src/components/shape_component.dart
```diff
@@ -17,19 +23,29 @@ class ShapeComponent extends PositionComponent {
   @override
   final Anchor anchor = Anchor.center;
 
+  @override
+  set angle(double a) {
+    super.angle = a;
+    shape.angle = a;
```
erickzanardo, line 29:
```suggestion
    super.angle = shape.angle = a;
```

packages/flame/lib/src/components/shape_component.dart
```diff
@@ -17,19 +23,29 @@ class ShapeComponent extends PositionComponent {
...
         super(
           position: shape.position,
           size: shape.size,
+          scale: scale,
           angle: shape.angle,
           anchor: Anchor.center,
           priority: priority,
         ) {
     shape.isCanvasPrepared = true;
+    position.addListener(() => shape.position.setFrom(position));
```
erickzanardo, line 47: We will need to clear those listeners when this component is removed right?

examples/lib/stories/collision_detection/only_shapes.dart
```diff
@@ -21,6 +22,12 @@ class OnlyShapes extends FlameGame with HasTappableComponents {
   final shapePaint = BasicPalette.red.paint()..style = PaintingStyle.stroke;
   final _rng = Random();
 
+  @override
+  Future<void> onLoad() async {
+    await super.onLoad();
+    camera.zoom = 2;
```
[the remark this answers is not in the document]
spydon: This is only for testing now, I don't think we need zoom on this example.
In the end maybe we should have zoom on all examples as a dashbook property?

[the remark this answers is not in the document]
erickzanardo: Up to you, just a small idea I had while reviewing this :)

Yeah, this could be potentially something cool to have on all examples

packages/flame/lib/src/components/shape_component.dart
spydon: Hmm, do we? Since this is listeners on itself, it shouldn't be a problem after it is removed right? And then if the component is re-added it will still have the listeners.

erickzanardo: Yeah, you are correct, I thought this was from the parent for some reason

erickzanardo (approved):

spydon:
This does not properly solve the issue, see #1075

packages/flame/lib/src/components/position_component.dart
```diff
@@ -164,6 +164,31 @@ class PositionComponent extends Component {
   Vector2 get scaledSize =>
       Vector2(width * scale.x.abs(), height * scale.y.abs());
 
+  final _totalScale = Vector2.zero();
+
+  /// The resulting scale after all the ancestors and the components own scale
+  /// has been applied.
+  Vector2 get absoluteScale {
```
st-pasha, line 171: I don't think this is well-defined.
If component N is rotated 90º and is inside another component that scales x-direction twice, then the "absolute scale" of N is that its y-direction is stretched twice. Of course, it's even more complicated when the rotation angles are not nice.

packages/flame/lib/src/geometry/shape.dart
```diff
@@ -116,25 +131,19 @@ abstract class Shape {
...
 
 mixin HitboxShape on Shape {
   late PositionComponent component;
 
   @override
-  Vector2 get size => component.scaledSize;
+  Vector2 get size => component.size;
+
+  @override
+  Vector2 get scale => component.absoluteScale;
```
st-pasha, line 143: I suspect that for non-symmetrical shapes taking absolute scale will result in wrong collision detection if the shape was flipped

packages/flame/lib/src/components/position_component.dart
spydon: That should be fine? I scale the vertices first and then rotate them as separate actions, and everything is defined around the center.

packages/flame/lib/src/geometry/shape.dart
spydon: True, I have not taken flipping into consideration yet (this is still a draft).
In the future (not in this PR) I will try to refactor `Polygon` to use `Path` instead, which supports a matrix transform method.

packages/flame/lib/src/components/position_component.dart
st-pasha: Rotations tend to turn X axes into Y and vice versa. So, scales from multiple transforms do not compose with each other.

spydon: Ah right, I see what you mean... Hmm, seems like we'll have to introduce a matrix that we can accumulate all operations on then, that should work right? This seems like it will be quite an expensive operation.

st-pasha: Indeed, rather expensive.

For a short-term solution we could use `PositionComponent.toAbsoluteRect()` which computes smallest bounding rect (in absolute coords) that circumscribes the component.

Alternatively, we could compute the product of `max(abs(scale.x), abs(scale.y))`, which will produce an over-estimation of the size of the component, but may be ok if we only need this in order to filter out objects that are sufficiently far away from each other.

st-pasha:
In general, it would be more efficient to find the common ancestor of all collidable components, and compute the common transforms only up to that point.

Alternatively, the common "ancestor" may not have to be the ancestor at all, but rather a component that minimizes the amount of computation needed to project every object into that space. For example:
```
    WorldRoot ---\
   /              \
BulletLayer -\    Player
|       \     \
Bullet Bullet Bullet ...
```
Here, if we only care about bullets colliding with a single player, making `BulletLayer` the "root" and projecting `Player` into that coordinate space would use less calculation then projecting every bullet into the `WorldRoot`.

packages/flame/lib/src/components/position_component.dart
spydon: Since the collision detection system isn't only for FCS `PositionComponent.toAbsoluteRect()` can't be used, and it wouldn't be accurate enough.

I think that as a short-term solution we'll just say in the docs that we don't support ancestors of the parent to the hitbox that are scaled, so the direct parent can be scaled, but none above that level. I think this will be more useful since nobody have actually complained about that usecase so far.

spydon:
> In general, it would be more efficient to find the common ancestor of all collidable components, and compute the common transforms only up to that point.
> 
> Alternatively, the common "ancestor" may not have to be the ancestor at all, but rather a component that minimizes the amount of computation needed to project every object into that space. For example:
> 
> ```
>     WorldRoot ---\
>    /              \
> BulletLayer -\    Player
> |       \     \
> Bullet Bullet Bullet ...
> ```
> 
> Here, if we only care about bullets colliding with a single player, making `BulletLayer` the "root" and projecting `Player` into that coordinate space would use less calculation then projecting every bullet into the `WorldRoot`.

This is related to #980 and the layer discussion going on there, but indeed we need a concept of layers in the long run.

doc/components.md
```diff
@@ -422,6 +422,99 @@ It is also possible to create custom renderers by extending the `ParallaxRendere
 Three example implementations can be found in the
 [examples directory](https://github.com/flame-engine/flame/tree/main/examples/lib/stories/parallax).
 
+## ShapeComponents
+
+There exists three basic components which can be used if you want to draw geometrical shapes as
```
st-pasha, line 427:
```suggestion
There exist three basic components that can be used if you want to draw geometrical shapes as
```

examples/lib/stories/collision_detection/circles.dart
```diff
@@ -7,11 +7,24 @@ import 'package:flame/geometry.dart';
 import 'package:flame/input.dart';
 import 'package:flutter/material.dart' hide Image, Draggable;
 
-const circlesInfo = '''
-This example will create a circle every time you tap on the screen. It will have
-the initial velocity towards the center of the screen and if it touches another
-circle both of them will change color.
-''';
+class Circles extends FlameGame with HasCollidables, TapDetector {
```
st-pasha, line 10: Rename into `CirclesExample`? The name `Circles` sounds like it's a class that somehow manages multiple circle shapes.

examples/lib/stories/collision_detection/circles.dart
```diff
@@ -7,11 +7,24 @@ import 'package:flame/geometry.dart';
...
+    add(ScreenCollidable());
+  }
+
+  @override
+  void onTapDown(TapDownInfo info) {
+    add(MyCollidable(info.eventPosition.game));
+  }
+}
 
 class MyCollidable extends PositionComponent
```
st-pasha, line 29: Change to `_MyCollidable`, in order to indicate that it's private to this file?

examples/lib/stories/collision_detection/multiple_shapes.dart
```diff
@@ -9,20 +9,94 @@ import 'package:flame/input.dart';
...
-const multipleShapesInfo = '''
-An example with many hitboxes that move around on the screen and during
-collisions they change color depending on what it is that they have collided
-with. 
+enum Shapes { circle, rectangle, polygon }
 
-The snowman, the component built with three circles on top of each other, works
-a little bit differently than the other components to show that you can have
-multiple hitboxes within one component.
+class MultipleShapes extends FlameGame
```
st-pasha, line 14: Similarly here: `MultipleShapesExample`

doc/components.md
```diff
@@ -422,6 +422,99 @@ It is also possible to create custom renderers by extending the `ParallaxRendere
...
+components on the screen. Since the `ShapeComponent`s are `PositionComponent`s you can use effects
+on them. All `ShapeComponent`s take a `Paint` as an argument and then arguments to define
+the shape of the specific component, it also takes all the arguments that can be passed to the
+`PositionComponent`.
+
+### CircleComponent
+A `CircleComponent` can be created only defining it's `radius`, but you most likely want to pass it
```
erickzanardo, line 434:
```suggestion
A `CircleComponent` can be created only by defining its `radius`, but you most likely want to pass it
```

doc/components.md
```diff
@@ -422,6 +422,99 @@ It is also possible to create custom renderers by extending the `ParallaxRendere
...
+a `position` and maybe `paint` (the default is white) too.
+
+Example:
+```dart
+final paint = BasicPalette.red.paint()..style = PaintingStyle.stroke;
+final circle = CircleComponent(radius: 200.0, position: Vector2(100, 200), paint: paint);
+```
+
+### RectangleComponent
+A `RectangleComponent` can be created in two ways, depending on if it is a square or not.
```
erickzanardo, line 444:
```suggestion
A `RectangleComponent` can be created in two ways, depending on if it's a square or not.
```

doc/components.md
```diff
@@ -422,6 +422,99 @@ It is also possible to create custom renderers by extending the `ParallaxRendere
...
+To create a `RectangleComponent` that is 300px wide and 200px in height you can do the following:
```
erickzanardo, line 445: I don't have a strong opinion about this, but maybe we should refrain from using px on docs as Flutter uses logical pixels instead. Not sure if having mo suffix or just dpi would be better or more accurate

examples/lib/stories/collision_detection/collidable_animation.dart
```diff
@@ -0,0 +1,116 @@
...
+    makes them turn. The birds have PolygonHitboxes which are marked with the
+    green lines and dots.
+  ''';
+
+  @override
+  Future<void> onLoad() async {
+    await super.onLoad();
+    add(ScreenCollidable());
+    // Top left component
+    add(AnimatedComponent(Vector2.all(200), Vector2.all(100))
```
erickzanardo, line 25: Nit: trailing comma on the component, I wonder why the lint didn't warned here

erickzanardo: Ahhh the examples have no flame_lint right? Hahha

packages/flame/lib/src/components/component.dart
```diff
@@ -176,6 +176,20 @@ class Component with Loadable {
     nextParent = component;
   }
 
+  final List<Component> _ancestors = [];
+
+  /// A list containing the current parent and its parent, and so on, until it
+  /// reaches a component without a parent.
+  List<Component> ancestors() {
+    _ancestors.clear();
```
erickzanardo, line 184: Why not just declare the list inside the method?

packages/flame/test/components/shape_component_test.dart
```diff
@@ -0,0 +1,420 @@
...
+  group('ShapeComponent.containsPoint tests', () {
+    test('Simple circle contains point', () {
+      final component = CircleComponent(
+        radius: 1.0,
+        position: Vector2(1, 1),
+        anchor: Anchor.center,
+      );
+      expect(
+        component.containsPoint(Vector2.all(1.5)),
+        true,
```
erickzanardo, line 18: Nit: for this line and the ones below we could use isTrue and isFalse for prettier messages.

Not blocker though

examples/lib/stories/collision_detection/circles.dart
spydon: Since this isn't library code I don't think this is necessary, and it would also break consistency with most of the other of the examples.

packages/flame/lib/src/components/component.dart
spydon: Because then we need to create multiple new objects on every tick when collision detection is used. This has shown to be quite expensive in Forge2D when you get some volume, probably because the GC has to run more often.

erickzanardo: Oh, that interesting

examples/lib/stories/collision_detection/circles.dart
spydon: Fixed!

doc/components.md
spydon: Makes sense, fixed

packages/flame/test/components/shape_component_test.dart
spydon: Aaah so that is what those are for, I think I removed them from some other place. 😅
Adding them now, and I will go over all the tests in a later PR where I can fix this too.

erickzanardo: There is a cool lint rule that can check that, maybe that is a good candidate to be added to flame lint

doc/components.md
```diff
@@ -422,6 +422,99 @@ It is also possible to create custom renderers by extending the `ParallaxRendere
 Three example implementations can be found in the
 [examples directory](https://github.com/flame-engine/flame/tree/main/examples/lib/stories/parallax).
 
+## ShapeComponents
+
+There exist three basic components that can be used if you want to draw geometrical shapes as
+components on the screen. Since the `ShapeComponent`s are `PositionComponent`s you can use effects
```
luanpotter, line 428: I think you should first say what ShapeComponent represents abstractly, and then mention there are 3 provided implementations ?

examples/lib/commons/square_component.dart
```diff
@@ -10,7 +10,7 @@ class SquareComponent extends RectangleComponent with EffectsHelper {
     Paint? paint,
     int priority = 0,
   }) : super(
-          Vector2.all(size),
+          size: Vector2.all(size),
           position: position,
```
luanpotter, line 14: nit: normally we do position then size?

doc/components.md
spydon: Good call, fixed.

examples/lib/commons/square_component.dart
spydon: Fixed!

packages/flame/lib/src/components/position_component.dart
```diff
@@ -163,8 +163,17 @@ class PositionComponent extends Component {
...
   /// modified by the user.
-  Vector2 get scaledSize =>
-      Vector2(width * scale.x.abs(), height * scale.y.abs());
+  Vector2 get scaledSize {
+    return Vector2(width * scale.x.abs(), height * scale.y.abs());
+  }
+
+  /// The resulting angle after all the ancestors and the components own angle
+  /// has been applied.
+  double get absoluteAngle {
```
luanpotter, line 172: should we consider caching this somehow? might be more effort than its worth though

spydon: I actually had caching first, but then considered it a premature optimization since this is only on primitive types, doesn't create any new objects and it is basically just going through a short linked list, I think it should be fine.

luanpotter (approved):
Wow, VERY optimized code!

PR #1082, opened 2021-11-08
Improvements to rendering pipeline

st-pasha:
# Description

1. `preRender()` is now called from `renderTree()` instead of from the `render()` method;
    - this makes the code less prone to bugs where either `preRender()` or `postRender()` are not called, or called out of order;
    - this ensures that `preRender` is always called **before** `render()`, not **within** it;
2. `postRender()` moved to be invoked at the end of the rendering stage, after all children were rendered;
3. `render()` is no longer marked as `@mustCallSuper`. The rendering system no longer relies on the user having to call `super.render()` as the very first statement in their `render()` implementation, so we no longer need to insist that they **must** call the super implementation at all.
4. `renderTree()` method no longer attempts to save/restore canvas when rendering each child. Instead, it is the job of each child to save/restore canvas if they perform any manipulation of it (which is actually very rare). This is a performance optimization: avoid unnecessary system calls when they are not needed.
5. `renderTree()` is no longer marked as `@mustCallSuper`. The cases where one would need to re-implement this method are quite rare, but when they do arise, the user should have the freedom to do a completely separate implementation of that method. Such potential examples include:
    - switch to different method of rendering depending on the zoom level;
    - implement custom logic for determining which of the children to render;
    - skip rendering those children that are known to be outside of viewport;
    - skip rendering if the object's opacity is 0;
    - avoid infinite recursion in the rendering pipeline;
    - etc.

## Checklist

## Breaking Change

- [x] Yes, this is a (minor) breaking change (please indicate a breaking change in `CHANGELOG.md`).


packages/flame/CHANGELOG.md
```diff
@@ -17,6 +17,8 @@
  - Create default implementations of `RectangleComponent`, `CircleComponent` and `PolygonComponent`
  - Streamlined the argument list for all components extending `PositionComponent`
  - Improved interaction between viewport and isHud components
+ - Calling super-method in `.render()` is now optional
+ - Rendering code that does canvas manipulation is now responsible for saving/restoring the canvas
```
erickzanardo, line 21: I think it is important to note that this changes are only related to the components, not the game.

st-pasha: ok, i'll re-phrase it

packages/flame/lib/src/components/component.dart
```diff
@@ -131,34 +131,28 @@ class Component with Loadable {
...
-      c.renderTree(canvas);
-      canvas.restore();
-    });
+    children.forEach((c) => c.renderTree(canvas));
 
     // Any debug rendering should be rendered on top of everything
     if (debugMode) {
       renderDebugMode(canvas);
     }
+    postRender(canvas);
```
spydon, line 145: I think this should be before `renderDebugMode`, so that the debug rendering always is on top.

spydon: Ahh, the `restore` runs in there now.

spydon (approved):
This is great, so many users have missed to call `super.render` or called it in the end of their `render` method.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -329,9 +329,15 @@ class PositionComponent extends Component {
 
   @override
   void preRender(Canvas canvas) {
+    canvas.save();
     canvas.transform(transformMatrix.storage);
   }
 
+  @override
+  void postRender(Canvas canvas) {
```
spydon, line 337: This and `preRender` should have `@mustCallSuper` now

st-pasha: done

packages/flame/lib/src/components/component.dart
```diff
@@ -131,34 +131,28 @@ class Component with Loadable {
...
-    children.forEach((c) {
-      canvas.save();
-      c.renderTree(canvas);
-      canvas.restore();
-    });
+    children.forEach((c) => c.renderTree(canvas));
 
     // Any debug rendering should be rendered on top of everything
     if (debugMode) {
       renderDebugMode(canvas);
```
spydon, line 143: If the canvas is flipped, won't the text in the debug mode also be flipped then?
Maybe that was the case before too.

st-pasha: It was the same before, only then `canvas.save()/restore()` were surrounding the `renderTree()` call, now they are inside the `renderTree()` call, but still surround the rest of the content there.

renancaraujo (approved):

wolfenrain (approved):

erickzanardo (approved):

luanpotter:
I love this! my only small concern is with:

> it is the job of each child to save/restore canvas if they perform any manipulation of it (which is actually very rare).

I think it can be confusing to the users, hence the reason why we had it be done by the engine. but I am ok with it, specially since we provide methods like renderAt, etc as extension function on canvas precisely to facilitate that.

packages/flame/lib/src/components/component.dart
```diff
@@ -131,34 +131,28 @@ class Component with Loadable {
...
 
-  @mustCallSuper
-  void render(Canvas canvas) {
-    preRender(canvas);
-  }
+  void render(Canvas canvas) {}
 
-  @mustCallSuper
   void renderTree(Canvas canvas) {
+    preRender(canvas);
```
luanpotter, line 137: we should probably add comments to preRender to let users know that they are responsible for saving/restoring the canvas

luanpotter (approved):
LGTM

packages/flame/lib/src/components/component.dart
st-pasha: ok, I added a mention about that in the doc-comment of `preRender()`.

spydon:
Formatting fails on `component.dart`, after that is fixed we can merge it.

PR #1085, opened 2021-11-10
Improving TimerComponent API

erickzanardo:
# Description

Our previous API on the `TimerComponent` was just a simple wrapper which didn't helped the user much other than just call the update method automatically, many times were easier to just user the `Timer` class directly.

This PR proposes an improved API for it.

## Checklist

## Breaking Change

## Related Issues


packages/flame/lib/src/timer.dart
```diff
@@ -1,8 +1,6 @@
 import 'dart:math';
 import 'dart:ui';
 
-import 'components/component.dart';
-
 /// Simple utility class that helps handling time counting and implementing
 /// interval like events.
```
st-pasha, line 5: The purpose is unclear: how does this help with interval-like events? Or is this class mixes two functionalities: keeping track of time and being an EffectController?

packages/flame/lib/src/components/timer_component.dart
```diff
@@ -0,0 +1,36 @@
...
+
+    if (autoStart) {
+      timer.start();
+    }
+  }
+
+  /// Called everytime the [timer] reached a tick
+  /// default implementaiton is a no-op, override this
+  /// to add custom logic.
+  void tick() {}
```
spydon, line 25: I think this is a bit of a confusing name for the callback.

spydon, line 25: I think this is a bit of a confusing name for the callback.

packages/flame/lib/src/timer.dart
spydon: You use the `callback` to help with "interval like events", it is more general than the effect controller. Say that you want to spawn some new components every 5 seconds, then this is useful.

packages/flame/lib/src/components/timer_component.dart
spydon: Maybe it is more useful if we can pass in the callback instead of having an implementation in the class, because now you have to extend this component to use it, but many times it is only the callback that you want to change, so maybe it would be easier to have it as an argument like for the timer class, wdyt?

packages/flame/lib/src/components/timer_component.dart
```diff
@@ -0,0 +1,36 @@
+import '../../components.dart';
+
+/// Simple component which wraps a [Timer] instance allowing it to be easily
+/// used inside a FlameGame game.
```
spydon, line 4:
```suggestion
/// A component that uses a [Timer] instance which you can react to when it has finished.
```

packages/flame/lib/src/components/timer_component.dart
erickzanardo: The idea of having a method here is so we can have a more "self contained" widget. Just like I refactored on the game in game example. Now all the logic is in the component.

But maybe we can have both things here, I will think on something and update here

spydon: The default callback argument can be null maybe and if it is specified both are called? :thinking: 
But maybe it is confusing to have two places for the callbacks.
It would be very nice to be able to just do `TimerComponent(limit: 5, callback: () => add(Enemy())` instead of having to create a new component for that.

erickzanardo: Yeah, that is what I am thinking of, have an argument and have the method. If the callback is specified we call it, otherwise the class method is called.

It is a shame that dart don't allow us to do method tear of on the constructors, because if it did, this would not be necessary at all.

I will implement this later.

erickzanardo: @spydon  Added the option to pass a callback as well, lmk what you think

packages/flame/lib/src/components/timer_component.dart
```diff
@@ -0,0 +1,36 @@
...
+class TimerComponent extends Component {
+  late final Timer timer;
+  final bool removeOnFinish;
+
+  TimerComponent({
+    required double limit,
```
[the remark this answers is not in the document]
erickzanardo: changed to period

packages/flame/lib/src/components/timer_component.dart
spydon: Looks good! I think the argument and the method should have the same name though, maybe rename `tick` to callback? Or if there is a better name we can give both.

packages/flame/lib/src/timer.dart
st-pasha: It still sounds like an effect, though. Call it `CallbackEffect`, and let it invoke the provided callback function when the progress reaches 1. The code would be very simple and similar to `RemoveEffect`.

But unlike the current `TimerComponent`, it will allow you to customize the timing more flexibly. For example, you can make the callback fire once, or `n` times, or keep firing infinitely; the intervals between invocations can be either the same, or random (once we have `RandomEffectController`), or custom (if you define your own controller).

erickzanardo: TimerComponent is just a wrapper on the Timer class which is a simpler, generic feature for running period things, it is simpler and it is available outside the FCS too.

Both things can exist, if what you need is too complex for the TimerComponent to be used, effects can be used instead.

packages/flame/lib/src/components/timer_component.dart
```diff
@@ -0,0 +1,50 @@
...
+  ///
+  /// [period] The period of time in seconds that the tick will be called
+  /// [repeat] When true, this will continue running after [period] is reached
+  /// [autoStart] When true, will start upon instantiation
+  /// [callback] When provided, will be called everytime [period] is reached. This
+  /// overrides the [tick] method
+  TimerComponent({
+    required double period,
+    bool repeat = false,
+    bool autoStart = false,
```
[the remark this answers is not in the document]
erickzanardo: Changed to true

packages/flame/lib/src/components/timer_component.dart
```diff
@@ -0,0 +1,50 @@
...
+    required double period,
+    bool repeat = false,
+    bool autoStart = true,
+    this.removeOnFinish = false,
+    VoidCallback? onTick,
+  }) {
+    timer = Timer(
+      period,
+      repeat: repeat,
+      onTick: onTick ?? onTick,
```
luanpotter, line 27: `onTick ?? onTick` === `onTick`?

packages/flame/lib/src/components/timer_component.dart
```diff
@@ -0,0 +1,50 @@
...
+
+    if (autoStart) {
+      timer.start();
+    }
+  }
+
+  /// Called everytime the [timer] reached a tick and
+  /// no onStep is provided on the component instantiation
+  /// default implementaiton is a no-op, override this
+  /// to add custom logic.
```
luanpotter, line 38:
```suggestion
  /// Called everytime the [timer] reached a tick (but
  /// only if onStep is provided on the component instantiation).
  /// The default implementaiton is a no-op; override this
  /// to add custom logic.
```

packages/flame/lib/src/components/timer_component.dart
```diff
@@ -0,0 +1,50 @@
...
+  void onTick() {}
```
luanpotter, line 39: would it be confusing to have this just not called if a lambda is provided on the constructor? what if the default onTick implementation called the lambnda. then people who override can decide if they wanna call super.onTick or not, and there is no magic effects

doc/util.md
```diff
@@ -135,12 +135,10 @@ class MyFlameGame extends FlameGame {
...
       TimerComponent(
-        Timer(
-          10,
-          callback: () => print("10 seconds elapsed"),
-          repeat: true,
-        )
-        ..start()
+        period: 10,
+        repeat: true,
+        onTick: () => print("10 seconds elapsed"),
```
spydon, line 140:
```suggestion
        onTick: () => print('10 seconds elapsed'),
```

examples/lib/stories/components/game_in_game.dart
```diff
@@ -13,6 +13,19 @@ changes its parent from its original game to the component that is rotating.
 After another 5 seconds it changes back to its original parent, and so on.
 ''';
 
+class GameChangeTimer extends TimerComponent with HasGameRef<GameInGame> {
+  GameChangeTimer() : super(period: 5, repeat: true, autoStart: true);
```
spydon, line 17:
```suggestion
  GameChangeTimer() : super(period: 5, repeat: true);
```

packages/flame/lib/src/components/timer_component.dart
spydon, line 27:
```suggestion
      onTick: onTick ?? this.onTick,
```

packages/flame/lib/src/timer.dart
```diff
@@ -32,15 +30,15 @@ class Timer {
       if (_current >= limit) {
         if (!repeat) {
           _running = false;
-          callback?.call();
+          onTick?.call();
           return;
         }
         // This is used to cover the rare case of _current being more than
-        // two times the value of limit, so that the callback is called the
+        // two times the value of limit, so that the onTick is called the
```
spydon, line 37:
```suggestion
// two times the value of limit, so that the onTick is called the
```

packages/flame_bloc/example/lib/src/game/components/enemy_creator.dart
```diff
@@ -1,35 +1,22 @@
...
-          EnemyComponent(
-            (gameRef.size.x - 25) * random.nextDouble(),
-            0,
-          ),
-        );
-      },
-    );
-    enemyCreator.start();
-  }
+  EnemyCreator() : super(period: 1, repeat: true, autoStart: true);
```
spydon, line 11:
```suggestion
  EnemyCreator() : super(period: 1, repeat: true);
```

spydon (approved):
Just some minor comments

packages/flame/lib/src/components/timer_component.dart
erickzanardo: Ops, I guess I trusted too much on the find and replace ahhaha

packages/flame/lib/src/timer.dart
```diff
@@ -1,18 +1,16 @@
...
-
 /// Simple utility class that helps handling time counting and implementing
 /// interval like events.
 class Timer {
   final double limit;
-  VoidCallback? callback;
+  VoidCallback? onTick;
   bool repeat;
   double _current = 0;
   bool _running = false;
```
spydon, line 11: Shouldn't this have the same behaviour as `TimerComponent` and start directly on creation?

doc/util.md
```diff
@@ -135,12 +135,10 @@ class MyFlameGame extends FlameGame {
...
-        Timer(
-          10,
-          callback: () => print("10 seconds elapsed"),
-          repeat: true,
-        )
-        ..start()
+        period: 10,
+        repeat: true,
+        onTick: () => print('10 seconds elapsed'),
+        autoStart: true,
```
spydon, line 141: Maybe a comment that this is the default?

packages/flame/lib/src/timer.dart
spydon: Like having the `bool autoStart = true,` as an argument?

spydon (approved):

packages/flame/lib/src/timer.dart
```diff
@@ -32,15 +37,15 @@ class Timer {
       if (_current >= limit) {
         if (!repeat) {
           _running = false;
-          callback?.call();
+          onTick?.call();
           return;
         }
         // This is used to cover the rare case of _current being more than
-        // two times the value of limit, so that the callback is called the
+// two times the value of limit, so that the onTick is called the
```
st-pasha, line 44: oops, a whitespace accident

luanpotter (approved):
LGTM

packages/flame/lib/src/timer.dart
spydon, line 44:
```suggestion
        // two times the value of limit, so that the onTick is called the
```

spydon: Good catch

PR #1092, opened 2021-11-12
Added a simple makefile for flame package

st-pasha:
# Description

Small Makefile to simplify common tasks during development.

If this proves useful, a similar makefile can be added in the future to other flame projects.

## Checklist

## Breaking Change

### DIFF
--- a/packages/flame/CHANGELOG.md
+++ b/packages/flame/CHANGELOG.md
@@ -22,6 +22,7 @@
  - Components that manipulate canvas state are now responsible for saving/restoring that state
  - Remove `super.render` calls that are no longer needed
  - Fixed typo in error message
+ - Added a Makefile for the `flame` repository
 
 ## [1.0.0-releasecandidate.16]
  - `changePriority` no longer breaks game loop iteration
--- a/packages/flame/Makefile
+++ b/packages/flame/Makefile
@@ -1,0 +1,14 @@
+
+.PHONY: test
+test:
+	@flutter test
+
+.PHONY: format
+format:
+	@flutter analyze
+	@flutter format .
+
+.PHONY: coverage
+coverage:
+	@flutter test --coverage
+	@genhtml coverage/lcov.info --output-directory=coverage/


spydon:
I vote no for this, imho Makefiles should be deprecated with their obscure syntax.
If it would have been a standard in the Flutter world I would have voted yes just to keep consistency, but luckily it is not. :smile:

erickzanardo:
> I vote no for this, imho Makefiles should be deprecated with their obscure syntax. If it would have been a standard in the Flutter world I would have voted yes just to keep consistency, but luckily it is not. 😄

I agree, it is just wrapping things that are already available on the Flutter CLI anyway.

st-pasha:
The purpose of this makefile is not to implement the build process (which was the original purpose of makefiles, now outdated), but to provide a simple way to perform common developer tasks from CLI.

For example, I need to run `flutter analyze` and `flutter format` before each push. They have to run in that particular order. It's much easier to say `make format` then. Similarly for coverage. It's easier to say `make coverage` then trying to remember and then typing out the necessary instructions.

spydon:
> For example, I need to run `flutter analyze` and `flutter format` before each push. They have to run in that particular order.
This sounds very much like a setup for individual developers, I think most have this integrated into their IDE.

> then trying to remember and then typing out the necessary instructions.
Now it will be two ways of doing things instead, going via the Makefile, or straight to the flutter CLI, and to keep consistency with other Flutter projects we shouldn't invent a new way of doing things I think.

> How do you guys usually generate coverage?
For me at least, I very rarely generate that myself, but let the GitHub action generate that. The few times I do generate that myself I simply press `ctrl+r` and type `cov` and it'll pop up.

st-pasha:
> This sounds very much like a setup for individual developers

Absolutely. Some developers will have powerful IDEs with everything integrated. Others won't be so lucky. This makefile helps those others, without negatively affecting the first group.

> Now it will be two ways of doing things instead

As you have mentioned before, there are already two ways of doing this: one is via the IDE magic button, another is via the flutter command in the terminal. So, this PR adds a third way: via the make command. I don't think having multiple ways to accomplish the same task is bad, if those ways have different sets of pros and cons.

> I simply press ctrl+r and type cov and it'll pop up.

This obviously only works on your computer.

I mean, we have the question "does your PR reduce coverage?" in the PR checklist, so there should be at least some easy way for a contributor to verify whether that's true or not.

spydon:
> Absolutely. Some developers will have powerful IDEs with everything integrated. Others won't be so lucky. This makefile helps those others, without negatively affecting the first group.

 We don't add IDE specific settings to the repo, so neither should we have something that is only targeted to the small group that is not running an IDE, after all the recommended way from Flutter when installing it is either Android Studio or VSCode. And I mean, even Erick that it running pure vim is voting this down.

> As you have mentioned before, there are already two ways of doing this: one is via the IDE magic button, another is via the flutter command in the terminal. So, this PR adds a third way: via the make command. I don't think having multiple ways to accomplish the same task is bad, if those ways have different sets of pros and cons.

The user knows that it goes through flutter when that happens, and it is usually done on save, so no active action to perform this. If I see a makefile in a project I would think that that is the project owners preferred way of running things.

> This obviously only works on your computer.
Yeah, it was only an example since you said you had a hard time remembering the command, but the point is that you shouldn't need to remember it, the GitHub action will do this for you.
 
> I mean, we have the question "does your PR reduce coverage?" in the PR checklist, so there should be at least some easy way for a contributor to verify whether that's true or not.
We could rephrase that to that the cicd doesn't report reduced coverage. We could even add coveralls like we have on `ordered_set` if we want it to be a bit more visual, but that should be discussed in an issue.

Anyhow, we are discussing something that hasn't been an issue for any contributor that we have heard of, if we would have seen that the community had problems with this, we'd probably bring back the script folder that we used to have and the coverage script that was in there. I'll close this since none in the team doesn't wants us to introduce a makefile.

PR #1095, opened 2021-11-13
Unify tests to use `FlameTester`

spydon:
# Description

Will do the rest of the tests when the structure is reviewed.

Unify tests to use FlameTester where possible.
Especially check the methods added in `FlameGame` that should only be used for testing.

## Checklist

## Breaking Change

## Related Issues


packages/flame/lib/src/game/flame_game.dart
```diff
@@ -118,4 +118,20 @@ class FlameGame extends Component with Game {
 
   @override
   Projector get projector => camera.combinedProjector;
+
+  /// Only use this method for tests, it makes sure that the [component]s
+  /// added to the tree if you wait for the returned future to resolve.
+  @visibleForTesting
+  Future<void> addUpdate(Component component) async {
```
erickzanardo, line 125: What if we add this to the tester instead? Maybe as an extension

spydon: Sounds good, but how can it access things from there inside the lamda in which we are running the test? Or you mean like a static function?

spydon: Added them as static functions now, but if there is a way to make the function accessible within the lambda without being static, please do tell. :smile:

erickzanardo: This is a method on the flame game class right? So it can be an extension method, but declared on flame test dependency. That way we don't need this comment and the visible for testing annotation

spydon: An extension on the game class? Then we need to add that extension every time we make a new game class within a test file?

spydon: Haha sorry, I'm tired, that is genius of course that works. I'll do that. :)

packages/flame_test/lib/src/flame_test.dart
```diff
@@ -11,6 +12,22 @@ extension FlameFinds on CommonFinders {
   }
 }
 
+extension FlameGameExtension on FlameGame {
+  /// Makes sure that the [component] is added to the tree if you wait for the
+  /// returned future to resolve.
+  Future<void> addUpdate(Component component) async {
```
erickzanardo, line 18: What if we call it ensureComponentAddition, or something like that?

erickzanardo (approved):
One additional suggestion, but LGTM already

packages/flame_test/lib/src/flame_test.dart
spydon: It's so long though, could we come up with a shorter name since it is going to be used everywhere?

spydon: `ensureAdd` or `fullAdd` maybe? Since it is doing `add` which is acting with the component I don't think the component part in the name is needed.

erickzanardo (approved):
LGTM

Issue #976, opened 2021-09-23
Remove and replace QueryableOrderedSet


st-pasha:
Our current implementation of the `ComponentSet` is built upon the class `QueryableOrderedSet`, which in turn extends `OrderedSet`, which is built on top of a `SplayTreeSet`.

The choice of the container structure used throughout the game engine is obviously a very important one. Such choice should first and foremost be based on the efficiency consideration, as well the set of features that various implementations support. Thus, I'd like to compare the current implementation versus a more simple one based on a plain list of Components.

1. The most common pattern of access to the components is iteration: we iterate through all of the components at least twice every tick: once for the update()s, and then once again for render()ing. It is dead simple to iterate through a plain list. Though one can argue that `SplayTreeSet` also provides a suitable iterator, so this is simple too. Still, in terms of the number of CPU instructions, iterating over a list is ought to be simpler than iterating over a tree, so it's probably a small win for the List approach.

2. The second most common pattern of access is an iteration in reverse order, which we perform for every pointer event. Many games rely on drag events to control the main character (for example via joystick), and these events fire every game tick. Thus, on every game tick, we need to iterate through the game components in reverse order. Once again, iterating in reverse over a list is very simple. For the current `QueryableOrderedSet`, it's not: in order to iterate over the tree backwards, we convert it to a list and then reverse that list. Needless to say, serializing the entire game tree into a list on every game tick is rather expensive and should be avoided. Thus, a clear win for the List approach.

3. Extendability. If `QueryableOrderedSet` was implemented in-house, it would be easy to extend it with new functionality. For example, it would be possible to add functionality to modify the set while iterating over it; or to replace an existing element with a new one. It is very challenging to do the same with the existing implementation since many of its internal functions are private and thus cannot be overridden. Again, this is a win for the List approach.

[...]

PR #1098, opened 2021-11-14
Added Component.onRenderTree

st-pasha:
# Description

This PR adds callback `.onRender` for the `Component` class. This callback is null by default, but when it is not null, it will be used instead of the regular `render()`.

Such callback allows implementing various effects that want to change the appearance of an existing object. Examples include:
  - replacing the component with a cached image;
  - generic opacity effect;
  - generic color tint effect;
  - explosion effect;
  - duplication effect;
  - shadow (on the ground);
  - edge detection effect?

## Checklist

## Breaking Change

## Related Issues

This allows some of the use cases described in #911


erickzanardo:
A method that runs instead of the component owns render method seems a little bit confusing for me.

If the exception of the first example of use case, it is not much clear to me how this enables the others.

spydon:
Hmm, can you explain a bit more the thought behind this and how it would work? I think a change like this probably should have been discussed in an issue first, but since it is so little code I guess not much time is wasted anyways so maybe it actually is even easier to discuss it like this.

I see how this can be used for `replacing the component with a cached image` and `explosion effect`, but both of those you could also just replace the component with another one, or control the state within the component like you usually would? And for the other ones, since the `onRender` replaces the `render` method call, is it meant that the `onRender` should then call `render` after it has performed the changes that it wants to do to the  canvas?

It feels like the flow becomes a bit too complicated and a bit overlapping with `preRender`.

st-pasha:
> Hmm, can you explain a bit more the thought behind this and how it would work?

Of course. So, the first observation is that it's not currently possible to replace a component with another one. You can remove a component and then attach another one to the same parent, but that would change their relative order. For example, if the parent has children `[A, B, C, D]`, and you want to replace `C` with `X`, you'd end up with `[A, B, D, X]` instead of the desired `[A, B, X, D]`.

We've discussed this `replaceChild` functionality before (e.g. #911 or #976), but it's not that easy.

Now, even if we did have the `replaceChild` functionality, there could still be problems with the "remove C and replace it with X" approach in certain cases. For once, removing a component triggers the `onRemove` callback which may be problematic. In addition, some components are tightly coupled to their parents (for example, they may read their parent's properties or try to affect the parent), so suddenly replacing their parent with `X` may break things.

> And for the other ones, since the onRender replaces the render method call, is it meant that the onRender should then call render after it has performed the changes that it wants to do to the canvas?

In the majority of implementations, yes. But one can easily imagine implementations that ignore the parent's render method altogether. For example, you can render a black "censored" box instead of the component. 

The caching effect is in the middle: it first calls parent's render to record the image on a new canvas, and then during subsequent calls it will ignore the parent's render and just use the generated image.

> It feels like the flow becomes a bit too complicated and a bit overlapping with preRender.

There are several differences compared to preRender:

1. The `onRender` callback can be changed at runtime, for example using an effect. This is powerful, because it allows you to apply changes to any other component.
2. Unlike `preRender/postRender`, the callback can not only change the canvas, but replace it entirely. Which is needed if you want to record an image, for example.


As for the examples given, here are more detailed explanations:

1. **replacing the component with a cached image** -- a common optimization technique, especially if the component is  complex and procedurally-generated -- this works by `onRender` creating a new canvas with a picture recorder, then calling the original `render` on that canvas, storing the result as an image, and finally using that image during all subsequent calls to `onRender`.
2. **generic opacity effect** -- similar to (1), but the final image is drawn with an opacity paint. Note that unlike current `opacityEffect`, this would apply to any component.
3. **generic color tint effect** -- same as (2), only use colored paint when drawing the image. Similar to current `ColorEffect`, but applies to any component.
4. **explosion effect** -- use Voronoi diagram to split the component's rect into multiple shards, call the original render once for every "shard", using that shard as a clip-mask, then make those shards move away from each other over time. This will create an effect where the underlying component breaks into pieces and "explodes". 
5. **duplication effect** -- simply render the component once normally and once more through a combination of other effects. Could be used for example to create a ghost exiting a body, or "drunken vision" effect.
6. **shadow on the ground** -- similar to duplication effect: this will render the component normally, but also render a separate copy through a combination of a color filter, a scale transform, and a skew transform.
7. **edge detection effect** -- based on (1), but then directly manipulate pixels of the resulting image in order to run a custom algorithm such as edge detection. Maybe this can be done easier through a 1-pixel shadow though. But still, there could be effects that you could accomplish with direct pixel manipulation.
8. **wave effect** -- if I want to make a character (or a scene) to become distorted in a wave-like manner, I would split the view into multiple stripes, then call the render method multiple times for every strip shifting and skewing it as appropriate.
9. **head roll effect** -- suppose your character hacks through an army of enemies, wouldn't it be nice to make their heads roll occasionally? The implementation is similar to the explosion effect, but now you have only 2 shards, and one of them falls to the ground and rolls away. (This is obviously a game-specific effect, not gonna add this into Flame).
10. **glow effect** -- similar to (7) but render the background copy with a "glow" blur.
11. any **background effect** -- if you want to draw something underneath a component, but still be logically tied to that component in the game tree. For a example, a selection indicator under character's feet.
12. etc? There are actually quite a lot of things you can do if you allow to hijack the rendering pipeline.

luanpotter:
Can't this easily be accomplished by, instead of,

```dart
component.onRender = <lambda>
```

doing,

```dart
component.flag = true
```

and on your component class:

```dart
void render(Canvas c) {
  if (flag) {
    <lambda>
    return;
  }

  // ... (normal render)
}
```

?

I understand it can be useful and would be a couple lines short, but seems to be a very specific thing and would feel a bit "magical" to have render replaced like that. Sometimes I tend towards slightly more verbose code if it makes things simpler to reason about.

st-pasha:
@luanpotter Your suggestion requires that 
1. the author of the component anticipated the need for that component to use alternative rendering and took steps towards implementing that. This would still require storing the lambda as `component.onRender`, but now it'll be explicit on every component. The problem is that you have to repeat that for every component where you want to apply an advanced effect (which may end up being almost every component);

2. it is still unclear how this can be achieved in a *reusable* way. That is, suppose you want to have a generic opacity effect -- how would such effect know that it needs to set `component.flag=true` and `component.onRender=<lambda>`, unless both `.flag` and `.onRender` are defined in the `Component` class, in which case why not use those properties from the `Component` class as well and make everyone's life easier?

spydon:
Could we instead have a component that wraps the component that we want to affect, which can change its own `renderTree` method and thus can affect the canvas before the child's `render` method?  Not completely satisfied with that solution either though.
I'm thinking if an effect, that needs access to the `Canvas`, is added to a component `c` with parent `p` it can do something like this:
```dart
p.add(wrapper);
c.changeParent(wrapper)
```
So the effect implements the wrapper component.
And then when the effect is done or removed the parent is changed back.

erickzanardo:
One important note about the use cases, we can't use `PictureRecorder` to create effects that will change a lot, like opacity, color, etc. `Picture` is very limited and transforming it to `Image` is an async operation, so doing that a lot in a small timeframe will not be good, not to mention that even if we find a good way to overcame that, we will basically be doing two renderings per frame, one for the base image and on with the applied effect.

That is why we built the paint effects the way we did, with just changing the paint object, we get the effect applied to the component on a single call.

But I see some good use for this idea of pre rendering the image on some other effects, where we need to generate it only once, like a destruction effect for example.

But I prefer the suggestion from @spydon , a wrapper component/effect.

st-pasha:
> So the effect implements the wrapper component. And then when the effect is done or removed the parent is changed back.

This is what was originally discussed in #911. The primary advantage of this approach is that it can control the original effect more fully, i.e. not only its rendering but also updating. I'm not sure what kind of effects would need to affect the component's `update()` method, but who knows? The disadvantages of this approach are more numerous:

1. Removing the component from parent and then re-adding it wrapped, will change the visual order of elements on the screen. For example, if your character is fighting a mob of enemies, then one of them dies and you want to apply some kind of death animation effect, you still wouldn't want that enemy to suddenly jump to the front of the crowd.
2. Adding/removing components may have unintended side effects through the `onRemove/onMount` callbacks. This of course depends on how those components are programmed, but I can imagine scenarios where this could become either a roadblock or a source of hard-to-catch bugs.
3. A component may be tightly coupled with its parent, so having a"wrong" parent may become problematic. For example, a [shield] component may need the `parent.shieldAttachPoint` property in order to position itself properly. When the parent becomes a wrapper, everything breaks.
4. A parent may be tightly coupled with its children. For example, it may want to iterate through its children and look for one with a <Weapon> trait, so that it can properly apply a Disarm spell.

Overall, I think that unexpected manipulations of the game tree can be way more disruptive and unexpected than sly replacement of the `renderTree()` method. While there could probably be scenarios where wrapping the component is exactly what's needed, it's better to make it explicit, so that the programmer can plan around any potential side-effects.

st-pasha:
> so doing that a lot in a small timeframe will not be good, not to mention that even if we find a good way to overcame that, we will basically be doing two renderings per frame, one for the base image and on with the applied effect.

I agree, there will definitely be some performance considerations there, so we will have to keep that in mind when implementing and using those effects. For example, `OpacityEffect`/`ColorEffect` can apply the current approach if they detect that the parent is `HasPaint`, but switch to a more generic approach if the parent is any other component. Similarly, using multiple effects like these, for multiple components, may be infeasible, but this will ultimately be up to the programmer to use their judgment, measure the performance on the target device, and decide where to draw the line.

luanpotter:
@st-pasha , my thoughts on your comments:

> the author of the component anticipated the need for that component

you can always change your components as your project evolves, right? if you are talking about a base component provided by flame, that you are using out-of-the-box, like SpriteComponent, you can always wrap it on your component, add the flag, and pass the method calls to the base sprite component.

> it is still unclear how this can be achieved in a reusable way

that depends on how you want to organize your code, but you can create a base component class for all your components that need to override the render, and put all the functionality with the flag and checks there. you could even have multiple versions of it for different use cases. you can even have a generic wrapper for any Component that adds a custom render behind a flag. so I don't think that needs to live on the core engine...

spydon:
> > So the effect implements the wrapper component. And then when the effect is done or removed the parent is changed back.
> 
> This is what was originally discussed in #911. The primary advantage of this approach is that it can control the original effect more fully, i.e. not only its rendering but also updating. I'm not sure what kind of effects would need to affect the component's `update()` method, but who knows? The disadvantages of this approach are more numerous:
> 
>     1. Removing the component from parent and then re-adding it wrapped, will change the visual order of elements on the screen. For example, if your character is fighting a mob of enemies, then one of them dies and you want to apply some kind of death animation effect, you still wouldn't want that enemy to suddenly jump to the front of the crowd.
> 
>     2. Adding/removing components may have unintended side effects through the `onRemove/onMount` callbacks. This of course depends on how those components are programmed, but I can imagine scenarios where this could become either a roadblock or a source of hard-to-catch bugs.
> 
>     3. A component may be tightly coupled with its parent, so having a"wrong" parent may become problematic. For example, a [shield] component may need the `parent.shieldAttachPoint` property in order to position itself properly. When the parent becomes a wrapper, everything breaks.
> 
>     4. A parent may be tightly coupled with its children. For example, it may want to iterate through its children and look for one with a  trait, so that it can properly apply a Disarm spell.
> 
> 
> Overall, I think that unexpected manipulations of the game tree can be way more disruptive and unexpected than sly replacement of the `renderTree()` method. While there could probably be scenarios where wrapping the component is exactly what's needed, it's better to make it explicit, so that the programmer can plan around any potential side-effects.

Yeah, I agree, that way is not feasible either when thinking about it.
I just really don't like the suggested way, it becomes so much spaghetti code between the effects system and the components. 
But maybe it is the only way. 

However, this doesn't seem to be needed to bring over the current effects to the new system and it won't be a breaking change later, right? If that is the case I vote that we delay this for post-v1 since we shouldn't take in any more features before the release. Then post-v1 we can discuss how to continue with this path.

st-pasha:
> you can create a base component class for all your components that need to override the render, and put all the functionality with the flag and checks there. you could even have multiple versions of it for different use cases. you can even have a generic wrapper for any Component that adds a custom render behind a flag.

@luanpotter  If I understand you correctly, you're suggesting creating some kind of mixin that can be applied to existing components, and which will replace the component's `renderTree` method to allow calling the `onRenderTree` callback function. Then, in order to apply this mixin, the user would have to replace every component in Flame with their own version that uses that mixin, e.g. `MyComponent`, `MyPositionComponent`, `MySpriteComponent`, etc. This would still not allow us to implement any advanced visual effects in Flame, since those effects have no way of knowing and using the programmer's personal mixin. Thus, the programmer would also need to implement all those effects themself, which is, again, quite challenging.

While I agree that it is, of course, possible to do what is described above, I do not see how it is better than what amounts to 2 lines of code as proposed in this PR.

> However, this doesn't seem to be needed to bring over the current effects to the new system and it won't be a breaking change later, right? If that is the case I vote that we delay this for post-v1 since we shouldn't take in any more features before the release. 

@spydon That's right, this is not a breaking change. It may not even qualify as a "feature" either. I would consider it more of a "meta-feature", i.e. functionality that enables creation of new features. So, while this doesn't affect any existing users, it will open up new opportunities for them to build new features (if they depend on the bleeding-edge version of flame).

spydon:
> @spydon That's right, this is not a breaking change. It may not even qualify as a "feature" either. I would consider it more of a "meta-feature", i.e. functionality that enables creation of new features. So, while this doesn't affect any existing users, it will open up new opportunities for them to build new features (if they depend on the bleeding-edge version of flame).

I mean that you would be able to bring over the current effects to the new effect system without this? And that when adding this post-v1 it wouldn't be a breaking change. We should only bring in code to `main` that will be needed for v1 now.

spydon:
Closing this, but opened #1111 where we can discuss this more post-v1.

PR #1099, opened 2021-11-14
Made knob as a non-required field

OlegNovosad:
# Description

Joystick might be loaded asynchronously while being required for other components. Making `knob` a non required make improve the possibility to extend from the `JoystickComponent` and let's say create own _left_ or _right_ joysticks with encapsulated logic. Currently knob is a required field.

## Checklist

- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors and are passing (See [Contributor Guide]). (let me know if I should add tests for missing knob behavior)
- [ x I removed the `Draft` status, by clicking on the `Ready for review` button in this PR.

## Breaking Change

## Related Issues

N/A


packages/flame/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## [next]
+ - Added possibility to extend `JoystickComponent` (made `knob` unrequired)
```
spydon, line 4: This should be in the bottom of the list

packages/flame/lib/src/components/input/joystick_component.dart
```diff
@@ -72,28 +73,30 @@ class JoystickComponent extends HudMarginComponent with Draggable {
   @override
   Future<void> onLoad() async {
     await super.onLoad();
-    knob.anchor = Anchor.center;
-    knob.position.add(size / 2);
-    _baseKnobPosition = knob.position.clone();
+    knob?.anchor = Anchor.center;
+    knob?.position.add(size / 2);
+    _baseKnobPosition = knob?.position.clone() ?? Vector2.all(0);
```
spydon, line 78:
```suggestion
    _baseKnobPosition = knob?.position.clone() ?? Vector2.zero();
```

packages/flame/lib/src/components/input/joystick_component.dart
```diff
@@ -72,28 +73,30 @@ class JoystickComponent extends HudMarginComponent with Draggable {
   @override
   Future<void> onLoad() async {
     await super.onLoad();
-    knob.anchor = Anchor.center;
-    knob.position.add(size / 2);
-    _baseKnobPosition = knob.position.clone();
+    knob?.anchor = Anchor.center;
+    knob?.position.add(size / 2);
```
spydon, line 77:
```suggestion
    knob?.position = size / 2;
```

packages/flame/lib/src/components/input/joystick_component.dart
```diff
@@ -20,7 +20,7 @@ enum JoystickDirection {
 }
 
 class JoystickComponent extends HudMarginComponent with Draggable {
-  late final PositionComponent knob;
+  late final PositionComponent? knob;
```
spydon, line 23: I don't think this should be nullable, but we have to fix that on all other components that have this pattern too. It should just be `late` and then we have to check in the constructor body whether the argument was null or not and if it was we'll assume that it is set in `onLoad`.

OlegNovosad: Should I update the same logic for the background property? Currently it is nullable so far.

packages/flame/CHANGELOG.md
OlegNovosad: Updated

packages/flame/lib/src/components/input/joystick_component.dart
OlegNovosad: After changing knob back to just `late` this change is gone.

packages/flame/lib/src/components/input/joystick_component.dart
OlegNovosad: Done

packages/flame/lib/src/components/input/joystick_component.dart
spydon: Keep them as nullable now, we'll go over all the components and fix this later.

spydon: Ah you already changed the knob, then you can keep that as non-nullable, but the background you can keep as nullable.

packages/flame/lib/src/components/input/joystick_component.dart
```diff
@@ -57,27 +57,36 @@ class JoystickComponent extends HudMarginComponent with Draggable {
...
-          knob.position.isZero() && (background?.position.isZero() ?? true),
+          (knob?.position.isZero() ?? true) &&
+              (background?.position.isZero() ?? true),
           'Positions should not be set for the knob or the background',
         ),
         super(
           margin: margin,
           position: position,
-          size: background?.size ?? Vector2.all(size ?? 0),
+          size: background?.size ?? Vector2.zero(),
```
spydon, line 67: This needs to take the `size` variable into account, so it can be changed back to what it was before the PR.

OlegNovosad: Agree, changed.

packages/flame/lib/src/components/input/joystick_component.dart
OlegNovosad: Made `background` nullable again.

packages/flame/lib/src/components/input/joystick_component.dart
```diff
@@ -45,7 +45,7 @@ class JoystickComponent extends HudMarginComponent with Draggable {
   late Vector2 _baseKnobPosition;
 
   JoystickComponent({
-    required this.knob,
+    PositionComponent? knob,
```
spydon, line 48: Line 87 can fail now:
https://github.com/flame-engine/flame/pull/1099/files#diff-6e372118ee33f5a48d17030bcf2ec58a7dee48383c3bb2c8077a2a9f5a854a57R87
That line should be moved to `onMount`, similarly to how it is done here:
https://github.com/flame-engine/flame/blob/main/packages/flame/lib/src/components/input/hud_button_component.dart#L46-L55

Actually everything in `onLoad` can be moved to `onMount`.

OlegNovosad: So generally I made it similar to `HudButtonComponent`. It makes more sense to have both `knob` and `background` as nullable and assert these in `onMount`.

packages/flame/lib/src/components/input/joystick_component.dart
```diff
@@ -70,29 +72,37 @@ class JoystickComponent extends HudMarginComponent with Draggable {
...
-    knob.position.add(size / 2);
-    _baseKnobPosition = knob.position.clone();
-    if (background != null) {
-      add(background!);
-    }
-    add(knob);
+  @mustCallSuper
+  void onMount() {
+    assert(
+      background != null,
```
spydon, line 78: Background can be null, it is only the knob that is mandatory.

spydon (commented):
Just the last fix on `background` and then I think this is ready. :)

packages/flame/lib/src/components/input/joystick_component.dart
OlegNovosad: Done.

spydon (approved):
Good job! :)

erickzanardo (approved):

PR #1122, opened 2021-11-19
Added the `SizeEffect` and a general `ComponentEffect`

spydon:
# Description

Since the `Transform2DEffect` could be used here I created a more general `ComponentEffect`.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/examples/lib/stories/effects/old_size_effect_example.dart
+++ b/examples/lib/stories/effects/old_size_effect_example.dart
@@ -1,15 +1,14 @@
 import 'package:flame/components.dart';
+import 'package:flame/effects.dart';
 import 'package:flame/extensions.dart';
 import 'package:flame/game.dart';
 import 'package:flame/input.dart';
 import 'package:flame/palette.dart';
-import 'package:flame/src/effects2/size_effect.dart'; // ignore: implementation_imports
-import 'package:flame/src/effects2/standard_effect_controller.dart'; // ignore: implementation_imports
 import 'package:flutter/material.dart';
 
 import '../../commons/square_component.dart';
 
-class SizeEffectExample extends FlameGame with TapDetector {
+class OldSizeEffectExample extends FlameGame with TapDetector {
   static const String description = '''
     The `SizeEffect` changes the size of the component, the sizes of the
     children will stay the same.
@@ -37,14 +36,11 @@ class SizeEffectExample extends FlameGame with TapDetector {
     final s = grow ? 300.0 : 100.0;
 
     grow = !grow;
-
     square.add(
-      SizeEffect.to(
-        Vector2.all(s),
-        StandardEffectController(
-          duration: 1.5,
-          curve: Curves.bounceInOut,
-        ),
+      SizeEffect(
+        size: Vector2.all(s),
+        speed: 250.0,
+        curve: Curves.bounceInOut,
       ),
     );
   }
--- a/examples/lib/stories/effects/old_size_effect_example.dart
+++ b/examples/lib/stories/effects/old_size_effect_example.dart
@@ -10,9 +10,9 @@ import '../../commons/square_component.dart';
 
 class OldSizeEffectExample extends FlameGame with TapDetector {
   static const String description = '''
-    The `SizeEffect` changes the size of the component, the sizes of the
+    The `SizeEffect` changes the size of the component, but the sizes of its
     children will stay the same.
-    In this example you can tap the screen and the component will size up or
+    In this example, you can tap the screen and the component will size up or
     down, depending on its current state.
   ''';
 
--- a/packages/flame/lib/src/effects2/component_effect.dart
+++ b/packages/flame/lib/src/effects2/component_effect.dart
@@ -4,7 +4,7 @@ import '../../components.dart';
 import 'effect.dart';
 import 'effect_controller.dart';
 
-/// Base class for effects that target a [Component].
+/// Base class for effects that target a [Component] of type [T].
 ///
 /// A general abstraction for creating effects targeting [Component]s, currently
 /// only used by `SizeEffect` and `Transform2DEffect`.
--- a/packages/flame/lib/src/effects2/component_effect.dart
+++ b/packages/flame/lib/src/effects2/component_effect.dart
@@ -25,9 +25,7 @@ abstract class ComponentEffect<T extends Component> extends Effect {
     if (parent is T) {
       target = parent! as T;
     } else {
-      throw UnsupportedError(
-        'Can only apply this effect to $T',
-      );
+      throw UnsupportedError('Can only apply this effect to $T');
     }
   }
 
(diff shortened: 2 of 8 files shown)


spydon:
Ping @st-pasha

erickzanardo (approved):

examples/lib/stories/effects/old_size_effect_example.dart
```diff
@@ -0,0 +1,47 @@
...
+import 'package:flame/game.dart';
+import 'package:flame/input.dart';
+import 'package:flame/palette.dart';
+import 'package:flutter/material.dart';
+
+import '../../commons/square_component.dart';
+
+class OldSizeEffectExample extends FlameGame with TapDetector {
+  static const String description = '''
+    The `SizeEffect` changes the size of the component, the sizes of the
```
luanpotter, line 13:
```suggestion
    The `SizeEffect` changes the size of the component, but the sizes of its
```

examples/lib/stories/effects/old_size_effect_example.dart
```diff
@@ -0,0 +1,47 @@
...
+    children will stay the same.
+    In this example you can tap the screen and the component will size up or
```
luanpotter, line 15:
```suggestion
    In this example, you can tap the screen and the component will size up or
```

luanpotter (approved):
LGTM

packages/flame/lib/src/effects2/transform2d_effect.dart
```diff
@@ -14,41 +12,14 @@ import 'effect_controller.dart';
 /// Currently this class only supports being attached to [PositionComponent]s,
 /// but in the future it will be extended to work with any [Transform2D]-based
 /// classes.
-abstract class Transform2DEffect extends Effect {
+abstract class Transform2DEffect extends ComponentEffect<PositionComponent> {
```
st-pasha, line 15: I don't think this is correct: a `Transform2DEffect` targets a `Transform2D` property, which doesn't have to be on a `PositionComponent`. For example, I want the same effects to be able to target a `Camera` (once we refactor it to use the `Transform2D` internally). There could be other use cases too down the line.

spydon: Right now the `Transform2DEffect` is using the `PositionComponent` to reach to the `Transform2D` though, and later we can refactor it to somehow differently reach the `Transform2D`. The most common use case will be through a component though, and later maybe the `Camera` will be a component that has a `Transform2D` too even.

st-pasha: I mean, currently (in the `main` branch), `Transform2DEffect` is a class that already does its job, and also can be trivially extended to support `Camera` and other use cases. 

With your suggested changes, it can no longer be extended to support `Camera` and other use cases without doing a substantial refactoring, which would probably involve reverting to the state that the class is currently in.

At a fundamental level, `Transform2DEffect` is not a `ComponentEffect` because it does not target a component -- it targets a specific property, and it just so happens that `PositionComponent` has that property.

spydon: As soon as you want to support `Transform2D` without a component, you just stop extending `ComponentEffect`. But right now it is tightly coupled with `PositionComponent` anyways, so why have code duplication?

spydon: @st-pasha does it make sense? On main `Transform2DEffect` has this right now:
```dart
    if (parent is PositionComponent) {
      target = (parent! as PositionComponent).transform;
    }
    // TODO(st-pasha): add Camera support once it uses Transform2D
    else {
      throw UnsupportedError(
        'Can only apply a Transform2DEffect to a PositionComponent class',
      );
    }
```

packages/flame/lib/src/effects2/component_effect.dart
```diff
@@ -0,0 +1,44 @@
+import 'package:flutter/cupertino.dart';
+
+import '../../components.dart';
+import 'effect.dart';
+import 'effect_controller.dart';
+
+/// Base class for effects that target a [Component].
```
st-pasha, line 7: `... that target components of type [T]`.

packages/flame/lib/src/effects2/size_effect.dart
```diff
@@ -0,0 +1,46 @@
+import '../../components.dart';
+import '../../extensions.dart';
+import 'component_effect.dart';
+import 'effect_controller.dart';
+
+/// Change the size of a component over time.
+///
+/// The following constructors are provided:
+///
+///   - [SizeEffect.by] will set the size in relation to it's current size;
```
st-pasha, line 10: `it's` -> `its`.
Also, this could use some more detailed explanation. Is the effect going to take the current size and add some additional size to it? Say, if the current size is `(100,100)` and you do `SizeEffect(10,-10)` -- will it end up having size `(110,90)`?

packages/flame/lib/src/effects2/size_effect.dart
```diff
@@ -0,0 +1,46 @@
+import '../../components.dart';
+import '../../extensions.dart';
+import 'component_effect.dart';
+import 'effect_controller.dart';
+
+/// Change the size of a component over time.
```
st-pasha, line 6: Will this work with a `ShapeComponent` too?

I wonder if `ComponentEffect<PositionComponent>` is the correct super-class even for this effect. One could easily imagine non-PositionComponent classes having the notion of size. Say, a `Viewport`?

packages/flame/lib/src/effects2/size_effect.dart
```diff
@@ -0,0 +1,46 @@
...
+
+  factory SizeEffect.to(Vector2 targetSize, EffectController controller) =>
+      _SizeToEffect(targetSize, controller);
+
+  Vector2 _offset;
+
+  @override
+  void apply(double progress) {
+    final dProgress = progress - previousProgress;
+    target.size += _offset * dProgress;
```
st-pasha, line 29: I wonder what would happen if the size becomes negative, and whether there is something we could do about it?
Obviously, reduction in component's size should be allowed, but this could easily drive the size into the negative range (especially if multiple effects are applied), and I wonder if it would break anything?

packages/flame/lib/src/effects2/size_effect.dart
spydon: It will work on the `ShapeComponent`, since it is a `PositionComponent`. Since right now the `size` is defined on `PositionComponent` I think it will be enough for now, we can evolve that later if we would see the need to have these effects on other things, like the viewport.

packages/flame/lib/src/effects2/size_effect.dart
spydon: It probably would have some weird consequences, I'll cap it at 0.

PR #1133, opened 2021-11-22
`OpacityEffect` in the new effects engine

spydon:
# Description

The `OpacityEffect` is a bit tricky to compose since the opacity only is stored in 8 bits, so it can get pretty big numeric discrepancies, so we have to accumulate the rounding errors. 

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/examples/lib/stories/effects/opacity_effect_example.dart
+++ b/examples/lib/stories/effects/opacity_effect_example.dart
@@ -31,8 +31,7 @@ class OpacityEffectExample extends FlameGame with TapDetector {
         position: Vector2(300, 100),
         size: Vector2(149, 211),
       )..add(
-          OpacityEffect.to(
-            0.0,
+          OpacityEffect.fadeOut(
             StandardEffectController(
               duration: 1.5,
               reverseDuration: 1.5,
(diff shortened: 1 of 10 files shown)


spydon:
@st-pasha any idea how we could handle rounding errors like the ones we will get in here when the affected storage is only 8 bits, so that the rounding errors can become quite big?

examples/lib/stories/effects/opacity_effect_example.dart
```diff
@@ -43,10 +46,10 @@ class OpacityEffectExample extends FlameGame with TapDetector {
   @override
   void onTap() {
     final opacity = sprite.paint.color.opacity;
-    if (opacity == 1) {
-      sprite.add(OpacityEffect.fadeOut());
-    } else if (opacity == 0) {
-      sprite.add(OpacityEffect.fadeIn());
+    if (opacity >= 0.5) {
+      sprite.add(OpacityEffect.fadeOut(StandardEffectController(duration: 1)));
```
erickzanardo, line 50: This line is a little bit long now, I wonder if we couldn't create shortcuts or something (not for this PR ofc)

erickzanardo (approved):

st-pasha:
> any idea how we could handle rounding errors like the ones we will get in here when the affected storage is only 8 bits, so that the rounding errors can become quite big?

We could store double-precision version of a variable within the effect, and then send into the component the rounded value. Also, we'll need to keep track of any external updates to the component's value, and apply it to our double-precision internal value.

spydon:
> We could store double-precision version of a variable within the effect, and then send into the component the rounded value. Also, we'll need to keep track of any external updates to the component's value, and apply it to our double-precision internal value.

If we just take the delta of the internal value each tick it should be fine without keeping track of other updates, right?

st-pasha:
> If we just take the delta of the internal value each tick it should be fine without keeping track of other updates, right?

There is always a possibility that another effects updates alpha, so need to keep track of that.

spydon:
> There is always a possibility that another effects updates alpha, so need to keep track of that.

So you mean the internal double wouldn't be on the effect, but somewhere else? I was thinking that out would be on the effect, and that we only update with the delta from the last internal double. Then they mix their updates, but maybe it wouldn't be any more acccurate than now, since rounding errors would still have to be taken into consideration when converting it to alpha. Maybe the current solution is good enough?

st-pasha:
We can merge it as-is, if it works; and then think about possible improvements later.

PR #1146, opened 2021-11-26
Added a ButtonComponent

spydon:
# Description

Since we have a `HudButtonComponent` I'm adding this one for consistency.
If you haven't used the `HudButtonComponent` it is simply two `PositionComponent`s, one that shows when the button is pressed and one that shows otherwise.

https://user-images.githubusercontent.com/744771/143618565-4e4df42a-4fed-4248-9a6d-770151386d66.mp4

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/lib/src/components/input/button_component.dart
+++ b/packages/flame/lib/src/components/input/button_component.dart
@@ -1,4 +1,3 @@
-import 'package:flutter/rendering.dart' show EdgeInsets, VoidCallback;
 import 'package:meta/meta.dart';
 
 import '../../../components.dart';
@@ -18,7 +17,7 @@ class ButtonComponent extends PositionComponent with Tappable {
   /// Callback for what should happen when the button is pressed.
   /// If you want to interact with [onTapUp] or [onTapCancel] it is recommended
   /// to extend [ButtonComponent].
-  VoidCallback? onPressed;
+  void Function()? onPressed;
 
   ButtonComponent({
     this.button,
--- a/packages/flame/lib/src/components/input/button_component.dart
+++ b/packages/flame/lib/src/components/input/button_component.dart
@@ -1,0 +1,86 @@
+import 'package:flutter/rendering.dart' show EdgeInsets, VoidCallback;
+import 'package:meta/meta.dart';
+
+import '../../../components.dart';
+import '../../../extensions.dart';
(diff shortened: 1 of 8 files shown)


packages/flame/lib/src/components/input/button_component.dart
```diff
@@ -0,0 +1,85 @@
...
+import '../../../input.dart';
+import '../../gestures/events.dart';
+
+/// The [ButtonComponent] bundles two [PositionComponent]s, one that shows while
+/// the button is being pressed, and one that shows otherwise.
+///
+/// Note: You have to set the [button] in [onLoad] if you are not passing it in
+/// through the constructor.
+class ButtonComponent extends PositionComponent with Tappable {
+  late final PositionComponent? button;
```
luanpotter, line 14: can this be non-nullable?

luanpotter (approved):
I'm a bit confused on whats the difference then between ButtonComp and HudButtonComp? maybe something to explain on their docs in case people find the wrong one

packages/flame/lib/src/components/input/button_component.dart
spydon: No, because then we can't do an assertion on `onMount`.

spydon:
> I'm a bit confused on whats the difference then between ButtonComp and HudButtonComp? maybe something to explain on their docs in case people find the wrong one

`HudButtonComponent` is a hud and its position will be static to the margin that you have set, or if you set a position it will be calculate the margin and use that if the screen resizes. Which is how all `HudMarginComponent`s work.
Meanwhile `ButtonComponent` is just a normal `PositionComponent`, and same with `SpriteButtonComponent`.

packages/flame/lib/src/components/input/button_component.dart
st-pasha: Do we need an assertion though? If a variable is declared `late final` and you use it before it is initialized, Dart will through an assertion error for you automatically.

spydon: Those runtime errors are hard for the user to understand though, but maybe they are enough.
If we decide those are enough, then we need to change on all other `PositionComponent`s that can be initialized like this too, so it is outside of the scope of this PR since it only follows the current structure.

PR #1148, opened 2021-11-27
Fix render order of components and add tests

luanpotter:
# Description

[I am porting](https://github.com/bluefireteam/gravitational_waves/pull/4) Gravity Runner (aka Gravitational Waves) to the latest Flame.
I think it will be very good to have a fully functional, complete big game example before release.
Most of it was painless or even deleting stuff that we had incorporated from GR to Flame (like camera, viewport, etc).
However, I did come out with a list of things that were easier to do before. So I am putting out a few PRs.
And also I found this bug where component rendering order was not respected for isHud components.
This fixes that but in order to add a test I am modifying `canvas_test` to better test save/restore: https://github.com/bluefireteam/canvas_test/pull/3

## Checklist

## Breaking Change

### DIFF
--- a/packages/flame/test/components/position_component_test.dart
+++ b/packages/flame/test/components/position_component_test.dart
@@ -632,7 +632,8 @@ void main() {
           ..drawLine(const Offset(0, -2), const Offset(0, 2))
           ..drawLine(const Offset(-2, 0), const Offset(2, 0))
           ..drawParagraph(null, const Offset(-30, -15))
-          ..drawParagraph(null, const Offset(-20, 10)),
+          ..drawParagraph(null, const Offset(-20, 10))
+          ..translate(0, 0), // canvas.restore
       );
     });
 
@@ -650,7 +651,8 @@ void main() {
           ..translate(18, 12)
           ..drawRect(const Rect.fromLTWH(0, 0, 10, 10))
           ..drawLine(const Offset(5, 3), const Offset(5, 7))
-          ..drawLine(const Offset(3, 5), const Offset(7, 5)),
+          ..drawLine(const Offset(3, 5), const Offset(7, 5))
+          ..translate(0, 0), // canvas.restore
       );
     });
   });
--- a/packages/flame/test/game/component_rendering_test.dart
+++ b/packages/flame/test/game/component_rendering_test.dart
@@ -14,7 +14,7 @@ class _MyComponent extends Component {
   bool isHud;
 
   _MyComponent(int priority, {this.isHud = false}) : super(priority: priority);
- 
+
   @override
   void render(Canvas canvas) {
     final p = Vector2Extension.fromInts(priority, priority);
@@ -49,7 +49,7 @@ void main() {
         );
       },
     );
-  
+
     flameGame.test(
       'only HUD components',
       (game) async {
@@ -71,7 +71,7 @@ void main() {
         );
       },
     );
-  
+
     flameGame.test(
       'mixed',
       (game) async {
--- a/packages/flame/test/game/component_rendering_test.dart
+++ b/packages/flame/test/game/component_rendering_test.dart
@@ -1,0 +1,105 @@
+import 'dart:ui';
+
+import 'package:canvas_test/canvas_test.dart';
+import 'package:flame/components.dart';
+import 'package:flame/game.dart';
+import 'package:flame/palette.dart';
+import 'package:flame_test/flame_test.dart';
+import 'package:flutter/rendering.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+class _MyComponent extends Component {
+  @override
+  bool isHud;
+
+  _MyComponent(int priority, {this.isHud = false}) : super(priority: priority);
+  @override
+  void render(Canvas canvas) {
+    final p = Vector2Extension.fromInts(priority, priority);
+    final s = Vector2.all(1.0);
+    canvas.drawRect(p & s, BasicPalette.white.paint());
+  }
+}
+
+void main() {
+  group('components are rendered according to their priorities', () {
+    flameGame.test(
+      'only camera components',
+      (game) async {
+        await game.ensureAddAll([
+          _MyComponent(4),
+          _MyComponent(1),
+          _MyComponent(2),
+        ]);
+
+        final canvas = MockCanvas();
+        game.camera.snapTo(Vector2(12.0, 18.0));
+        game.render(canvas);
+
+        expect(
+          canvas,
+          MockCanvas()
+            ..translate(-12.0, -18.0)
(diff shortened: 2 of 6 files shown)


spydon (approved):
Lgtm, what happens when a hud component is a child of a non-hud component?

luanpotter:
@spydon that is a good question. I believe isHud is only considered for top level components. but that is existing behaviour, that we have to consider when we fully fix the camera. other point is that there is no way to bypass the viewport

luanpotter:
(this requires the new version of canvas_test to pass)

packages/flame/test/game/component_rendering_test.dart
```diff
@@ -0,0 +1,105 @@
...
+import 'package:flame_test/flame_test.dart';
+import 'package:flutter/rendering.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+class _MyComponent extends Component {
+  @override
+  bool isHud;
+
+  _MyComponent(int priority, {this.isHud = false}) : super(priority: priority);
```
erickzanardo, line 16:
```suggestion
  _MyComponent(int priority, {this.isHud = false}) : super(priority: priority);
 
```

packages/flame/test/game/component_rendering_test.dart
```diff
@@ -0,0 +1,105 @@
...
+          canvas,
+          MockCanvas()
+            ..translate(-12.0, -18.0)
+            ..drawRect(const Rect.fromLTWH(1, 1, 1, 1))
+            ..drawRect(const Rect.fromLTWH(2, 2, 1, 1))
+            ..drawRect(const Rect.fromLTWH(4, 4, 1, 1))
+            ..translate(0.0, 0.0),
+        );
+      },
+    );
```
erickzanardo, line 50:
```suggestion
    );
  
```

packages/flame/test/game/component_rendering_test.dart
```diff
@@ -0,0 +1,105 @@
...
+
+        expect(
+          canvas,
+          MockCanvas()
+            ..drawRect(const Rect.fromLTWH(1, 1, 1, 1))
+            ..drawRect(const Rect.fromLTWH(2, 2, 1, 1))
+            ..drawRect(const Rect.fromLTWH(4, 4, 1, 1)),
+        );
+      },
+    );
```
erickzanardo, line 71:
```suggestion
    );
  
```

erickzanardo (approved):

st-pasha:
I am not sure what bug is being fixed here?

The `isHud` components are deliberately rendered after the non-hud components, so that they would visually appear over of the game world.

packages/flame/test/components/position_component_test.dart
```diff
@@ -632,7 +632,8 @@ void main() {
           ..drawLine(const Offset(0, -2), const Offset(0, 2))
           ..drawLine(const Offset(-2, 0), const Offset(2, 0))
           ..drawParagraph(null, const Offset(-30, -15))
-          ..drawParagraph(null, const Offset(-20, 10)),
+          ..drawParagraph(null, const Offset(-20, 10))
+          ..translate(0, 0), // canvas.restore
```
st-pasha, line 636: why is this needed?

luanpotter:
@st-pasha  `isHud` just means do not respect camera. they should be rendered according to priority order, which was broken at some point, causing issues with hud components on Gravity.

packages/flame/test/components/position_component_test.dart
luanpotter: after the render cycle our implementation resets the camera, so this represents that change. we could ignore a final translate(0, 0) but I think it makes sense to add it to the tests as it does happen in reality

st-pasha:
@luanpotter There was a bug not too long ago (#934) about the fact that `isHud` components must be rendered on top of the game world ignoring the priority, and they weren't. So this PR would re-introduce that bug.

I understand that there could be compelling reasons to render hud components inter-mixed with world components, but I cannot come up with such scenarios. Could you please clarify the use-case that you had in mind for this PR?

spydon:
> @luanpotter There was a bug not too long ago (#934) about the fact that `isHud` components must be rendered on top of the game world ignoring the priority, and they weren't. So this PR would re-introduce that bug.
> 
> I understand that there could be compelling reasons to render hud components inter-mixed with world components, but I cannot come up with such scenarios. Could you please clarify the use-case that you had in mind for this PR?

It was my fault opening that bug, `isHud` should really be named `ignoreCamera` and later we should introduce `ignoreViewport` too. The parallax is an example for when it is needed.

st-pasha:
> isHud should really be named ignoreCamera and later we should introduce ignoreViewport too.

"isHud" means the component is part of a "heads-up display", a UI that renders on top of the world. This is how it is currently defined, and used. Modifying that definition is a huge breaking change at this point.

Now, I agree with you that "isHud" is not a very good property. In particular, it only works for top-level components, which is a sure sign that it cannot be a generic property defined for every component.

However, I don't agree about renaming it, or introducing `ignoreViewport`: such properties would again suffer from the same problem that they won't work anywhere except at the top level of the game tree. Instead of encumbering `Component` class with more and more stuff, we should explore alternative solutions tied to the Camera / Viewport class.

> The parallax is an example for when it is needed.

`ParallaxComponent` is a very special component indeed. And it is long overdue for refactoring to make it more useful. It is definitely not a Hud component. However, it does interact with a camera in its own special way. It doesn't mean that the Camera should be aware of the ParallaxComponent though, instead the ParallaxComponent should be aware of the camera.

luanpotter:
`isHUD` might be a bad naming, but the intent was always to just ignore the camera. That is what it should do. We already have priority to define priorities. It makes no sense to couple this concepts in an irreparable way, and it causes issues such as hacks in our camera system to account for parallax.

If you look at Gravity there are a handful of examples of HUD components that should definitely NOT be on top; thus upgrading flame broke all this use cases. If do you want your HUDs it to be on top just set the priority (we have that on Gravity too); the other way around, it's impossible to work around.

I agree with @spydon the name is not ideal and should be replaced with respectCamera and a respectViewport should be introduced, which is another issue that I pointed out (no way to do it right now).

But for now, I believe this PR as-is is a step forward as it's fixing a current bug.

We can consider the followups later, as I also agree with @st-pasha that the respect* properties are not perfect bc they are only relevant to top level components, but not sure what change we can easily do instead to properly fix this. Since we have a very tight timeline for v1 we might wanna go with that unless we can think of a better solution that is achievable rn.

st-pasha:
> If you look at Gravity there are a handful of examples of HUD components that should definitely NOT be on top.

Could you please post some links to screenshots or videos of how this looks like? I'd like to understand the use case(s) here better.

PR #1150, opened 2021-11-28
[DO NOT MERGE] Broken test framework!

luanpotter:
This is not a PR to be merged; rather it's a MWE reproducing a terrible, terrible bug.

I don't know the reason, but if we have several await's inside a test, the test framework is not awaiting until the end.
Runt the test added:

1. via `f test test/timer_test.dart`
2. via `f test` (run all tests)
3. via IDE
4. github actions are green

In all cases, the test go up to a certain number, never past 8. And therefore the test passes, because the assertions in the end never run.

Unless I am doing something really stupid, this could be a MASSIVE bug with flame_test (but more likely with flutter_test itself), that compromises our entire test suit and stability.

I'm using flutter stable btw. I will try to repro this with flutter test directly, bypassing flame_test -- I imagine the issue has to come from below as we don't seem to do anything at the flame_test level that could cause this.

### DIFF
--- a/packages/flame/test/timer_test.dart
+++ b/packages/flame/test/timer_test.dart
@@ -1,4 +1,6 @@
+import 'package:flame/components.dart';
 import 'package:flame/timer.dart';
+import 'package:flame_test/flame_test.dart';
 import 'package:test/test.dart';
 
 void main() {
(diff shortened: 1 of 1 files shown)


erickzanardo:
Can you reproduce the same thing using flameWidgetTest?

luanpotter:
Do you mean `widgetTest`? that one seems to work, as far as I am testing

erickzanardo:
I think so, The one that uses GameWidget, and not only the Game Class, I don't quite remember the right method. 

This maybe be worth having an issue open, I would enjoy investigating this.

luanpotter:
Please do, I am at odds.

I wonder if `test` simply doesn't support async. Looking at the signature turns out dart is not super type safe wrt to whether lambdas are async or not. Test just expect a lambda that returns dynamic instead of a non-future T. but looking at the code for it, it does await the lambda and seem to be build in a way that would implicate this should work.

Also testWidget (which does explicitly expect a future lambda) just calls test just like we do.

Finally, it seems to be random how long it rusn for; different methods give different results and even the same method varies. I even got it to run to the end and fail once (not sure how).

erickzanardo:
Closing in favor of #1166

Issue #1154, opened 2021-11-29
Parallax Component does not work with more than 3 images


tiofabby:
Hi,
I am running flame: ^1.0.0-releasecandidate.17 and implementing parallax component with 3 images with following code, it works fine. Images are running faster from background to front.

If I uncomment commented lines and add one or 2 images, velocity gets weird and background goes faster than front images!

Code here:
```
class MyParallaxComponent extends ParallaxComponent {
  @override
  Future<void> onLoad() async {
    parallax = await gameRef.loadParallax([
      ParallaxImageData('parallax-background/spring/bgr_spring1.png'),
      ParallaxImageData('parallax-background/spring/bgr_spring2.png'),
      ParallaxImageData('parallax-background/spring/bgr_spring3.png'),
      // ParallaxImageData('parallax-background/spring/bgr_spring4.png'),
      // ParallaxImageData('parallax-background/spring/bgr_spring5.png')
    ],
        baseVelocity: Vector2(10, 0),
        velocityMultiplierDelta: Vector2(1.3, 1.0));
  }
}

class MyGame extends FlameGame {
  @override
  Future<void> onLoad() async {
    await super.onLoad();
    add(MyParallaxComponent());
  }
}
```

spydon:
Hi,
the examples are also running with 5 images and I can't see any problems in those ones:
https://examples.flame-engine.org/#/Parallax_Basic

Are your images of the same size?

tiofabby:
Hi @spydon
I tried with the images from the example you provided: (https://github.com/flame-engine/flame/tree/main/examples/assets/images/parallax) and I saw that the 2 first background images are half width compared to the 3 other front images.
So I tried to do the same for my own images: I cut half width my 2 first background images and... it worked!
So I guess this could be a temporary workaround to fix such issue... as all images should be same size and velocity should work correctly.
Thank you
Fabrice

spydon:
You shouldn't need to cut the first images, I just tried with 5 identical images, and it works as intended.
Could you share a minimal reproducible example in a repository that I can run?

tiofabby:
Hi @spydon , First, Thank you for spending time on such issue. I created then a very small repo running parallax for the 5 images I have on my side:

https://github.com/tiofabby/debugflame.git

When I run it, background 3 and 4 are going slower than 1, 2 and 5 which makes parallax effect kind of weird.
Thank you

spydon:
Alright, I think I know what is happening.
[...]

PR #1155, opened 2021-11-30
Fix parallax loading with different loading times

spydon:
# Description

Fix parallax loading with different loading times

## Checklist

## Breaking Change

## Related Issues

#1154 

### DIFF
--- a/packages/flame/lib/src/parallax.dart
+++ b/packages/flame/lib/src/parallax.dart
@@ -485,18 +485,18 @@ class Parallax {
     var depth = 0;
     final layers = await Future.wait<ParallaxLayer>(
       dataList.map((data) async {
-        final renderer = await data.load(
-          repeat,
-          alignment,
-          fill,
-          images,
-        );
         final velocityMultiplier =
             List.filled(depth, velocityDelta).fold<Vector2>(
           velocityDelta,
           (previousValue, delta) => previousValue.clone()..multiply(delta),
         );
         ++depth;
+        final renderer = await data.load(
+          repeat,
+          alignment,
+          fill,
+          images,
+        );
         return ParallaxLayer(
           renderer,
           velocityMultiplier: velocityMultiplier,
--- a/packages/flame/lib/src/parallax.dart
+++ b/packages/flame/lib/src/parallax.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:ui';
 
+import 'package:collection/collection.dart';
 import 'package:flutter/painting.dart';
 
 import '../game.dart';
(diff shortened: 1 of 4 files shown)


packages/flame/lib/src/parallax.dart
```diff
@@ -485,18 +485,18 @@ class Parallax {
...
-          fill,
-          images,
-        );
         final velocityMultiplier =
             List.filled(depth, velocityDelta).fold<Vector2>(
           velocityDelta,
           (previousValue, delta) => previousValue.clone()..multiply(delta),
         );
         ++depth;
+        final renderer = await data.load(
```
erickzanardo, line 494: how come this fixes the issue? You just moved the code, am I missing something?

spydon: Because now it gets its velocity before the loading of the image, so the time it takes to load the image won't have any effect on which velocity it gets.

spydon: Since the function is async it will continue with the other layers when it has started the loading, so layers out of order could get their velocity before an earlier layer, if that makes it clearer?

erickzanardo: but nothing inside this closure mutates anything on the outer scope, the only thing outside is `velocityDelta`, which is final (or no modified). It shouldn't matter if you await on the beginning or on the end of the method.

spydon: It does, it mutates `depth`, which is used while calculating the `velocityMultiplier`.

spydon: Made it cleaner, so that it uses `mapIndexed` instead.

erickzanardo (approved):
Maybe we could add a test for the change?

spydon:
> Maybe we could add a test for the change?

Is it worth it? We'll have to emulate different loading times etc.

erickzanardo:
> > Maybe we could add a test for the change?
> 
> Is it worth it? We'll have to emulate different loading times etc.

Since this is fixing an actual bug, I do find it worth, so we don't fall into this again.

Either way I already approved so not a block, just something good to have.

packages/flame/lib/src/parallax.dart
```diff
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:ui';
 
+import 'package:collection/collection.dart';
```
luanpotter, line 4: is this a new dependency? I am weary of adding deps for `flame` itself just for sugars, as flame itself is a library used by other people. also if we were to I would prefer to add one, I would prefer dartlin :p

spydon: No, this is part of dart. It's weird that it is not part of the core collection lib, but I think it has something to do with the null-safe transition. There are all kinds of great extension collection methods in there.

spydon: Ah, it's part of Flutter, not dart.

luanpotter: oh wow, that is really useful to know, I had no idea that was available. cool stuff!

PR #1158, opened 2021-12-02
Introduce updateTree

### LINKED ISSUE #1115: Add method Component.updateTree
Method `updateTree` will relate to `update` as `renderTree` relates to `render` right now.

Reasons:
1. Symmetry. It's easier to understand the component's lifecycle if `render` and `update` are used in exactly the same manner.
2. No need to mark the update method as `@mustCallSuper` anymore. Thus, no need to write `super.update(dt)` every time.
3. More straightforward to implement methods that need to manipulate the update pipeline of their children.

spydon:
# Description

Since we have `renderTree`, this PR introduces `updateTree`.
I'm not fully happy with the solution, but it is better than changing `GameRenderBox` to call `updateTree` and `renderTree`.
Any other ideas of how to solve this is appreciated.

## Checklist

## Breaking Change

## Related Issues

Fixes #1115 

### DIFF
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -124,9 +124,14 @@ class Component with Loadable {
   /// each one takes to update adds up to the next update cycle.
   void update(double dt) {}
 
-  void updateTree(double dt, {bool callUpdate = true}) {
+  /// This method traverses the component tree and calls [update] on all its
+  /// children according to their [priority] order, relative to the
+  /// priority of the direct siblings, not the children or the ancestors.
+  /// If you call this method from [update] you need to set [callOwnUpdate] to
+  /// false so that you don't get stuck in an infinite loop.
+  void updateTree(double dt, {bool callOwnUpdate = true}) {
     children.updateComponentList();
-    if (callUpdate) {
+    if (callOwnUpdate) {
       update(dt);
     }
     children.forEach((c) => c.updateTree(dt));
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -122,10 +122,14 @@ class Component with Loadable {
   /// your state considering this.
   /// All components in the tree are always updated by the same amount. The time
   /// each one takes to update adds up to the next update cycle.
-  @mustCallSuper
-  void update(double dt) {
+  void update(double dt) {}
+
+  void updateTree(double dt, {bool callUpdate = true}) {
     children.updateComponentList();
-    children.forEach((c) => c.update(dt));
+    if (callUpdate) {
+      update(dt);
+    }
+    children.forEach((c) => c.updateTree(dt));
   }
 
   void render(Canvas canvas) {}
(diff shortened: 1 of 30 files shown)


packages/flame/lib/src/game/flame_game.dart
```diff
@@ -78,7 +75,7 @@ class FlameGame extends Component with Game {
   @override
   @mustCallSuper
   void update(double dt) {
-    super.update(dt);
+    super.updateTree(dt, callUpdate: false);
```
st-pasha, line 78: Instead of having a special parameter `callUpdate`, which is used just in this class, it's cleaner to simply re-define `updateTree()` within this class.

packages/flame/lib/src/components/component.dart
```diff
@@ -122,10 +122,14 @@ class Component with Loadable {
...
-  void update(double dt) {
+  void update(double dt) {}
+
+  void updateTree(double dt, {bool callUpdate = true}) {
     children.updateComponentList();
-    children.forEach((c) => c.update(dt));
+    if (callUpdate) {
+      update(dt);
+    }
+    children.forEach((c) => c.updateTree(dt));
```
st-pasha, line 132: The question is whether the current component should be updated before or after the children components.

packages/flame/lib/src/components/component.dart
```diff
@@ -122,10 +122,14 @@ class Component with Loadable {
   /// your state considering this.
   /// All components in the tree are always updated by the same amount. The time
   /// each one takes to update adds up to the next update cycle.
-  @mustCallSuper
-  void update(double dt) {
+  void update(double dt) {}
+
+  void updateTree(double dt, {bool callUpdate = true}) {
```
st-pasha, line 127: We should definitely have a doc-comment here, because the purpose of this method (and how it is different from `update()`) will not be immediately clear to a reader.

st-pasha:
I think you need to write `Closes #1115` or `Fixes #1115` in the PR description in order for the issue to be closed automatically when the PR is merged.

packages/flame/lib/src/components/component.dart
spydon: Before I would say, since it is more likely that the children depend on the state of their parent than the other way around..

packages/flame/lib/src/game/flame_game.dart
spydon: That won't work if the game is not the root component, realized that this won't work in that case either since `updateTree` will then be called twice.

st-pasha: I think this would work then:
```dart
  @override
  void update(double dt) {
    if (parent == null) {
      children.updateComponentList();
      children.forEach((c) => c.updateTree(dt));
    }
    _cameraWrapper.update(dt);
  }

  @override
  void updateTree(double dt) => update(dt);
```

spydon: Isn't it nicer to not duplicate the code, in case we would change component.updateTree in the future? Did you see the update? This comment is on an outdated file.

erickzanardo (approved):
LGTM

Issue #1151, opened 2021-11-28
Broken test framework


luanpotter:
Writing this just to keep track of this issue.
All details I have so far are here: https://github.com/flame-engine/flame/pull/1150

PR #1166, opened 2021-12-05
Fixing tests with multiple futures finishing abruptly

erickzanardo:
# Description

Fixes an issue where our test framework sometimes finished abruptly, without running the whole test when there were too many futures on it.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame_test/lib/src/flame_test.dart
+++ b/packages/flame_test/lib/src/flame_test.dart
@@ -30,7 +30,7 @@ extension FlameGameExtension on Component {
 }
 
 typedef GameCreateFunction<T extends Game> = T Function();
-typedef VerifyFunction<T extends Game> = void Function(T);
+typedef VerifyFunction<T extends Game> = dynamic Function(T);
 
 typedef GameWidgetCreateFunction<T extends Game> = GameWidget<T> Function(
   T game,
@@ -94,7 +94,7 @@ class GameTester<T extends Game> {
   ) {
     flutter_test.test(description, () async {
       final game = await initializeGame();
-      verify(game);
+      await verify(game);
     });
   }
 
--- a/packages/flame_test/test/flame_async_test.dart
+++ b/packages/flame_test/test/flame_async_test.dart
@@ -1,0 +1,35 @@
+import 'package:flame/components.dart';
+import 'package:flame_test/flame_test.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  var instructions = 0;
+
+  tearDown(() {
+    assert(instructions == 9);
+  });
+  flameGame.test(
+    'runs all the async tests',
+    (game) async {
+      await game.ensureAdd(Component());
+      instructions++;
+      await game.ensureAdd(Component());
+      instructions++;
+      await game.ensureAdd(Component());
+      instructions++;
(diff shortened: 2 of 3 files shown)


packages/flame_test/lib/src/flame_test.dart
```diff
@@ -30,7 +30,7 @@ extension FlameGameExtension on Component {
 }
 
 typedef GameCreateFunction<T extends Game> = T Function();
-typedef VerifyFunction<T extends Game> = void Function(T);
+typedef VerifyFunction<T extends Game> = dynamic Function(T);
```
erickzanardo, line 33: Note: I don't like dynamics, but here I followed the signature that dart test does, that way we can await for it without compiler errors.

packages/flame_test/test/flame_async_test.dart
```diff
@@ -0,0 +1,35 @@
+import 'package:flame/components.dart';
+import 'package:flame_test/flame_test.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  var instructions = 0;
+
+  tearDown(() {
+    assert(instructions == 9);
```
erickzanardo, line 9: Note: Only way I could think on verifying that all of the method ran.

erickzanardo:
Fixes #1151

packages/flame_test/test/flame_async_test.dart
spydon: Don't you want to do the `instructions++` from within the `async` context?

spydon (approved):

packages/flame_test/test/flame_async_test.dart
erickzanardo: But they are on the async context, here we just verify with the right number of instructions ran after the test is finished.

spydon: I commented on the wrong line accidentally, I meant on line 15 and the likes. Very unlikely, but if `game.ensureAdd` returns without actually running the code in there `instructions` would still be bumped, but if you did `instructions++` from within the async operation that is running you would know that it has actually run.

packages/flame_test/lib/src/flame_test.dart
luanpotter: I think that is fine in this case, good catch!

luanpotter (approved):

packages/flame_test/test/flame_async_test.dart
erickzanardo: I still not sure I understood, is game.ensureAdd never completes, instructions will never run

spydon: If the test would run some fake async where it returns directly on the `await`, they would still run `instructions++` but not necessarily the asynchronous code. But yeah, it would be unlikely that it would happen all of a sudden, so not worth to care about.

erickzanardo: Ah, got it, but AFAIK just by being async, is enough for the issue to arise, Component.onLoad don't run real async code, but without my fix, this test would break.

spydon: Yeah, merge eeet!

PR #1193, opened 2021-12-09
feat: `Component.childrenFactory` can be used to set up a global `ComponentSet` factory

st-pasha:
# Description

Currently, the only way to override creation of a `ComponentSet` in a component is to create a derived class and override method `createComponentSet` within that class. However, this may not be always feasible, since Flame has many components that the user may want to use as-is.

In this PR I am adding an alternative approach: static global factory function `Component.childrenFactory`. Replacing this method with a custom implementation will make all subsequently created components use that custom factory function. The original `createComponentSet` function remains, both for backward-compatibility, and to make it possible to override creation of a component set within a particular subclass.

## Checklist

## Breaking Change

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ make integrations seamless. Currently, we have bridge libraries to the following
   [FireAtlas](https://github.com/flame-engine/fire-atlas): Create texture atlases for games.
 - [flame_forge2d](https://github.com/flame-engine/flame/tree/main/packages/flame_forge2d) for
   [Forge2D](https://github.com/flame-engine/forge2d): A Box2D physics engine.
-- [flame_lint](https://github.com/flame-engine/flame/tree/main/packages/flame_forge2d) -
+- [flame_lint](https://github.com/flame-engine/flame/tree/main/packages/flame_lint) -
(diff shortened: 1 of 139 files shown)


spydon (approved):
Lgtm, can you explain a bit about the usecase that you want to use it for, because I guess you have something in mind?

For the next release we are going to use Melos hopefully, so can you name your PR in a conventional commit way and skip the changelog entry? I'm going to update the guide today.

conventionalcommits.org/

st-pasha:
Use case, for example: create all `ComponentSet`s in non-strict mode; or use custom implementation of a `ComponentSet` everywhere.

Not sure what you mean by "name your PR in a conventional commit", but I can remove the changelog entry.

spydon:
> Not sure what you mean by "name your PR in a conventional commit", but I can remove the changelog entry.

Since we squash the commits, the name of the final commit will be the title of the PR, so the title should be a conventional commit name. You could basically just add `feat: ` in front of your current title.

spydon:
@st-pasha I'll break out the analyze fixes to a new PR so that the feature isn't hidden within all that noise.

erickzanardo (commented):
Can we add a small test for this new feature?

PR #1201, opened 2021-12-10
feat: Effect.onComplete callback as an alternative to onFinish()

st-pasha:
# Description

`Effect.onFinish` is now a property that can be directly set by the user. This allows an alternative way to chain multiple effects, and to perform some special action when an effect completes.

Previously, `Effect.onFinish()` was a method, and the only way to use it was to subclass a particular effect and create an overridden `.onFinish` method -- which is very cumbersome.

Note the asymmetry between `onStart()` and `onFinish()`: the first still remains an overridable method. This is because there are many effects that are interested in performing a custom action at start, and for them using a callback function would be more difficult. At the same time, there doesn't appear to be a clear use case when the user would want to provide a custom action at the start of an event.

### DIFF
--- a/doc/effects.md
+++ b/doc/effects.md
@@ -80,7 +80,7 @@ functionality inherited by all other effects. This includes:
     removed from the game tree and garbage-collected once the effect completes. Set this to false
     if you plan to reuse the effect after it is finished.
 
-  - Optional user-provided `onComplete` callback, which will be invoked when the effect has just
+  - Optional user-provided `onFinishCallback`, which will be invoked when the effect has just
     completed its execution but before it is removed from the game.
 
   - The `reset()` method reverts the effect to its original state, allowing it to run once again.
@@ -248,7 +248,7 @@ final effect = ColorEffect(
 The `Offset` argument will determine "how much" of the color that will be applied to the component,
 in this example the effect will start with 0% and will go up to 80%.
 
-__Note :__Due to how this effect is implemented, and how Flutter's `ColorFilter` class works, this
+__Note:__ Due to how this effect is implemented, and how Flutter's `ColorFilter` class works, this
 effect can't be mixed with other `ColorEffect`s, when more than one is added to the component, only
 the last one will have effect.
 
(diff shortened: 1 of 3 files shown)


packages/flame/lib/src/effects/effect.dart
```diff
@@ -44,6 +44,15 @@ abstract class Effect extends Component {
   /// in order to make it run once again.
   bool removeOnFinish;
 
+  /// This method is called once when the effect is about to finish, but before
+  /// it is removed from parent. The notion of "about to finish" is defined by
```
spydon, line 48:
```suggestion
  /// it is removed from its parent. The notion of "about to finish" is defined by
```

spydon (approved):

erickzanardo (commented):
We shouldn't introduce breaking changes with V1, an alternative here could probably be keep the method, receive a callback and the default implementation of the method, calls that callback.

st-pasha:
> We shouldn't introduce breaking changes with V1, an alternative here could probably be keep the method, receive a callback and the default implementation of the method, calls that callback.

So, would you suggest that we mark the `onFinish` method as deprecated, or we'd rather always have both the method and the callback simultaneously?

erickzanardo:
I would rather have the method and the callback.

doc/effects.md
```diff
@@ -79,6 +79,9 @@ functionality inherited by all other effects. This includes:
     removed from the game tree and garbage-collected once the effect completes. Set this to false
     if you plan to reuse the effect after it is finished.
 
+  - Optional user-provided `onComplete` callback, which will be invoked when the effect has just
```
spydon, line 82: It is a bit confusing to have one function that is called `onComplete` and one that is called `onFinish`, not sure how to solve it though.

st-pasha: Yeah, I guess that's why Flutter style guide insists on naming convention "onComplete" for callbacks, and "handleComplete" for methods.

erickzanardo: We could just name it `onFinishCallback`

luanpotter (approved):
LGTM

spydon (approved):

PR #1203, opened 2021-12-10
feat: Add RandomEffectController

st-pasha:
# Description

Add `RandomEffectController` -- a controller that makes the duration of some other effect controller random. This could be useful if the user wants to inject some amount of unpredictability into the effect.

## Checklist

## Breaking Change

### DIFF
--- a/doc/effects.md
+++ b/doc/effects.md
@@ -31,7 +31,6 @@ the final value is provided by the user explicitly, and progression over time is
 
 There are multiple effects provided by Flame, and you can also
 [create your own](#creating-new-effects). The following effects are included:
-- [`ColorEffect`](#coloreffect)
 - [`MoveEffect.by`](#moveeffectby)
 - [`MoveEffect.to`](#moveeffectto)
 - [`MoveAlongPathEffect`](#movealongpatheffect)
@@ -42,6 +41,7 @@ There are multiple effects provided by Flame, and you can also
 - [`SizeEffect.by`](#sizeeffectby)
 - [`SizeEffect.to`](#sizeeffectto)
 - [`OpacityEffect`](#opacityeffect)
+- [`ColorEffect`](#coloreffect)
 - [`RemoveEffect`](#removeeffect)
 
 An `EffectController` is an object that describes how the effect should evolve over time. If you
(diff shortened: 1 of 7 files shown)


doc/effects.md
```diff
@@ -42,6 +41,7 @@ There are multiple effects provided by Flame, and you can also
 - [`SizeEffect.by`](#sizeeffectby)
 - [`SizeEffect.to`](#sizeeffectto)
 - [`OpacityEffect`](#opacityeffect)
+- [`ColorEffect`](#coloreffect)
```
spydon, line 44: They were ordered alphabetically before, isn't that better, or what is the reason for this change?

spydon (approved):

doc/effects.md
st-pasha: Huh? They were never alphabetic: for example, Opacity < Size.
They are roughly ordered by category: first, transform effects, then visual effects, then miscellaneous effects. Maybe we should have a two-level list here to make the structure more obvious.

spydon: Ah alright, they were pretty close to alphabetical so that's why I assumed they were. Your ordering is fine.

spydon (approved):

erickzanardo (approved):

packages/flame/lib/src/effects/controllers/random_effect_controller.dart
```diff
@@ -0,0 +1,132 @@
...
+      'Random generator produced a negative value: $duration',
+    );
+    child.duration = duration;
+  }
+}
+
+/// [RandomVariable] is an object capable of producing random values with the
+/// prescribed distribution function. Each distribution is implemented within
+/// its own derived class.
+abstract class RandomVariable {
```
luanpotter, line 100: I have a feeling this could be useful in a general context, not only to random effect controller;
For example, on Gravity I introduced a [method similar to Uniform Random Variable](https://github.com/bluefireteam/gravitational_waves/blob/master/game/lib/game/collections.dart#L4)
If so, we could move it to its own file and add factory methods to create both private implementations.
Thoughts?

luanpotter (approved):
LGTM

packages/flame/lib/src/effects/controllers/random_effect_controller.dart
spydon: Probably as a follow-up though?

PR #1206, opened 2021-12-11
feat: Adding filterQuality override to parallax

newntech:
# Description

Updates to Parallax to allow overriding of the filterQuality parameter passed to the paintImage method from render.

This allows the client to specify the override as such:
 `   var img = await loadParallaxImage('background.png',
        filterQuality: FilterQuality.none,
        );`

Change is not ideal as the paintImage method supplies the default value of FilterQuality.low, and doesn't allow a null parameter, meaning the same value has to be defaulted  as a non-nullable field of the ParallaxRenderer.

`        filterQuality = filterQuality ?? FilterQuality.low;`

### DIFF
--- a/packages/flame/lib/src/parallax.dart
+++ b/packages/flame/lib/src/parallax.dart
@@ -38,6 +38,7 @@ extension ParallaxExtension on Game {
     ImageRepeat repeat = ImageRepeat.repeatX,
     Alignment alignment = Alignment.bottomLeft,
     LayerFill fill = LayerFill.height,
+    FilterQuality? filterQuality,
   }) {
     return ParallaxImage.load(
       path,
@@ -45,6 +46,7 @@ extension ParallaxExtension on Game {
       alignment: alignment,
       fill: fill,
       images: images,
+      filterQuality: filterQuality,
     );
   }
 
@@ -93,13 +95,17 @@ abstract class ParallaxRenderer {
   /// How to fill the screen with the image, always proportionally scaled.
   final LayerFill fill;
 
+  final FilterQuality filterQuality;
+
   ParallaxRenderer({
     ImageRepeat? repeat,
     Alignment? alignment,
     LayerFill? fill,
+    FilterQuality? filterQuality,
   })  : repeat = repeat ?? ImageRepeat.repeatX,
         alignment = alignment ?? Alignment.bottomLeft,
-        fill = fill ?? LayerFill.height;
+        fill = fill ?? LayerFill.height,
+        filterQuality = filterQuality ?? FilterQuality.low;
 
   void update(double dt);
   Image get image;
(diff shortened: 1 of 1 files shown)


packages/flame/lib/src/parallax.dart
```diff
@@ -93,13 +95,17 @@ abstract class ParallaxRenderer {
...
   ParallaxRenderer({
     ImageRepeat? repeat,
     Alignment? alignment,
     LayerFill? fill,
+    FilterQuality? filterQuality,
   })  : repeat = repeat ?? ImageRepeat.repeatX,
         alignment = alignment ?? Alignment.bottomLeft,
-        fill = fill ?? LayerFill.height;
+        fill = fill ?? LayerFill.height,
+        filterQuality = filterQuality ?? FilterQuality.low;
```
spydon, line 108: Maybe we should default to `none` or `high` instead?

spydon (commented):
Looks good! Just some thoughts, what is the difference between `FilterQuality.high` and `FilterQuality.none`? And maybe we should default to `none` or `high` instead, or does it have a big performance impact?

packages/flame/lib/src/parallax.dart
newntech: For me, high was not appropriate, as seems to be just a more detailed version of scaling, i.e. the pixelated appearance is still lost.

I was looking for the nearest neighbour scaling, since I want to maintain the pixelated appearance. 

I just left it as low as that's what the existing flutter method defaults to. I guess this hits the right balance for performance/appearance for most needs. 

From the comments on the method:

///  * `filterQuality`: Use this to change the quality when scaling an image.
///     Use the [FilterQuality.low] quality setting to scale the image, which corresponds to
///     bilinear interpolation, rather than the default [FilterQuality.none] which corresponds
///     to nearest-neighbor.

packages/flame/lib/src/parallax.dart
```diff
@@ -93,13 +95,17 @@ abstract class ParallaxRenderer {
   /// How to fill the screen with the image, always proportionally scaled.
   final LayerFill fill;
 
+  final FilterQuality filterQuality;
+
   ParallaxRenderer({
     ImageRepeat? repeat,
     Alignment? alignment,
     LayerFill? fill,
+    FilterQuality? filterQuality,
```
st-pasha, line 104: wouldn't it be simpler to say `this.filterQuality = FilterQuality.low`?

packages/flame/lib/src/parallax.dart
st-pasha: if the default value is `.low`, then we should keep that, otherwise this would become a breaking change.

packages/flame/lib/src/parallax.dart
spydon: We should have all of the arguments in the same fashion I think, I have no idea why the rest of the arguments don't have a default instead of accepting null but now it would be a breaking change to change them.

spydon (approved):

erickzanardo (approved):

renancaraujo (approved):

luanpotter (approved):
LGTM

PR #1134, opened 2021-11-23
Effect controllers restructuring

st-pasha:
# Description

Given the new requirement that effect controllers must be able to run backwards in time, and owing to the considerable complexity this entails, this PRs decomposes effect controllers into multiple small building blocks. The main API of `EffectController` remains mostly unchanged (with few new properties added), but the implementation of various effect controllers changes radically. In particular, there is no more class `StandardEffectController`. Instead, there are several smaller classes:
 - `LinearEffectController`,
 - `CurvedEffectController`,
 - `ReverseLinearEffectController`,
 - `ReverseCurvedEffectController`,
 - `DelayedEffectController`,
 - `RepeatedEffectController`,
 - `InfiniteEffectController`,
 - `PauseEffectController`,
 - `SequenceEffectController`.

In addition, factory constructor `EffectController()` has exactly the same API as old `StandardEffectController`, and can be used in its place.

This new structure gives several advantages:
 - Implementing "backwards in time" functionality is much easier for each small building block. My attempts at implementing the same functionality within the old `StandardEffectController` were producing an incomprehensible mess.
 - `SimpleEffectController` is no longer needed, the `EffectController()` constructor will use as little or as many building blocks as necessary to create the effect controller.
 - The new system allows to build custom effect controllers very easily.

## Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] My PR includes unit or integration tests for *all* changed/updated/fixed behaviors and are passing (See [Contributor Guide]).
- [x] My PR does not decrease the code coverage, or I have __a very special case__ and explained on the PR description why this PR decreases the coverage.
- [x] I updated/added relevant documentation (doc comments with `///`) and updated/added examples in `doc/examples`.
- [x] I have formatted my code with `flutter format` and the `flutter analyze` does not report any problems.
- [x] I read and followed the [Flame Style Guide].
- [x] I have added a description of the change under `[next]` in `CHANGELOG.md`.
- [x] I removed the `Draft` status, by clicking on the `Ready for review` button in this PR.
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

- [x] Yes, this is a breaking change (please indicate a breaking change in `CHANGELOG.md`).

[issue database]: https://github.com/flame-engine/flame/issues
[Contributor Guide]: https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md
[Flame Style Guide]: https://github.com/flame-engine/flame/blob/main/STYLEGUIDE.md

PR #1218, opened 2021-12-14
feat(effects)!: Added SequenceEffect

st-pasha:
# Description

Added SequenceEffect, which performs a series of other effects. 

The biggest challenge in implementing this feature came from the need to run the sequence in reverse, due to the `alternate` flag. This required that every effect and every controller supported running "back in time", which is not as simple as it sounds.

The following breaking changes were introduced:
 - The `Effect` class no longer supports `.reverse()` method and `.isReversed` flag.
 
    This flag was added only 2 weeks ago (#1134), with the idea that it will be necessary for the SequenceEffect. However, as it turned out, this flag is not as helpful as I thought it would be. In fact, given the user's ability to change it any point, it makes the implementation very error-prone.

    To be clear, the ability for each effect to run in reverse remains -- only now it can no longer be triggered by the user manually. Instead, SequenceEffect triggers that ability itself at the alternation point. If there is demand in the future to manually force any effect to run backwards, we could restore this flag, but this would require thorough testing to make it work correctly.

- Infinite effects now return `duration = double.infinity` instead of `null`, which seems more appropriate.

## Checklist

## Breaking Change


examples/lib/stories/effects/sequence_effect_example.dart
```diff
@@ -0,0 +1,63 @@
...
+      Player()
+        ..position = Vector2(200, 300)
+        ..add(
+          SequenceEffect(
+            [
+              MoveEffect.to(Vector2(400, 300), duration(0.7)),
+              RotateEffect.by(tau / 4, duration(0.5)),
+              MoveEffect.to(Vector2(400, 400), duration(0.7)),
+              ScaleEffect.by(Vector2.all(1.5), duration(0.7)),
+              MoveEffect.to(Vector2(400, 500), duration(0.7)),
```
spydon, line 30: It would be good having one of these effects with `alternate: true` here.

spydon (approved):
Lgtm, good job! I remember this being the hardest effect to create in the old effects system too. 
It would be good with some tests that has alternating effects withing an alternating `SequenceEffect`.

I think duration shouldn't have been made a named constructor (not from this PR, and too late to change it for now), it becomes so verbose to do simple effects.

st-pasha:
> It would be could with some tests that has alternating effects within an alternating SequenceEffect.

The last test in `sequence_effect_test.dart` has an alternating SequenceEffect, and within it another alternating repeating SequenceEffect, and one other non-alternating repeating SequenceEffect.

spydon:
> The last test in `sequence_effect_test.dart` has an alternating SequenceEffect, and within it another alternating repeating SequenceEffect, and one other non-alternating repeating SequenceEffect.

Right, but none of the "leaves" are alternating? I'm sure it works from just having that test, that's why I approved, but it would be a little bit different with just having another type of alternating effect inside the `SequenceEffect`?

st-pasha:
I don't understand - which structure do you have in mind?

spydon:
> I don't understand - which structure do you have in mind?

Like a simpler test with a `SequenceComponent` just having an alternating `MoveEffect` for example (or whichever other effect really).

st-pasha:
> Like a simpler test with a SequenceComponent just having an alternating MoveEffect for example (or whichever other effect really).

Ok, i added another test like that.

packages/flame/lib/src/effects/sequence_effect.dart
```diff
@@ -0,0 +1,203 @@
...
+/// reverse after it ran forward.
+///
+/// Parameter `repeatCount` will make the sequence repeat a certain number of
+/// times. If `alternate` is also true, then the sequence will first run
+/// forward, then back, and then repeat this pattern `repeatCount` times in
+/// total.
+///
+/// The flag `infinite = true` makes the sequence repeat infinitely. This is
+/// equivalent to setting `repeatCount = infinity`. If both the `infinite` and
+/// the `repeatCount` parameters are given, then `infinite` takes precedence.
```
luanpotter, line 23: we should just throw in these cases, better let the user know explicitly rather than trying to assume

packages/flame/test/effects/sequence_effect_test.dart
```diff
@@ -0,0 +1,278 @@
...
+                  : time <= 6
+                      ? 10 - 10 * (time - 3) / 3
+                      : 30 * (time - 6) / 4;
+          final y = time <= 1
+              ? 0
+              : time <= 3
+                  ? 10 * (time - 1) / 2
+                  : time <= 6
+                      ? 10
+                      : 10 + 30 * (time - 6) / 4;
```
luanpotter, line 104: some comments would be good explaining what this means

luanpotter (approved):

packages/flame_test/lib/flame_test.dart
```diff
@@ -1,3 +1,4 @@
+export 'src/close_to_vector.dart';
 export 'src/expect_double.dart';
 export 'src/expect_vector2.dart';
```
luanpotter, line 3: why do we need close_to_vector and expect_vector2 ? I think we should have only one of them

st-pasha: I created an issue for that: #1273

packages/flame/lib/src/effects/sequence_effect.dart
st-pasha: ok

packages/flame/test/effects/sequence_effect_test.dart
st-pasha: Ok, i simplified it a bit by using `expectedPositions` list.

packages/flame/test/effects/sequence_effect_test.dart
```diff
@@ -0,0 +1,281 @@
...
+          for (var i = 0; i < 100; i++) Vector2(10 + i * 0.1, 10 - i * 0.1),
+          for (var i = 0; i < 100; i++) Vector2(20 + i * 0.1, i * 0.1),
+          for (var i = 100; i > 0; i--) Vector2(20 + i * 0.1, i * 0.1),
+          for (var i = 100; i > 0; i--) Vector2(10 + i * 0.1, 10 - i * 0.1),
+          for (var i = 0; i < 200; i++) Vector2(10, 10 + i * 0.1),
+          for (var j = 0; j < 5; j++)
+            for (var i = 0; i < 200; i++)
+              Vector2(i < 100 ? i * 0.01 : 1, i < 100 ? 0 : i * 0.01 - 1)
+                ..add(Vector2(10 + j * 1.0, 30 + j * 1.0)),
+        ];
```
spydon, line 266: These are a lot of magic numbers for someone not familiar with the code, Maybe some of the numbers could be broken out to named variables that makes it easier to read?

spydon: And maybe a comment explaining it too.

packages/flame/test/effects/sequence_effect_test.dart
```diff
@@ -0,0 +1,281 @@
...
+        final game = FlameGame()..onGameResize(Vector2.all(1000));
+        await game.ensureAdd(component);
+        game.update(0);
+
+        // Each point here is spaced 0.1 seconds apart
+        final expectedPositions = [
+          for (var i = 0; i < 10; i++) Vector2(i * 1.0, 0),
+          for (var i = 0; i < 20; i++) Vector2(10, i * 0.5),
+          for (var i = 0; i < 30; i++) Vector2(10 - i / 3, 10),
+          for (var i = 0; i < 40; i++) Vector2(i * 0.75, 10 + i * 0.75),
```
spydon, line 105: I'd use `List.generate` here instead (and on the other places).

packages/flame/test/effects/sequence_effect_test.dart
st-pasha: Ok, i've turned the magic numbers into constants

packages/flame/test/effects/sequence_effect_test.dart
st-pasha: I tried, and it doesn't look any better or worse than the for-loops:
```dart
        final expectedPositions = <Vector2>[
          ...List.generate(10, (i) => Vector2(i * 1.0, 0)),
          ...List.generate(20, (i) => Vector2(10, i * 0.5)),
          ...List.generate(30, (i) => Vector2(10 - i / 3, 10)),
          ...List.generate(40, (i) => Vector2(i * 0.75, 10 + i * 0.75)),
          Vector2(30, 40),
        ];
```
In other places, though, we need loops running with decrementing counters, so `List.generate` won't work there (or at least it won't be pretty).

spydon: Looks a bit cleaner I think

st-pasha: ok, i've updated the test then

PR #1224, opened 2021-12-14
refactor: Removed parameter Component.updateTree({callOwnUpdate})

### LINKED ISSUE #1219: Remove {callOwnUpdate} parameter in updateTree()
## What could be improved
Function `Component.updateTree()` takes an optional parameter `bool calOwnUpdate = true`, which can be removed.

## Why should this be improved
1. The parameter was added very recently, just before the 1.0 release, and there wasn't enough time to carefully consider its implications. We can say that because of its novelty, removing the parameter does not constitute a breaking change.
2. The documentation for the parameter says that "if you call this method (updateTree) from update you need to set callOwnUpdate to false" -- however there is never a need (outside of `FlameGame`) to call `updateTree` from `update`, since the game structure is such that for every component `updateTree()` is called first. The only reason `FlameGame` is an exception is because of an unfortunate fact that both `Game` and `Component` define `update()` to mean different things, and `FlameGame` inherits from both...
3. The real problem with `callOwnUpdate` is that it imposes high burden on anyone who needs to override `updateTree()`, because this parameter has to be propagated everywhere. Then, the implementation of `updateTree` needs to think about what to do if this parameter is either true or false, and that may create impossible challenges. For example, what if the parameter says `callOwnUpdate = true` but the component doesn't implement its own update? Or if the parameter says `callOwnUpdate = false` but the component would not work without calling its `update()` method?

# Any risks?
Minimal. Since the parameter was added very recently, and it is unlikely that anyone created new components that override `updateTree()`, it shouldn't break anyone's code.

# More information

st-pasha:
# Description

- `Component.updateTree()` method will no longer have the optional parameter `callOwnUpdate`. 
- The `Game` class now has method `.updateTree()`, which by default redirects to `Game.update()`, but alternative implementations may be provided by the derived class.
- Method `FlameGame.update()` now works according to how `Component.update()` is expected to work, not how `Game.update()` is expected to work. If you need `Game.update()` functionality for `FlameGame`, call `FlameGame.updateTree()`.

## Checklist

## Breaking Change

## Related Issues

Closes #1219


spydon:
I can't say I'm a big fan of this solution, having overlapping functionality on different methods becomes confusing I think. The current solution isn't nice either, but I hope that we can come up with something else than this.

I would actually rather expose `updateTree` than doing this.

st-pasha:
> I would actually rather expose updateTree than doing this.

Alright, I renamed `tick()` into `updateTree()`.

st-pasha:
>> I would actually rather expose updateTree than doing this.
>
> Alright, I renamed tick() into updateTree().

Turns out this won't work:
  - If `Game` class implements `updateTree(dt) => update(dt)`, then `FlameGame` will have inherited this implementation and has no means of reverting it. Indeed, in order to get back to normal Component's `updateTree`, the FlameGame would have to write `updateTree(dt) => super.super.updateTree(dt)`, but that's not valid in Dart;
  - If `Game` class leaves `updateTree` unimplemented, then it becomes a breaking change, since all other classes that derive from `Game` would no longer work without implementing this method.

spydon:
> 
> Turns out this won't work:
> 
>     * If `Game` class implements `updateTree(dt) => update(dt)`, then `FlameGame` will have inherited this implementation and has no means of reverting it. Indeed, in order to get back to normal Component's `updateTree`, the FlameGame would have to write `updateTree(dt) => super.super.updateTree(dt)`, but that's not valid in Dart;

I don't think I follow, why can't `FlameGame` just override the `Game` implementation of `updateTree`?

st-pasha:
> I don't think I follow, why can't FlameGame just override the Game implementation of updateTree?

I mean, we could just copy-paste the code from `Component.updateTree`, but as far as I can see, there is no way for FlameGame to call that implementation directly. Or is there?

st-pasha:
> there is no way

Ok, scratch that. There is a way, and it's even a very nice way, but it necessitates a breaking change.

So, the trick is to replace
```dart
FlameGame extends Component with Game
```
with
```dart
FlameGame extends Game with Component
```

So, this is a big conceptual change. But it's a nice change: it's much easier to think of `FlameGame` or `OxygenGame` as extending the abstract class `Game` rather than mixing it in.

It would also be a breaking change, because in order to be able to use `Component` as a mixin, it must not have a constructor. Currently, it has a constructor that optionally takes the priority argument. The constructor can be removed in lieu of setting the priority value after construction, but this would probably be breaking for a small number of users.

I wouldn't recommend such change just for the sake of being able to call the method `updateTree` instead of `tick`, but it seems like it could be a great change on its own -- WDYT @spydon?

spydon:
> I wouldn't recommend such change just for the sake of being able to call the method `updateTree` instead of `tick`, but it seems like it could be a great change on its own -- WDYT @spydon?

If we are having a breaking change here, I think it would be nicer to just add `updateTree` unimplemented to `Game`. Then `FlameGame` will also keep consistency with how the other components are implemented.

> It would also be a breaking change, because in order to be able to use Component as a mixin, it must not have a constructor. Currently, it has a constructor that optionally takes the priority argument. The constructor can be removed in lieu of setting the priority value after construction, but this would probably be breaking for a small number of users.

We want to keep priority on the constructor here, since that is consistent from `Component` to all `PositionComponent`s (which are most of our components). This change would be a breaking change affecting more users than the actual change with mixing in `Component` instead of extending it on the `Game` (since not many write their own game classes).

erickzanardo:
> I think it would be nicer to just add updateTree unimplemented to Game.

`updateTree` is a FCS concept though, so it don't make much sense to have it on game.

Honestly, I prefer the way we have right now, I do agreed that it has problems, but I think I prefer those problems that any other alternative so far. We could mitigate those problems by having some section on our docs deeply explaining what callOwnUpdate is and why it exists.

st-pasha:
> We want to keep priority on the constructor here, since that is consistent from Component to all PositionComponents (which are most of our components). 

PositionComponent won't be affected here:
```dart
class PositionComponent extends Component {
  PositionComponent({..., int? priority}) {
    changePriorityWithoutResorting(priority); // or we could make a simple setter in Component
  }
}
```
So the only users that would be affected are those that inherit from Component directly and call its constructor explicitly -- which, in my opinion, a small minority.

That being said, I'm wary in general about creating methods in `Game` that have exactly the same name as methods in `Component`, because each such method spells trouble for `FlameGame` class. Sure, those methods might have same purpose now; but nobody guarantees that that purpose won't change in the future (as already happened with `update`). It is safer to have non-overlapping method names in two classes.

spydon:
> So the only users that would be affected are those that inherit from Component directly and call its constructor explicitly -- which, in my opinion, is a small minority.

Quite a lot of people use `Component` directly I would say, for building levels and screens, most of them probably don't use `priority` on that level though. But it is about consistency, just like we want to be consistent of how the components are created, so that it is not done like a mixin in `FlameGame`, but inheritance in other components.

> That being said, I'm wary in general about creating methods in `Game` that have exactly the same name as methods in `Component`, because each such method spells trouble for `FlameGame` class. Sure, those methods might have same purpose now; but nobody guarantees that that purpose won't change in the future (as already happened with `update`). It is safer to have non-overlapping method names in two classes.

We don't guarantee that, but we are in control of it. If we want to make breaking changes on that level we should of course be wary. But I think I'm leaning more to Erick's solution than the last one I suggested too, that we just keep it the way it currently is and document `callOwnUpdate` better (we haven't had any questions about it so far as far as I'm aware).

st-pasha:
@spydon I hope we can at least agree on the fact that our current class hierarchy is ungood. We have `Game` class, which is great, and `Component` which is also fine -- but then `FlameGame` tries to be both a `Game` and a `Component` at the same time. Now, Dart doesn't allow multi-inheritance (and for a good reason), so we tried to circumvent that by declaring `Game` as a mixin. However, that's cheating: Game is conceptually the abstract base class, not a mixin. And FlameGame is conceptually closer to Game than to Component. So, we sacrificed the hierarchical integrity of our class system in order to make FlameGame work -- and it does work, with some warts.

And one of those warts is the fact that both Game and Component have the `update()` method, but they do different things. So we sort of have to make another hack in order to fix that: the `callOwnUpdate` parameter. 

The problem then is that this hack propagates into the rest of the Component hierarchy. And that's bad, because it makes the original code smell endemic, and much harder to fix in the future.

> we haven't had any questions about it so far as far as I'm aware

Well, except for #1224, obviously. I mean, just like `renderTree`, it's a very advanced feature needed for not so many users. But I did ran in this problem working on #1218, so it's not an abstract consideration.

>  we just keep it the way it currently is and document callOwnUpdate better 

I do not see how this solves the original concern that the parameter cannot be respected by the implementations at all. And if the documentation says "ok you have this parameter, but feel free to ignore it" -- then what kind of parameter is it?

----
That being said, there are other methods for dealing with the `FlameGame.update()` method that do not involve adding extra function in Game. So maybe we should explore those?

spydon:
> Well, except for #1224, obviously. I mean, just like `renderTree`, it's a very advanced feature needed for not so many users. But I did ran in this problem working on #1218, so it's not an abstract consideration.

#1224 is this PR, I guess you meant another issue/PR?

> That being said, there are other methods for dealing with the `FlameGame.update()` method that do not involve adding extra function in Game. So maybe we should explore those?

Absolutely, if you have more ideas, put them on the table. :)

spydon:
Maybe the best would be to just duplicate the lines from `Component` to `FlameGame`, like I think you suggested in the beginning @st-pasha? So that it would be something like this in `FlameGame`:

```dart
  @override
  @mustCallSuper
  void update(double dt) {
    super.update(dt);
    _cameraWrapper.update(dt);
    if (parent == null) {
      updateTree(dt);
    }
  }
  
  @override
  void updateTree(double dt) {
    children.updateComponentList();
    if (parent != null) {
      update(dt);
    }
    children.forEach((c) => c.updateTree(dt));
  }
```

st-pasha:
^ Yes, this should work. And, I mean, it's not such a big code chunk anyways...

Another possibility would be to add private variable `bool _preventUpdate` in `FlameGame`, so that it would look something like this:
```dart
class FlameGame {
  bool _preventUpdate = false;
  void update(double dt) {
    if (_preventUpdate) return;
    _cameraWrapper.update(dt);
    if (parent == null) {
      _preventUpdate = true;
      super.updateTree(dt);
      _preventUpdate = false;
    }
  }
}
```

spydon:
> ^ Yes, this should work. And, I mean, it's not such a big code chunk anyways...
> 
> Another possibility would be to add private variable `bool _preventUpdate` in `FlameGame`, so that it would look something like this:
> ```dart
> class FlameGame {
>   bool _preventUpdate = false;
>   void update(double dt) {
>     if (_preventUpdate) return;
>     _cameraWrapper.update(dt);
>     if (parent == null) {
>       _preventUpdate = true;
>       super.updateTree(dt);
>       _preventUpdate = false;
>     }
>   }
> }
> ```

The variable way would call an overridden version of `update` twice, even if super was called.

st-pasha:
> The variable way would call an overridden version of update twice, even if super was called.

The idea here is that first the user calls `update()`, then `update()` sets flag `preventUpdate` to true and calls `updateTree()`. The method `updateTree()` would then call `update()` again, as it normally does, but the first statement in `update()` checks the flag `preventUpdate`, sees that it is true, and quickly exits as if the call didn't happen. The `updateTree` then finishes normally, the execution returns back to the original `update`, at which point the flag `preventUpdate` is set to false again. Thus, next time the system calls `update()`, it works normally again.

It might sound complicated, but think of it as passing the parameter `callOwnUpdate=false` not as an actual parameter, but as an environment variable.

spydon:
> > The variable way would call an overridden version of update twice, even if super was called.
> 
> The idea here is that first the user calls `update()`, then `update()` sets flag `preventUpdate` to true and calls `updateTree()`. The method `updateTree()` would then call `update()` again, as it normally does, but the first statement in `update()` checks the flag `preventUpdate`, sees that it is true, and quickly exits as if the call didn't happen. The `updateTree` then finishes normally, the execution returns back to the original `update`, at which point the flag `preventUpdate` is set to false again. Thus, next time the system calls `update()`, it works normally again.
> 
> It might sound complicated, but think of it as passing the parameter `callOwnUpdate=false` not as an actual parameter, but as an environment variable.

Sure, I understand how it works, but imagine this scenario:

```dart
  @override
  void update(double dt) {
    super.update(dt);
    print(1);
  }
}
```

When that is added directly to the game widget `update` will be called, and then for each `update` call, since `updateTree` also calls `update`, the `print(1)` statement will be called again, but the original `super.update` code will be skipped due to the variable.

st-pasha:
I see, makes sense.
In that case we could declare `update` as `@nonVirtual` and make the user override `updateTree` instead, but this does feel kinda heavy-handed. Let's try going with your suggestion [above](https://github.com/flame-engine/flame/pull/1224#issuecomment-996146499) and see how it shapes out.

packages/flame_test/lib/src/flame_test.dart
```diff
@@ -19,13 +19,15 @@ extension FlameGameExtension on Component {
   Future<void> ensureAdd(Component component) async {
     await add(component);
     updateTree(0);
+    update(0);
```
spydon, line 22: Future "Game in game" tests could be affected by this, since `update` then would be called twice. We`ll probably need a check of the parent here too.

spydon (approved):
Just a minor comment, otherwise lgtm!

luanpotter (approved):

packages/flame_test/lib/src/flame_test.dart
st-pasha: In theory, calling `update(0)` multiple times should not make a difference.

But in general, I was thinking that maybe `ensureAdd`/`ensureAddAll` methods can be moved into the `Component` class, in which case `FlameGame` would be able to redefine them to call `update` instead of `updateTree`.

spydon: Hmm, don't think we'd want to add that directly in `Component` since this only should be used for tests.
But you are right, since two `0`s shouldn't make a difference we can just as well have it like this for now.

spydon: Wait, but why is `update(0)` called here? This function only exist so that the component will be added directly to the tree, and that will happen already from `updateTree`?

st-pasha: Because without it, when we call `flameGame.ensureAdd(component)` -- it would invoke `invokeTree`, which would in turn skip the `update()`. So, if there is a `MyFlameGame` that defines its own functionality in `update()`, then that functionality won't be invoked from `ensureAdd`.

st-pasha: If the extension is specific to Component, then why is it called `FlameGameExtension`, and not `ComponentExtension`?

spydon: Because we simply forgot to rename it when changing from having it specific for `FlameGame` to `Component`, it has `on Component` so can only be used on `Component`s. And even if it was specific for `FlameGame`, `update` wouldn't have had to be called here.

erickzanardo (approved):

packages/flame_test/lib/src/flame_test.dart
spydon, line 22:
```suggestion
```

packages/flame_test/lib/src/flame_test.dart
```diff
@@ -19,13 +19,15 @@ extension FlameGameExtension on Component {
...
   }
 
   /// Makes sure that the [components] are added to the tree if you wait for the
   /// returned future to resolve.
   Future<void> ensureAddAll(Iterable<Component> components) async {
     await addAll(components);
     updateTree(0);
+    update(0);
```
spydon, line 30:
```suggestion
```

PR #1234, opened 2021-12-16
docs: Added documentation for GameLoop class

st-pasha:
# Description

Added documentation for the GameLoop class.

## Checklist

- [x] The title of my PR starts with a [Conventional Commit] prefix (`fix:`, `feat:`, `doc:` etc).

## Breaking Change


spydon (approved):
LGTM, but I wonder if there was a specific reason to have the separate methods, since they weren't doing exactly the same thing.

packages/flame/lib/src/game/game_loop.dart
```diff
@@ -1,48 +1,79 @@
...
+  /// do not have to stop the game loop before disposing of it.
   void dispose() {
     _ticker.dispose();
   }
 
-  void pause() {
-    _ticker.muted = true;
-    previous = Duration.zero;
-  }
+  @Deprecated('Use stop() instead')
```
erickzanardo, line 74: Why refactor these?

Either way a change like this should be on its own PR

erickzanardo (changes requested):
Docs LGTM, but this PR is addressing multiple things under the same PR, I don't understood the need to refactor pause and resume, but this should be treated on a separate PR.

st-pasha:
> I don't understood the need to refactor pause and resume

As I was writing the documentation for each method in the class, I found that it's hard to describe what those methods do as compared to start and stop. I mean, sure they had their own quirks, but it's better to fix those quirks than document them.

packages/flame/lib/src/game/game_loop.dart
st-pasha: I'll split them out into another PR.

packages/flame/lib/src/game/game_loop.dart
```diff
@@ -1,48 +1,82 @@
...
+  /// do not have to stop the game loop before disposing of it.
   void dispose() {
     _ticker.dispose();
   }
 
-  void pause() {
-    _ticker.muted = true;
-    previous = Duration.zero;
-  }
+  @Deprecated('Internal variable')
```
wolfenrain, line 74: Why use `Deprecated`? There are specialized decorators for this like `Protected`.

st-pasha: Because we don't use that variable anywhere. It was never intended to be public.
So, now that we make it private, the getter is provided in case anyone's code was reading that variable. The getter is marked deprecated, as a way to signal that this variable really shouldn't be used, and that the getter will be removed after a while.

erickzanardo (approved):

wolfenrain (approved):

PR #1240, opened 2021-12-17
fix: Deprecate pause and resume in GameLoop

st-pasha:
# Description

`GameLoop` methods `pause()` and `resume()` are marked as deprecated, with a recommendation to use `start()` and `stop()` instead. The reason they are deprecated is because their functionality is identical to that of the other pair of methods, so no reason to keep them around... Also, property `previous` was made final (with a deprecated getter for backward-compatibility), since changing it externally would break GameLoop's functionality.

Here's the summary of behavior changes:
- `start()` can now be called multiple times; previously calling `start()` twice would produce an error
- `stop()` now works the same way as `pause()`; previously calling `stop()` and then `start()` would produce negative times
- property `.previous` is no longer settable

## Checklist

## Breaking Change

## Related Issues

This is split out of #1234


st-pasha:
Note: we may also make pause/resume the main methods, and redirect/deprecate start+stop.

packages/flame/lib/src/game/game_loop.dart
```diff
@@ -2,47 +2,41 @@ import 'package:flutter/scheduler.dart';
...
   void stop() {
     _ticker.stop();
+    _previous = Duration.zero;
   }
 
   void dispose() {
     _ticker.dispose();
   }
 
-  void pause() {
```
luanpotter, line 35: I don't think they were meant to be the same, IIRC @erickzanardo added those to fix a bug or something

erickzanardo: Yes, I do remember something like this too, but I can remember exactly what but and why, we should hold this until we can remember

spydon: I also though we added these, but apparently these were added by someone called @erf 20 months ago, maybe they could explain?

spydon: Here is the PR: https://github.com/flame-engine/flame/pull/302

erf: @spydon hi, i took a look now at my PR where i simplified the game loop to use a Ticker. It looks like `start` and `stop` replaced the `scheduleTick` and `unscheduleTick` calls. The only different between those were that `previous` Duration were set to zero in `pause`. It think it would be fine to combine these calls. Not sure if there would be a case where e.g. the user would e.g. want to pause the callbacks, but still make the timer move forward  or something like that.

spydon: Thanks for the input @erf!

erf: No problem! I would perhaps also take a look at the `Ticker` inner workings, and see if you want to expose e.g. `isActive` or `hidden` to `GameLoop`.

st-pasha: > The only different between those were that previous Duration were set to zero in pause.

Note that NOT setting duration to zero in stop, would have produced negative `dt` after the subsequent call to `start`, which is clearly a bug. The only reason it wasn't noticed is because `stop` was never used.

spydon (approved):
Lgtm, I don't think there ever was a reason to have both and everything seems to work fine without `pause` and `resume`.

packages/flame/lib/src/game/game_loop.dart
luanpotter: the bug was: whenever the widget was recreated, since we were using the engine state as the pause/unpause, it would restart the engine. we needed a difference between user manually calling pause (persistent) and engine being paused/unpaused by the framework.

can we test this new version under that scenario?

spydon: @luanpotter any suggestion for how this test should look?

luanpotter: I didn't mean an automated test, just manually really!

st-pasha: Interesting... So I tried the following manual test: launch examples/System/Pause-resume-engine, turn on a device frame, click on the helicopter to pause the engine, then click the button to rotate the device frame (which involves rebuilding the widget).

The helicopter starts spinning again, the engine doesn't stay paused. So I guess the test fails? I also tried the same test in the current `main`, and the outcome is the same: the engine restarts, test fails.

It looks like the reason is that the "ticker" of the engine belongs to the `GameLoop` class, which in turn belongs to the `GameRenderBox`, which in turns gets re-created via the `_GameRenderObjectWidget` whenever the `GameWidget` is rebuilt. So, when `GameWidget` is rebuilt, the `GameLoop` gets recreated, and it starts running.

In short, the bug that you describe currently exists in `main`, and this PR neither fixes it nor makes it worse. Should we perhaps open a separate issue for the bug?

spydon: Yeah, then I think we can just as well merge this and file a bug for that behavior.

spydon: Can you file an issue for that @st-pasha since you have the most context?

spydon: Btw, in your test, did you declare the game outside of the `GameWidget`?

st-pasha: I've tested with the standard example from our Dashbook, I'll have to check if it recreates the game instance upon frame rotation.

st-pasha: Ok, I checked again -- after making sure that the game instance is declared only once in the examples app, the "bug" disappears (i.e. the game state keeps being paused after rotating the frame or changing window size, or opening the side menu). This happens both in the main branch and in this branch.

spydon: Super, then it works as intended!

PR #1243, opened 2021-12-17
refactor: Loadable mixin no longer declares onMount and onRemove

### LINKED ISSUE #1210: Remove onMount/onRemove from Loadable
## What could be improved

The `Loadable` mixin is required for the generic `Game` class; at the same time methods `onMount`/`onRemove` are only suitable for the `Component` class and the derived `FlameGame`.

These methods should, therefore, be moved out of `Loadable` and into the `Component` class.

- [x] Yes, I can work on a PR for this

st-pasha:
# Description

`Loadable` mixin is currently used for 2 classes: `Game` and `Component`. Its purpose is to provide the caching layer for the future returned by the `onLoad()` method. Having empty methods `onMount()` / `onRemove()` in that mixin does not fit its overall purpose. (Plus, in general, the purpose of any mixin is to add some additional **concrete** functionality into an existing class).

In addition to the scope mismatch, these methods perform similar yet not quite the same function in `Game` and `Component`. Having these methods declared directly within `Game` and `Component` allows to document them more precisely, and help the user understand their functionality/purpose better.

## Checklist

## Breaking Change

## Related Issues

Closes #1210


st-pasha:
Note: I wasn't quite able to figure out the difference between `onMount` and `onAttach` for the `Game` class -- which I guess only stresses the need for clear and precise documentation for that function.

spydon:
> Note: I wasn't quite able to figure out the difference between `onMount` and `onAttach` for the `Game` class -- which I guess only stresses the need for clear and precise documentation for that function.

`onAttach` is when the Flutter widget is attached to the Flutter widget tree, `onMount` is when the component is added to a component tree (after `onLoad`).

packages/flame_bloc/test/flame_bloc_game_test.dart
```diff
@@ -96,6 +96,7 @@ void main() {
         game.add(component);
 
         expect(component.state, equals(InventoryState.sword));
+        game.remove(component);
```
spydon, line 99: Why was this line added?

spydon (approved):
Lgtm

packages/flame_bloc/test/flame_bloc_game_test.dart
st-pasha: In order to get 100% coverage for flame_bloc repository.

st-pasha:
> onAttach is when the Flutter widget is attached to the Flutter widget tree, onMount is when the component is added to a component tree (after onLoad).

Exactly: this is where it gets confusing. I understand `onMount` for the `Component` class. But the Game class is a different story. So the question is: what is `onMount` in the `Game` class, and how is it different from the `onAttach` handler?

The situation with `onRemove` / `onDetach` is similar, only I believe the former is called _before_ the game is removed from the widget, while the latter called _after_. We should consider at some point to rename them to improve clarity. For example: `onBeforeDetach` / `onAfterDetach`.

spydon:
> Exactly: this is where it gets confusing. I understand `onMount` for the `Component` class. But the Game class is a different story. So the question is: what is `onMount` in the `Game` class, and how is it different from the `onAttach` handler?

Yeah, I see the confusion. We should remove `onMount` and `onRemove` from the `game` mixin (I must have accidentally put them there) and in `FlameGame` we should call `onMount` from `onAttach` and `onRemove` from `onDetach` probably. Since a game can be added as a child to another component, there will still be a difference to them.

packages/flame_bloc/test/flame_bloc_game_test.dart
spydon: But it is kind of cheating, this test doesn't test anything regarding removal?

st-pasha: Fair, i've expanded this to a bigger test

packages/flame/lib/src/game/mixins/game.dart
```diff
@@ -95,6 +95,8 @@ mixin Game on Loadable {
   /// Use for calculating the FPS.
   void onTimingsCallback(List<FrameTiming> timings) {}
 
+  void onMount() {}
```
spydon, line 98: I don't think `onMount` and `onRemove` is needed in this mixin since it is meant for games that live can live in the component tree, for other games `onAttach` and `onDetach` can be used.

st-pasha: Ok, great, I'll remove them then.

st-pasha: I just tried removing them, and it turns out to be a breaking change -- even 2 tests start to fail. 
So, probably such breaking change should be in its own PR.

PR #1252, opened 2021-12-20
feat(collision detection)!: Use a broadphase to make collision detection more efficient

spydon:
# Description

This is the first step in making the collision detection more efficient. This uses AABBs (axis aligned bounding box) for checking whether a component is possibly colliding with another component. The AABB is currently recalculated for each component in each tick, but cached between ticks. In this PR a broadphase step is also added to the collision detection system, the only implementation right now is sweep-and-prune, but it's easy to extend it with new broadphase algorithms later.

Two small things left to do that I intend to finish this evening:
 - [ ] Fix polygon normals bug.
 - [ ] Write migration docs.

## Checklist

## Breaking Change

## Related Issues

#980 


spydon:
Possibly I'll build the quadtree from the AABBs already in this PR, and then I won't have to do the super call to `HasHitboxes`.

packages/flame/lib/src/components/mixins/has_hitboxes.dart
```diff
@@ -1,12 +1,38 @@
...
+  Aabb2 get aabb {
+    if (_validAabb) {
+      return _aabb;
+    }
+    final size = scaledSize;
+    // This has +1 since a point on the edge of the bounding box is currently
+    // counted as inside.
+    final maxHalfExtent = (max(size.x, size.y) / 2) + 1;
+    _maxHalfExtents.setValues(maxHalfExtent, maxHalfExtent);
+    _aabb.setCenterAndHalfExtents(absoluteCenter, _maxHalfExtents);
```
st-pasha, line 32: What if the component is rotated/scaled?

packages/flame/lib/src/components/mixins/has_hitboxes.dart
```diff
@@ -1,12 +1,38 @@
...
   final List<HitboxShape> _hitboxes = <HitboxShape>[];
+  final Aabb2 _aabb = Aabb2();
+  bool _validAabb = false;
+  final Vector2 _maxHalfExtents = Vector2.zero();
+
+  @override
+  @mustCallSuper
+  void update(double dt) {
+    super.update(dt);
+    _validAabb = false;
```
st-pasha, line 20: Looks like this invalidates the computed bounding rect on every game tick... 
However, if an object isn't moving then its bounding rect can be kept in cache.

st-pasha (commented):
The mixins `HasHitboxes` and `HitboxShape` are missing the documentation. 
I know, it wasn't there to begin with, but still it would be easier to understand code structure if there was a clear description of what each class is doing.

Oh and btw - have we considered implementing collidables as components rather than mixins?

packages/flame/lib/src/components/mixins/has_hitboxes.dart
spydon: That is what line 30 is for, it grows the box to the maximum that it should have to cover.
We could also get the absolute angle and "rotate" the AABB, I will see what is cheaper later.

packages/flame/lib/src/components/mixins/has_hitboxes.dart
spydon: Indeed, we have no cheap way to check whether the component, or one of its ancestors, has gained a different size, scaled or moved.
If you know of a way that is cheaper than just re-creating the AABB I'm all ears, this will be even more important later when I turn it into a quad tree.

packages/flame/lib/src/components/mixins/has_hitboxes.dart
st-pasha: For example, if the component is a square (size.x == size.y), and rotated 45º, then its bounding rect should be sqrt(2) times larger than the size.

spydon: Ah, true, good catch. I was meaning to take the length of the size/2.

spydon: But then just rotating the AABB is probably cheaper anyways, so I'll do that.

packages/flame/lib/src/components/mixins/has_hitboxes.dart
st-pasha: I think for a start we could compute all bounding rects in world coordinates rather than screen coordinates. After all, there are plenty of objects that are stationary with respect to the world, but not so many that are stationary relative to the screen (other than HUD elements, but those don't usually collide :)

spydon:
> The mixins `HasHitboxes` and `HitboxShape` are missing the documentation. I know, it wasn't there to begin with, but still it would be easier to understand code structure if there was a clear description of what each class is doing.
> 
> Oh and btw - have we considered implementing collidables as components rather than mixins?

Will add docs to them. 

Since the `Collidable`s live in the `collidables` list and are trying to be outside of the component tree to make them more efficient I don't see what we would gain by having them as components, but definitely open for that option if we could make it efficient.

packages/flame/lib/src/components/mixins/has_hitboxes.dart
spydon: All bounding rects are already in world coordinates.

st-pasha:
We could start with gathering all collision-related functionality into a single folder, say `flame/lib/src/collision`. Because right now it's even hard to find where it is.

packages/flame/lib/src/components/mixins/has_hitboxes.dart
```diff
@@ -1,12 +1,40 @@
...
+    // This has +1 since a point on the edge of the bounding box is currently
+    // counted as inside.
+    _halfExtents.setValues(size.x + 1, size.y + 1);
+    _rotationMatrix.setRotationZ(absoluteAngle);
+    _aabb
+      ..setCenterAndHalfExtents(absoluteCenter, _halfExtents)
+      ..rotate(_rotationMatrix);
+    _validAabb = true;
+    return _aabb;
+  }
```
st-pasha, line 37: Hmm, I think there is still a problem with the scale of all ancestors being ignored.

Also, note that in `PositionComponent` there is method `toAbsoluteRect()` which can probably be used here.

spydon: We never came up with a solution for adding up scales right? It is documented that it doesn't work if you ancestors are scaled, but if you know how to solve it, it would of course be good to do so.
`toAbsoluteRect` is more expensive than this solution since it will have to project two vectors, and also we want an `Aabb2`, not a `Rect`, since we later can interact with the rest of the classes in `vector_math` with that.

packages/flame/lib/src/components/mixins/has_hitboxes.dart
```diff
@@ -15,9 +15,14 @@ mixin HasHitboxes on PositionComponent {
...
-  void update(double dt) {
-    super.update(dt);
-    _validAabb = false;
+  void onMount() {
+    super.onMount();
+    void addCallback(PositionComponent c) {
+      c.transform.addListener(() => _validAabb = false);
+    }
+
+    ancestors().whereType<PositionComponent>().forEach(addCallback);
```
st-pasha, line 24: How about adding a flag to `ancestors()` such that
```dart
if (includeSelf) {
  yield this;
}
```

spydon: Good idea, it doesn't really follow the function name, but since the named variable makes it quite clear I changed to using that.

packages/flame/lib/src/components/screen_collidable.dart
```diff
@@ -0,0 +1,29 @@
...
+import '../collision_detection/collision_callbacks.dart';
+import '../collision_detection/hitboxes/hitbox_rectangle.dart';
+
+/// This component is used to detect hitboxes colliding into the viewport of the
+/// game.
+class ScreenCollidable<T extends FlameGame> extends PositionComponent
+    with CollisionCallbacks, HasGameRef<T> {
+  @override
+  Future<void> onLoad() async {
+    await super.onLoad();
```
st-pasha, line 12:
```suggestion
```

doc/flame/collision_detection.md
```diff
@@ -9,129 +10,169 @@ If you have the following needs you should at least consider to use
...
-### Collidable
-The `Collidable` mixin is added to a `PositionComponent` that has a `HasHitboxes` and it is used for
-detecting collisions with other `Collidable`s. If you do not add a shape to your `HasHitboxes` 
-component it will never collide with anything. If you want the component to have a default 
-rectangular shape that fills the size of your component you can simply do 
-`addHitbox(HitboxRectangle())`.
-
-To make your component collidable you would start off something like this:
 
+### GestureHitboxes
```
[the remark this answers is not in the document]
spydon: The whole document should maybe have another name, since the `GestureHitboxes` are dependant on hitboxes I think it fits in here, but maybe it should be under the gestures section and link to here instead.

packages/flame/lib/src/collision_detection/broadphase.dart
```diff
@@ -0,0 +1,33 @@
...
+
+  Broadphase({List<T>? items}) : items = items ?? [];
+
+  Set<Potential<T>> query();
+}
+
+/// A [Potential] is a tuple that is used to contain two potentially colliding
+/// hitboxes.
+@immutable
+class Potential<T> {
```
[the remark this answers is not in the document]
spydon: `Pair` is more vague though, it should be a word that describes a pair in a potential collision.t
Maybe `CollisionProspect` or something like that?

packages/flame/lib/src/collision_detection/hitboxes/hitbox.dart
```diff
@@ -0,0 +1,30 @@
+import '../../../collision_detection.dart';
+import '../../../extensions.dart';
+
+/// The [Hitbox] is the default building block to determine whether two objects
+/// have collided with each other. [HitboxShape] is the default implementation
+/// used in FCS.
+mixin Hitbox<T extends Hitbox<T>> implements GenericCollisionCallbacks<T> {
```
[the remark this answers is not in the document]
spydon: > Is this supposed to be mixed in or implemented?

You can do both, depending on how you implement you collision detection system.

 >  Also it is not very clear what Hitbox<T> represents.

It represents the type of hitboxes that are going to be used, I'll add that to the dartdoc.

packages/flame/lib/src/collision_detection/hitboxes/hitbox.dart
```diff
@@ -0,0 +1,30 @@
...
+  ///   collide with hitboxes that are active, but not other passive ones).
+  ///   * [CollidableType.inactive] - not collide with any other hitboxes.
+  CollidableType get collidableType;
+
+  /// The axis-aligned bounding box of the [Hitbox], this is used to make a
+  /// rough estimation of whether two hitboxes can possibly collide or not.
+  Aabb2 get aabb;
+
+  /// Checks whether the [Hitbox] contains the [point].
+  bool containsPoint(Vector2 point);
```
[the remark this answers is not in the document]
spydon: That is the point, since then it can be used in FCS directly by using this mixin on a component, but it can at the same time be used outside of FCS knowing that the `containsPoint` method will exist.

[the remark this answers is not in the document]
st-pasha: The `point` is under-specified here: it doesn't say what the coordinate system is. Is it in local coordinates, parent's coordinates, world coordinates, canvas coordinates, device coordinates, or something else? The real danger then comes from when `Component` uses one coordinate system but the `Hitbox` mixin uses another. Remember how we had a trouble with `update()` in `Component`/`Game`?

packages/flame/lib/src/collision_detection/broadphase.dart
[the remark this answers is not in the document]
st-pasha: The way the class is written right now, it could be a completely generic utility class that lives somewhere in `utils/`. Or do you envision that we'll need some collision-specific features in here?

packages/flame/lib/src/collision_detection/hitboxes/hitbox.dart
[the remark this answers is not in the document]
spydon: It is in the global coordinate system for `HitboxShape`, just like it is for `PositionComponent`, I'll make sure to add that in the dartdocs for `HitboxShape` too.
The `Hitbox` is general enough to not have to care about the coordinate system, that is left up to the implementer of the new type of hitbox.

packages/flame/lib/src/collision_detection/broadphase.dart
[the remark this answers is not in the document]
spydon: I don't think we should create a general tuple class, there are reasons that the Dart team haven't added it, I think mainly because it easily leads to anti-patterns among folks who don't know what they are doing.

erickzanardo (commented):
This seems to be a big breaking change, What do you think about having a migration guide on the docs?

doc/flame/collision_detection.md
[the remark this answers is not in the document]
spydon: Moved it to `gesture-input.md`.

doc/flame/collision_detection.md
```diff
@@ -141,71 +182,84 @@ from the game.
...
-mandatory argument is a list of `Vector2` which defines how your polygon should look, but not the
-size or position for it (since that sill be defined by the component that you attach it to).
-For example you could create a diamond like in the [](#collidable) example like this:
+### HitboxCircle
+
+The `HitboxCircle` has the same constructors as the ()[#CircleComponent], see that section for
+documentation regarding those.
+
+
+## ShapeComponents
```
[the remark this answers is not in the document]
spydon: Moved this section to `components.md`

doc/flame/collision_detection.md
```diff
@@ -9,129 +10,169 @@ If you have the following needs you should at least consider to use
...
 being called with a large delta time (for example if your app is not in the foreground). This
 behaviour is called tunneling, if you want to read more about it.
 
-Also note that the collision detection system doesn't work properly if you scale ancestors of the
-component that is `Collidable`.
+Also note that the collision detection system has a limitation that makes it not work properly if
+you have certain types of combinations of flips and scales of the ancestors of the hitboxes.
+
+
+## Adding hitboxes
```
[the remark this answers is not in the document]
spydon: Done

packages/flame/lib/collision_detection.dart
```diff
@@ -0,0 +1,11 @@
+export 'src/collision_detection/broadphase.dart';
```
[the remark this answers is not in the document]
spydon: Done

doc/flame/inputs/gesture-input.md
```diff
@@ -381,14 +381,21 @@ The provided event info is from the mouse move that triggered the action (enteri
...
-Then you can use the `HasHitboxes` mixin to define a more accurate polygon for which the input 
-should be within for the event to be counted on your component.
+You can add new hitboxes to the component that has the `GestureHitboxes` mixin just like they are
+added in the below `Collidable` example.
 
-An example of you to use it can be seen
+More information about how to define hitboxes can be found in the hitbox section of the 
+[collision detection](../collision_detection.md#adding-hitboxes) docs.
+
+// TODO(spydon): Add correct example
```
wolfenrain, line 399: TODO still needed?

packages/flame/lib/src/collisions/collision_detection.dart
```diff
@@ -0,0 +1,123 @@
+import '../../collisions.dart';
+import '../../components.dart';
+
+/// [CollisionDetection] is the foundation of the collision detection system in
+/// Flame. If the [HasCollisionDetection] mixin is added to the game, [run] is
```
wolfenrain, line 5:
```suggestion
/// Flame. 
///
/// If the [HasCollisionDetection] mixin is added to the game, [run] is
```

packages/flame/lib/src/components/position_component.dart
```diff
@@ -164,14 +164,33 @@ class PositionComponent extends Component {
...
+      height * absoluteScale.y.abs(),
+    );
+  }
+
   /// The resulting angle after all the ancestors and the components own angle
   /// has been applied.
   double get absoluteAngle {
-    return ancestors()
-        .whereType<PositionComponent>()
+    // TODO(spydon): take scale into consideration
```
wolfenrain, line 180: Will this be done in a follow up?

doc/flame/collision_detection.md
```diff
@@ -1,137 +1,135 @@
...
-The `Collidable` mixin is added to a `PositionComponent` that has a `HasHitboxes` and it is used for
-detecting collisions with other `Collidable`s. If you do not add a shape to your `HasHitboxes` 
-component it will never collide with anything. If you want the component to have a default 
-rectangular shape that fills the size of your component you can simply do 
-`addHitbox(HitboxRectangle())`.
-
-To make your component collidable you would start off something like this:
 
+### HasCollisionDetection
+If you want to use collision detection in your game you have to add the `HasCollisionDetection`
```
wolfenrain, line 29:
```suggestion

If you want to use collision detection in your game you have to add the `HasCollisionDetection`
```

doc/flame/collision_detection.md
```diff
@@ -1,137 +1,135 @@
...
+mixin to your game so that the game knows that it should use a collision detection system to keep
+track of which components that can collide.
+
+Example:
 ```dart
```
wolfenrain, line 34:
```suggestion

```dart
```

doc/flame/collision_detection.md
```diff
@@ -1,137 +1,135 @@
...
 
-#### CollidableType
-By default the `CollidableType` is `active` on your `Collidable`, but there are two other types that
-you can set your collidable to too if you need to optimize the collision detection.
+## HitboxShape
+
+The `HitboxShape`s are normal components, so you add them to the component that you want to add
+hitboxes to just like any other component:
+
+`component.add(HitboxRectangle());`
```
wolfenrain, line 93: Would still put this in a code block personally

doc/flame/collision_detection.md
```diff
@@ -141,124 +139,98 @@ from the game.
...
-on the top fourth of the component you would write something like this:
+It is a good idea to just use the Flame collision detection system if you on the other hand only
+need some of the following things (since it is simpler to not involve Forge2D):
+- The ability to act on some of your components colliding
+- The ability to act on your components colliding with the screen boundaries
+- Complex shapes to act as a hitbox for your component so that gestures will be more accurate
+- Hitboxes that can tell what part of a component that collided with something
 
-`HitboxCircle(definition: 0.5)..relativePosition = Vector2(0, 0.5)`
+## Broad phase
```
wolfenrain, line 203: Would it make sense to move this one up? So before the comparison with forge2d?

wolfenrain (approved):
LGTM, few nits

Awesome work @spydon!

packages/flame/lib/src/components/position_component.dart
spydon: Yes, this can't be handled with our current nested system, we'd have to know in each step what transformations that has been done before it and in which order.

doc/flame/collision_detection.md
```diff
@@ -1,4 +1,5 @@
 # Collision detection
+
```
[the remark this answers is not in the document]
spydon: Better now?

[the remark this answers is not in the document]
st-pasha: :+1:

packages/flame/lib/src/collisions/collision_callbacks.dart
```diff
@@ -0,0 +1,138 @@
...
+/// The default implementation used with FCS is [CollisionCallbacks].
+mixin GenericCollisionCallbacks<T> {
+  Set<T>? _activeCollisions;
+
+  /// The objects that the object is currently colliding with.
+  Set<T> get activeCollisions => _activeCollisions ??= {};
+
+  /// Whether the object is currently colliding or not.
+  bool get isColliding {
+    return _activeCollisions != null && _activeCollisions!.isNotEmpty;
```
st-pasha, line 30:
```suggestion
    return _activeCollisions?.isNotEmpty ?? false;
```

doc/flame/collision_detection.md
```diff
@@ -1,137 +1,143 @@
...
-More about how the different shapes are defined in the [Shapes](#shapes) section.
+Now when you add `HitboxShape`s to components that are then added to the game, they will
+automatically be checked for collisions.
 
-Remember that you can add as many `HitboxShape`s as you want to your `Collidable` to make up more
-complex hitboxes. For example a snowman with a hat could be represented by three `HitboxCircle`s and
-a `HitboxPolygon` as its hat.
 
-To react to a collision you should override the `collisionCallback` in your component.
+### CollisionCallbacks
```
st-pasha, line 46: This used to be called `Collidable`, isn't it? Do you think this new name is better? I kinda liked how a component could be "Tappable", "Draggable", "Collidable".

doc/flame/collision_detection.md
```diff
@@ -1,137 +1,143 @@
...
-  ...
-
+class MyCollidable extends PositionComponent with CollisionCallbacks {
   @override
-  void onCollision(Set<Vector2> points, Collidable other) {
-    if (other is CollidableScreen) {
-      ...
-    } else if (other is YourOtherCollidable) {
-      ...
+  void onCollision(Set<Vector2> points, PositionComponent other) {
```
st-pasha, line 54: `onCollisionStart`?

Also: I wonder if we should wrap these into an "Event" object, similar to how we have events for taps, drags, key presses, and so on.

doc/flame/collision_detection.md
```diff
@@ -1,137 +1,143 @@
...
 
-It is a good idea to just use the Flame collision detection system if you on the other hand only
-need some of the following things (since it is slightly simpler to not involve Forge2D):
- - The ability to act on when some of your components collide
- - The ability to act on when your components collide with the screen boundaries
- - Complex shapes to act as a hitbox for your component so that gestures will be more accurate
- - Hitboxes that can tell what part of a component that collided with something
+In most collision detection systems you use something called hitboxes to create more precise
+bounding boxes of your components. In Flame the hitboxes are areas of the component that can react
+to collisions (and make [gesture input](inputs/gesture-input.md#GestureHitboxes) more accurate.
```
st-pasha, line 8:
```suggestion
to collisions.
```
The mentions of gestures distracts from the main idea of the document here (the reader now feels like they need to go and read what "gesture input" is).

doc/flame/collision_detection.md
```diff
@@ -1,137 +1,143 @@
...
 
-In this example it can be seen how the Dart `is` keyword is used to check which other `Collidable`
-that your component collided with. The set of points is where the edges of the hitboxes collided.
-Note that the `onCollision` method will be called on both collidable components if they
-have both implemented the `onCollision` method, and also on both shapes if they have that method
-implemented. The same goes for the `onCollisionStart` and `onCollisionEnd` methods, which are
-called when two components or shapes starts or stops colliding with each other.
+In this example it can be seen how the Dart `is` keyword is used to check which other
+`PositionComponent` that your component collided with. The set of points is where the edges of the
+hitboxes collided.
```
st-pasha, line 75:
```suggestion
In this example we use Dart's `is` keyword to check what kind of component we collided with. The set of points is where the edges of the
hitboxes collided.
```

packages/flame/lib/src/collisions/collision_callbacks.dart
```diff
@@ -0,0 +1,138 @@
...
+    implements GenericCollisionCallbacks<PositionComponent> {
+  @override
+  Set<PositionComponent>? _activeCollisions;
+  @override
+  Set<PositionComponent> get activeCollisions => _activeCollisions ??= {};
+
+  @override
+  bool get isColliding {
+    return _activeCollisions != null && _activeCollisions!.isNotEmpty;
+  }
```
st-pasha, line 84: This seems to be exactly same as in the parent class.

packages/flame/lib/src/collisions/collision_callbacks.dart
```diff
@@ -0,0 +1,138 @@
...
+  Set<T> get activeCollisions => _activeCollisions ??= {};
+
+  /// Whether the object is currently colliding or not.
+  bool get isColliding {
+    return _activeCollisions != null && _activeCollisions!.isNotEmpty;
+  }
+
+  /// Whether the object is colliding with [other] or not.
+  bool collidingWith(T other) {
+    return _activeCollisions != null && activeCollisions.contains(other);
```
st-pasha, line 35:
```suggestion
    return _activeCollisions?.contains(other) ?? false;
```

packages/flame/lib/src/collisions/collision_callbacks.dart
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /// [onCollision] is called in every tick when this object is colliding with
+  /// [other].
+  @mustCallSuper
+  void onCollision(Set<Vector2> intersectionPoints, T other) {
```
st-pasha, line 41: I would call this `onCollisionUpdate`, similar to how we have `onDragStart`, `onDragUpdate` and `onDragEnd`.

Which also makes me wonder if there is such thing as `onCollisionCancel`?

doc/flame/collision_detection.md
spydon: I think it is important to give the connection to the reader here though, and it's just a parenthesis.

doc/flame/collision_detection.md
spydon: It is quite different though, previously you had to add the `Collidable` mixin for the collision detection to work, now you only have to add it if you need to act on the callbacks, hence the name.

packages/flame/lib/src/collisions/collision_callbacks.dart
spydon: `Cancel` comes from Flutter's gesture system, I don't think we have to involve cancel here.

packages/flame/lib/src/collisions/collision_callbacks.dart
spydon: Yes, but since mixins can't extend we'll have to re-implement it here.

doc/flame/collision_detection.md
```diff
@@ -1,137 +1,143 @@
...
- - The ability to act on when your components collide with the screen boundaries
- - Complex shapes to act as a hitbox for your component so that gestures will be more accurate
- - Hitboxes that can tell what part of a component that collided with something
+In most collision detection systems you use something called hitboxes to create more precise
+bounding boxes of your components. In Flame the hitboxes are areas of the component that can react
+to collisions (and make [gesture input](inputs/gesture-input.md#GestureHitboxes) more accurate.
 
 The collision detection system supports three different types of shapes that you can build hitboxes
-from, these shapes are Polygon, Rectangle and Circle. A hitbox can be represented by many shapes to
+from, these shapes are Polygon, Rectangle and Circle. Multiple hitbox can be added to a component to
```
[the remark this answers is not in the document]
spydon: `CollisionPassthrough` can be used for this, but I created `CompositeHitbox` to help with this too now.

st-pasha:
Will this support alternative implementations of Hitboxes, those which use a GJK or MPR algorithms to detect collisions? Some of those algorithms may be able to detect collisions, but not the points of collisions.

spydon:
> Will this support alternative implementations of Hitboxes, those which use a GJK or MPR algorithms to detect collisions? Some of those algorithms may be able to detect collisions, but not the points of collisions.

Currently they'd have to return a non-empty set of intersection points, but they could just return the positions of the colliding components for example and it would work.

doc/flame/components.md
```diff
@@ -535,104 +554,165 @@ It is also possible to create custom renderers by extending the `ParallaxRendere
...
-between -1.0 and 1.0 that describes the ration of the length from the center to the edge of the size
-of the component. So
-`[Vector2(1.0, 1.0), Vector2(1.0, -1.0), Vector2(-1.0, -1.0), Vector2(-1.0, 1.0)]`
-would describe a rectangle that fills the full size of the component. Remember to define the list in
-a counter clockwise manner (if you think in the screen coordinate system where the y-axis is
-flipped, otherwise it is clockwise).
-
-So to create a diamond shaped `PolygonComponent` which is slightly smaller than the defined size you
-would do this:
+You can also create a `RectangleComponent` from a normal, or use the default constructor to build
```
st-pasha, line 654: What does "normal" mean here?

packages/flame/lib/src/collisions/hitboxes/hitbox.dart
```diff
@@ -0,0 +1,30 @@
+import '../../../collisions.dart';
+import '../../../extensions.dart';
+
+/// The [Hitbox] is the default building block to determine whether two objects
+/// have collided with each other. [ShapeHitbox] is the default implementation
+/// used in FCS.
+mixin Hitbox<T extends Hitbox<T>> implements GenericCollisionCallbacks<T> {
```
st-pasha, line 7: This looks more like an abstract class to be implemented, not a mixin.
Conceptually, a mixin cannot be "a building block".

doc/flame/components.md
spydon: True, that has to be explained more. It is the point but in relation to the parent size instead of in an absolute position.

packages/flame/lib/src/collisions/hitboxes/hitbox.dart
spydon: Will convert it to an abstract class, didn't know you could mixin classes just like mixins (`HitboxShape` needs to use it as a mixin).

spydon: Fixed

doc/flame/components.md
spydon: Fixed

st-pasha: But why is it called "normal"? I only heard of "normal" meaning "standard" or "orthogonal", neither of which seem to fit here.

packages/flame/lib/src/collisions/hitboxes/hitbox.dart
st-pasha: :+1:

doc/flame/components.md
spydon: I found that name fitting when researching it earlier, but I can't find out where I got it from again and I agree it isn't a very good name. Do you have any suggestions for the name?

st-pasha: Perhaps "relative"? I'm not sure what the intended use-case for this constructor, though.

erickzanardo (approved):

doc/flame/components.md
spydon: It is used to create a hitbox which size is relative to the intended parent's size, it was quite useful in the last collision detection system. 

Now I remember where `normal` came from, it came from that you can normalize a `Vector2`. Still a really bad name though since the normal of a vector is something completely different than a normalized vector. I'll change it to something else, maybe relative.

spydon: The named constructors are now called `.relative` and the argument that refines the relation is called `relation`, wdyt?

st-pasha: :+1:

PR #1253, opened 2021-12-20
example: Add Raycast example for flame_forge2d

ufrshubham:
# Description

This PR adds a new sample in flame_forge2d for Raycast. It contains a sample implementation for finding nearest and farthest fixtures along the casted ray. Here is how it works, two rays (red and blue) are fired from same starting point with mouse position as the target. Red ray always stops at the nearest fixture and blue ray always stop at the farthest fixture. Paint of fixtures are changed depending on which ray hits it. If both rays hit the same fixture, it turns yellow.

https://user-images.githubusercontent.com/33748002/146823360-4e8b30c1-0d0e-44b0-8d25-61a8fb42bc0d.mp4

## Checklist

## Breaking Change

  e.g. `feat!:`, `fix!:`).

## Related Issues

NA

### DIFF
--- a/packages/flame_forge2d/example/lib/main.dart
+++ b/packages/flame_forge2d/example/lib/main.dart
@@ -12,6 +12,7 @@ import 'draggable_sample.dart';
 import 'joint_sample.dart';
 import 'mouse_joint_sample.dart';
 import 'position_body_sample.dart';
+import 'raycast_sample.dart';
 import 'sprite_body_sample.dart';
 import 'tappable_sample.dart';
 import 'widget_sample.dart';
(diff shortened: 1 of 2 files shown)


spydon:
Nice! Could you add a short screen capture of the example to the description so that reviewers don't have to checkout your branch to see it?

ufrshubham:
Added a video showing the demo of new sample.

spydon (approved):

st-pasha:
Why is it that the ray sometimes stops short (for example, at 0:14 in the video)?

spydon:
> Why is it that the ray sometimes stops short (for example, at 0:14 in the video)?

The example is using the `onMouseMove` detector and the rays are going to where the mouse is, with a small offset.

ufrshubham:
Oh, that is because the target for raycast is mouse position. If no fixture is hit within start point to mouse position range, it stops at the mouse position. I forgot to turn on mouse capture while recording the video, that is why it looks a bit confusing.😅

st-pasha:
Ah, I see, makes sense.

What about 0:17 when the blue ray goes through the white rectangle without hitting it?

PR #1260, opened 2021-12-21
feat(effects): Added SpeedEffectController

### LINKED ISSUE #1239: Add support for effects with fixed speed
# Problem to solve
This existed in old effect system, but not in the new one: create an effect that proceeds with a fixed speed, rather than with a fixed duration.

+assign-me

st-pasha:
# Description

This adds `SpeedEffectController` -- new kind of effect controller that can be used to create effects that proceed with a predefined speed. In addition, the factory constructor `EffectController()` now has 2 more parameters `speed` and `reverseSpeed` for easier creation of speed-based effects. The argument `duration` is now optional.

The base `EffectController` class now has `onMount()` callback, which is invoked when the effect controller is attached to an `Effect`.

## Checklist

## Breaking Change

## Related Issues

Closes #1239


packages/flame/lib/src/effects/controllers/effect_controller.dart
```diff
@@ -78,35 +86,115 @@ abstract class EffectController {
...
         reverseCurve == Curves.linear || ((reverseCurve == null) && isLinear);
+    final reverseHasDuration =
+        (reverseDuration != null) || (reverseSpeed == null && duration != null);
+
     final items = [
-      if (isLinear) LinearEffectController(duration),
-      if (!isLinear) CurvedEffectController(duration, curve),
-      if (atMaxDuration != 0) PauseEffectController(atMaxDuration, progress: 1),
+      // FORWARD
+      if (isLinear)
```
spydon, line 148: All if-else statements should have an explicit body `{}`, I thought we had this in the analyzer already...

packages/flame/lib/src/effects/controllers/speed_effect_controller.dart
```diff
@@ -0,0 +1,89 @@
...
+  double? get duration => _initialized ? child.duration : double.nan;
+
+  @override
+  double get progress => child.progress;
+
+  @override
+  double advance(double dt) {
+    if (!_initialized) {
+      _initialized = true;
+      final measure = _parentEffect.measure();
```
spydon, line 49: This must become quite expensive?

packages/flame/lib/src/effects/move_effect.dart
```diff
@@ -17,22 +19,26 @@ import 'transform2d_effect.dart';
...
-      : _offset = offset.clone(),
+      : vector = offset.clone(),
         super(controller);
 
   factory MoveEffect.to(Vector2 destination, EffectController controller) =>
       _MoveToEffect(destination, controller);
 
-  Vector2 _offset;
+  @protected
+  Vector2 vector;
```
spydon, line 31: This could have a more descriptive name

spydon (commented):
Lgtm, just some small comments

packages/flame/lib/src/effects/controllers/effect_controller.dart
st-pasha: This is inside the list constructor `[ ... ]`, so curly braces aren't even supported there...

packages/flame/lib/src/effects/controllers/speed_effect_controller.dart
st-pasha: This runs only once at the start of the effect (guarded by `_initialized` flag).

packages/flame/lib/src/effects/move_effect.dart
st-pasha: This is the best I could come up with. It's `vector` not because the type is `Vector2`, but because the amount of movement is literally described by a vector from the point of origin to the point of destination.

Do you have any better ideas for how to name it?

packages/flame/lib/src/effects/controllers/effect_controller.dart
spydon: Ah right, it should be broken up into smaller parts then, so that this much logic isn't inside the list construction.

packages/flame/lib/src/effects/move_effect.dart
spydon: Hmm, isn't the previous `offset` name better then?

packages/flame/lib/src/effects/controllers/effect_controller.dart
st-pasha: Ok then

packages/flame/lib/src/effects/move_effect.dart
st-pasha: Looks like it doesn't even have to be protected, it's just a remnant from my older attempt at trying to build this feature another way. So i'll just revert it to how it was before.

spydon (approved):

luanpotter (approved):

PR #1261, opened 2021-12-21
feat(effects): Added ZigzagEffectController

st-pasha:
# Description

Simple alternating effect controller. Over the course of one `period`, this controller will proceed
linearly from 0 to 1, then to -1, and then back to 0. Use this for oscillating effects where the
starting position should be the center of the oscillations, rather than the extreme (as provided
by the standard alternating `EffectController`).

```dart
final ec = ZigzagEffectController(period: 2);
```

https://user-images.githubusercontent.com/4231472/146996943-60a76bae-5e3c-42b8-bf69-548fd5ba400e.mov

## Checklist

- [x] The title of my PR starts with a [Conventional Commit] prefix (`fix:`, `feat:`, `docs:` etc).
- [x] I have read the [Contributor Guide] and followed the process outlined for submitting PRs.
- [x] I have updated/added tests for ALL new/updated/fixed functionality.
- [x] I have updated/added relevant documentation in `docs` and added dartdoc comments with `///`.
- [x] I have updated/added relevant examples in `examples`.

## Breaking Change

- [x] No, this is *not* a breaking change.

## Related Issues

[issue database]: https://github.com/flame-engine/flame/issues
[Contributor Guide]: https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md
[Flame Style Guide]: https://github.com/flame-engine/flame/blob/main/STYLEGUIDE.md
[Conventional Commit]: https://conventionalcommits.org

PR #1262, opened 2021-12-21
feat(effects): Added SineEffectController

st-pasha:
# Description

An effect controller that represents a single period of the sine function. Use this to create
natural-looking harmonic oscillations. Two perpendicular move effects governed by
`SineEffectControllers` with different periods, will create a [Lissajous curve].

```dart
final ec = SineEffectController(period: 1);
```

https://user-images.githubusercontent.com/4231472/147003315-c5e6a2da-8e3a-4c1a-a2db-b826cdc6dc94.mov

## Checklist

## Breaking Change

## Related Issues

Similar to #1261


examples/lib/stories/effects/effect_controllers_example.dart
```diff
@@ -0,0 +1,84 @@
...
+          Vector2(0, 20),
+          InfiniteEffectController(ZigzagEffectController(period: 8/7)),
+        ),
+        MoveEffect.by(
+          Vector2(10, 0),
+          InfiniteEffectController(ZigzagEffectController(period: 2/3)),
+        ),
+      ]),
+    );
+    */
```
spydon, line 53: You forgot to uncomment this

spydon (approved):

examples/lib/stories/effects/effect_controllers_example.dart
st-pasha: This is code from Zigzag branch -- I can only uncomment it once #1261 is merged.

luanpotter:
Maybe I am misunderstanding, but don't we have a `CurvedEffectController` that can be created for any desired curve?

st-pasha:
> Maybe I am misunderstanding, but don't we have a `CurvedEffectController` that can be created for any desired curve?

A `CurvedEffectController` uses a `Curve` object, which "bends" time as it goes from 0 to 1. The documentation for the `Curve` class says as much: _"A [Curve] must map t=0.0 to 0.0 and t=1.0 to 1.0."_.

Here time goes from 0 to 1 then to 0 then to -1 and finally back to 0. Thus, this doesn't satisfy the requirement of the Curve class.

packages/flame/test/effects/controllers/sine_effect_controller_test.dart
```diff
@@ -0,0 +1,41 @@
...
+      expect(ec.started, true);
+      expect(ec.completed, false);
+      expect(ec.progress, 0);
+      expect(ec.isRandom, false);
+    });
+
+    test('progression', () {
+      final ec = SineEffectController(period: 3);
+      final expectedProgress = [
+        for (var i = 0; i <= 100; i++) sin(i * 0.01 * 2 * pi),
```
spydon, line 21: Why not use `List.generate` instead of all these for loops inside of lists?

st-pasha: yeah, in this case List.generate might be nicer

luanpotter:
I see, makes sense. I wonder if we could simplify the creation of controllers based on arbitrary functions (more generically than Curve allows). Imagine someone might want cos, 2*sin(x), sin(x + 1), etc. But that is just ideas for the future

luanpotter (approved):

st-pasha:
> I see, makes sense. I wonder if we could simplify the creation of controllers based on arbitrary functions (more generically than Curve allows). Imagine someone might want cos, 2*sin(x), sin(x + 1), etc. But that is just ideas for the future

Yeah, this sounds like a good idea. Then we could have sine and zigzag as special cases.

Issue #396, opened 2020-06-26
ShakeEffect


spydon:
An effect that inserts some noise to shake the component.

PR #1263, opened 2021-12-22
feat: Added 1d+2d Perlin and Simplex noise functions

st-pasha:
# Description

These functions are used in games for a variety of purposes: terrain generation, visual effects, randomized motion, textures, enemy spawns, etc, etc.

This implementation is ported from the original code by Stefan Gustavson, and manually verified that it produces the same values as his C code. The original library had more varieties of noise functions, including 3D- and 4D- noise, as well as noise with gradients and periodic noise. If needed, those other functions can be ported as well.

## Checklist

## Breaking Change

## Related Issues

WIP for #396


packages/flame/lib/algos.dart
```diff
@@ -0,0 +1 @@
+export 'src/noise.dart' show noise1, noise2, snoise1, snoise2;
```
spydon, line 1: The name of this file should be the full word.

packages/flame/lib/src/noise.dart
```diff
@@ -0,0 +1,231 @@
...
+  final nx10 = _grad2(_perm[ix1 + _perm[iy0]], fx1, fy0);
+  final nx11 = _grad2(_perm[ix1 + _perm[iy1]], fx1, fy1);
+  final n1 = _lerp(t, nx10, nx11);
+  return 0.507 * _lerp(s, n0, n1);
+}
+
+//#endregion
+
+//#region Simplex noise
+// This is a clean, fast, modern and free Perlin Simplex noise function.
```
spydon, line 59: Isn't this dartdoc?

spydon (commented):
Maybe create a new directory called algorithms or something? Or we could do that when there are more of them instead.

packages/flame/lib/src/noise.dart
st-pasha: I don't know, this is a comment that relates to all functions in that section, and it was copied from the original Gustavson's implementation

spydon: Pretty funny that it says modern since the comment was written like 20 years ago 😄

spydon (approved):

spydon:
On a second thought; do we really gain anything on maintaining this inside the engine, since there are already good dart libraries for this? Example: https://pub.dev/packages/fast_noise

st-pasha:
They don't have 1d noise

wolfenrain:
As there are already other packages who provide this I dont see why we should also maintain another version. I have used the one @spydon mentioned in Flame many times both for 2d and 1d noise (you can just hardcode one of the files). 

And even if there were no, or bad, packages I dont think this should be a core flame feature. We should then opt for just having a perlin-like package. It has zero connections to Flame itself and could be useful outside of the flame ecosystem as well.

st-pasha:
> you can just hardcode one of the files

This is exactly what this pr is doing

st-pasha:
In order to avoid the [status quo bias](https://en.wikipedia.org/wiki/Status_quo_bias), we should ask ourselves the following: if this feature was already merged long time ago, would we be arguing in favor of removing it and replacing with an external dependency? How would we go about implementing the 1d noise functionality that we actually need?

spydon:
> In order to avoid the [status quo bias](https://en.wikipedia.org/wiki/Status_quo_bias), we should ask ourselves the following: if this feature was already merged long time ago, would we be arguing in favor of removing it and replacing with an external dependency? How would we go about implementing the 1d noise functionality that we actually need?

The status quo bias doesn't apply here, new information was provided from DevKage that there already are great libraries for this, which I didn't know, and that's why I approved the PR at first. We don't need the 1d noise functionality for anything specific directly in the engine right? It is not hard to take just one of the dimensions from a 2d noise.

st-pasha:
> We don't need the 1d noise functionality for anything specific directly in the engine right?

I need noise specifically in order to implement #396, and computing 2d noise then taking 1d slice of it is much less performant: just compare the amount of calculations in `noise1()` and `noise2()` functions.

spydon:
> > We don't need the 1d noise functionality for anything specific directly in the engine right?
> 
> I need noise specifically in order to implement #396, and computing 2d noise then taking 1d slice of it is much less performant: just compare the amount of calculations in `noise1()` and `noise2()` functions.

Why can't you just use `Random()` for that?

st-pasha:
> Why can't you just use Random() for that? 

All articles that I've seen on shaking in games recommend using some kind of noise function (perlin, simplex, value, sum of sines, etc), and warn against using pure Random() because it produces unnatural-looking effect.

spydon:
> > Why can't you just use Random() for that? 
> 
> All articles that I've seen on shaking in games recommend using some kind of noise function (perlin, simplex, value, sum of sines, etc), and warn against using pure Random() because it produces unnatural-looking effect.

It must depend on how you use random? You could just generate a delta each tick and not the absolute position from random? 

> it produces unnatural-looking effect

I think this makes more sense for other movement effects, the shake is supposed to be disruptive anyways?

st-pasha:
There is no single "right" way to shake things up. The shake from an earthquake is different from the shake from an explosion, which is different from the shake from a Big Guy walking, which is different from the shake from bullet recoil, etc. I imagine developers would want to experiment and see what works best in the context of their games. Our job then is to provide them with the necessary tools.

 > You could just generate a delta each tick and not the absolute position from random?

This would produce the [random walk](https://en.wikipedia.org/wiki/Random_walk) behavior. I know its mathematical properties, but I don't know how it will actually _feel_. Still, it could be just another tool in the toolbox.

spydon:
> I imagine developers would want to experiment and see what works best in the context of their games. Our job then is to provide them with the necessary tools.

If this is for users of Flame, then this shouldn't be in the engine I think, possibly a PR (with 1d) to one of the libraries already providing this. If we need it internally in the engine we should have it though, since we don't want a new external dependency. As of now, I don't see much use for it internally if it is only for the shake effect, but maybe you have more usecases?

st-pasha:
> If this is for users of Flame, then this shouldn't be in the engine I think

Everything in the engine is for users of Flame, isn't it? I mean, sure, perlin-noise-based shake effects aren't the only kind of shakes possible, but they are quite common ones. And giving the user ability to choose among several shake effects that better suite their needs is good.

I mean, there is no actual cost of including it into Flame -- the tree-shaking algorithm ensures that whatever functions are not used by the user will not be included into the final executable.

> but maybe you have more use cases?

Besides **shake effects**, noise can also be used in **particle systems**. Noise is also great for terrain generation and texture generation, but I'm not sure we're going to include that functionality into the engine anytime soon.

spydon:
> Everything in the engine is for users of Flame, isn't it? 

It is in some sense, but if it is already done well by another package in the dart ecosystem which can be used by our users we shouldn't reinvent the wheel. This is why the question whether this is something that is needed internally or not is important.

> I mean, there is no actual cost of including it into Flame -- the tree-shaking algorithm ensures that whatever functions are not used by the user will not be included into the final executable.

The cost is an unnecessary maintenance cost, not a performance cost.

st-pasha:
> The cost is an unnecessary maintenance cost

Well, this is a well-established stable algorithm. You probably won't find any other functionality that requires less maintenance than this.

spydon:
> Well, this is a well-established stable algorithm. You probably won't find any other functionality that requires less maintenance than this.

In the future there might be more noise algorithms that people want to add, you even spoke about some yourself, and then it is better to focus our efforts on helping an already established library in the ecosystem than to duplicate the code.

st-pasha:
> and then it is better to focus our efforts on helping an already established library in the ecosystem than to duplicate the code.

Well, there are several counterindications to this approach:
1. We don't want to burden Flame with external dependencies; so even if we helped another package, that would not be helpful to us.
2. We would have to contend with the pace of development of the other package. From what I see, `fast_noise` in its 4 years of existence have had only 1 PR and 2 issues (both still open).

spydon:
> Well, there are several counterindications to this approach:
> 
>     1. We don't want to burden Flame with external dependencies; so even if we helped another package, that would not be helpful to us

But none of your examples so far has been anything that we plan to do within the engine, and for shaking I don't think we will need noise. So the users could just use any other library when creating the examples you had together with flame in their own games, so we will never have to add the dependency to flame.

>     2. We would have to contend with the pace of development of the other package. From what I see, `fast_noise` in its 4 years of existence have had only 1 PR and 2 issues (both still open).

We could contact the person and see what the state of things are, they have moved over to null safety this year at least. If it's not possible, we could fork it and provide it under bluefire, since they have several nice algorithms in there.

erickzanardo:
Given that there is already a package for that, I don't see the point on adding this code source to Flame. If the lib is lacking something PRs should be added to it.

spydon:
I'll open an issue for how we should solve this instead, and if we come to the conclusion that this is something we'd want directly in the engine we can re-open this PR.

PR #1266, opened 2021-12-22
refactor!: Separate ComponentSet from the Component

st-pasha:
# Description

Currently, `ComponentSet` is tightly coupled with the `Component` class and is partially responsible for handling the `Component`'s lifecycle.

This PR breaks the two classes apart: `Component` will now perform its own lifecycle management, and `ComponentSet` will focus on being a simple container.

Several optional breaking changes introduced here. "Optional" because they can be reverted, at the cost of making the code less nice. However, since `ComponentSet` is mostly internal class, I deem the risk to end users low:
  - `Component.reAddChildren()` was made private. Calling it by end users was dangerous.
  - `Component.prepare()` declared protected. Likewise, it was never intended to be user-callable.
  - `ComponentSet` no longer has field `parent`, as it is no longer needed.
  - `ComponentSet` constructor no longer has argument `Component parent`, same for `ComponentSet.createDefault`.
  - Since we're changing `ComponentSet`'s constructor anyway, I use this opportunity to also change the `comparator` into a named argument, which allows it to be optional, so that no-argument `ComponentSet()` would create the component set with default comparator function. This also allows `ComponentSet.createDefault` to be greatly simplified, and in fact we will be able to remove it in Dart 2.15.
  - `ComponentSet.addChildren()` removed.
  - `ComponentSet.add()` and `ComponentSet.addAll()` are now prohibited (thus, the user can no longer say `myComponent.children.add(newComponent)`). In fact, it would be nice to prohibit direct manipulation of `ComponentSet` by the user altogether, but this PR doesn't go that far.

Out of scope for this PR: methods for removal of components.

## Checklist

## Breaking Change

  e.g. `feat!:`, `fix!:`).


spydon (approved):
Lgtm!

spydon (approved):

packages/flame/lib/src/components/component_set.dart
```diff
@@ -18,6 +19,19 @@ import '../../game.dart';
...
+  ComponentSet({
+    int Function(Component e1, Component e2)? comparator,
+    bool? strictMode,
+  }) : super(
+          comparator: comparator ?? Comparing.on<Component>((c) => c.priority),
+          strictMode: strictMode ?? defaultStrictMode,
+        );
+
+  // When we switch to Dart 2.15 this can be replaced with constructor tear-off
+  static ComponentSet createDefault() => ComponentSet();
```
luanpotter, line 33: Is this needed/used anywhere? The reason to have it before was bc it was actually configuring the created set. Now you could just call ComponentSet() directly?

st-pasha: Yes, this is used by `Component.childrenFactory`. Somehow using a simple function comprehension `() => ComponentSet()` produces an analyzer warning there.

st-pasha: Apparently, they fixed the warning just 8 days ago: https://github.com/dart-lang/linter/pull/3118, but it will probably take time until we can actually use that.

Issue #1164, opened 2021-12-05
Artificial intelligence for game development


nmarafo:
It would be great if flame engine included an artificial intelligence framework such as [gdx-ai](https://github.com/libgdx/gdx-ai).

spydon:
This sounds like something we could have as a bridge library, do you know any good Dart packages that could be bundled into this? I only know about the a-star lib.

nmarafo:
I've been looking and can't find anything intended to dart. You could do something similar to what was done with flame_forge2d and box2d.

Gdx-ai seems to me to be quite robust.

Therefore, if the conversion from java to dart is considered feasible and guidelines are set, the community could collaborate in that regard.

spydon:
> I've been looking and can't find anything intended to dart. You could do something similar to what was done with flame_forge2d and box2d.
> 
> Gdx-ai seems to me to be quite robust.
> 
> Therefore, if the conversion from java to dart is considered feasible and guidelines are set, the community could collaborate in that regard.

It is a looot of work to port big java libs to dart, but it can be done, so if anyone does it, I'd be happy to help with the bridge package.

nmarafo:
Well, I think it's wonderful that a line of work could be opened in this regard.
I'm going to study the gdx-ai structure a bit and from my humble knowledge see if a simple assimilation for Flame engine is feasible.

nmarafo:
After an arduous research task I have encountered several problems translating the library from java to dart. I don't know what type of Vector to use. I have seen that the flame-forge2d library uses "vector_math" but I have a problem about how to transcribe "class <T extends Vector <T>>", since I cannot find "Vector <T extends Vector>". I would also like to know from which library to import a Math class for Math.sqrt.
On the other hand, I think there will be no problem with the "Steering Behaviors" adapter, which at first seems to me to be one of the most useful functionalities to get started.

spydon:
For Vector, we use the `Vector2` class which is from `vector_math`, I don't think you'd have to generics on it.
`sqrt` exists in `dart:math`, so no need for a dependency there.

nmarafo:
I would need help again. Apologies in advance. 

If I extend Vector2 (flame / extensions.dart) this class uses a constructor with "factory" so when I transcribe the gdx-ai classes that extend Vector2, it forces me to build with factory.

The problem is that I can't call "super" if I build the new class with factory.

So I would need some way to make Vector2 not build with "factory" or skip that rule somehow when I create a class that extends Vector2. 

[...]

PR #1267, opened 2021-12-23
feat(ai): Add Pursue behavior from gdx-ai

st-pasha:
# Description

This is the beginning of the work on porting the gdxAI library into dart and integrating it into the Flame engine.

This PR adds the `Pursue` behavior, and implements some of the basic interface classes such as `Steerable`, `Limiter`, `Location`, `SteeringBehavior`, and `SteeringAcceleration`.

I have also made some modifications to the original Pursue algorithm, for comparison the timings of the old and the new algorithm are as follows:
| test name     | original time | current time |
|---------------|---------------|--------------|
| pursuit exact | 7.0           | 7.0          |
| pursuit 1     | 4.34          | 4.28         |
| pursuit 2     | 1000          | 14.34        |
| pursuit 3     | 1000          | 2.6          |
| pursuit 4     | 12.90         | 4.8          |
| pursuit 5     | 20.26         | 10.84        |
| pursuit 6     | 1.18          | 1.18         |
| chase 1       | 305.6         | 302.8        |
| chase 2       | 252.6         | 250.4        |
| chase 3       | 17.3          | 16.2         |

https://user-images.githubusercontent.com/4231472/147298375-8a5de9b3-7a18-41e9-bf1d-9707963347fd.mov

## Checklist

## Breaking Change

## Related Issues

WIP for #1164


spydon:
Maybe this should be `flame_ai`? Do we need to include the licence in the header of each file, can it not be in its separate licence file?

st-pasha:
> Maybe this should be flame_ai bridge package?

Well, there is nothing to bridge to. The gdx-ai library is in Java, so cannot be referenced directly. Plus, this library will eventually be deeply integrated with main Flame; for example all logic for camera motion are the examples of behaviors (seek/pursuit), so we'll refactor them later to make Camera steerable.

> Do we need to include the licence in the header of each file, can it not be in its separate licence file?

Here's how I understand the conditions imposed upon us by the Apache-2 license:
  1. _"You must give any other recipients of the Work or Derivative Works a copy of this License;"_ -- this copy is in the readme file;
  2. _"You must cause any modified files to carry prominent notices stating that You changed the files;"_ -- this is why every file has "translated into Dart with modifications" in capital letters;
  3. _"You must retain, in the Source form of any Derivative Works that You distribute, all copyright, patent, trademark, and attribution notices from the Source form of the Work, excluding those notices that do not pertain to any part of the Derivative Works"_ -- this is why every file carries the copy of the boilerplate note taken from the original file;
  4. _"If the Work includes a "NOTICE" text file as part of its distribution ..."_ -- not applicable since they don't have a NOTICE file.

I've also included a line stating that the translated file's copyright really belongs to the Blue Fire project, not to the original gdxAI team.

packages/flame/lib/src/ai/steer/steerable.dart
```diff
@@ -0,0 +1,45 @@
...
+  Vector2 get linearVelocity;
+
+  /// The angular velocity in radians of this Steerable.
+  double get angularSpeed;
+
+  /// The bounding radius of this Steerable.
+  double get boundingRadius;
+
+  /// Generic flag utilized in a variety of ways.
+  bool tagged = false;
```
spydon, line 44: I'm guessing this comment was just taken from the source, but it should definitely not be this vague. If it can only be this vague, it feels like it is a quite strange design to rely on this seemingly arbitrary variable for different things.

packages/flame/lib/src/components/steerable_component.dart
```diff
@@ -0,0 +1,117 @@
...
+import '../ai/steer/steering_acceleration.dart';
+import '../ai/steer/steering_behavior.dart';
+import '../anchor.dart';
+import '../extensions/vector2.dart';
+import 'position_component.dart';
+
+class SteerableComponent extends PositionComponent implements Steerable {
+  SteerableComponent({
+    Vector2? velocity,
+    double? angularSpeed,
```
spydon, line 15: `angularVelocity`?

packages/flame/lib/src/ai/steer/steerable.dart
st-pasha: Yeah, it felt weird for me too. I guess we can remove it now, and then add back later when and if it is actually needed.

packages/flame/lib/src/components/steerable_component.dart
st-pasha: _Velocity_ is a vector quantity, whereas _speed_ is scalar. Whereas in 3d space rotation would be represented as a vector, in 2d rotation is a simple scalar, and hence its rate of change should be called speed.

spydon: Speed doesn't have a direction though, and I guess this can be negative?

st-pasha: hmm, fair point. I guess i can rename it into velocity then

doc/toc.rst
```diff
@@ -36,6 +36,12 @@
     Particles                       <particles.md>
     Layers                          <layers.md>
 
+.. toctree::
+	:caption: AI
+	:hidden:
+
+	Behaviors      <behaviors.md>
```
spydon, line 43: Wrong indentation

packages/flame/lib/src/ai/README.md
```diff
@@ -0,0 +1,189 @@
+
+Most of the code in this directory was ported from [gdxAI] project by libGDX, licensed under
```
spydon, line 2: Maybe this should be a license file instead of a readme

packages/flame/lib/src/ai/README.md
```diff
@@ -0,0 +1,189 @@
...
+Apache v2.0. The text of the license is given below.
+
+For now, all ai-related code in Flame should be considered experimental, and its API may change
```
spydon, line 5: This would be an argument to have it in a separate package until it's stable and can be moved in to here? I do see the value of having it directly in here once it's stable.

packages/flame/lib/src/ai/steer/steerable.dart
```diff
@@ -0,0 +1,42 @@
...
+import '../utils/location.dart';
+import 'limiter.dart';
+
+/// A [Steerable] is a [Location] that gives access to the character's data
+/// required by steering system.
+///
+/// Notice that there is nothing to connect the direction that a Steerable is
+/// moving and the direction it is facing. For instance, a character can be
+/// oriented along the x-axis but be traveling directly along the y-axis.
+abstract class Steerable extends Location with Limiter {
```
spydon, line 33: This is a strange class hierarchy, `Steerable` being a type of `Location` sounds strange to me at least. I guess this is from the ported library and not something you came up with? It could have `implements Location` instead maybe?

packages/flame/lib/src/ai/steer/steering_acceleration.dart
```diff
@@ -0,0 +1,67 @@
...
+
+  /// Adds the [other] steering acceleration to this steering acceleration.
+  void add(SteeringAcceleration other) {
+    linearAcceleration.add(other.linearAcceleration);
+    angularAcceleration += other.angularAcceleration;
+  }
+
+  /// First apply [scale] to the supplied [steering], then add it to this
+  /// steering acceleration.
+  void mulAdd(SteeringAcceleration steering, double scale) {
```
spydon, line 56: No short forms of words, and strange to bake the two operations together into one function? Or is this used a lot and it would be too verbose to have it as two calls?

packages/flame/lib/src/extensions/vector2.dart
```diff
@@ -75,6 +75,15 @@ extension Vector2Extension on Vector2 {
     }
   }
 
+  /// Ensure that the vector is no larger than [max] in magnitude. Scales the
+  /// vector down without changing its direction if necessary.
+  void clampMagnitude(double max) {
```
spydon, line 80: Is this needed when `clampScalar` exists in the original class?

spydon (commented):
Good job, will be cool to use this.

doc/toc.rst
st-pasha: right, tabs sneaked in.

packages/flame/lib/src/ai/steer/steerable.dart
st-pasha: Originally all of them are interfaces, so `Steerable` simply extends both `Location` and `Limiter`

packages/flame/lib/src/ai/steer/steering_acceleration.dart
st-pasha: I think they merely follow naming convention of Java's Vector class. So we should probably do the same and rename it into `addScaled()`, following the convention from our `Vector2`.

packages/flame/lib/src/ai/README.md
st-pasha: It can't be in the LICENSE file, because then it would apply to all code in the directory and make it seem as if the code is owned by libGDX. 

So what we could do instead is to create the directory `flame/licenses/`, and put that license file there, together with the readme explaining how it applies.

packages/flame/lib/src/ai/steer/steerable.dart
spydon: Then why isn't the implements keyword used, instead of extends?

packages/flame/lib/src/ai/README.md
st-pasha: I don't see any advantages to the separation. Clearly, the code is very new and may need to change somewhat in the future, but it doesn't destabilize the rest of the system. So it should be fine, as long as we make it abundantly clear that this part of the code doesn't have the same stability guarantees as the rest the system. I moved the warning into the main documentation page, so that it can't be missed.

In general, we've seen in the past (e.g. `mock_canvas`) that moving things into separate packages, creates a significant burden for development.

packages/flame/lib/src/ai/steer/steerable.dart
st-pasha: We could make `Steerable` implement `Location` and `Limiter` -- but that would require copy-pasting code from those classes into here. Another alternative is to use both Location and Limiter as mixins, so that it becomes `abstract class Steerable with Location, Limiter`. WDYT, which way is better?

packages/flame/lib/src/extensions/vector2.dart
st-pasha: `clampScalar` will constrain the vector to lie in the box `[min, max] x [min, max]`. So this does not ensure maximum length of the vector, and will also cause the vector to change the direction.

But perhaps `clampLength` might be a better name here.

spydon: https://github.com/flame-engine/flame/blob/main/packages/flame/lib/src/extensions/vector2.dart#L71

st-pasha: `scaleTo` scales the vector to the desired length, this function scales the vector only if its length already exceeds `max`. So, it is equivalent to `if (vector.length > max) { vector.scaleTo(max); }`, except that it avoids computing the length of the vector twice.

luanpotter:
Porting an AI library should be done in a bridge package, like flame_gdx_ai. Similar to how we do flame_box/forge2d.
If our core Flame code is not flexible enough to allow any couplings we want to be done from the flame_x package side, we should, instead, improve it so it does.

st-pasha:
> Porting an AI library should be done in a bridge package, like flame_gdx_ai. Similar to how we do flame_box/forge2d.

There is `forge2d` package, and `flame_forge2d` serves as a bridge, allowing to easier use `forge2d` library together with the core Flame. Similarly, there is `rive` package, and then `flame_rive` serves as a bridge to bring Rive animations into Flame. But there is no `gdx-ai` package (not in Dart ecosystem at least). There is nothing to bridge to. 

In addition, this functionality does belong to the Flame core. We already have some rudimentary steering behaviors, although they are done in a bespoke non-reusable way. So, this is going to be a massive upgrade of the existing functionality.

luanpotter:
Despite the name, the bridge packages don't necessarily need to "bridge" to anything external, even though most do. For example, flame_splash_screen doesn't bridge to a 3rd package; rather, it contains within itself all of its functionality, that we thought to separate from Flame just for separation of concerns. The forge2d package exists because we decided to created it, as a place to port box2d (existing non-dart lib) into dart. Similarly, if we are porting an external library from another language, it would be much more valuable to encapsulate it in a dart package, so that anyone can use it, even if they don't use Flame itself, or even Flutter. Then, the flame bridge package would just bridge that into Flame components and reusable pieces. Similar to how we extracted canvas_test and that package is valuable to anyone testing canvas, idependently of Flame. Regardless of the 3rd package being made (to provide the implementation to the dart community outside of Flame), I think the code should be better separated into its own package. I am not aware of any AI related coded currently inside Flame, sounds like a very extensive set of tools that would be great to have but feel like could be properly separated. I am not sure what you mean by steering behaviors.

st-pasha:
Separation of concerns is great when done on the level of methods, or classes, or even sub-packages. But separating things into different libraries is a step too far. Just because effects can, in theory, be torn away and put into a separate library -- doesn't mean they should. Just because particle systems, or collision detection, or some other components like text or parallax can be separated into their own tiny libraries -- doesn't mean they should.

There is very little in terms of benefits that can be achieved there, whereas the cost is quite immediate and real. These costs are: (1) the need to maintain another 2 separate libraries, (2) the need to build a separate documentation system, (3) the need to create a new test harness, (4) the need to build a separate example library, (5) the cost of synchronizing APIs between libraries, (6) separate release schedule, etc.

Taken together, that's a big cost. And I'm not saying this theoretically: from the very practical standpoint, none of the libraries that we released as separate packages achieve the same level of quality as the main Flame library.

In short, I'm not in favor of shooting a project in the foot before it even had a chance to make its first baby steps.

spydon:
I agree with this, we shouldn't make bridge libraries just because we can, it will be less easy for the user to use too. But I think we need a plan for what should be ported here, if there will be a lot of changes I think this should start off as a separate package (not two) with a version <1.0.0 and then simply move it in when it is done. If this feat is expected to have a fairly stable api within 2 months (next stable release) then I think it can be kept in the main package.

st-pasha:
It is difficult to anticipate the amount of change that will be needed. On one hand, the source library is quite stable. On the other, there's been quite a few non-trivial adjustments to their code just in making this PR. 

We could simply avoid exporting this into main Flame namespace. Thus, if someone wants to use it, they'd have to access it via `package:flame/lib/src/ai/...`.

erickzanardo:
I agree with people that this should be a separate package.

spydon:
> We could simply avoid exporting this into main Flame namespace. Thus, if someone wants to use it, they'd have to access it via `package:flame/lib/src/ai/...`.

I think that sounds like a pretty good compromise, until it becomes more stable and we'll export it.

packages/flame/lib/src/ai/steer/steerable.dart
spydon: Yeah, I think the mixin way is definitely better.

st-pasha:
> Possibly rename the directory to something else than AI, I think this is much more generic than that?

What do you have in mind?

To me, "AI" sounds generic enough: _artificial intelligence_ -- any functionality that helps make your game more smart. In addition to behaviors, we'll also have reactive actions, movement in formations, path-finding, behavior trees, etc.

spydon:
We talked in the team yesterday about having this as a separate library or not, since Erick and Luan seem to prefer that.
And I had a more thorough look at gdx-ai today, and it is massive!

The pros and cons I see are the following (please fill in the list with things I miss):

As two packages `gdx-ai` (maybe named something fitting our "naming convention") and `flame_gdx_ai`:
 + The code in `gdx-ai` can be re-used in any dart project.
 + It can be released whenever deemed necessary.
 - More maintenance overhead since two more libraries need to be released.
 - Possibly docs, but the plan is to have all docs in the flame docs I think.

Inside of Flame:
 + Easier to use for our users.
 - Possibly a lot more code in the main package.

If we have a pure Dart `gdx-ai` port we can have you as the maintainer of that lib, if that is something you'd be interested in @st-pasha?

st-pasha:
Here's how I view the breakdown of various factors:

## Inside of Flame
- **`+` Easier to use for our users**
  This is a somewhat weak benefit: sure, it's much easier to depend on a single package than on three different ones, while  also keeping up-to-date with upgrading their versions, etc. But it's not too much work.
- **`+(+)` Discoverability**
  If the code is inside Flame, then it will be documented and the users will know that it's available for use. If the code is outside then it's much harder to discover. This warrants 1 `+` if we advertise the library on the main page of Flame docs, and 2 `+`s if we don't.
- **`++` Trust**
  Any external package that the user might want to use, needs to be vetted. Is it trustworthy? Is it actively maintained? Is it supported by a single developer or a team (bus factor)? Is it well documented? etc. Being inside Flame, all of these qualities are automatic.
- **`o` A lot more code in the main package**
  In some languages, having lots of code in the library is problematic, because the library is compiled into a single executable, and that executable is then used by the downstream users. However, Dart [_is not_](https://github.com/dart-lang/sdk/issues/33920) such a language. The tree-shaking algorithm ensures that the users who don't use these features will not be burdened by them.
- **`o` The code will need to be maintained**
  I understand how this could be perceived as a `-`, but in reality the code will need to be maintained anyway -- whether it'll be in Flame or in some other repo. And maintaining it inside Flame is easier, because there will be less temptation to "sweep it under the rug".
- **`+++` Ease of development**
  This is a huge benefit from the developer standpoint. Flame has existing system for publishing examples/documentation. The Flame engine is needed for running tests. Compared to the alternatives, it's much easier to make changes in one package that are needed in another (this is currently a major pain point with our existing packages `mock_canvas` and `ordered_set`).
- **`++` Flame is perceived as a more powerful library**
  The users will appreciate that Flame can fulfill all or most of their needs for game development, without them having to "shop for extra parts". This is like a difference between renting a fully-furnished apartment, and a bare-walls one. In practice you can think of it like this: in the popularity context between different libraries, people try to use at easily available metrics: stars on github, mentions on StackOverflow, downloads on pub.dev etc -- spreading the functionality across many small packages makes each of them appear small and insignificant, whereas taken together they are much more powerful.

## Externally

- **`+` The code can be reused in other Dart projects**
  This is a weak benefit: the code is useful for games only. And it can be reused when it is inside Flame too, though there could be a perception that it is somehow "too heavy". It is not, because of the tree-shaking.
- **`-` Separate release process**
  I'm putting this as a `-` because Flame is already committed to regular releases, so compared to the first option there is little to be gained by having a separate release process. However, release process has its costs, and these costs will now have to be paid in full.
- **`--` Documentation**
  Currently we have a working system for publishing Flame documentation for each version (including examples). There is no comparable system for all other packages (even those in the mono-repo). Until such system is developed, this would be a major drawback.
- There could also be different variants for the dependence structure:
  - **Flame depends on this ai library**
    - **`-` Adds an external dependency to Flame** 
      External dependencies should generally be avoided, since they increase perceived size and risk profile of the main library. Say, a user may trust the Flame team, but how sure can they be in our 3rd-party vetting process?
  - **Add a bridge library between flame and the ai library**
    - **`--` The need to support another library**
      Supporting a new library has substantial costs. In this scenario we'll be supporting one additional library, so that's extra costs to the maintainers.
    - **`---` Code duplication**
      Since main Flame has the need for some of the ai behaviors (specifically for Camera movement), we'll end up creating the same functionality twice. And because this functionality will leave in separate classes, the end user will have to pay this cost too, as the functions will end up included twice in the final executable.
 
## Summary
In my view, the benefits of including this functionality into Flame far outweigh the alternative of releasing the package separately. Of course, many of these arguments are somewhat speculative -- but many are drawn on the existing examples of packages that we have as externals, and which present significant challenges to the development.

If anything, we should start thinking of how we can bring our many disparate packages under a single umbrella (at least into the mono-repo).

spydon:
> Here's how I view the breakdown of various factors:
> ## Inside of Flame
>     * **`++` Trust**

The package would live under the same organization, and that would be clear on pub, so I don't think this is an issue.

>     * **`o` The code will need to be maintained**
>       I understand how this could be perceived as a `-`, but in reality the code will need to be maintained anyway -- whether it'll be in Flame or in some other repo. And maintaining it inside Flame is easier, because there will be less temptation to "sweep it under the rug".

If the full gdx-ai library should be ported it is probably close to 10k lines of code, it feels like that deserves its own pipeline, but I agree, the code does still need to be maintained.

>     * **`+++` Ease of development**
>       This is a huge benefit from the developer standpoint. Flame has existing system for publishing examples/documentation. The Flame engine is needed for running tests. Compared to the alternatives, it's much easier to make changes in one package that are needed in another (this is currently a major pain point with our existing packages `mock_canvas` and `ordered_set`).

The plan is to merge all examples and docs, so I don't think this deserves three `+`. I have not had any pain points regarding this when developing things in `ordered_set`, quite the opposite, since it is contained in its own repo and doesn't need Flame the pipeline it is ten times faster. Can you elaborate on why you think it is a pain point?

>     * **`++` Flame is perceived as a more powerful library**
>       The users will appreciate that Flame can fulfill all or most of their needs for game development, without them having to "shop for extra parts". This is like a difference between renting a fully-furnished apartment, and a bare-walls one. In practice you can think of it like this: in the popularity context between different libraries, people try to use at easily available metrics: stars on github, mentions on StackOverflow, downloads on pub.dev etc -- spreading the functionality across many small packages makes each of them appear small and insignificant, whereas taken together they are much more powerful.

If we document it well I don't think that Flame will be perceived as less powerful, and I don't think we'd get less stars on github on the monorepo because we have this separation. I mean, we don't even have audio support in the core library.

> ## Externally
> 
>     * **`+` The code can be reused in other Dart projects**
>       This is a weak benefit: the code is useful for games only. And it can be reused when it is inside Flame too, though there could be a perception that it is somehow "too heavy". It is not, because of the tree-shaking.

A lot of people have used `Forge2D` outside of Flame and I think this would fall under many of the same use cases as Forge2D.
We of course know about tree-shaking, what we don't want is a big monolith. Even gdx-ai doesn't require libgdx to be used.

>     * **`-` Separate release process**
>       I'm putting this as a `-` because Flame is already committed to regular releases, so compared to the first option there is little to be gained by having a separate release process. However, release process has its costs, and these costs will now have to be paid in full.

It is not hard to do a release to pub, you simply write `pub publish` (if you don't use melos) and then it is done. With a separate release schedule it can be released faster than we'd do flame releases, so I would definitely put this as a `+`. It will be much harder to cherry pick patches and do fast releases when it is within the core.

>     * **`--` Documentation**
>       Currently we have a working system for publishing Flame documentation for each version (including examples). There is no comparable system for all other packages (even those in the mono-repo). Until such system is developed, this would be a major drawback.

See reply on "Ease of development".

>     * There could also be different variants for the dependence structure:
>       
>       * **Flame depends on this ai library**

This is not an option for us.

> ## Summary
> 
> ... which present significant challenges to the development.

What are these significant challenges?

> If anything, we should start thinking of how we can bring our many disparate packages under a single umbrella (at least into the mono-repo).

The umbrella is blue-fire and flame-engine, packages that can be used without Flame should not live within the monorepo.

Before I knew that the gdx-ai package was so big I thought it was a pretty good plan to put it within the core, but know I'm leaning towards @erickzanardo's and @luanpotter's standpoint to have it as a separate package.

st-pasha:
> The package would live under the same organization, and that would be clear on pub, so I don't think this is an issue.

Maybe it's just me. But it took me more than 2 months to realize that `ordered_set` belongs to BlueFire team; and I just learned _yesterday_ that `audioplayers` does too.

> The plan is to merge all examples and docs, 

Obviously this is a great plan, deserves a :+1: or even :+2:. But for now it is still more of an idea than an actual plan. I mean, there is no tracking issue, or a project, or an approximate estimate for how long it would take, or how much effort will be required, or commitment from specific people to do it. Surely, it can all magically materialize before the New Year, but it could also take several months or even quarters.

> I have not had any pain points regarding this when developing things in ordered_set, quite the opposite, since it is contained in its own repo and doesn't need Flame the pipeline it is ten times faster. Can you elaborate on why you think it is a pain point?

Here's an actual example from `canvas_test` that I've encountered. Suppose I'm working on a new Flame feature, adding a test to draw something on a canvas -- and it turns out that there is no `MockCanvas.drawCircle` method. Ideally, if `MockCanvas` belonged to the `flame_test` repository, I would have went in and added the missing method in 5 minutes. 

Instead, I'd need to (1) fork the canvas_test repository, (2) set up Android Studio for the new project, (3) implement the `drawCircle` functionality, including tests, (4) make a PR, (5) wait for the PR to be approved and merged, (6) make another PR for the new release of `canvas_test` (actually I'm not entirely sure how to do this part), (7) wait for the PR to be approved and merged (optionally: spend some time arguing whether a single function is release-worthy or maybe it should be released later), (8) make a third PR in the main flame repository bumping the version of the `canvas_test` dependency, (9) wait for that PR to be approved and merged (hopefully quicker than the first two), (10) pull these changes into the feature branch that you were originally working on -- only now I'm finally unblocked to work on my feature.

Note that most of these concerns wouldn't apply if the ai library was inside the mono-repo and flame depended on it via relative import.

> A lot of people have used Forge2D outside of Flame and I think this would fall under many of the same use cases as Forge2D. We of course know about tree-shaking, what we don't want is a big monolith. Even gdx-ai doesn't require libgdx to be used.

I wonder what's their use-case though? Developing game engines alternative to Flame? Do we even want to encourage that?

For gdx-ai I suspect their prime motivator is to make their engine jar files smaller, so that developers who don't need AI capabilities wouldn't have to carry it along. After all, Java doesn't have tree-shaking like Dart.

> It is not hard to do a release to pub, you simply write `pub publish` and then it is done. 

There is always some hidden cost of publishing, and in general maintaining the CI infrastructure. Creating release branches and tags, making GitHub release, resolving deprecation events, updating readme, creating changelog, re-building documentation, making public announcements, etc.

st-pasha:
> The umbrella is blue-fire and flame-engine, packages that can be used without Flame should not live within the monorepo.

We can always design the package in such a way that it cannot be used without Flame. Ultimately, this AI engine is for use in games, it has no external purpose like, say, `audioplayers`.

------
Ultimately, though, is the question of scope: is this functionality within the scope of Flame engine?

Suppose, for example, that there was no gdxAI project. I came in and say: look, we have position-components which can be placed anywhere, and we also have move effects that can move them in a predefined way, and even a "speed" parameter that can change the effect based on the initial position of the target. But wouldn't it be nice if we also added more complicated types of behavior, where targets update their trajectory continuously, where they can react to other targets or to the environment?

How do we decide what's in Flame's scope and what's outside?

If the concern is that the code originates from gdxAI -- then we can develop it independently. It would still be easier than maintaining a separate project.

If the concern is code size (and btw, Java is about twice more verbose than Dart), then we can limit the amount of things we are going to implement.

If the concern is that the code can in theory be separated, then I don't understand what's the scope of the Flame library. Because it has plenty of things that could have been in separate packages (like particles, effects, collisions, etc).

erickzanardo:
For me the reason for this being a package on its own is pretty simple, this is a port of another lib, better keep its code base apart from Flame.

st-pasha:
> For me the reason for this being a package on its own is pretty simple, this is a port of another lib, better keep its code base apart from Flame.

So, would you be ok with including this (or equivalent) functionality into Flame, if it was developed independently from gdx-ai or any other existing library?

spydon:
> Maybe it's just me. But it took me more than 2 months to realize that `ordered_set` belongs to BlueFire team; and I just learned _yesterday_ that `audioplayers` does too.

They were moved over quite recently from Luan's personal account, if anyone is actually concerned about the origin of a package it is very easy to see that they come from us now though. 

> Obviously this is a great plan, deserves a +1 or even :+2:. But for now it is still more of an idea than an actual plan. I mean, there is no tracking issue, or a project, or an approximate estimate for how long it would take, or how much effort will be required, or commitment from specific people to do it. Surely, it can all magically materialize before the New Year, but it could also take several months or even quarters.

It can be started for new projects already, so all gdx-ai docs would be written just like if they would have been included in the core.

> Instead, I'd need to (1) fork the canvas_test repository, (2) set up Android Studio for the new project, (3) implement the `drawCircle` functionality, including tests, (4) make a PR, (5) wait for the PR to be approved and merged, (6) make another PR for the new release of `canvas_test` (actually I'm not entirely sure how to do this part), (7) wait for the PR to be approved and merged (optionally: spend some time arguing whether a single function is release-worthy or maybe it should be released later), (8) make a third PR in the main flame repository bumping the version of the `canvas_test` dependency, (9) wait for that PR to be approved and merged (hopefully quicker than the first two), (10) pull these changes into the feature branch that you were originally working on -- only now I'm finally unblocked to work on my feature.

I can see your point, 6 and 7 doesn't work like that though, we publish a new version and then we do the PR afterwards. And 8-10 could be done in the same PR as where you need the changes and you could have your PR depend on your branch before it is merged. For these kind of libraries which can develop on their own we definitely don't want them inside of the monorepo, these can live completely without flame and if someone wants to continue work on them they can simply fork only the small repo instead of the full monorepo.

> I wonder what's their use-case though? Developing game engines alternative to Flame? Do we even want to encourage that?

Of course, anything that gives our users freedom to come up with whatever crazy ideas they have. One person for example created a site with directed graphs and each entity was controlled with a Forge2D body and springs, completely outside of Flutter. We don't want to lock in people to Flame and Flutter when we don't need to, if it isn't making our lives too hard, so that is  what we are really discussing. A pure dart library with gdx-ai could be used with so much more than just Flame, compare with the a-star library for example.

> There is always some hidden cost of publishing, and in general maintaining the CI infrastructure. Creating release branches and tags, making GitHub release, resolving deprecation events, updating readme, creating changelog, re-building documentation, making public announcements, etc.

Almost all of that should be automated though, and just as much work outside the monorepo as within it. We have generalized most of the pipeline actions so that they can be used in different repositories.

> How do we decide what's in Flame's scope and what's outside?

This is a very good question! I'll make sure that we discuss that on our next meeting!
(Everyone has been a bit off now during the holiday, but we should be having a meeting soon)
It might sound like I'm completely against having this in the mainrepo, but I'm not, I just want to weigh all the pros and cons.

st-pasha:
Ok, I surrender. I'll create a new package.

Issue #1293, opened 2022-01-04
FlameGame.onGameResize - Zero value when game is closed and restarted


couturmi:
# Current bug behaviour
I'm seeing this issue on iOS (physical device only) when the app is closed/dismissed then quickly reopened.

On the first run the `onGameResize` will return a nonzero `Vector2`.
After closing & reopening, the `onGameResize` returns (0,0).

The below images show the different states of printing the `Vector2` passed to `onGameResize` on startup:
| First Run| After Closing and Restarting |
| ------- | ------- |
| <img src="https://user-images.githubusercontent.com/9330448/148103606-4eeffad6-64e0-4fc7-80cd-cd667c37f95e.PNG" height=400> | <img src="https://user-images.githubusercontent.com/9330448/148103570-4299692c-6655-48b6-8c10-69f80ae2fc40.PNG" height=400> |

This issue is causing any components that depend on game size (whether it be size or position) to be incorrectly loaded on startup. In my case, many components are all appearing in the top left corner rather than the center since the gameSize is 0.

# Expected behaviour
`onGameResize` should always return a nonzero size when the app is started.

# Steps to reproduce
I have only noticed this issue on physical iOS devices (simulator does not have this issue).

This occurs with only this small amount of code in a new Flutter project:
```
import 'package:flame/components.dart';
import 'package:flame/game.dart';
import 'package:flutter/material.dart';

class MyGame extends FlameGame {
  String text = "";
  @override
  Future<void> onLoad() async {
    text += "onLoad occurred\n";
    await super.onLoad();
    add(TextComponent(text: text, position: Vector2(50, 50)));
  }

  @override
  void onGameResize(Vector2 canvasSize) {
    text += "onGameResize occurred\n\tCanvas Size: $canvasSize\n";
    super.onGameResize(canvasSize);
  }
}

main() {
  final myGame = MyGame();
  runApp(
    GameWidget(
      game: myGame,
    ),
  );
}

```
1. On a physical device, open the app and confirm that the size text displayed is a nonzero Vector2.
2. Swipe up and dismiss/close the app.
3. Quickly reopen the app and confirm that the size text is now (0,0). _(If this did not occur, repeat steps 2-3 a few times until it does)_

# More environment information
* Flame version: 1.0.0
* Flutter version: 2.8.1
* Dart version: 2.15.1

# Log information
n/a

spydon:
Strange, is there another resize after the (0, 0) resize?

couturmi:
[...]

PR #1308, opened 2022-01-12
refactor: Resize logic in GameRenderBox

st-pasha:
# Description

The documentation of `RenderBox` says the following:
```dart
/// When implementing a [RenderBox] subclass, one must make a choice. Does it
/// size itself exclusively based on the constraints, or does it use any other
/// information in sizing itself? An example of sizing purely based on the
/// constraints would be growing to fit the parent.
///
/// Sizing purely based on the constraints allows the system to make some
/// significant optimizations. Classes that use this approach should override
/// [sizedByParent] to return true, and then override [computeDryLayout] to
/// compute the [Size] using nothing but the constraints, e.g.:
```
This PR changes `GameRenderBox` to adhere to this advice and set `sizedByParent` to true. 

## Checklist

## Breaking Change

## Related Issues

Maybe helps with #1293 ?

### DIFF
--- a/packages/flame/lib/src/game/game_render_box.dart
+++ b/packages/flame/lib/src/game/game_render_box.dart
@@ -19,10 +19,10 @@ class GameRenderBox extends RenderBox with WidgetsBindingObserver {
   bool get isRepaintBoundary => true;
 
   @override
-  void performResize() {
-    super.performResize();
-    game.onGameResize(constraints.biggest.toVector2());
-  }
+  bool get sizedByParent => true;
+
+  @override
+  Size computeDryLayout(BoxConstraints constraints) => constraints.biggest;
 
   @override
   void attach(PipelineOwner owner) {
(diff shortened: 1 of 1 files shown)


spydon (approved):

packages/flame/lib/src/game/game_render_box.dart
```diff
@@ -58,11 +57,6 @@ class GameRenderBox extends RenderBox with WidgetsBindingObserver {
     markNeedsPaint();
   }
 
-  @override
-  void performLayout() {
```
luanpotter, line 62: don't you still need to update the size variable when flutter changes our size?

luanpotter (approved):
Nice!

spydon:
Seems like @renancaraujo changed it the other way around here:
https://github.com/flame-engine/flame/pull/845

renancaraujo (approved):
Makes sense now.

packages/flame/lib/src/game/game_render_box.dart
st-pasha: That's what I thought at first too, but it turns out https://github.com/flame-engine/flame/blob/main/packages/flame/lib/src/game/game_widget/game_widget.dart#L319 already does that.

PR #1317, opened 2022-01-14
fix!: Change Tiled batched rendering to batched rendering per layer

### LINKED ISSUE #1264: [Tiled] tiles in different layers are conflict each other
You can clone my reproduction repo and try it

# Current bug behaviour
tiles in tileset are conflicting with other tiles in other tileset
For example, I have Bending machine tile(furniture tileset) and wall tile(walls tileset) or floor tile(floors tileset) in background layer behind the former.
But the bending machine tile does not rendered.
In my case, I found following rules
1. floors tileset with furnitures or exterior tileset are renders only floor tile
2. floors with interior tileset works OK
3. walls tileset works with nothing
Screenshot_20211222-160229

# Expected behaviour
Renders map as same as the official map editor
스크린샷 2021-12-22 오후 4 02 18

# Steps to reproduce
Compare between Tiled(Official map editor) program and rendered by flame_tiled

- flame_tiled
1. Clone [reproducible repo](https://github.com/Hwan-seok/flutter-flame-tiled-iss-repro) -> Resources in repo are only for reproduction and must not use it for other purposes.
2. flutter pub get
3. flutter run

- Tiled
1. Find tmx map file from cloned repo - `assets/tiles/map.tmx`
2. Open it with Tiled map editor

# Flutter doctor output
``` flutter doctor -v                                           ✔
[✓] Flutter (Channel stable, 2.8.0, on macOS 11.6 20G165 darwin-x64, locale
    ko-KR)
    • Flutter version 2.8.0 at /Users/lafity111/flutter
    • Framework revision cf44000065 (13 days ago), 2021-12-08 14:06:50 -0800
    • Engine revision 40a99c5951
    • Dart version 2.15.0

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at /Users/lafity111/Library/Android/sdk
    • Platform android-31, build-tools 30.0.3
    • Java binary at: /Applications/Android
      Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.8+10-b944.6916264)

[✓] Xcode - develop for iOS and macOS (Xcode 13.0)
    • CocoaPods version 1.11.2

[✓] Android Studio (version 4.2)
    • Java version OpenJDK Runtime Environment (build 11.0.8+10-b944.6916264)

[✓] VS Code (version 1.63.2)
    • Flutter extension version 3.29.0

[✓] Connected device (1 available)
    • SM G977N (mobile) • R3CM6093EAL • android-arm64  • Android 11 (API 30)

• No issues found!
```

* Flame version: 1.0.0

# Log information

```
No abnormal logs
```

# More information

lfraker:
# Description

Rendering a layered tiled map ran into an issue where sub layers would overwrite the tiles placed on top (in a higher layer), due to the map being rendered in the arbitrary order in which the SpriteBatches were loaded. This fix introduces a list of SpriteBatch maps corresponding to the visible TileLayers. The BatchMaps are then rendered in order of the layers.

This was written to not be a breaking change for the existing RenderableTiledMap, although there is no good way to UnitTest this, without writing a separate constructor path or replicating/exposing private methods to avoid the new code. I was only able to manually test backwards compatibility.
**Update:** Per code review, this is now a breaking change.

Wasn't able to test performance implications, but unless there is a significant number of spritesheets and layers, perf probably shouldn't be much of an issue (going from a few iterations through number of spritesheets --> iterations through number of spritesheets * number of laters).

## Checklist

## Breaking Change

  e.g. `feat!:`, `fix!:`).

Per code review comments, updated to be a breaking change and remove use of batches property.

## Related Issues

Fixes #1264


packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -80,23 +90,48 @@ class RenderableTiledMap {
...
+  }
+
+  static Future<Map<String, SpriteBatch>> _deepCopySpriteBatchMap(
+    Map<String, SpriteBatch> copyFrom,
+  ) async {
+    final result = <String, SpriteBatch>{};
+
+    await Future.forEach(copyFrom.keys, (String src) async {
+      result[src] = await SpriteBatch.load(src);
+    });
```
spydon, line 104: This could probably be written a bit simpler I think?
Something like:
```dart
    return {
      for (var key in copyFrom.keys) key: await SpriteBatch.load(key)
    };
```

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -23,12 +23,16 @@ class RenderableTiledMap {
   /// The size of tile to be rendered on the game.
   final Vector2 destTileSize;
 
+  /// Cached list of [SpriteBatch]es, ordered by layer.
+  final List<Map<String, SpriteBatch>>? batchesByLayer;
```
spydon, line 27: Can't the `final Map<String, SpriteBatch> batches;` map be removed and only this map be populated instead?
Either the `batches` could be turned into a getter, or just removed completely, it should mostly be used internally right?

spydon: Then we get rid of a lot of nullchecks too.

spydon (commented):
Good job! Some small comments.

packages/flame_tiled/lib/src/renderable_tile_map.dart
lfraker: I figured that would make this a breaking change, since RenderableTiledMap exposes batches in the public constructor, so I left it in for backwards compatibility. If you're confident that no one is using that constructor externally, then I can make the change. please let me know?

If we make the change, the getter would also be a dummy getter and just point to the first map in the list.

packages/flame_tiled/lib/src/renderable_tile_map.dart
lfraker: Great point, just updated!

packages/flame_tiled/lib/src/renderable_tile_map.dart
spydon: > If we make the change, the getter would also be a dummy getter and just point to the first map in the list.

Sgtm!

lfraker: @spydon, sorry but before pushing the update, I want to confirm how much of a breaking change you're ok with me making:

**Option 1 [Breaking Change]** - Remove the field and remove the parameter entirely from the constructor. Replace it with the list as a required parameter (changes a public-facing contract). Add a ```@Deprecated``` batches getter that returns first in the list or an empty map.

**Option 2 [Backwards Compatible Contract]** - Remove the field, but keep the public facing contract the same with the optional, additional list parameter (batches is just an unused parameter passed in, not stored). Add a ```@Deprecated``` batches getter that returns first in the list or an empty map. I'll need to ignore the linter rule for unused parameters here.
It's worth noting that if anyone is directly using the RenderableTiledMap constructor, and passing in pre-loaded batches, their map would not render. This is still a breaking change from a functional perspective.

Let me know which is preferable and I will push. Thanks!

spydon: Sorry for the slow reply, I had to double check with the rest of the team first. Since this is for a bridge library and not the core library you can go with option 1 (this library was accidentally put at version 1.0.0, it really should be <1 still).

lfraker: Thanks for checking, just updated to proceed with Option 1.

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -49,11 +57,20 @@ class RenderableTiledMap {
...
+
+    return RenderableTiledMap(
+      map,
+      batchesByLayer,
+      destTileSize,
+    );
   }
 
+  static Iterable<TileLayer> _renderableTileLayers(TiledMap map) =>
+      map.layers.where((layer) => layer.visible).whereType<TileLayer>();
```
spydon, line 72: All functions that spread across more than one line in our code base should use the `{ }` notation.

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -80,23 +97,31 @@ class RenderableTiledMap {
...
 
-    map.layers
-        .where((layer) => layer.visible)
-        .whereType<TileLayer>()
+    var layerInd = 0;
+    _renderableTileLayers(map)
         .map((e) => e.tileData)
         .whereType<List<List<Gid>>>()
-        .forEach(_renderLayer);
+        .forEach(
```
spydon, line 113: You could use `forEachIndexed` here from `'package:collection/collection.dart'`, that collection package is in flutter, but you have to add it to the pubspec file if it isn't there already:
```yaml
  collection: ^1.15.0
  ```

spydon (approved):
Nice, totally worth the breaking change, so much cleaner now! :)

packages/flame_tiled/lib/src/renderable_tile_map.dart
lfraker: Wasn't sure if it'd be ok to add another dependency for that, but it's much cleaner with it. Thanks!

spydon: Yeah, normally not, but this collection package is already in Flame so we can just as well have it here too.

luanpotter (approved):
LGTM

spydon:
flame_tiled version 1.1.0 is now released containing this fix :tada:

Issue #1321, opened 2022-01-17
The contact callback will be called even the object is not contacted.


BlockRen:
There will be a problem when two same class A in the listener list, because the forEach and _mayCallback will callback the both but only one is in the contact.
The callback process is no need to do 'register' and 'forEach' loop to fetch.

alestiago:
I think this has now been solved with #1547.

Can this be closed? (cc: @spydon )

PR #1322, opened 2022-01-17
fix: Optimize the contact callback process.

BlockRen:
Sorry for late. Yes, this is a breaking change, it's my mistake for description.

The original code may cause a bug when register more than one same class, such as I register two Football. Then the two class both will be selected in the functions of 'ContactTypes' and both will be callback. So maybe the two balls will both move if we don't do anything else to make the two Football different.

About "why there are no need to register the callbacks". Because the 'Contact' call back from the forge2d (beginContact, endContact) already tell us both the contacted objects. 'Contact' includes the two bodies and the two fixtures that contacted. And we also can get the contacted object using the 'userData' of 'BodyDef'.

Using my modified code, the listening callback can be like below:

class Football extends PositionBodyComponent with ContactCallback {
  ...
  @override
  void begin(a, b, Contact contact) {
    // Here a is the Football, b is another contacted object.
  }

  @override
  void end(a, b, Contact contact) {}
}

---------------- new above --- old below --------------

# Description

The callback process is no need to do 'register' and 'forEach' loop to fetch.
There will be a problem when two same class A in the listener list, because the forEach and _mayCallback will callback the both but only one is in the contact.

## Checklist

- [x] I have did tests for ALL new/updated/fixed functionality.
- [/] I have updated/added relevant documentation in `docs` and added dartdoc comments with `///`.
- [/] I have updated/added relevant examples in `examples`.

## Breaking Change

YES, this is a breaking change.

## Related Issues

#1321

### DIFF
--- a/packages/flame_forge2d/lib/contact_callbacks.dart
+++ b/packages/flame_forge2d/lib/contact_callbacks.dart
@@ -2,25 +2,9 @@ import 'package:forge2d/forge2d.dart';
 
 import 'body_component.dart';
 
-class ContactTypes<T1, T2> {
-  // If o1 is, or inherits from, T1 or T2
-  bool has(Object o1) => o1 is T1 || o1 is T2;
-  bool hasOne(Object o1, Object o2) => has(o1) || has(o2);
-
-  // Only makes sense to call with objects that you know is in [T1, T2]
-  bool inOrder(Object o1, Object o2) => o1 is T1 && o2 is T2;
-
-  // Remember that this is not symmetric, it checks if the types in `o1` and
-  // `o2` are the same or inherits from the types in `other`
-  bool match(Object o1, Object o2) =>
-      (o1 is T1 && o2 is T2) || (o2 is T1 && o1 is T2);
-}
-
-typedef ContactCallbackFun = void Function(Object a, Object b, Contact contact);
+typedef ContactCallbackFun = void Function(ContactCallback a, ContactCallback b, Contact contact);
 
 abstract class ContactCallback<Type1, Type2> {
-  ContactTypes<Type1, Type2> types = ContactTypes<Type1, Type2>();
-
   void begin(Type1 a, Type2 b, Contact contact);
   void end(Type1 a, Type2 b, Contact contact);
   void preSolve(Type1 a, Type2 b, Contact contact, Manifold oldManifold) {}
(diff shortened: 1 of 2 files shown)


spydon:
You'll have to explain a bit more what you are doing here, why is there no need to register the callbacks?

And also you have marked this as a non-breaking change, but since you are removing functions from `Forge2DGame` it is a breaking change. (Breaking changes are okay though)

spydon:
@BlockRen it is failing the analyzer and the formatter (I can help you with those issues if you want), but more importantly I think I need the change explained to me to be able to properly review this.

abgutierrez (approved):
LGTM. 

Edit: I cannot remove my verification, apologies. Didn't see pending comments from @spydon, I agree with him.

st-pasha:
@spydon Have you seen https://github.com/marketplace/actions/dart-analyze-with-annotations? It's similar to our analyzer, but reports directly into the PR.

spydon:
> @spydon Have you seen https://github.com/marketplace/actions/dart-analyze-with-annotations? It's similar to our analyzer, but reports directly into the PR.

I have indeed, we have discussed adding it, don't remember what our conclusion was but we should definitely try it. And probably run it as a second non-mandatory step, since I guess that one can't be as nicely parallelized.

wolfenrain:
> > @spydon Have you seen https://github.com/marketplace/actions/dart-analyze-with-annotations? It's similar to our analyzer, but reports directly into the PR.
> 
> I have indeed, we have discussed adding it, don't remember what our conclusion was but we should definitely try it. And probably run it as a second non-mandatory step, since I guess that one can't be as nicely parallelized.

Conclusion was: lets try it. Or in Lukas, leettsss dooo iiiiit

spydon:
> LGTM.
> 
> Edit: I cannot remove my verification, apologies. Didn't see pending comments from @spydon, I agree with him.

@abgutierrez have you tried it?

BlockRen:
> 

More explanation added in the comment. Thank you.

BlockRen:
> @BlockRen it is failing the analyzer and the formatter (I can help you with those issues if you want), but more importantly I think I need the change explained to me to be able to properly review this.

More explanation added in the comment. Thank you.

spydon:
> > @BlockRen it is failing the analyzer and the formatter (I can help you with those issues if you want), but more importantly I think I need the change explained to me to be able to properly review this.
> 
> More explanation added in the comment. Thank you.

Thanks! Could you update the example that is using contact callbacks in `flame_forge2d`?
https://github.com/flame-engine/flame/blob/main/packages/flame_forge2d/example/lib/contact_callbacks_sample.dart

spydon:
Closing this in favor of #1547

PR #1324, opened 2022-01-17
feat: modifiable map

Schnurber:
The map should be more flexible.  For some game concepts, it could be useful to change the map at runtime.
In the proposed version, you can show and hide layers.
```dart
var layer = tiles.tileMap.map.layers[1] as TileLayer;
layer.visible = false;
```
In addition, you can change the tile data and the cache can be updated afterwards.

```dart
layer.tileData![3,2] = const Gid(0, Flips.defaults());
tiles.tileMap.update();
```
[See small example](https://github.com/meillermedia/hello_flame/tree/dev)


packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -146,9 +147,12 @@ class RenderableTiledMap {
   }
 
   /// Render [batchesByLayer] that compose this tile map.
-  void render(Canvas c) {
+void render(Canvas c) {
+    var i = 0;
     batchesByLayer.forEach((batchMap) {
```
spydon, line 152: This could use `forEachIndexed`

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -103,9 +103,10 @@ class RenderableTiledMap {
     return result;
   }
 
-  void _fillBatches() {
+ /// Uptates the cache
+  void update() {
```
spydon, line 107: This should have a more specific name to not be confused with a `Component.update` method

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -103,9 +103,10 @@ class RenderableTiledMap {
     return result;
   }
 
-  void _fillBatches() {
+ /// Uptates the cache
```
spydon, line 106: Typo

spydon (commented):
Looks good to me, just a few fixes that needs to be done.

packages/flame_tiled/lib/src/renderable_tile_map.dart
Schnurber: OK?

packages/flame_tiled/lib/src/renderable_tile_map.dart
Schnurber: OK now?

spydon (commented):
Looks good! But you haven't run `flutter format` on it.

spydon:
@lfraker could you review this small change too? :)

Schnurber:
OK!

spydon (commented):
It would be good to have a function that updates the visibilty of a layer and then calls `refreshCache`, because right now it will be a bit confusing for the user that doesn't know that `updateCache` needs to be called after a change to the map.

And some more dartdocs on the `refreshCache` function about when it should be used would be good too.

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -147,8 +147,10 @@ class RenderableTiledMap {
 
   /// Render [batchesByLayer] that compose this tile map.
   void render(Canvas c) {
-    batchesByLayer.forEach((batchMap) {
-      batchMap.forEach((_, batch) => batch.render(c));
+    batchesByLayer.forEachIndexed((i, batchMap) {
+      if (map.layers[i].visible) {
```
lfraker, line 151: Would the index be mismatched here? The batchesByLayer list is only a list of visible layers (see _renderableTileLayers), whereas map layers is all layers. Imagine you had 2 visible layers, 1 invisible layer, then 1 visible layer. BatchesByLayer[2] would point to the invisible layer.

It's a bit less efficient, but would it make sense to require a refreshCache call for every change, and just render the batchesByLayer list without any checks?

Schnurber: You are right! It's also much easier and rendering is more efficient because the `if` is not needed. I have changed it.
Visibility can now be changed as follows:
```dart
var tl = tiles.tileMap.map.layers[x] as TileLayer;
tl.visible = false;
tiles.tileMap.refreshCache();
```

spydon: Can you add a convenience function for it too?
Just something like:
```dart
void changeVisibility(int layerId, bool visibility) {}
```

Schnurber:
Well, I also added convenience functions to change the tileData and read all values: `setLayerVisibility`, `getLayerVisibility`, `setTileData`, `getTileData`

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -41,6 +41,45 @@ class RenderableTiledMap {
...
+
+  /// Gets the Gid  of the corresponding layer at the given position
+  Gid? getTileData({required int layerId, required int x, required int y}) {
+    final layer = map.layers[layerId] as TileLayer;
+    final td = layer.tileData;
+    if (td != null) {
+      return td[y][x];
+    } else {
+      return null;
+    }
```
spydon, line 80:
```suggestion
    return td?[y][x];
```
Should be possible to write like that I think?

spydon (approved):

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -41,6 +41,51 @@ class RenderableTiledMap {
...
+          td[y][x].flips.diagonally != gid.flips.diagonally) {
+        td[y][x] = gid;
+        refreshCache();
+      }
+    }
+  }
+
+  /// Gets the Gid  of the corresponding layer at the given position
+  Gid? getTileData({required int layerId, required int x, required int y}) {
+    final layer = map.layers[layerId] as TileLayer;
```
st-pasha, line 80: How do you know that this particular `Layer` is a `TileLayer`?

Schnurber: Better now?

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -41,6 +41,57 @@ class RenderableTiledMap {
...
+  /// Gets the Gid  of the corresponding layer at the given position
+  Gid? getTileData({required int layerId, required int x, required int y}) {
+    final layer = map.layers[layerId];
+    if (layer is TileLayer) {
+      final td = layer.tileData;
+      if (td != null) {
+        return td[y][x];
+      } else {
+        return td?[y][x];
+      }
```
spydon, line 89:
```suggestion
      return layer.tileData?[y][x];
```

packages/flame_tiled/lib/src/renderable_tile_map.dart
spydon: I think you misunderstood what I meant, sent a new suggested change. :)

packages/flame_tiled/lib/src/renderable_tile_map.dart
Schnurber: Here, `layer` ist not of type `TileLayer`, so it has no member `tileData`.

packages/flame_tiled/lib/src/renderable_tile_map.dart
Schnurber: Ah, OK, next time I simply click "Commit suggestion"! :-))

packages/flame_tiled/lib/src/renderable_tile_map.dart
spydon: Yes it is? You do the check in the `if` outside of this change.

st-pasha (commented):
It looks to me that all 4 new methods operate on a layer, selected by its layerId. In this case it is better to have one method in `RenderableTileMap` to retrieve a layer with the given id, and then all other methods would belong to that layer class. Thus, instead of `.setLayerVisibility(layerId, value)`, we'd have `.layer(id).visibility = value;`.

Schnurber:
> It looks to me that all 4 new methods operate on a layer, selected by its layerId. In this case it is better to have one method in `RenderableTileMap` to retrieve a layer with the given id, and then all other methods would belong to that layer class. Thus, instead of `.setLayerVisibility(layerId, value)`, we'd have `.layer(id).visibility = value;`.

`Layer` is from the tiled package and has no reference to `Flame`. Thus, the cache cannot be refreshed from there.

st-pasha:
Unless you wrap it into your own `Layer` object (similarly how `RenderedTiledMap` wraps `TiledMap`).

spydon:
> It looks to me that all 4 new methods operate on a layer, selected by its layerId. In this case it is better to have one method in `RenderableTileMap` to retrieve a layer with the given id, and then all other methods would belong to that layer class. Thus, instead of `.setLayerVisibility(layerId, value)`, we'd have `.layer(id).visibility = value;`.

How do you propose this should be done without creating new objects, an extension? We also need a reference to the `RenderableTileMap` in there so that the cache can be updated after a change. I agree that they all could be grouped as a layer, but I think the current solution works too.

Schnurber:
@st-pasha I agree with @spydon That would be difficult. I would prefer a simple solution.
The situation is as follows, 

In package tiled there is:
```dart
abstract class Layer {
	...
	bool visible;
	...
}

class TileLayer extends Layer {...}

class TiledMap {
...
	List<Tileset> tilesets;
	List<Layer> layers;
...
}

```
and in flame:
```dart
class RenderableTiledMap {
...
	final TiledMap map;

	void refreshCache() {...}

	void setLayerVisibility(int layerId, bool visibility) {
		...
		refreshCache();
		... 
	}
...
}
```
Unfortunately, it is not possible to work with extension methods, as the references to `RenderableTileMap` are not possible this way. 
You would have to create subclasses `RenderableLayer` and `RenderableTilset` in flame, change `RenderableTiledMap` and define it as a subclass of `TiledMap`. It would then be complicated because you would have to convert the nested list entries ` List<Tileset>` and `List<Layer> layers` into lists filled with subclasses of type `RenderableTilset` and `RenderableLayer`. You would actually have to generate everything new to provide it with references to the `RenderableTileMap`.

spydon:
@Schnurber did you see my response on the suggestion with the null check?

packages/flame_tiled/lib/src/renderable_tile_map.dart
Schnurber:
```dart

final layer = map.layers[layerId];
if (layer is TileLayer) {
  final td = layer.tileData;
  if (td != null) {
    return td[y][x];
  } else {
    return td?[y][x];
  }
  return layer.tileData?[y][x]; // Dead Code Here
}  
return layer.tileData?[y][x]; // layer not of type TileLayer here  
```

spydon: That line is something you accidentally added when you didn't commit my full suggestion the first time around. ;)
```dart
final layer = map.layers[layerId];
if (layer is TileLayer) {
  return layer.tileData?[y][x];
}  
return null;
```

Schnurber: Great! Should I change this or should I press the botton above?

spydon: The change above will be wrong still since it has the line that you mentioned, so probably easiest to just add those lines manually.

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -41,6 +40,57 @@ class RenderableTiledMap {
...
+    } else {
+      return null;
+    }
```
spydon, line 91:
```suggestion
   final layer = map.layers[layerId];
   if (layer is TileLayer) {
     return layer.tileData?[y][x];
   }  
   return null;
```

packages/flame_tiled/lib/src/renderable_tile_map.dart
spydon: Added a new suggestion here that you can just directly press commit on if you want:
https://github.com/flame-engine/flame/pull/1324/files#r787768163

packages/flame_tiled/pubspec.yaml
```diff
@@ -8,7 +8,8 @@ environment:
   flutter: ">=2.5.0"
 
 dependencies:
-  flame: ^1.0.0
+  flame:
+    path: ../flame/
```
spydon, line 12:
```suggestion
  flame: ^1.0.0
```

spydon: This should not be on path, melos is handling all of those things now, check here: :slightly_smiling_face: 
https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md#environment-setup

erickzanardo (approved):
LGTM

PR #1329, opened 2022-01-21
refactor: Loadable no longer declares onGameResize

st-pasha:
# Description

Mixin `Loadable` no longer declares method `onGameResize()`. Instead, this method is now defined by classes `Component` and `Game` separately (which they were already doing before).

Resizing has nothing to do with loading, so it is not proper to declare this method inside `Loadable`. I presume it existed there by accident. 

## Checklist

## Breaking Change

### DIFF
--- a/packages/flame/lib/src/game/flame_game.dart
+++ b/packages/flame/lib/src/game/flame_game.dart
@@ -106,7 +106,11 @@ class FlameGame extends Component with Game {
   @mustCallSuper
   void onGameResize(Vector2 canvasSize) {
     camera.handleResize(canvasSize);
-    super.onGameResize(canvasSize);
+    super.onGameResize(canvasSize); // Game.onGameResize
+    // [onGameResize] is declared both in [Component] and in [Game]. Since
+    // there is no way to explicitly call the [Component]'s implementation,
+    // we propagate the event to [FlameGame]'s children manually.
+    children.forEach((child) => child.onGameResize(canvasSize));
   }
 
   /// Whether a point is within the boundaries of the visible part of the game.
--- a/examples/lib/stories/system/without_flamegame_example.dart
+++ b/examples/lib/stories/system/without_flamegame_example.dart
@@ -8,7 +8,7 @@ class NoFlameGameExample with Loadable, Game, KeyboardEvents {
   static const String description = '''
     This example showcases how to create a game without the FlameGame.
     It also briefly showcases how to act on keyboard events.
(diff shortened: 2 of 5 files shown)


packages/flame/lib/src/game/flame_game.dart
```diff
@@ -106,7 +106,11 @@ class FlameGame extends Component with Game {
   @mustCallSuper
   void onGameResize(Vector2 canvasSize) {
     camera.handleResize(canvasSize);
-    super.onGameResize(canvasSize);
+    super.onGameResize(canvasSize); // Game.onGameResize
+    // [onGameResize] is declared both in [Component] and in [Game]. Since
+    // there is no way to explicitly call the [Component]'s implementation,
+    // we propagate the event to [FlameGame]'s children manually.
+    children.forEach((child) => child.onGameResize(canvasSize));
```
spydon, line 113: It feels like this makes it worse since this code duplication wasn't needed before.
It is resized when loaded, therefore one could argue that it can live in `Loadable`. A bit confusing maybe, but I think still better than code duplication.

st-pasha: One line of code is barely "code duplication", and it's only needed because of the unfortunate fact that FlameGame attempts to inherit both from Game and from Component. If you're worried about lines of code, you'll notice that this PR removes 6 lines of code while adding only 1.

> It is resized when loaded, therefore one could argue that it can live in Loadable.

It is resized after being loaded too, so one could argue it must not live in Loadable. Regardless, resizing is an implementation detail. We've discussed elsewhere about changing or removing these events completely. While we have not decided whether we want to actually do that, the fact that it is perfectly possible demonstrates that resizing has nothing to do with loading.

st-pasha:
The important benefit that this brings is it makes the code understandable.
Before, we had `Component.onGameResize()` calling `super.onGameResize`, which was empty -- but removing that super call would break the engine. This kind of shady code execution pattern that an IDE cannot follow, should be avoided.

packages/flame/lib/src/game/flame_game.dart
spydon: I guess it is fine since you explain it very well in the comment above, and that line should very seldom change.

spydon (approved):
This is one of the key points where OOP fails I think, that you want multiple inheritance, but you can't have pure multiple inheritance so it becomes this non-linear pattern which can be hard to follow. I think you are right, this does improve the code, approved!

PR #1333, opened 2022-01-22
refactor: Remove Loadable, optional onLoads

st-pasha:
# Description

1. `Game` no longer depends on `Loadable`, that is, when creating a new `Game` subclass you only need to say `class MyGame with Game` instead of `class MyGame with Loadable, Game`. This is done by moving the 2 lines of code that are inside `Loadable` into the `Game` class. The benefit here is that it simplifies the usage of the `Game` class.

2. `Component` no longer depends on `Loadable`. This brings in several advantages:
   - Conceptually, `Component` is the most "base" class in the Flame engine. Having it being derived from some other helper class is not ideal.
   - The `onLoad` method is now defined within the `Component` class, **alongside all other lifecycle methods**, which makes it easier to document.
   - The `onLoadCache` variable is removed -- the fact that the component is loaded only once is guarded by the `isLoaded` flag, so carrying around that `Future` was unnecessary and was only increasing the size of the `Component` object.

3. `Loadable` mixin is now empty and declared `Deprecated`, to be removed in 1.1.0. Anyone who uses a custom `Game` class via `with Loadable, Game` can simply remove `Loadable` from the declaration.

4. The default `onLoad` methods are no longer declared as `@mustCallSuper`, since they have no functionality. This simplifies user code: there is no longer need to say `await super.onLoad()` at the start of your `onLoad()` implementation. A derived class can always re-add this annotation if it overrides `onLoad` and if it intends to be derived further.

## Checklist

## Breaking Change

- [x] No, this is not a breaking change.


wolfenrain:
We could remove the breaking change by having an empty `Loadable` mixin right? And then just deprecated it in the next breaking release?

st-pasha:
> We could remove the breaking change by having an empty Loadable mixin right? And then just deprecated it for the next breaking release?

Yeah, makes sense, let's do that.

spydon (approved):
Lgtm! Don't remember why I created that mixin from the beginning, but this is clearly nicer.

packages/flame/lib/src/game/mixins/loadable.dart
```diff
@@ -1,36 +1,2 @@
...
-  Future<void>? onLoad() => null;
-
-  /// Since [onLoad] only should run once throughout a the lifetime of the
-  /// implementing class, it is cached so that it can be reused when the parent
-  /// component/game/widget changes.
-  @internal
-  @nonVirtual
-  late Future<void>? onLoadCache = onLoad();
-}
+@Deprecated('This mixin will be removed in 1.1.0')
```
erickzanardo, line 1: I think 1.1.0 will be our next version, so this should remain on that version and the removal can happen on the next version, otherwise people will not have in between versions to update their code.

```suggestion
@Deprecated('This mixin will be removed after 1.1.0')
```

wolfenrain (approved):

erickzanardo (approved):

PR #1337, opened 2022-01-24
feat: Components are now always added in the correct order

### LINKED ISSUE #1222: Components may be added into game tree out of order
## Current bug behavior
As noted in a [Discord discussion](https://discord.com/channels/509714518008528896/516639688581316629/920074132488609822), when multiple components are added into a game, they may end up being in a different order from how the user intended. The primary cause for this is that each component is only truly added into the game tree once its `onLoad` has finished its job, which means a component with longer-running onLoad may end up being pushed "to the end of the line".

## Expected behavior
The components should appear exactly in the same order as they were added by the user.

### LINKED ISSUE #1293: FlameGame.onGameResize - Zero value when game is closed and restarted

### LINKED ISSUE #1305: Android render measuring problem
# Current bug behaviour

When i run my game for the first time or in the release build, some images are rendered in wrong size. 
This happens because of system animation rotating the screen from the portrait mode to landscape when the app is open. onGameResize called several times and it seems that components do not handle this resize. So at the first resize, components are measured for the portrait mode, but when the screen rotated, the game do not change it's measuring. If i set some delay before launching the game widget, everything is ok.

# Steps to reproduce
Run the app for the first time or in the release mode. [Here is a little working example](https://github.com/mancj/flame_sizing_mwe)

# Log output
```dart
game resized to [360.0,785.0] // onGameResize of the flame game
parallax component resized: [360.0,785.0] // onGameResize of the parallax component
parallax onLoad size: [360.0,785.0] // onLoad of the parallax component
game resized to [773.3333333333334,345.0] // onGameResize of the parallax component
// as we see, the component didn't get the last onGameResized, so maybe this is the reason
```

# Flutter doctor output

```
[✓] Flutter (Channel stable, 2.8.1, on macOS 12.1 21C52 darwin-x64, locale ru)
    • Flutter version 2.8.1 at /Users/mancj/fvm/versions/2.8.1
    • Framework revision 77d935af4d (3 weeks ago), 2021-12-16 08:37:33 -0800
    • Engine revision 890a5fca2e
    • Dart version 2.15.1

[!] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
    • Android SDK at /Users/mancj/Library/Android/sdk

[✓] Xcode - develop for iOS and macOS (Xcode 13.1)
    • CocoaPods version 1.11.2

    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7281165)

[✓] IntelliJ IDEA Community Edition (version 2021.2.4)
    • IntelliJ at /Applications/IntelliJ IDEA CE.app
    • Flutter plugin version 63.3.3
    • Dart plugin version 212.5632

[✓] VS Code (version 1.63.2)
    • Flutter extension version 3.29.0

    • SM G991B (mobile) • R5CRA1J2VWH • android-arm64  • Android 12 (API 31)
    • macOS (desktop)   • macos       • darwin-x64     • macOS 12.1 21C52 darwin-x64
    • Chrome (web)      • chrome      • web-javascript • Google Chrome 97.0.4692.71

```

* Flame version: 1.0.0
* Platform affected: android
* Platform version affected: android 12

# More information
[Sample code to reproduce the bug](https://github.com/mancj/flame_sizing_mwe)
[Discord discussion of the bug with videos and some production code](https://discord.com/channels/509714518008528896/929095243238895626)

### LINKED ISSUE #1335: Improving Flame testing by allowing users to wait for game to load
# What could be improved
Right now, when unit testing a flame game, it is very hard to ensure that the game is loaded before triggering events, making assertions, this makes testing quite challenging sometimes.

We have however an easy way to provide this for users. the FlameGame class already have an attribute called `onLoadCache`, which holds a future of the loading of the game, by awaiting that future, we can be sure that the game was completely loaded.

The problem is that this attribute is marked with `@internal`.

My suggestion is to keep that attribute internal, but offer a public getter for it, so people can write their tests like this example:

```dart
          tester.widgetTest(
            'pick up the item when it is close enough',
            (game, tester) async {
              await tester.pump();

              await game.ensureAdd(
                Pickupable(item: GameItem.sword)
                  ..position = Vector2(
                    0,
                    40,
                  ),
              );

              await game.gameLoading();
              await tester.pump();

              final event = MockRawKeyDownEvent();
              when(() => event.logicalKey).thenReturn(LogicalKeyboardKey.space);

              game.player.onKeyEvent(event, {});

              verify(
                () => inventoryBloc.add(
                  const GameItemPickedUp(GameItem.sword),
                ),
              ).called(1);
            },
          )
```

Maybe this is something that cold be handled on #1333 
 
# Why should this be improved
To make testing easier

# Any risks?
None come to mind

# More information

### LINKED ISSUE #1338: Forge2DGame + PositionComponents inconsistent onGameResize()
I'm trying to add simple PositionComponent to the Forge2DGame and respond to the onGameResize calls.

# Current bug behaviour

```dart 
import 'package:flame_forge2d/forge2d_game.dart';

class TestComponent extends PositionComponent {
  @override
  void onGameResize(Vector2 canvasSize) {
    print("child::gameSize: $canvasSize");
    super.onGameResize(canvasSize);
  }
}

class MyGame extends Forge2DGame {
  MyGame()
      : super(
          gravity: Vector2(0, -1.0),
          zoom: 10.0,
        );

  @override
  Future<void>? onLoad() {
    add(TestComponent());
    return super.onLoad();
  }

  @override
  void onGameResize(Vector2 canvasSize) {
    print("game::gameSize: $canvasSize");
    super.onGameResize(canvasSize);
  }
}

main() {
  final myGame = MyGame();
  runApp(
    GameWidget(
      game: myGame,
    ),
  );
}
```
If I build and run the code above for the macosx target and try to resize game window the  code produces this debug output: 
```
flutter: game::gameSize: [1937.0,1089.0]
flutter: child::gameSize: [193.70000000000002,108.9]
flutter: game::gameSize: [1929.0,1082.0]
flutter: child::gameSize: [1929.0,1082.0]
```

# Expected behaviour

As you can see the first onGameResize() call passes scaled (zoomed) size, but all following updates uses real window size.  It's expected that output be consistent either scaled or not.

# Steps to reproduce

build and run the code above for the macosx target and try to resize game window 

# Flutter doctor output

```
Output of: flutter doctor -v
[✓] Flutter (Channel stable, 2.8.1, on macOS 12.1 21C52 darwin-x64, locale en-ES)
    • Flutter version 2.8.1 at /Volumes/Home/dmp/flutter
    • Framework revision 77d935af4d (6 weeks ago), 2021-12-16 08:37:33 -0800
    • Engine revision 890a5fca2e
    • Dart version 2.15.1

[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
    • Android SDK at /Volumes/Home/dmp/Library/Android/sdk/
    • Platform android-31, build-tools 31.0.0
    • ANDROID_HOME = /Volumes/Home/dmp/Library/Android/sdk/
    • ANDROID_SDK_ROOT = /Volumes/Home/dmp/Library/Android/sdk/
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7281165)

[✓] Xcode - develop for iOS and macOS (Xcode 13.2.1)
    • CocoaPods version 1.11.2

    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7281165)

[✓] VS Code (version 1.63.2)
    • Flutter extension version 3.32.0

[✓] Connected device (7 available)
```

* Flame version: 1.0.0
* Platform affected: desktop, probably more
# Log information

```
```

# More information

### LINKED ISSUE #1342: Remove prepareComponent
> The plan was to remove `prepareComponent` completely after v1 actually, we never made a ticket for it but we have it on our internal planning board. 

_Originally posted by @spydon in https://github.com/flame-engine/flame/issues/1340#issuecomment-1021046421_

See also discussions in #992

st-pasha:
# Description

- #1222: Main change is that the system now ensures that when components are added into the game tree, they will be always added in the order that the user requested. For example, if the user says `game..add(c1)..add(c2)`, then components `c1` and `c2` will appear in that order in game's children list, regardless of how long it took for them to load.

- New mixin `SingleGameInstance` allows `onLoad()` to be always called when the component is added to a parent, even if the parent is not mounted yet.

- #1335: Added `game.ready()` function, which can be awaited if you need to resolve all pending changes to the game tree (mostly useful for testing).

- #1342: Lifecycle methods `prepare` and `prepareComponent` were removed. Existing logic in `prepare` can be merged with `onMount`, and functionality of `prepareComponent` can always be replicated within the `Component` itself.

- Some of the functionality from `ComponentSet` was moved out into the `_LifecycleManager` class. This has several advantages: first, the `Component` is no longer forced to carry temporary lifecycle queues forever, which reduces overall memory consumption; second, implementing alternative `ComponentSet`s now becomes easier, as it is no longer required to cater to the `Component`s lifecycle peculiarities.

- From testing standpoint, the main difference is that now at least _some_ game instance is required in each test which involves more than one component. The idiom `await game.ready()` can now be used to ensure that all components were fully mounted. In most cases it could replace the `game.update(0)` call, which is less reliable.

## Checklist

## Breaking Change

- [x] No, this is not a breaking change.
  - Except if you were using `prepare()` / `prepareComponent()`, which we believe to be extremely rare. But if you did, then the recommendation is to merge that functionality into the `onMount` callback.

## Related Issues

Closes #1222 
Closes #1293
Closes #1305
Closes #1335
Closes #1338
Closes #1342


spydon:
Since we are already having an ordering function in the `OrderedSet`, can't we make use of that one?
We would only have to keep track of an integer for which order the component was added in and within the same priority layer that order index would decide which order it would be in in the `ComponentSet`.

So the comparator would be something like this:
```dart
Comparing.join<Component>([
  Comparing.on<Component>((c) => c.priority),
  Comparing.on<Component>((c) => c.addIndex),
]);
```
(`addIndex` would probably have a different name, but you probably understand the gist at least.)

This could also be made a non-breaking change in terms of API.

Another small thing is that we probably want to keep the `Future` on `add`, in almost all cases it will work to `await` that one, the only case I can come up with is when you add a component to a component that doesn't have a root yet, but since subcomponents mostly are added in either `onLoad` or as reactions to input that `await` almost always work. Or do you have another common usecase in mind? If we don't remove it we do need better documentation for the case where it won't work however. Or if we can come up with another way of waiting for specific components to be loaded.

st-pasha:
> Since we are already having an ordering function in the OrderedSet, can't we make use of that one?

Can you please elaborate what benefit do you see for this approach? What it does is that it adds component unordered, but keeps an extra variable `addIndex` around that helps with putting all components back in order. This PR, on the other hand, makes sure that everything is put in order in the first place.

The approach that you suggest also leads to a significant increase in memory size of the component tree. This is due to the fact of how `OrderedSet` is implemented: it puts all components with same priorities in a single list, and components with different priorities into different lists. Thus, in the common case when there are `n` components with the default priority of 0, the OrderedSet will put them into a single list of length `n` in a tree of size 1. However, with the introduction of `addIndex`, the priorities will now always be different, which means `n` lists of size 1 living in a tree of size `n`.

I'm not even sure how solid the statement "we are already having an ordering function in the OrderedSet" is. I mean, surely, we do, but eventually we want to support different implementations for `ComponentSet`s. For example, for some simple leaf-level components like sprites you might have a simpler ComponentSet based on a list. For a large-size game you may want to have a ComponentSet based on a quadtree or a similar structure. For an isometric game you could have a ComponentSet that automatically sorts the components according to their distance from the camera. And so on.

Another advantage this PR brings is that it reduces the size of the `Component`: it no longer needs to carry the `Set _addLater`, instead these queues are handled at the Game level.


> Another small thing is that we probably want to keep the Future on add, in almost all cases it will work to await that one

Sure, we can. After all, it's already there: `onLoadFuture`, and it's just the question of whether we want to return it or not. However, I find that future very confusing. Without consulting documentation, it's easy to assume looking at the function signature `Future<void> add(Component c)` that when you wait on that future you wait for `c` to be added. It's not even true that awaiting on that future will guarantee that `c` is loaded, after all `c`'s loading can be delayed if it cannot be prepared.

So, having a future there that sometimes works and sometimes doesn't, just doesn't sound like a solid API to me. Though I do agree with you that scenarios where that future is meaningful are quite common. Still, they aren't _overwhelmingly_ common.

On the other hand, what if we removed the restriction that the parent needs to be mounted before we can proceed with loading? That would remove the "add" queue altogether and make the future far better defined.

spydon:
> Can you please elaborate what benefit do you see for this approach? What it does is that it adds component unordered, but keeps an extra variable `addIndex` around that helps with putting all components back in order. This PR, on the other hand, makes sure that everything is put in order in the first place.

It makes the code here a lot simpler, there is no extra cost to add them in their correct place in the OrderedSet even if they aren't added in the add-order.

> The approach that you suggest also leads to a significant increase in memory size of the component tree. This is due to the fact of how `OrderedSet` is implemented: it puts all components with same priorities in a single list, and components with different priorities into different lists. Thus, in the common case when there are `n` components with the default priority of 0, the OrderedSet will put them into a single list of length `n` in a tree of size 1. However, with the introduction of `addIndex`, the priorities will now always be different, which means `n` lists of size 1 living in a tree of size `n`.

That's true, that's a good argument.

> Another advantage this PR brings is that it reduces the size of the `Component`: it no longer needs to carry the `Set _addLater`, instead these queues are handled at the Game level.

It does remove `_addLater`, but the queue solution comes with quite some complexity that wasn't needed here before.

> On the other hand, what if we removed the restriction that the parent needs to be mounted before we can proceed with loading?

Many games depend on the game `size` and the game caches in their `onLoad` methods, and we'd need to be mounted to have access to those. But it is an intriguing idea.

> That would remove the "add" queue altogether and make the future far better defined.

How could it remove the queue though, since there would still be different loading times between components?

st-pasha:
> Many games depend on the game size and the game caches in their onLoad methods, and we'd need to be mounted to have access to those. But it is an intriguing idea.

What if we handle this at Game's lifecycle? That is, first `GameWidget` waits for the size to be available (and non-zero), and only _then_ it launches `Game.onLoad`. This way any component that is added within `onLoad` is guaranteed to be able to know the size of the canvas. If a component needs to have some additional game resources, like images, it can await those inside the `onLoad`, exactly as it is done right now. 

> It does remove _addLater, but the queue solution comes with quite some complexity that wasn't needed here before.

I agree that it does look somewhat more complex, but I wonder if maybe that's just a matter of perception? Because what I did is I gathered code that was spread over `prepare` and `prepareComponent` and put it all into `add()`, then added several asserts -- so now the code looks much bigger, but it's actually doing all the same things. If anything, it's easier to understand now that it's all in one place.

As for the queue, it replaces the `_addLater` set. So, likewise, there was some code in the `ComponentSet` that was handling that set, and now that code transformed into the queue-processing code. It's probably a bit more code, but only because we're taking care of removing queues as soon as they empty up, which reduces memory footprint.

So, what I am saying is that the overall complexity remains approximately the same, only now it is concentrated in a single place, which makes it appear like it got more complex.

> How could it remove the queue though, since there would still be different loading times between components?

We could remove the "add" queue only, the "children" queue will remain.

spydon:
> What if we handle this at Game's lifecycle? That is, first `GameWidget` waits for the size to be available (and non-zero), and only _then_ it launches `Game.onLoad`. 

I wonder how much overhead that will add to the loading time, if it wouldn't be much that sounds like a good plan to make the lifecycle much clearer.

> I agree that it does look somewhat more complex, but I wonder if maybe that's just a matter of perception?

Possibly, I've had a look at it a few times though, but I'll go through it again tomorrow.
 
> As for the queue, it replaces the `_addLater` set. So, likewise, there was some code in the `ComponentSet` that was handling that set, and now that code transformed into the queue-processing code. It's probably a bit more code, but only because we're taking care of removing queues as soon as they empty up, which reduces memory footprint.

> So, what I am saying is that the overall complexity remains approximately the same, only now it is concentrated in a single place, which makes it appear like it got more complex.

It feels a bit like this might belong in the `ComponentSet` though, to get some separation of concerns, or that was the point from the beginning at least if I remember correctly (ping @luanpotter, I think you created that?).

> We could remove the "add" queue only, the "children" queue will remain.

That makes sense. 

One change that will come with this is that fast loading components would be "stuck" behind slower ones, even though they are already loaded, right? It might be good, or maybe even necessary, to have an optional way to add slow loading components that you don't need from the start to not congest the queue. Not sure if there are any usecases that need that, but we should give it a thought at least.

st-pasha:
> I wonder how much overhead that will add to the loading time, if it wouldn't be much that sounds like a good plan to make the lifecycle much clearer.

I think it'll be in milliseconds. After all, the standard Flutter pipeline builds an entire widget tree, then creates rendering tree, then lays them out, and then renders -- all within the span of 10ms. Here we'd be waiting for the "lay out" stage, so it'll probably be just a few milliseconds before we can begin Game's onLoad.

> It feels a bit like this might belong in the ComponentSet though, to get some separation of concerns

There's still some degree of separation: queue-handling logic is inside of the `ComponentTreeRoot` mixin. I look at this more from the code re-use standpoint: when we implement a new `ComponentSet` class, would we need to duplicate this logic? The approach taken in this PR allows maximum flexibility: one can build alternative `ComponentSet`s, or even alternative `Game`s that are use components but not derived from the standard `FlameGame`.

> It might be good, or maybe even necessary, to have an optional way to add slow loading components that you don't need from the start to not congest the queue. 

This is an interesting challenge. I suppose we could have an additional boolean flag for `add()` which would delay putting the component into the `childrenQueue` until after it was loaded/mounted. This way it won't appear in any queue until it has been fully loaded, which means the other components will not be hindered.

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -316,7 +316,11 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
             color: game.backgroundColor(),
             child: LayoutBuilder(
               builder: (_, BoxConstraints constraints) {
-                game.onGameResize(constraints.biggest.toVector2());
+                final size = constraints.biggest.toVector2();
+                if (size.isZero()) {
+                  return Container();
```
spydon, line 321: I guess this will quickly flicker a black screen? Could we somehow make the `loadingBuilder` widget run until this is non-zero?

packages/flame/test/components/collidable_type_test.dart
```diff
@@ -8,8 +8,9 @@ class _HasCollidablesGame extends FlameGame with HasCollidables {}
 
 class _TestBlock extends PositionComponent with HasHitboxes, Collidable {
   final List<Collidable> collisions = List.empty(growable: true);
+  String? name;
 
-  _TestBlock(Vector2 position, Vector2 size, CollidableType type)
+  _TestBlock(Vector2 position, Vector2 size, CollidableType type, {this.name})
```
spydon, line 13: Why was this changed in this PR?

packages/flame_test/lib/src/flame_test.dart
```diff
@@ -17,15 +17,15 @@ extension FlameGameExtension on Component {
   /// Makes sure that the [component] is added to the tree if you wait for the
   /// returned future to resolve.
   Future<void> ensureAdd(Component component) async {
-    await add(component);
-    updateTree(0);
+    add(component);
+    await Component.root!.ready();
```
spydon, line 21: Shouldn't it be enough to just await this `add` now if a future is returned from `add`?

spydon (commented):
This starts to look really good now I think!

packages/flame/lib/src/components/component.dart
```diff
@@ -384,43 +399,118 @@ class Component {
...
+/// lifecycle management. These include: lifecycle event queues, and the canvas
+/// size for [Component.onGameResize].
+///
+/// In order to use this mixin in a custom [Game], do the following:
+///   - add this mixin to your game class;
+///   - set it as the default [Component.root];
+///   - call [processComponentQueues] on every game tick.
+mixin ComponentTreeRoot on Game {
+  final Map<Component, Queue<Component>> childrenQueue = {};
+  final Map<Component, Queue<Component>> addQueue = {};
```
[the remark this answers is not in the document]
st-pasha: Indeed! I was planning to add them, but then forgot :)

packages/flame/lib/src/components/component.dart
```diff
@@ -251,67 +267,78 @@ class Component {
...
+      'components from the constructor? Use the onLoad() method instead.',
+    );
+    if (!isMounted) {
+      (root!.addQueue[this] ??= Queue()).addLast(component);
+      return;
+    }
+
+    component._parent = this;
+    component.debugMode |= debugMode;
+    component.onGameResize(root!.canvasSize);
```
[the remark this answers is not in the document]
st-pasha: To me, canvasSize is just something more stable and well-defined. Consider this:
  - for top-down or side-scroller games, the _game size_ simply refers to the size of the world visible through the viewport (if the game has parallax, then the `z=0` portion of the world). But, what do you do in an isometric game?
  - what about top-down games where the camera can rotate (e.g. racing games)? If you define "game size" as AABB, then it's going to change on every game tick -- do we really want a resize event on every game tick?
  - game size can even change through internal mechanisms, such as changing zoom level. Do we want game resize events whenever camera zoom changes?  

Also, these resize events are mostly useful to components that are independent from the camera, such as HUD elements. They would need raw canvas size (or viewport size), not the game size.

packages/flame/test/components/collidable_type_test.dart
st-pasha: For testing purposes: I was getting error messages that `Instance of _TestBlock` was not equal to an `Instance of _TestBlock`, and that just wasn't very helpful. This doesn't affect the nature of the test, only what message it emits when it fails.

packages/flame_test/lib/src/flame_test.dart
st-pasha: Not really: the future from `add` can only take you as far as to the end of the `onLoad` stage. But if you need to make sure that the component is actually mounted into the tree, `add()` cannot provide that.

packages/flame/lib/src/game/game_widget/game_widget.dart
st-pasha: Oh yeah, actually we can, i'll do that.

packages/flame/lib/src/components/component.dart
```diff
@@ -384,43 +399,118 @@ class Component {
...
+///
+/// The purpose of this mixin is to collect common facilities for component
+/// lifecycle management. These include: lifecycle event queues, and the canvas
+/// size for [Component.onGameResize].
+///
+/// In order to use this mixin in a custom [Game], do the following:
+///   - add this mixin to your game class;
+///   - set it as the default [Component.root];
+///   - call [processComponentQueues] on every game tick.
+mixin ComponentTreeRoot on Game {
```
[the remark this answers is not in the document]
st-pasha: Usually I'm all in favor of one-class-per-file policy; but in this case I feel that it strongly depends on the `Component` class, so it's better to have them side-by-side. And by "depends" I mean that the mixin supports `Component`'s lifecycle, so the code in this mixin can best be understood when consider together with the `Component` class.

packages/flame/lib/src/components/component.dart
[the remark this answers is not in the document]
spydon: Since it was the `canvasSize` before (except in one accidental place) let's keep it like that and then potentially change it at a later stage.

packages/flame/lib/src/components/component.dart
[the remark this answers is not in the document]
spydon: Does it really make it easier to see that just because they are in the same file? I think even if you'd want to look on the component lifecycle at the same time as you look at this you'd need two editor tabs/windows, same as if they were in two different files?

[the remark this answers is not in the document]
st-pasha: I guess it's kinda subjective. To me, it feels nice to have all lifecycle-related code in one place, where "one place" means "in same file", since it can't be in a single class. However, I don't feel that strongly about this. 

WDYT, can you explain your thought process behind recommending this to be in a separate file?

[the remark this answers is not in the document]
spydon: My thought process is simply that if I'm looking for a mixin for the game I'll look in the directory for game-mixins. I don't feel very strongly about it either, but I'd prefer a separate file for sure.

doc/components.md
```diff
@@ -34,17 +34,16 @@ it is only run once even if the component is removed both by using the parents r
...
-`late` variables here, since this method might run several times throughout the component's
-lifetime. This method is executed after the initial "preparation" of the component is done and after
-`onGameResize` and `onLoad`, but before the inclusion of the component in the parent's list of
-components.
+like loading an image for example. This method is executed after `onGameResize`, but before the
+`onMount`. This method is guaranteed to execute only once during the lifetime of the component, so
+you can think of it as an "asynchronous constructor". 
+
+The `onMount` method runs every time when the component is mounted into a game tree. This means that 
+you should not initialize `late` variables here, since this method might run several times
```
spydon, line 42: This is only valid for `late final` right? You can re-assign a `late` variable.

packages/flame/lib/src/components/component.dart
```diff
@@ -255,63 +265,80 @@ class Component {
...
+  /// This method returns a future that completes when the component is done
+  /// loading, and mounting if the parent is currently mounted. However, this
+  /// future will not guarantee that the component will become "fully mounted":
+  /// it still needs to be added to the parent's children list, and that
+  /// operation will only be done on the next game tick.
+  ///
+  /// A component can only be added to one parent at a time. It is an error to
+  /// try to add it to multiple parents, or even to the same parent multiple
+  /// times. If you need to change the parent of a component, use [changeParent]
+  /// method.
```
spydon, line 307:
```suggestion
  /// times. If you need to change the parent of a component, use the
  /// [changeParent] method.
```

packages/flame/lib/src/components/component.dart
```diff
@@ -255,63 +265,80 @@ class Component {
...
 
-  /// The children are added again to the component set so that [prepare],
-  /// [onLoad] and [onMount] runs again. Used when a parent is changed
-  /// further up the tree.
-  Future<void> _reAddChildren() async {
-    if (_children != null) {
-      await Future.wait(_children!.map(add));
-      await Future.wait(_children!.addLater.map(add));
+  /// Adds this component to the provided [parent] (see [add] for details).
+  Future<void> addToParent(Component parent) async {
```
spydon, line 316: Maybe this should be private? So that we try to get people to use one way of adding children, since both methods require both the child and the parent anyways?

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -316,13 +316,23 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
...
                   builder: (_, snapshot) {
                     if (snapshot.hasError) {
                       final errorBuilder = widget.errorBuilder;
                       if (errorBuilder == null) {
+                        // @Since('2.16')
+                        // throw Error.throwWithStackTrace(
+                        //   snapshot.error!,
+                        //   snapshot.stackTrace,
+                        // )
+                        print(snapshot.stackTrace);
```
spydon, line 335: I guess this was accidentally left?

packages/flame/lib/src/game/mixins/component_tree_root.dart
```diff
@@ -0,0 +1,153 @@
...
+  /// finish loading in arbitrary order.
+  final Map<Component, Queue<Component>> childrenQueue = {};
+
+  /// Global queue of components that need to be mounted.
+  ///
+  /// Occasionally, a component may be added to a parent that hasn't been
+  /// mounted yet. In such a case, the child component must wait until its
+  /// parent mounts before it can mount too. Such components are placed into
+  /// this queue.
+  final Queue<Component> mountQueue = Queue();
```
spydon, line 46: When a parent is mounted, couldn't that just propagate to all it's children that their `onMount` methods should run?

spydon (approved):
Good job! Just some minor comments

packages/flame/lib/src/components/component.dart
st-pasha: My thinking was that since we have both `remove` and `removeFromParent`, there should also be `add` and `addToParent` -- for symmetry.

packages/flame/lib/src/game/game_widget/game_widget.dart
st-pasha: No, this was deliberate. Right now if an error occurs during loading, the stacktrace is lost. With `print`, it would at least be in the console. The best solution is to re-throw with stacktrace, but that function is only available in Dart 2.16. So I left the comment here to remember to improve this part once we upgrade to a newer Dart version.

packages/flame/lib/src/components/component.dart
spydon: Hmm, I see your point, but the `removeFromParent` is actually making thinga simpler in some cases since you don't need to know the parent, this one just flips around the argument and the instance it called on. Let's see what the others say.

packages/flame/lib/src/game/mixins/component_tree_root.dart
st-pasha: I'll need to think about that.

packages/flame/lib/src/components/component.dart
st-pasha: Well, this function can be useful too. For example, you can write
```dart
SomeComponent()
  ..position = Vector2(x, y)
  ..scale = Vector2.all(1.5)
  ..addToParent(level);
```

packages/flame/lib/src/game/mixins/component_tree_root.dart
st-pasha: Yes, I was able to remove this queue.

packages/flame/lib/src/components/component.dart
```diff
@@ -326,6 +353,57 @@ class Component {
     _children?.removeAll(cs);
   }
 
+  @internal
+  void mount() {
+    assert(_parent != null);
+    assert(_loaded && !_mounted && !_prepared);
+    if (_parent!.isMounted) {
+      onMount();
+      _prepared = true;
```
spydon, line 362: What is the difference between `_mounted` and `_prepared` now?

spydon (approved):

packages/flame/lib/src/game/mixins/component_tree_root.dart
spydon: Nice!

packages/flame/lib/src/components/component.dart
st-pasha: `_prepared` is set to true after `onMount` runs (I know, sounds weird). Then, at some later time, the component is picked up from the "children queue" and added to the `ComponentSet` of its parent. At that moment, when the component is finally added to its parent's children, the component is considered fully mounted and is marked with `_mounted = true`. 

Immediately after that we perform `remount()` on all children of the component (if any), in order to execute `onGameResize/onMount` for them. By the time this method finishes, both the original component, and all its children will be fully mounted.

spydon: Right, and in what situations do we need to check `_prepared`?

spydon: That queue is being traversed when the parent is being mounted right? Couldn't `mount` run right before it is being added there instead and then the `_parent!.isMounted` check and the `_prepared` variable wouldn't be needed here?

packages/flame/lib/src/components/component.dart
```diff
@@ -384,33 +462,14 @@ class Component {
...
-    }
-  }
+  /// Component at the root of the component tree. This node serves as the
+  /// central place for resolving lifecycle event queues, and also facilitates
+  /// `onGameResize` events. This variable MUST be initialized before components
+  /// may be combined into a tree.
+  ///
+  /// Usually, [FlameGame] is the root of the component tree, and it declares
+  /// itself as such automatically.
+  static ComponentTreeRoot? root;
```
erickzanardo, line 472: How will this work when we have multiple different games on the screen at the same time?

packages/flame/lib/src/game/game_widget/game_widget.dart
erickzanardo: I would at least use `log` from `dart:developer` here instead

packages/flame/lib/src/game/mixins/component_tree_root.dart
```diff
@@ -0,0 +1,116 @@
...
+  /// of adding the components into the tree are fully materialized.
+  ///
+  /// Warning: awaiting on a game that was not fully connected will result in an
+  /// infinite loop. For example, this could occur if you run `x.add(y)` but
+  /// then forget to mount `x` into the game.
+  Future<void> ready() {
+    return Future.doWhile(() async {
+      processComponentQueues();
+      // Give chance to other futures to execute too
+      await Future<void>.delayed(const Duration());
```
erickzanardo, line 60: Can't we use a completer here somehow? instead

packages/flame/lib/src/components/component.dart
st-pasha: Each game is a component, and they all can be mounted to a single "master" component and shown on the screen simultaneously.
Or do you have in mind some specific example that you're worrying about?

erickzanardo: Yeah, on that case it will work, but since this is a static variable, and there will be only one value during the entire app runtime. So I imagine that if we have two games running on different game widgets, I wonder if this will work fine

packages/flame/lib/src/game/game_widget/game_widget.dart
wolfenrain: We should also add a `NOTE:` in the code then to keep track of it OR make a follow up issue for it :D

spydon: Or update to Dart 2.16...

packages/flame/lib/src/game/mixins/component_tree_root.dart
st-pasha: The idea here is that when this function is called, there are multiple components that are being loaded (i.e. they are executing their async onLoad methods). We need to wait for these microtasks to finish, and in order to do that we need to periodically make this async break, giving those microtasks an opportunity to run. This has to be in a loop, because on the first async break there is no guarantee that all onLoads will finish their jobs.

I'm not sure how a completer could work here: after all, a completer simply waits until you declare it as complete.

packages/flame/lib/src/components/component.dart
st-pasha: You're right, this does make the code simpler.
The only trouble was that when a component starts adding new components in its `onMount`, and it modifies the `childrenQueue` while it is being iterated -- however, I added a check to guard against that.

packages/flame/lib/src/game/game_widget/game_widget.dart
st-pasha: Current Flutter version uses Dart 2.15; but Flutter 2.10 (currently in beta) will have Dart 2.16. So, hopefully, we'll be able to upgrade soon. In the meanwhile, i'll post an issue.

packages/flame/lib/src/components/component.dart
st-pasha: Yes, since this is a static variable, there could be some difficulties. It does help though that `ComponentTreeRoot` is designed in such a way that it doesn't _have_ to be physically the root-most component of the tree. So, in theory, multiple independent component trees can share the same "root". We have the "game-in-game" example to demonstrate this.

Ultimately, this use case seems to be purely hypothetical, though. I asked Lukas about this, and neither he nor I were able to come up with a single game example where this could possibly be needed. Can you think of any?

spydon: The situation Erick is talking about is different than the one we were talking about, in Erick's sitation there are multiple `GameWidget`s in the same Fluttet widget tree, not multiple games in a component tree. That usecase is much more common since people don't only use Flame for full blown games, but it could be for only having sprites animated somewhere in their normal app for example.

I think this could be a real problem because one `GameWidget` could be disposed before the other and if that one has the root I guess things could blow up?

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -105,6 +90,10 @@ class FlameGame extends Component with Game {
   @override
   @mustCallSuper
   void onGameResize(Vector2 canvasSize) {
+    if (!isMounted) {
+      // TODO(st-pasha): remove this hack, which is for test purposes only
```
wolfenrain, line 94: Should this not be removed then?

packages/flame/lib/src/game/mixins/component_tree_root.dart
```diff
@@ -0,0 +1,115 @@
...
+    for (final queue in childrenQueues.values) {
+      while (queue.isNotEmpty) {
+        final mounted = queue.first.tryMounting();
+        if (mounted) {
+          queue.removeFirst();
+        } else {
+          break;
+        }
+      }
+      // Check if [childrenQueue] was modified, in which case we need to stop
```
wolfenrain, line 107: Not really a comment, but dartdoc references like this one dont work if it just a normal code comment right?

wolfenrain (commented):
Left a few comments, I also think we should write some docs about testing to help people out a bit more (the `game.ready` is now not "explained" in our docs). But that is for a follow up issue

packages/flame/lib/src/components/component.dart
st-pasha: Well, I imagine in such a situation one would have to be careful to create a single root at the top level, and then not to reassign it when creating internal `Game`s. We could even have an extra parameter `root=false` to `FlameGame` that will prevent it from acquiring the root.

@spydon Would this be ok, or did you have something else in mind?

packages/flame/lib/src/game/flame_game.dart
st-pasha: We have many tests that create a game instance via
```dart
final game = FlameGame()..onGameResize(Vector2.all(100));
```

Of course, this is not a valid initialization of a Game instance, but we use it because it kinda works. However, after this PR, a game will also need to be "mounted" in order to work properly, so all those tests would have to be updated in order to mount the Game instance too.

For now, a simple hack to perform mounting during onGameResize solves the problem, but going forward we'd need to update tests so that they only use properly initialized game instances.

packages/flame/lib/src/game/mixins/component_tree_root.dart
st-pasha: Probably not, at least they aren't highlighted in the IDE.

packages/flame/lib/src/components/component.dart
spydon: > Well, I imagine in such a situation one would have to be careful to create a single root at the top level.

Imho users that create apps just using small parts of Flame shouldn't have to consider the widget lifecycle differently just because they happen to have `GameWidget`s in their widget tree.

> We could even have an extra parameter root=false to FlameGame that will prevent it from acquiring the root.

I don't think it would help, the developer doesn't necessarily know in which order the widgets will live.

> did you have something else in mind?

I don't, but I think we have to come up with another way.

packages/flame/lib/src/game/flame_game.dart
spydon: What needs to be done to fix the tests?

st-pasha: Replace all cases where we create FlameGame manually with some standard way of creating+initializing the game. There could probably be a simple function `initGame()`, so that you could create a game via
```dart
final game = initGame(FlameGame());
```

packages/flame/lib/src/components/component.dart
st-pasha: > but I think we have to come up with another way.

Do you want to do it within this PR, or leave it as a follow-up issue for the future?

st-pasha: We could have a function like `Flame.initializeMultiGameSupport()` that needs to be called at the start of the application, and which creates and stores a single "master" game instance, while prohibiting all subsequent games from overwriting it. Would this approach work?

erickzanardo: It may work, but I don't like that approach, Flame games should be self contained and they shouldn't have global stuff IMO

st-pasha: I believe the main principle was that "common tasks should be easy, and uncommon feasible". The overwhelmingly common case here is to use Flame engine to build games, using FlameGame class or its derivatives. When one builds a game, there is always only one game (it is impossible to play two games at a time). Thus, in practice, a Game is always a singleton, whether we recognize it officially as such or not.

It is only when one tries to use FlameGame class to build non-games, this becomes a concern. But these use-cases are extraordinarily rare. I don't think we ever mention them in documentation or tutorials. So IMHO it's ok for them to be special cases.

spydon: But this is still a common case already today, so we can't overlook it. `Game` is not a singleton, and should not be a singleton. It is pretty much one of our first points every time we present Flame; that it is just another widget.

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -85,6 +67,8 @@ class FlameGame extends Component with Game {
 
   @override
   void updateTree(double dt) {
+    // processComponentQueues();
```
spydon, line 70: Accidentally left I guess

packages/flame/lib/src/game/mixins/single_game_instance.dart
```diff
@@ -0,0 +1,16 @@
+import '../../components/component.dart';
+import 'game.dart';
+
+mixin SingleGameInstance on Game {
```
spydon, line 4: This will differentiate cases with a single and multiple `GameWidget`s in the Flutter tree, this should not be the case.
I think we'll just have to make the `staticGameInstance` non-static, and then traverse the tree each time that a component is added.

packages/flame/test/components/component_lifecycle_test.dart
```diff
@@ -35,74 +38,111 @@ void main() {
...
+        final parent = Component();
+        final child = SlowComponent(0.01);
+        final future = child.addToParent(parent);
+        expect(parent.isMounted, false);
+        expect(parent.isLoaded, false);
+        expect(child.isMounted, false);
+        expect(child.isLoaded, false); // not yet..
+        await future;
+        expect(parent.isMounted, false);
+        expect(child.isLoaded, true);
```
spydon, line 125: How can the child be loaded before the branch has been added to the root?

packages/flame/lib/src/components/component.dart
```diff
@@ -326,6 +366,71 @@ class Component {
...
+  void processPendingLifecycleEvents() {
+    if (_lifecycleManager != null) {
+      _lifecycleManager!.processChildrenQueue();
+      if (!_lifecycleManager!.hasPendingEvents) {
+        _lifecycleManager = null;
+      }
+    }
+  }
+
+  static Game? staticGameInstance;
```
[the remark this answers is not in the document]
st-pasha: It is controlled by the `SingleGameInstance` mixin. So, if a user explicitly adds that mixin to their game, it allows that instance to be found statically, and thus have faster onLoads. A mixin should only be used in traditional game applications where there is a single game.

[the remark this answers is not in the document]
spydon: Ah, now I understand how it works. Maybe that solution isn't so bad then!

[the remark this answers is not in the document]
spydon: It would be good to have a small test for lifecycles where there are multiple game instances.

spydon (approved):
Good stuff, just minor things left now!

packages/flame/lib/src/game/mixins/single_game_instance.dart
st-pasha: This is an _optional_ mixin: if the user knows that the application that they are building only ever uses a single `Game` instance, within a single `GameWidget`, they can opt into using this mixin.

spydon, line 4: Would be good to have this documented in the docs (and with dartdocs)

packages/flame/lib/src/components/component.dart
```diff
@@ -326,6 +366,71 @@ class Component {
...
+  /// and later re-mounted. For these components we need to run [onGameResize]
+  /// (since they haven't passed through [add]), but we don't have to add them
+  /// to the parent's children because they are already there.
+  @internal
+  void mount({bool existingChild = false}) {
+    assert(_loaded && !_mounted && _parent!._mounted);
+    if (existingChild || !_loading) {
+      onGameResize(findGame()!.canvasSize);
+    }
+    _loading = false;
```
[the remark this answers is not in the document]
st-pasha: We need to be able to distinguish between the cases when `onLoad` had run just now, vs the component was loaded a long time ago. In the latter case we have to call `onGameResize` before mounting, in the former case `onGameResize` was already called before loading, so no need to do that again.

The flag `_loading` helps distinguish between these two cases.

packages/flame/lib/src/components/component.dart
```diff
@@ -326,6 +366,71 @@ class Component {
...
+    onMount();
+    _mounted = true;
```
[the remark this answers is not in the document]
st-pasha: I consider `setMounted` to be a temporary function, which can be removed after we refactor our test to only use properly initialized game instances.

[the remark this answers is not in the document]
spydon: Ah, I see, maybe put a `TODO` on that one and open an issue for that instead then?

packages/flame/test/components/component_lifecycle_test.dart
st-pasha: This particular test uses a game with `SingleGameInstance` mixin, which preserves the eager loading mechanics.

packages/flame/lib/src/game/flame_game.dart
st-pasha: yep

packages/flame/lib/src/components/component.dart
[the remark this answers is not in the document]
st-pasha: ok

packages/flame/lib/src/game/mixins/single_game_instance.dart
st-pasha: done

packages/flame/lib/src/game/flame_game.dart
st-pasha: I created issue #1351 for this.

packages/flame/lib/src/components/component.dart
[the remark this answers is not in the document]
st-pasha: added

packages/flame/lib/src/components/component.dart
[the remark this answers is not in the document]
spydon: Hmm, I think the variable should probably be renamed then, it's a bit strange that `_loaded` here needs to be `true`, and `_loading` can be also be `true`. Maybe rename to something that is related to whether `onGameResize` should run or not?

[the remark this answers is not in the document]
st-pasha: Yeah, the variable serves dual purpose: first it tells you whether the component is currently being loaded, then it stays on to signal that the component was actually loaded _just now_, not some time ago. I can't figure out a good name for it.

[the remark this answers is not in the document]
spydon: Hmm, maybe we could have some form of internal state variable instead of the booleans?
Doesn't have to be a full blown state-machine, but just a variable that takes an enum value, wdyt?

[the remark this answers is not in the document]
spydon: Or we keep some booleans, but as getters that check the state variable.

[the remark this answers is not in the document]
st-pasha: ok, done

spydon (approved):
Lgtm, the state variable was a real improvement!

packages/flame/lib/src/components/component.dart
```diff
@@ -326,6 +371,73 @@ class Component {
...
+  void processPendingLifecycleEvents() {
+    if (_lifecycleManager != null) {
+      _lifecycleManager!.processChildrenQueue();
+      if (!_lifecycleManager!.hasPendingEvents) {
+        _lifecycleManager = null;
+      }
+    }
+  }
+
+  static Game? staticGameInstance;
```
erickzanardo, line 435: Can we mark this as @internal?

packages/flame/lib/src/components/component_set.dart
```diff
@@ -55,7 +44,7 @@ class ComponentSet extends QueryableOrderedSet<Component> {
   /// `updateComponentList()`.
   @internal
   void addChild(Component component) {
```
erickzanardo, line 46: Maybe just remove this if it just calls the super.add, it is marked as internal, so no breaking changes will be introduced

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -105,12 +93,42 @@ class FlameGame extends Component with Game {
...
+  /// This is mainly intended for testing purposes: awaiting on this future
+  /// ensures that the game is fully loaded, and that all pending operations
+  /// of adding the components into the tree are fully materialized.
+  ///
+  /// Warning: awaiting on a game that was not fully connected will result in an
+  /// infinite loop. For example, this could occur if you run `x.add(y)` but
+  /// then forget to mount `x` into the game.
+  Future<void> ready() async {
+    var repeat = true;
+    while (repeat) {
```
erickzanardo, line 119: I am failing to see how this will ensure that the game will be loaded once this future is completed.

`processPendingLifecycleEvents` do call load from children but it don't await for their completion.

Can you shed some light here?

packages/flame/lib/src/game/game_widget/game_widget.dart
erickzanardo, line 335: Lets use at least the log method from the developer package here

packages/flame_bloc/pubspec.yaml
```diff
@@ -2,6 +2,7 @@ name: flame_bloc
 description: Bloc integration to Flame games
 version: 1.1.0
 homepage: https://github.com/flame-engine/flame/tree/main/packages/flame_bloc
+publish_to: none
```
erickzanardo, line 5: I don't think these changes are correct

cc @spydon

packages/flame_bloc/test/flame_bloc_game_test.dart
```diff
@@ -27,8 +26,15 @@ class MyBlocGame extends FlameBlocGame {}
 class InventoryComponent extends Component
     with BlocComponent<InventoryCubit, InventoryState> {}
 
-// ignore: implicit_dynamic_type
-class MockInventoryComponent extends Mock implements BlocComponent {}
+class MockInventoryComponent extends Component
```
erickzanardo, line 29: Why not using mocktail here anymore?

packages/flame_test/example/pubspec.yaml
```diff
@@ -11,12 +11,14 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  flame: ^1.0.0
+  flame:
```
erickzanardo, line 14: Same here, melos will take care of this.

packages/flame_test/lib/src/flame_test.dart
```diff
@@ -81,6 +81,9 @@ class GameTester<T extends Game> {
 
     await game.onLoad();
     game.update(0);
+    if (game is FlameGame) {
```
erickzanardo, line 84: I am not sure if we should hard code this, there may be situations where the user will want to do assertions before the game is ready and after.

packages/flame_test/pubspec.yaml
```diff
@@ -2,19 +2,21 @@ name: flame_test
 description: A package with classes to help testing applications using Flame
 version: 1.1.0
 homepage: https://github.com/flame-engine/flame/tree/main/packages/flame_test
+publish_to: none
```
erickzanardo, line 5: Here as well, I don't think these changes are correct (except the test package bump)

packages/flame/lib/src/components/component.dart
st-pasha: yes, makes sense.

packages/flame/lib/src/components/component_set.dart
st-pasha: ok, will do

packages/flame/lib/src/game/game_widget/game_widget.dart
st-pasha: done

packages/flame_bloc/pubspec.yaml
st-pasha: I just tried undoing these changes in [e49616](https://github.com/flame-engine/flame/pull/1337/commits/e49616130bafbf384dd73bb9d4e9a5bc4ff256e9), but it immediately led to [analyzer failure](https://github.com/flame-engine/flame/runs/5080118337?check_suite_focus=true) in the CI. So I believe these will have to stay until #1354 is working correctly.

erickzanardo: I would rather wait for that PR to be merged than to merge this changes as it is, because then we will have to go through the places and revert

packages/flame/lib/src/game/flame_game.dart
st-pasha: `processChildrenQueue` method removes components from the queue only after they finished loading+mounting. Thus, while a component is loading, the getter `.hasPendingLifecycleEvents` will be returning `true`.

Then, in each iteration of this loop we check if there are any queues that has pending lifecycle events. If such queues are found, we make a short async break and start polling all the queues again.

packages/flame_test/lib/src/flame_test.dart
st-pasha: `await game.ready()` serves the same purpose as `game.update(0)`, only it's more reliable: the queues are flushed not once but repeatedly until they're empty. Thus, this line improves upon the existing behavior without fundamentally altering it.

> there may be situations where the user will want to do assertions before the game is ready and after.

I agree, this could be useful in limited number of circumstances, such as for example when testing the `FlameGame` class itself. But in those tests we would probably want to avoid using `GameTester` and invoke all relevant lifecycle methods manually.

packages/flame_bloc/pubspec.yaml
st-pasha: I guess it would depend on how close #1354 is to completion, so probably @spydon knows best. I notice that that PR already fixes some of the `pubspec.yaml` files, so it's probably not a big deal to fix few more within that same PR.

packages/flame_test/lib/src/flame_test.dart
erickzanardo: Still I would prefer to give the user the option. I think that an optional book argument, defaulting to true would gives us the best of both worlds

st-pasha: Is `book` the name of the argument?
Also, should it guard only `game.ready` or also `game.update`, `game.onLoad`, `game.onMount` and `game.onGameResize` ?

packages/flame/lib/src/game/flame_game.dart
spydon: Is the loop used here? It looks like it is set to `repeat = false` in the first iteration, and never set to true?

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -316,13 +318,27 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
             color: game.backgroundColor(),
             child: LayoutBuilder(
               builder: (_, BoxConstraints constraints) {
-                game.onGameResize(constraints.biggest.toVector2());
+                final size = constraints.biggest.toVector2();
+                if (size.isZero()) {
+                  return widget.loadingBuilder?.call(context) ?? Container();
```
spydon, line 323: I have little knowledge of how the `loadingBuilder` works, will this return the currently running loading widget, or will it create a new one here?

packages/flame_rive/pubspec.yaml
```diff
@@ -18,4 +17,5 @@ dev_dependencies:
   dartdoc: ^4.1.0
   flutter_test:
     sdk: flutter
+  dart_code_metrics: ^4.10.0
```
spydon, line 20: This should not be re-added

spydon (approved):
Some final small comments, other than those I think this is ready to get merged!

packages/flame_rive/pubspec.yaml
st-pasha: yeah, looks like it got left accidentally after a merge

packages/flame/lib/src/game/flame_game.dart
st-pasha: Yes, it's set to true below in line 126 whenever a child says that it still has some pending lifecycle events.

packages/flame_test/lib/src/flame_test.dart
erickzanardo: Book was a typo, it was supposed to be `bool`. I think it could guard only the ready, as it is the one being introduced here

packages/flame/lib/src/game/game_widget/game_widget.dart
st-pasha: `loadingBuilder` is a function that creates and returns a Flutter widget on each invocation. This is perfectly normal from Flutter's standpoint: everything starting from line 270 above will create a new Flutter widget every time the build method is called. If you need some persistency in widget's data, you'd declare it as a `StatefulWidget` and keep its state as a separate object whose lifetime is managed more carefully.

This is of course very different from what we do in FCS, where components are added and removed from the game explicitly.

packages/flame/lib/src/components/component.dart
```diff
@@ -255,63 +277,86 @@ class Component {
...
+  /// Adds this component to the provided [parent] (see [add] for details).
+  Future<void>? addToParent(Component parent) {
+    assert(
+      _parent == null,
+      '$this cannot be added to $parent because it already has a parent: '
+      '$_parent',
+    );
+    assert(
+      _state == LifecycleState.uninitialized ||
+          _state == LifecycleState.removed,
```
luanpotter, line 342: pretty message for this assert? or this one that "should never happen AFAWK"

packages/flame/lib/src/game/mixins/single_game_instance.dart
```diff
@@ -0,0 +1,29 @@
...
+/// The effect of using this mixin is that it allows component lifecycle events
+/// to be resolved faster in certain situations. For example, [Component]'s
+/// `onLoad` will be able to run even if its parent is currently detached.
+///
+/// Using this mixin in applications where there are multiple [Game] instances
+/// attached to multiple GameWidgets, is invalid.
+mixin SingleGameInstance on Game {
+  @override
+  void onMount() {
+    Component.staticGameInstance = this;
```
luanpotter, line 20: we could add some checks like: `if staticGameInstance != null && != this` then throw pretty error message saying you cannot use SingleGameInstance like that

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -1,9 +1,10 @@
+import 'dart:developer';
```
luanpotter, line 1: wow what class requires this? never seem it before

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -316,13 +318,27 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
...
+                  return widget.loadingBuilder?.call(context) ?? Container();
+                }
+                game.onGameResize(size);
                 return FutureBuilder(
                   future: loaderFuture,
                   builder: (_, snapshot) {
                     if (snapshot.hasError) {
                       final errorBuilder = widget.errorBuilder;
                       if (errorBuilder == null) {
+                        // @Since('2.16')
```
luanpotter, line 332: can we just add dart>=2.16 to the pubspecs? otherwise I would just put out an issue with this code to tackle later, rather than commented code

luanpotter (approved):
This is awesome!

packages/flame/lib/src/components/component.dart
st-pasha: This is just a sanity check on the internal state: I do not think it is possible to violate this assert in the current code. But, if in the future we need to refactor this again (say, adding more states), this assert may trigger.

packages/flame/lib/src/game/mixins/single_game_instance.dart
st-pasha: There is an assert like that in `FlameGame` class:
```dart
assert(
      Component.staticGameInstance == null,
      '$this instantiated, while another game ${Component.staticGameInstance} '
      'declares itself to be a singleton',
    );
```
I think it's better to have it in FlameGame because then not only we protect against two `SingleGameInstance` objects, but we also prevent a `SingleGameInstance` to coexist with a regular `FlameGame`.

packages/flame/lib/src/game/game_widget/game_widget.dart
st-pasha: This is for logging the stack-trace in line 337. It won't be needed after we upgrade to Dart 2.16 (which is in Flutter 2.10.0, released on 2022-02-03).

packages/flame/lib/src/game/game_widget/game_widget.dart
st-pasha: I'd love to! We have issue #1346 so that we won't forget to do that.

packages/flame_test/lib/src/flame_test.dart
st-pasha: Ah I see, I was at a loss what did "book" mean :) 
Anyways, added the parameter `makeReady` now, which is true by default but can be set to false if the user so desires.

packages/flame/lib/src/components/component.dart
```diff
@@ -23,35 +25,38 @@ class Component {
   /// to the root `FlameGame`.
   PositionType positionType = PositionType.game;
 
-  /// Whether this component has been prepared and is ready to be added to the
-  /// game loop.
-  bool isPrepared = false;
+  LifecycleState _state = LifecycleState.uninitialized;
```
erickzanardo, line 28:
```suggestion
  var _state = LifecycleState.uninitialized;
```

erickzanardo (approved):
One small additional nit, but LGTM!

packages/flame/lib/src/components/component.dart
wolfenrain: Isn't this something we can enforce using lint rules?

wolfenrain: Like the [omit_local_variable_type](https://dart-lang.github.io/linter/lints/omit_local_variable_types.html)

erickzanardo: Yeah, I thought we already had that rule, but apparently not

spydon: We have that rule, but I don't think this variable counts as local since it's a top-level class variable, even though it is private?

erickzanardo: That is true, so we need a different rule for this?

spydon: Isn't it good to have the type on top-level since it is living in a bit larger context? And then it will be consistent with public class variables too.

spydon: Yeah true true, I'm still stuck in the mindset that it is nice to see it directly to the left.

erickzanardo: Either way, this is just a nit, totally not a blocker.

PR #1355, opened 2022-02-05
feat: Camera as a component

### LINKED ISSUE #1161: [FR] Camera
This is a specific proposal for camera design, following the discussion in #931.

## Overview
The following top-level classes to be added:
* `World`: a simple container class which has most other game elements as children. Its main functionality is that it suppresses rendering of its children, however updates are working normally. The elements inside the World can only be viewed via a `Camera`.
* `Camera` is a component that carries a reference to the world that it views, and a `Viewport` instance. Thus, a `Camera` works as a component-wrapper for a `Viewport` class. Camera's `render` method applies the viewport transform before rendering its children. Effects that are applied to the `Camera` should affect the viewport.
  - `Viewport` describes a "window" through which the game world is observed. This window has a specific size, shape, and a position on the canvas. 
* `Viewfinder` is a component that keeps track of the game world's current position, rotation angle, and scale (zoom). Thus, while `Viewport` represents a window that can be moved around the screen, the job of the `Viewfinder` is to move the world within that window. Viewfinder's `render()` method will apply the clip mask from the viewport, its own transform, and then render the game world. Effects that are applied to the `Viewfinder` will move the current point of view within the world.
* `CameraController` is a component that controls the behavior of the camera (the viewfinder). It can do things like follow a component, restrict camera's movement to within a certain global boundary, etc. There could be multiple controllers attached to a camera, and they can be inter-mixed with effects.
```
┌────────────┐
│ FlameGame  │
└─┬──────────┘
  │ ┌───────────────┐
  ├─┤ World         │◄─ ─ ─ ─ ┐
  │ └─┬─────────────┘         |
  │   └── World elements...   |
  │                           |
  │ ┌────────────────┐        |
  └─┤ Camera         │        |
    │   .viewport    │        |
    │   .world       │─ ─ ─ ─ ┘
    └─┬──────────────┘
      │  ┌──────────────┐
      ├──┤ ViewFinder   │
      │  └─┬────────────┘
      │    │  ┌───────────────────┐
      │    └──┤ CameraController  │
      │       └───────────────────┘
      └── HUD elements...
```
Additional notes:
* Elements that are attached to the `Camera` will be subject to the viewport transform, but not viewport clip mask. Such elements are what we commonly refer to `.isHud` (or `.ignoreCamera`) components.
* Elements that are added to the root FlameGame will not be subject to either the viewport or viewfinder transforms. These components are equivalent to current `.ignoreViewport` components.
* This design allows having multiple cameras, viewing either the same or different worlds.

## Transition
The following is the outline for how to introduce these changes in a non-breaking way:
1. Create class `FlameGame2` which derives from `FlameGame` and overrides `add*()` methods so that the components would be added to an internal `.world` instance. Encourage users to try this new class and test how it works.
2. Once we sufficiently sure that `FlameGame2` is working, rename `FlameGame` into `FlameGame1`, and `FlameGame2` into `FlameGame`.
3. At a later release, mark `add*()` methods as `@deprecated` and instruct the users to add their components to `game.world` or `game.camera` directly. Components that need to be added outside of the game world will go through the `addToRoot()` method.
4. At a later release, remove `add*()` method overrides. At this point whoever ignored warnings at stage 3 will have their game broken. Mark `addToRoot` method deprecated and instruct the users to use regular `add()` method.
5. Remove the `addToRoot` method, and the original `FlameGame1` class.

## Special cases
* Some components want to be aware of the current rendering context (clip rectangle and zoom level) for optimization purposes. In order to accommodate such needs, I suggest a simple static property `Camera.currentContext` which will be set whenever a camera begins its rendering pipeline, and reset when it finishes.
  - an alternative is to pass the context as `render(canvas, context)`, but this is much more intrusive to current users;
  - an alternative where the camera is discovered by traversing up the tree is not feasible, since there could be more than one camera watching the same world (for example: split-screen view);
  - `currentContext` will also protect us against infinite rendering loops when a camera watches itself.
* The `ParallaxLayer` component will be a regular component, belonging to the game world. When rendering, it will look at the camera's context and partially undo the translation that was applied by the camera. It will also need to modify the clip mask in the camera's context so that the children of the parallax layer can render correctly.
* Touch events. In a sense, such events are the inverse of `render`, and have to be treated accordingly. This area needs more exploration, but touch events are not working properly even with the current implementation (in a sense that if a component modifies `renderTree`, the touch events do not respect this).

st-pasha:
# Description

This PR adds a new (experimental) component `Camera2`, which implements the camera functionality. The new camera contains `Viewport` and `Viewfinder`, which are also components. Several viewport implementations are included:
  - `MaxViewport` 
  - `FixedAspectRatioViewport`
  - `FixedSizeViewport`
  - `CircularViewport`

Also:
 - implemented styling for admonitions in the documentation;
 - fixed a bug with re-adding a component with children.
 

## Example

https://user-images.githubusercontent.com/4231472/152636447-734531d2-6553-4d73-9d21-49b146725938.mov

## Checklist

## Breaking Change

## Related Issues

Closes #1161
Closes #950


packages/flame/lib/src/experimental/viewfinder.dart
```diff
@@ -0,0 +1,126 @@
...
+import 'camera.dart';
+import 'viewport.dart';
+
+/// [Viewfinder] is a part of a [Camera2] system that controls which part of
+/// the game world is currently visible through a viewport.
+///
+/// The viewfinder contains the game point that is currently at the
+/// "cross-hairs" of the viewport ([position]), the [zoom] level, and the
+/// [angle] of rotation of the camera.
+///
```
luanpotter, line 18:
```suggestion
```

packages/flame/lib/src/experimental/max_viewport.dart
```diff
@@ -0,0 +1,23 @@
+import 'dart:ui';
+
+import 'package:vector_math/vector_math_64.dart';
+
+import 'viewport.dart';
+
+/// The default viewport, which is as big as the game canvas allows.
+///
+/// This viewport does not perform any clipping. (?)
```
luanpotter, line 9:
```suggestion
/// This viewport does not perform any clipping.
```
I think that is totally correct given the impl & desc

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -323,6 +323,7 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
                     if (snapshot.hasError) {
                       final errorBuilder = widget.errorBuilder;
                       if (errorBuilder == null) {
+                        print(snapshot.stackTrace);
```
luanpotter, line 326:
```suggestion
```

packages/flame/lib/src/experimental/viewfinder.dart
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  /// Rotation angle of the game world, in radians.
+  ///
+  /// The rotation is around the axis that is perpendicular to the screen.
+  double get angle => -_transform.angle;
+  set angle(double value) => _transform.angle = -value;
+
+  /// Reference to the parent camera.
+  Camera2 get camera => parent! as Camera2;
```
luanpotter, line 50: what prevents users from manually adding a Viewfinder to other components? I wonder if there is a neat place to add a helpful error

packages/flame/lib/src/experimental/viewfinder.dart
```diff
@@ -0,0 +1,126 @@
...
+import 'viewport.dart';
+
+/// [Viewfinder] is a part of a [Camera2] system that controls which part of
+/// the game world is currently visible through a viewport.
+///
+/// The viewfinder contains the game point that is currently at the
+/// "cross-hairs" of the viewport ([position]), the [zoom] level, and the
+/// [angle] of rotation of the camera.
+///
+class Viewfinder extends Component {
```
luanpotter, line 19: I think a very useful feature of the current impl is the "follow" behaviour (including boundaries). how easy would it be to hook that up onto the viewfinder? might be even worth an example and/or helper methods

packages/flame/lib/src/experimental/world.dart
```diff
@@ -0,0 +1,23 @@
...
+
+import '../components/component.dart';
+import 'camera.dart';
+
+/// The root component for all game world elements.
+///
+/// The primary feature of this component is that it disables regular rendering,
+/// and allows itself to be rendered through a [Camera2] only. The updates
+/// proceed through the world tree normally.
+class World extends Component {
```
luanpotter, line 13: I wonder if "World" is sufficiently clear that it is part of the camera system, but I can't think of anything better at the moment

packages/flame/lib/src/experimental/viewfinder.dart
st-pasha: Yes, that would be the next step (can't fit everything into one PR!)
I want effects system to recognize the `Viewfinder` and `Viewport` classes, and then you can attach various move/rotate/scale/shake/whatever effects to either of them. "Follow" and "respect boundary" aren't effects though, but they can be implemented as independent components. In the end it should have a greater set of capabilities than the original `Camera`.

packages/flame/lib/src/experimental/world.dart
st-pasha: Eventually we can extend the functionality here.

For example, I was thinking that this could be a nice place to add the ability to stop or slow down the time.

packages/flame/lib/src/experimental/viewfinder.dart
st-pasha: There is an assert in `onMount`, I think that should suffice.

spydon (approved):

examples/lib/stories/camera_and_viewport/camera2_example.dart
```diff
@@ -0,0 +1,522 @@
...
+        dir = bendDirection ? -1 : 1,
+        path = Path(),
+        foot = Vector2.zero() {
+    final ok = placeFoot(Vector2(x1, y1));
+    assert(ok);
+  }
+
+  final double x0, y0;
+  final double x1, y1;
+  final double l1, l2, l3, r;
```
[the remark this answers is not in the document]
st-pasha: I've added comments explaining what those variables are

packages/flame/lib/src/experimental/camera.dart
```diff
@@ -0,0 +1,92 @@
...
+/// create a camera that looks at itself.
+///
+/// Since [Camera2] is a [Component], it is possible to attach other components
+/// to it. In particular, adding components directly to the camera is equivalent
+/// to adding them to the camera's parent. Components added to the viewport will
+/// be affected by the viewport's position, but not by its clip mask. Such
+/// components will be rendered on top of the viewport. Components added to the
+/// viewfinder will be rendered as if they were part of the world. That is, they
+/// will be affected both by the viewport and the viewfinder.
+class Camera2 extends Component {
```
spydon, line 30: We had a discussion in the meeting now, maybe this can be named `AdvancedCamera` or something in the lines of that and live alongside the other simpler camera system? (We are going to try to get this merged in the beginning of this week)

spydon (approved):

packages/flame/lib/src/experimental/camera.dart
st-pasha: The new camera isn't that much more advanced than the old one; the main difference is that it is now a fully-fledged Component, so it can be added anywhere. So, perhaps call it `CameraComponent` then?

spydon: Sounds good to me, I'll check with the rest.

spydon: Some docs explaining the difference of the camera's would be good, maybe some docs about the other new concepts too, even though it is in the experimental folder.

spydon: The others agreed to the `CameraComponent` name.

An export file called `experimental.dart` or such would be good to have too, and an explanation in the docs of what it means for a feature to be in there, so that people trying it don't have to do the internal imports.

My idea of the meaning of the experimental directory is to have things that we are not certain about the API etc, so that we can more easily make breaking changes regarding the features in there in the future. What do you think?

spydon (approved):
Good job!

Issue #1351, opened 2022-02-02
Tests should use only properly initialized game instances


st-pasha:
Currently there are multiple tests which create a `FlameGame` instance manually and partially initialize it.

In order to reduce the amount of test breakage when the initialization process of a game changes, we should change those tests to only use a game instance acquired from properly initialized sources (such as flame_test).

idlehart:
Is this something i can do for hacktoberfest?

spydon:
> Is this something i can do for hacktoberfest?

Go for it! I'm not sure how many unconverted tests that there are left, but I guess that is part of the ticket to find out. :)

spydon:
Do we have to create a new function that replaces the widget tests too?

PR #1358, opened 2022-02-06
test: Add testWithFlameGame and testWithGame helper functions

st-pasha:
# Description

Add function `withFlameGame()` that can supply a prepared `game` instance to the test body. 

The main advantage over the `flameGame.test` is that it doesn't confuse Android Studio over whether something is a test or not. In particular, it allows each test to be run individually if needed: 

Screen Shot 2022-02-05 at 8 25 00 PM

Note: the changes to `camera_and_viewport_test.dart` are purely cosmetic, even though GitHub shows a large diff.

## Checklist

## Breaking Change

## Related Issues

WIP for #1351


luanpotter (commented):
I remember that I annotated the `flameGame.test` so it would be recognized as a test function, at least on vscode. Interesting to see it didn't work, I wonder if we can do something to create custom test functions.

Regardless of that, I don't see any reason not to change the api slightly to your suggestion, it looks equally clean to me. I imagine the goal is to replace the old one entirely right?

I leave it for @erickzanardo to ponder as he originally designed this particular API

st-pasha:
Yes, I was wondering about that too...
My suspicion is that `@isTest` only works with pure functions, not with methods.

So I can actually change this current function into `@isTest`-declared `flameGameTest()`, in which case it would be much closer to the original, while still maintaining the ability to be run individually:

Screen Shot 2022-02-09 at 8 43 11 AM

What do you think?

spydon (approved):

st-pasha:
@spydon Do you think it's better to have
```dart
test(
  'test name',
  withFlameGame((game) async {
    ...
  }),
);
```
or
```dart
testFlameGame(
  'test name',
  (game) async {
    ...
  },
);
```
?

spydon:
> @spydon Do you think it's better to have
> ```dart
> test(
>   'test name',
>   withFlameGame((game) async {
>     ...
>   }),
> );
> ```
> or
> ```dart
> testFlameGame(
>   'test name',
>   (game) async {
>     ...
>   },
> );
> ```
> ?

Not sure, maybe the first? 🤔
Is there any standard for how others are doing this?

luanpotter:
@st-pasha I see, that must be the case, good catch. mystery solved aside, as for which is better, I don't have any strong preference. maybe @erickzanardo can chime in as he designed the original api (before my refactor)

st-pasha:
> I do prefer this: `testFlameGame( ...`

:+1:
I'll switch the API then.

packages/flame_test/lib/src/test_flame_game.dart
```diff
@@ -0,0 +1,73 @@
...
+///     expect(component.isMounted, true);
+///   },
+/// );
+/// ```
+///
+/// The `game` instance supplied by this function to your test [testBody] is a
+/// standard [FlameGame]. If you want to have any other game instance, use the
+/// [testUserGame] function.
+@isTest
+Future<void> testFlameGame(
```
erickzanardo, line 27: We should add the equivalent of these methods to the game widget test method, like we have on the Flame Tester class, wdyt?

st-pasha: Probably, though I'm not very familiar with how the widget tests work.

erickzanardo: you could base on the widgetTest on the FlameTester class.

But we could also leave that to a follow up too

packages/flame_test/pubspec.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
   flutter_test:
     sdk: flutter
   meta: ^1.7.0
-  test: ^1.19.3
+  test: ^1.17.12
```
erickzanardo, line 17: Hmm, not sure about downgrading this, any reason for that?

erickzanardo: Ahhh, got it, this is conflicting with flutter's sdk test package.

We can revisit this later

erickzanardo (approved):

PR #1360, opened 2022-02-07
feat: Added `transform` to `Rect`

wolfenrain:
# Description

Noticed there wasn't a method for transform rects. This method is kinda biased as it will transform from "center".

## Checklist

## Breaking Change

  e.g. `feat!:`, `fix!:`).

## Related Issues

*Replace this paragraph with a list of issues related to this PR from the [issue database].
Indicate which of these issues are resolved or fixed by this PR, i.e. Fixes #xxxx*


packages/flame/lib/src/extensions/matrix4.dart
```diff
@@ -78,6 +80,13 @@ extension Matrix4Extension on Matrix4 {
     return transform2(out);
   }
 
+  /// Transform [rect] of type [Rect] using the transformation defined by this.
+  Rect transformRect(Rect rect) {
```
spydon, line 84: Maybe this could take an `Anchor` too?

packages/flame/lib/src/extensions/matrix4.dart
```diff
@@ -78,6 +80,13 @@ extension Matrix4Extension on Matrix4 {
     return transform2(out);
   }
 
+  /// Transform [rect] of type [Rect] using the transformation defined by this.
```
st-pasha, line 83: This method works only with translation transforms and non-negative scale transforms, right?

wolfenrain: Correct, should add that to the dartdoc

packages/flame/lib/src/extensions/matrix4.dart
spydon, line 84: Maybe this live in a `RectExtension` and take a transform as the argument instead? I think that feels a bit more natural.

wolfenrain: Agreed! That makes more sense to me as well

packages/flame/lib/src/extensions/matrix4.dart
wolfenrain: If you have any suggestions, I am all ears.

erickzanardo (approved):

packages/flame/lib/src/extensions/rect.dart
```diff
@@ -52,6 +53,19 @@ extension RectExtension on Rect {
...
   }
 
+  /// Transform Rect using the transformation defined by [matrix].
+  ///
+  /// **Note:** Rotation matrices will increase the size of the [Rect] but they
+  /// will not rotate it as [Rect] does not have any rotational values.
+  ///
+  /// **Note:** Only non-negative scale transforms are allowed, if a negative
+  /// scale is applied it will return a zero-based [Rect].
+  Rect transform(Matrix4 matrix) {
```
spydon, line 63: It should probably be mentioned that it transforms from the center, since no optional anchor was added.

st-pasha: I don't think anchor would do anything here. If the transform is only limited to translations and scales, then it does not depend on what the anchor is.

packages/flame/lib/src/extensions/rect.dart
```diff
@@ -52,6 +53,19 @@ extension RectExtension on Rect {
     ];
   }
 
+  /// Transform Rect using the transformation defined by [matrix].
+  ///
+  /// **Note:** Rotation matrices will increase the size of the [Rect] but they
+  /// will not rotate it as [Rect] does not have any rotational values.
```
luanpotter, line 59: I wonder if it would be best to validate and throw errors for invalid matrixes.

We could (in the future) also have a version that fully works with any matrix and returns a [Polygon](https://github.com/flame-engine/flame/blob/main/packages/flame/lib/src/geometry/polygon.dart) to represent non-AA rectangles

packages/flame/test/extensions/rect_test.dart
```diff
@@ -36,5 +37,14 @@ void main() {
       expect(r2.position.x, r1.left);
       expect(r2.position.y, r1.top);
     });
+
+    test('test transform', () {
```
luanpotter, line 41: should we test the edge cases mentioned on the comments?

luanpotter (approved):
LGTM with a couple comments

packages/flame/lib/src/extensions/rect.dart
wolfenrain: I could, but not sure how to validate that on a matrix. So if anyone has any pointers, I am all ears!

packages/flame/lib/src/extensions/rect.dart
wolfenrain: Added a comment for it regardless, but yea adding an anchor isn't possible unless we go for polygons in the future

packages/flame/lib/src/extensions/rect.dart
```diff
@@ -52,6 +53,22 @@ extension RectExtension on Rect {
...
+  ///
+  /// **Note:** Only non-negative scale transforms are allowed, if a negative
+  /// scale is applied it will return a zero-based [Rect].
+  ///
+  /// **Note:** The transformation will always happen from the center of the
+  /// `Rect`.
+  Rect transform(Matrix4 matrix) {
+    final topLeft = matrix.transform2(this.topLeft.toVector2());
+    final bottomRight = matrix.transform2(this.bottomRight.toVector2());
```
spydon, line 68: Is it possible to calculate these values without creating new vectors? Since `transform2` also creates a new vector there will be four new vectors for every call to this method.

wolfenrain: I can do it without creating any vectors at all!

spydon (approved):

PR #1354, opened 2022-02-04
chore: Add melos-action to cicd

spydon:
# Description

Adding the `melos-action` to the cicd workflow so that `melos bootstrap` is run before analyze and test.

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the
relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] The title of my PR starts with a [Conventional Commit] prefix (`fix:`, `feat:`, `docs:` etc).
- [x] I have read the [Contributor Guide] and followed the process outlined for submitting PRs.
- [x] I have updated/added tests for ALL new/updated/fixed functionality.
- [x] I have updated/added relevant documentation in `docs` and added dartdoc comments with `///`.
- [x] I have updated/added relevant examples in `examples`.

## Breaking Change

Does your PR require Flame users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change. (Indicate it in the [Conventional Commit] prefix with a `!`,
  e.g. `feat!:`, `fix!:`).
- [x] No, this is *not* a breaking change.

## Related Issues

*Replace this paragraph with a list of issues related to this PR from the [issue database].
Indicate which of these issues are resolved or fixed by this PR, i.e. Fixes #xxxx*

[issue database]: https://github.com/flame-engine/flame/issues
[Contributor Guide]: https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md
[Flame Style Guide]: https://github.com/flame-engine/flame/blob/main/STYLEGUIDE.md
[Conventional Commit]: https://conventionalcommits.org

PR #1365, opened 2022-02-10
docs: Upgrade documentation

st-pasha:
# Description

This is a major overhaul of the documentation theme. The primary driving factor here is that we want to significantly expand the documentation site to include docs for all small "side-packages" (such as `flame_bloc`, `flame_rive`, etc) that together constitute the larger Flame framework. But before we can do that, the documentation site must be able to support this large number of docs in a way that wouldn't be overwhelming to the user. In short, we need to abandon the flat list of documents, and instead structure everything into appropriate folders.

In this PR, I do the following:
  - Switch from our previous heavily-modified `sphinx-book-theme` to a completely custom hand-made theme, code-named "flames". This was needed because we reached a point where modifying the theme via CSS was no longer sufficient.
    - The additional advantage here is that upstream changes to the external theme will no longer be able to break our docs site (as already happened once before). 
  - Updated look-and-feel of the website so that it would look more tidy and professional. Color palette switched towards warmer, "flame-like" colors.
  - The menu on the LHS can now be arbitrarily nested.
  - The docs site now renders much better on mobile devices.
  - Added Discord button in the top menu.
  - Added UI to dismiss highlighted search terms after a search.
  - External links now have visual indicator that they are external.
  - Organized the existing documents into several folders to improve the overall structure.

This last point can be elaborated. I imagine that in the future we will have the documentation structure that roughly follows the principle of "one top level folder per package". That is:
  - flame
  - flame_audio
  - flame_bloc
  - flame_fire_atlas
  - flame_flare
  - etc.
  
It is NOT the purpose of this PR to create documentation for all those missing packages. Instead, the goal is to create the structure where such expanded documentation can be easily added. I moved the existing documents in anticipation of this structure, and also in order to verify that documentation works for such nested use-cases.

## Samples

**Before**:
Screen Shot 2022-02-10 at 2 17 22 AM

**After**:
Screen Shot 2022-02-10 at 2 16 24 AM

**Before (mobile)**
Screen Shot 2022-02-10 at 2 44 38 AM

**After (mobile)**
Screen Shot 2022-02-10 at 2 44 27 AM

## Checklist

## Breaking Change


doc/flame/structure.md
```diff
@@ -1,14 +1,16 @@
 # Structure
 
-Flame has a proposed structure for your project that includes the standard Flutter `assets` directory in addition to two children: `audio` and `images`.
+Flame has a proposed structure for your project that includes the standard Flutter `assets` 
+directory in addition to two children: `audio` and `images`.
 
 If using the following example code:
-
 ```dart
+void main() {
```
spydon, line 8: Is it complaining without `main` here?

packages/flame_test/pubspec.yaml
```diff
@@ -2,13 +2,15 @@ name: flame_test
...
+publish_to: none
 
 environment:
   sdk: ">=2.14.0 <3.0.0"
   flutter: ">=2.5.0"
 
 dependencies:
-  flame: ^1.0.0
+  flame:
+    path: ../flame
```
spydon, line 13: These can be set back to what they were before once this PR is rebased, now when #1354 is merged.

doc/flame/structure.md
st-pasha: Yes, it complains in pretty much every snippet, I just didn't have a stomach to fix them all :)
I think the analyzer in general only accepts snippets that can be valid dart files. So, since a statement like `FlameAudio.play('explosion.mp3');` can only appear inside a function/method, it won't be happy until there's some function there.

doc/other/other.md
```diff
@@ -0,0 +1,14 @@
+# Other Modules
+
+```{eval-rst}
+.. toctree::
+    :hidden:
+    
+    Util           <util.md>
+    Widgets        <widgets.md>
```
luanpotter, line 8: these are part of flame-core, should they be on the main flame section?

luanpotter: Also debugging. They could be under flame > other
but then this "other" means: other packages

doc/_sphinx/theme/layout.html
```diff
@@ -0,0 +1,152 @@
...
+</div>
+
+<div class="main-area">
+  <div class="document-wrapper">
+    <div class="document" role="main">
+      {% block body %} {% endblock %}
+    </div>
+    <div class="copyright">
+      The content on this page is licensed under the <a href="https://creativecommons.org/licenses/by/4.0/">CC BY 4.0 License</a>,
+      and code samples under the <a href="https://opensource.org/licenses/MIT">MIT License</a>.
```
luanpotter, line 113: should we just link to our own license file here?

erickzanardo (approved):

doc/other/other.md
st-pasha: Sure, I can move those.

st-pasha:
> I think in an ideal world we could consider moving stuff that is only used for building the docs site to flame-engine/flame-docs-site, and keep only the docs themselves here (to be browsed via github). also, then, things like fontawesome could be downloaded from npm by the script that puts everything together. of course that would have to be a long term change not related to this PR

I believe a more pressing issue, in terms of future development, is to create a GitHub action to build the documentation with every PR. This would ensure that nobody can accidentally commit broken links or malformed Markdown into the docs.

We may even be able to find a Markdown linter and run through that.

spydon (approved):

st-pasha:
That content remains to be written. It should be like a high level overview of the module.

PR #1368, opened 2022-02-11
feat: exporting cache classes

erickzanardo:
# Description

Flame has a couple of useful classes that can be used to cache stuff in memory, but they were internal and not exported.

I plan on using these classes to help with perfomance on the flame svg package, so this PR simply move those classes to a more generic folder and export them and fix the places that was importing those classes.

## Checklist

## Breaking Change

  e.g. `feat!:`, `fix!:`).

## Related Issues

*Replace this paragraph with a list of issues related to this PR from the [issue database].
Indicate which of these issues are resolved or fixed by this PR, i.e. Fixes #xxxx*


st-pasha (commented):
1. Should the AssetsCache class and Images also be moved into the `cache` folder?

2. Looks like there is a bug in MemoryCache: when `setValue` is called but the key is already there, then the value won't be set.

packages/flame/lib/src/cache/cache.dart
```diff
@@ -0,0 +1,2 @@
+export 'memory_cache.dart';
+export 'value_cache.dart';
```
spydon, line 2: Is this file needed? Can't they be exported directly in the other file?

spydon (approved):

luanpotter (approved):
I'm always weary of hiding things from users, you never know what can be useful!

packages/flame/lib/src/cache/cache.dart
erickzanardo: They could, but I think it is good to have this barrel file, so we can use it internally.

I don't mind removing it if you feel too strong about it though.

spydon: I don't know how it usually is done, but can't we just use the other exported file internally?

erickzanardo:
> 1. Should the AssetsCache class and Images also be moved into the `cache` folder?
> 2. Looks like there is a bug in MemoryCache: when `setValue` is called but the key is already there, then the value won't be set.

1. I don't think so, they are more related to assets than cache
2. I think we did that intentionally, can't remember why now, if anyone feel this needs to be changed it could be addressed in a different PR

packages/flame/lib/src/cache/cache.dart
erickzanardo: Yeah, but having a barrel we can import everything related to cache on a single import, otherwise we will need to individually import each file

spydon: If the exports moved to `packages/flame/lib/cache.dart` and this file removed, wouldn't it have the same effect?

spydon: Or shouldn't the externally exported files be used internally? There is something I'm missing I think. :sweat_smile:

erickzanardo: > If the exports moved to `packages/flame/lib/cache.dart` and this file removed, wouldn't it have the same effect?

Yes! And I would honestly prefer that. But we have been doing relatively imports internationally, so that is why I created this internal barrel

st-pasha:
>> Should the AssetsCache class and Images also be moved into the cache folder?
>
> I don't think so, they are more related to assets than cache

Still, isn't it a **cache** of assets, or a **cache** of images?

I personally find directory name `assets` to be somewhat confusing, because in all other cases that directory would contain the assets of the project, i.e. images, sprites, animations, etc. But in our case the `assets` directory contains code, so it's confusing.

PR #1376, opened 2022-02-16
docs: Added tutorial for creating a bare Flame project

st-pasha:
# Description

Tutorial for how to start a new Flame project. The tutorial is targeted towards people who may have little or no experience with Flutter or Android development.

## Checklist

## Breaking Change


doc/tutorials/bare_flame_game.md
```diff
@@ -0,0 +1,186 @@
...
+The default project created by Flutter is not very useful for making a Flame
+game, so we should get rid of it.
+
+First, open the file `pubspec.yaml` and replace it with the following code
+(adjusting the `name` and `description` to match your project):
+
+```yaml
+name: syzygy
+description: Syzygy flame game
+version: 0.0.0
```
spydon, line 97: 0.1.0

st-pasha: `0.0.0` is symbolic of this being the "zero state" project. But ultimately it's up to the developer to decide on how to do versioning for their game.

st-pasha:
> but I think we should focus the rest of the tutorials on https://tutorials.flame-engine.org/ so I don't think we need a full tutorials section in the docs.

What is the benefit of having tutorials on a separate website? I feel they are an important part of the documentation, so should be kept on the main documentation site.

doc/tutorials/bare_flame_game.md
spydon: I thought `0.1.0` was the default from `flutter create`, but it's `1.0.0+1`, I don't see a reason to change from the default when it is a game and not a library?

spydon:
> > but I think we should focus the rest of the tutorials on https://tutorials.flame-engine.org/ so I don't think we need a full tutorials section in the docs.
> 
> What is the benefit of having tutorials on a separate website? I feel they are an important part of the documentation, so should be kept on the main documentation site.

The tutorial site is much more interactive than the docs are currently, it could be nice to integrate them eventually.

doc/tutorials/bare_flame_game.md
st-pasha: To me, `1.0.0` sounds like the first version of the product. So, if the intended product is the counting app (as `flutter create` makes), then it's perhaps reasonable to call it version 1. But if the intended product is a Flame game, then we need to remove the parts that make the counting app, and start from a clean slate. That clean slate is not a finished product, not even a started product, so I think it deserves version `0.0.0`. But then of course the developer is free to choose their own versioning scheme.

This version string will need to be updated when it's time to publish the game, but until then I feel designating it as `1.0.0+1` is a bit inappropriate.

It would be great to have a separate tutorial for the publishing process, which could go over these questions in more detail.

st-pasha:
> The tutorial site is much more interactive than the docs are currently, it could be nice to integrate them eventually.

Yes, that was my thinking too. If we could have a running Flame app on tutorials.flame-engine.org, then there is no reason why we can't make the same app run within the docs site, for example via an iframe.

doc/tutorials/bare_flame_game.md
spydon: It creates it with the same version when you run it with `flutter create . --template=skeleton` (which definitely is not something done), but I agree, it is a strange version to start on so let's go with `0.0.0`.

> It would be great to have a separate tutorial for the publishing process, which could go over these questions in more detail.

A publishing tutorial is probably something that would fit better in the docs than on the tutorials page too.

erickzanardo (approved):

PR #1385, opened 2022-02-19
refactor: Parent change and component removal logic

### LINKED ISSUE #1380: Deprecate `shouldRemove`
Currently there are two different ways to remove a component: either by calling the `remove()` method on its parent, or by setting the component's `shouldRemove` flag to true. These two approaches use different execution paths, which is error-prone.

I suggest to refactor this functionality so that:
- both methods of removal worked the same way;
- deprecate the `shouldRemove` flag (this is efficiency improvement: we won't have to check this flag on every game tick, and it reduces memory size of `Component`);
- Use Component's `LifecycleManager` to handle deletion instead of the `ComponentSet` -- this cleanup improves encapsulation of the Component class as ComponentSet will no longer need to handle this aspect of lifecycle.

st-pasha:
# Description

This PR further cleans up lifecycle management within `Component` by refactoring the handling of component removal. Specifically,
 - `Component.remove()` and `Component.removeFromParent()` are now the main methods for removing a component from its parent.
 - `Component.shouldRemove` flag is now deprecated. Instead of setting this flag, one should simply call `component.removeFromParent()` instead.
 - `ComponentSet.remove()` was marked as internal, so that removing a component directly from the component set is no longer allowed (similar how adding components to a component set is also not allowed) -- the `ComponentSet`, even though accessible to the user, should be considered a read-only container.
 - Removal is now handled through the component's lifecycle manager instead of via the ComponentSet. This results in better encapsulation and less memory usage in cases when nothing needs to be removed.
 - Changing a parent is now also handled via a lifecycle queue. This allows us to remove the `Component.nextParent` property.
 - Added internal lifecycle state `.removing`.
 - `Component.onRemove()` is no longer `@mustCallSuper`. This brings it in line with `onLoad` and `onMount`; derived components that implement `onRemove` and are designed to be derived further are encouraged to mark their `onRemove` methods as `@mustCallSuper`.

## Checklist

## Breaking Change

## Related Issues

Closes #1380 
WIP for #1426


doc/flame/components.md
```diff
@@ -44,6 +34,11 @@ throughout the component's lifetime. This method will only run if the parent is
 If the parent is not mounted yet, then this method will wait in a queue (this will have no effect
 on the rest of the game engine). 
 
+The `respectCamera` variable can be overridden or set to `false` (defaults to `true`) to make the
```
spydon, line 37: Didn't we refactor away this variable? :thinking:

packages/flame/lib/src/components/component.dart
```diff
@@ -74,7 +76,16 @@ class Component {
   ///
   /// It will be checked once per component per tick, and if it is true,
   /// FlameGame will remove it.
-  bool shouldRemove = false;
+  @nonVirtual
+  bool get shouldRemove => _state == LifecycleState.removing;
+
+  /// Legacy component removal mechanism.
+  @nonVirtual
+  @Deprecated('Use `removeFromParent()` instead. This will be removed in 1.1.0')
```
spydon, line 84: If we remove it in `1.1.0` it will never have a release in-between where it is just deprecated, was that intentional?

packages/flame/lib/src/components/component.dart
```diff
@@ -570,16 +599,31 @@ class _LifecycleManager {
   /// finish loading in arbitrary order.
   final Queue<Component> _children = Queue();
 
+  /// Queue for removing children from a component.
+  ///
+  /// Components that were placed into this queue will be removed from [parent]
+  /// when the pending events are resolved.
+  final Queue<Component> _dying = Queue();
```
spydon, line 606: I think this should have a clearer name to better reflect the state that the components are in.
Like always, hard with naming, maybe `_toBeRemoved`, `_removals` or something?

spydon (commented):
Lgtm, just some minor comments!

packages/flame/lib/src/components/component.dart
st-pasha: Haven't you argued for exactly this approach in #1316? Ever since then this is how we've been handling deprecation (searching for `@Deprecated` in the code shows that in all instances we're announcing deprecation in 1.1.0).

It's not too late to change the policy though and revisit our current deprecations.

doc/flame/components.md
st-pasha: Lol, true, this variable is now gone too. I'll remove this paragraph then

packages/flame/lib/src/components/component.dart
st-pasha: I like `_removals`

packages/flame/lib/src/components/component.dart
spydon: I'll have a check with the team on the meeting tomorrow, but I think for these cases that are easy to put a deprecation marker on and at the same time have the new functionality it would be nice to have it out for one release.

erickzanardo: Since this is a setter now, I don't see the need to deprecate and eventually remove it, we could keep it.

I see value having this attribute it allows things like

'''dart
comp.shouldRemove = position.x < 0
'''

doc/flame/components.md
spydon: Very out of the scope for this PR, but maybe you could remove this line mentioning it too?
https://github.com/flame-engine/flame/blob/main/packages/flame/lib/src/game/mixins/game.dart#L202

packages/flame/lib/src/components/component.dart
st-pasha: > I don't see the need to deprecate and eventually remove it, we could keep it.

I feel that a setter that works as a method is a bit of an anti-pattern. If you think that this is a useful method to have (and it kinda does look like a useful thing), then we could replace the setter with an actual method such as
```dart
void removeIf(bool shouldRemove) {
  if (shouldRemove) removeFromParent();
}
```

erickzanardo: I like the removeIf method, but idk, this is a breaking change that we could avoid, so I will have to put some thought on this

spydon: I see it like this:

`shouldRemove`
Pro: doesn't have to be called manually in `update`.
Con: adds another way to remove.

I don't see the use for `removeIf`, the `if` logic could just be directly in the update-loop in that case?

I think I'm leaning towards the option of deprecating it to minimize the amount of ways to do removals, but I don't have any strong opinion on the matter.

spydon (approved):

packages/flame/lib/src/components/component.dart
spydon: We still haven't decided whether `shouldRemove` should be deprecated or not, we'll probably have to vote about it.
But meanwhile, this method does not do the same thing as it did before this PR. Previously you could for example override the getter with logic for removal and you could also set it to true later in the same tick even if it had been set to false earlier.

st-pasha: @spydon The use case that you're describing, even though it's very rare, is still possible to implement with this PR. All you need is to create your own flag `toRemove`, then set it to true/false as many times as needed, then at the end of the `update()` method check the flag and call `.removeFromParent()` if it's true. That would do exactly what the current functionality is doing, but explicitly.

In general, I feel explicit deletion is better than implicit because 
 1) it's more clear what is going on and when the deletion happens;
 2) fewer things to document for us, fewer things for the user to get confused about;
 3) components that don't need to be removed don't pay the cost of querying the flag on every game tick.

spydon: I'm also on the side of deletion. But we should still try to avoid a direct breaking change here if we can.

> The use case that you're describing, even though it's very rare, is still possible to implement with this PR

It's not about whether you can or can't do it, it's that it changes the current behavior. So right now it will be a breaking change even though it is just deprecated. Might not matter though, I mean we just have to call it.

> The use case that you're describing, even though it's very rare.

It's not that rare sadly (overriding the getter), I've seen several projects with that logic embedded in the getter (I think it must have been in some external tutorial at some point).

```dart
@override
bool get shouldRemove => "not within screen" or "life below 0";
```

st-pasha: Ah, I see your point. My "very rare" comment was about deleting something and then deciding to undelete it in the very same tick. But what you were thinking is overriding the `shouldRemove` getter. That would indeed be more common, so we should probably call it out explicitly.

Right now the getter is marked as `@nonVirtual`, so the users will at least be aware that something in their code needs to change.

spydon: It was decided in a very even vote today that `shouldRemove` shouldn't be deprecated.

st-pasha: @spydon Is there any specific reason? The cost of `shouldRemove` is that every component has to check this flag on every single game tick.

spydon: I voted for removing it, but the argument for keeping it was that we should minimize breaking changes. Just checking a bool shouldn't be a performance concern though, if you can show that it is I'm sure they'd change their minds, but I would also be very worried about Dart as a language if it was true. 😅

st-pasha: It's just not a very nice property. For example, if you override its getter then it'll become inconsistent with the setter. Maybe we should at least override the setter then? It's equivalent to the `remove` method anyway.

spydon: > Maybe we should at least override the setter then?

Sounds good to me, how were you thinking that the override would look?

st-pasha: I meant, deprecate the setter. Because having both the setter and an overridable getter makes no sense to me.

spydon: Ah, I'll ask the team.

spydon: Can't we just set `@nonVirtual` on the setter?
The team didn't think it made much sense to only keep the getter since that would only be useful when overridden then.

st-pasha: In order to understand whether any kind of API is good or not, we need to think about how we're going to document it. And by "document" I don't mean to write a one-liner that just repeats the name of the function like "set whether this component should be removed" -- that's not a proper documentation. A great documentation follows the principles outlined here: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#writing-prompts-for-good-documentation

So, a good documentation for setter will look like this:
```
/// Setting this property to true will remove the component soon (before the next 
/// game cycle), thus it is equivalent to `component.removeFromParent()`. However,
/// if you override `.shouldRemove` getter, then there is no guarantee that setting
/// this property to true will have any effect.
```

And when you spell it out, it kinda becomes clear that it's not a very good API. Here, by "not good" I mean that it (1) has no real purpose (the same functionality can be achieved via more straightforward call to `.remove`), (2) does not work reliably.

But ok, if we want to keep it, then let's keep it.

spydon: I agree that it is not optimal, but the vote is done. It will be reliable though, since they'll be marked as `@nonVirtual` and can't be overridden.

st-pasha: Alright, i marked the getter and the setter `.shouldRemove` as `@nonVirtual`, and removed deprecation annotation.

PR #1405, opened 2022-02-26
feat!:  Added getImageLayer to flame_tiled

munsterlander:
# Description

This is a very simple add to flame_tile to allow getting image layers.

## Checklist

- [X] I have updated/added relevant examples in `examples`. 

## Breaking Change

- [X] Yes, this is a breaking change. (Indicate it in the [Conventional Commit] prefix with a `!`,
  e.g. `feat!:`, `fix!:`).


ufrshubham (commented):
I was thinking should we make it a generic method

```dart
T getLayer<T extends Layer>(String name){
    final t = map.layers.firstWhere((layer) {
      return layer is T && layer.name == name;
    });
    return t as T;
}
```

munsterlander:
Which I concur with and if @spydon wants to implement this, then we can go in and change all the docs, examples, and tests.  It is a breaking change though, so I don't know if a phased approach is better, i.e. have all 5 methods co-exist and mark the specific implementations as deprecated (even though 3 would be new), have just 2 methods and mark the old as deprecated with the generic one there, or just break it and only have 1.

munsterlander:
Documenting implementation in case there is a question: `getLayer<ObjectLayer>(”myObjectLayer");`

st-pasha (commented):
1. Looks like the three functions can be combined into a single template function.
2. Please use doc-comments (`///`) for documentation. You may read more about them in [this guide](https://dart.dev/guides/language/effective-dart/documentation#doc-comments).
3. Consider what would happen when the map doesn't have the requested layer.
4. Perhaps add a few tests?

munsterlander:
@st-pasha Yes, ideally it can return null to avoid errors with calls to non-existent layers. Ok, I will update this pr with the relevant change. Do we need to leave the existing method and just label it as deprecated?

st-pasha:
> Do we need to leave the existing method and just label it as deprecated?

yes, breaking changes should be avoided if possible. So, leaving the existing method as-is and marking it as `@Deprecated()` is the right idea.

munsterlander:
Ok, I think that does it.  We added the code, tests, examples, doc block, and docs.  I updated the PR to indicate blocking and remove unnecessary comments.

spydon (approved):

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -202,10 +202,27 @@ class RenderableTiledMap {
...
+  /// Returns a layer of type [T] with given [name] from all the layers
+  /// of this map. If no such layer is found, null is returned.
+  T? getLayer<T extends Layer>(String name) {
+    T? foundLayer;
+
+    try {
+      foundLayer = map.layers.firstWhere((layer) {
+        return layer is T && layer.name == name;
+      }) as T;
+    } on StateError {
```
st-pasha, line 222: Usually, it is not recommended to use exceptions for anything other than exceptional situations. This is because creating an exception object is a fairly inefficient process.
So, instead of a try-catch this could be rewritten as a simple for-loop, which will be simpler code too.

ufrshubham: This is now resolved with [70963eb](https://github.com/flame-engine/flame/pull/1405/commits/70963eb3db1c3d70dd59415b02cbcda0ce9c5eae)

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -202,10 +202,27 @@ class RenderableTiledMap {
...
+    for (var i = 0; i < map.layers.length; ++i) {
+      final layer = map.layers.elementAt(i);
+      if (layer is T && layer.name == name) {
+        foundLayer = layer;
+        break;
+      }
+    }
+
+    return foundLayer;
+  }
```
spydon, line 227:
```suggestion
  T? getLayer<T extends Layer>(String name) {
    T? foundLayer;

    for (final layer in map.layers.length {
      if (layer is T && layer.name == name) {
        foundLayer = layer;
        break;
      }
    }

    return foundLayer;
  }
```

ufrshubham: Thanks @spydon! I wasn't even aware of `firstWhereOrNull` 😄

Issue #1417, opened 2022-03-04
Adding hasMount future to Component


erickzanardo:
# Problem to solve

Right now, Flame components have an attribute to know is that component is already mounted, but there is no way for someone to await for its mount to happen.

# Proposal

This could be simply solved by added a completer on the component base class, expose its future and complete it once it has finished mounted.

# More information

spydon:
Should there be completers for all lifecycle events maybe?
The name should probably be `hasMounted`?
Other than that I think it sounds good. 🙂

alestiago:
@spydon, sounds superb! 

@erickzanardo and I did a POC to see if this was feasible. And turns out we used the same name "hasMounted".

So definitely agree! 🚀

spydon:
> @spydon, sounds superb!
> 
> @erickzanardo and I did a POC to see if this was feasible. And turns out we used the same name "hasMounted".
> 
> So definitely agree! 🚀

Nice! Did you also try for other lifecycle events?

st-pasha:
We need to be careful not to add any unnecessary data members into the core `Component` class though.

spydon:
> We need to be careful not to add any unnecessary data members into the core `Component` class though.

True! Maybe this could be added with a mixin?

erickzanardo:
I am going to start with the hasMounted which is something that already has use for, if we identify more lifecycle that would benefit from it, we can do follow ups.

st-pasha:
what if we modify component's `.add()` so that it returns a future that returns once the component is mounted, instead of only loaded?

erickzanardo:
> what if we modify component's `.add()` so that it returns a future that returns once the component is mounted, instead of only loaded?

That would be confusing IMO, adding and mounting the component are different part of the lifecycle.

st-pasha:
> That would be confusing IMO, adding and mounting the component are different part of the lifecycle.

You probably meant loading and mounting, because adding is not part of a lifecycle -- it's just a method that schedules lifecycle events of loading and mounting. And if the component is already loaded, then it's just mounting.

Of course, right now `.add()` returns the future from the loading part, and changing that to the mounting future may have bad consequences since mounting happens after a considerable delay (or sometimes never). Still, there could be a room for an optional flag to `.add()` so that it would return the mounting future instead. So that the user may write
```dart
await component.addToParent(game, waitToMount: true);
```

alestiago:
I'm somewhat against the flag for the `add`. 

[...]

PR #1418, opened 2022-03-04
feat: adding has mounted to component

erickzanardo:
# Description

Solves #1417 

## Checklist

## Breaking Change


packages/flame/lib/src/components/component.dart
```diff
@@ -53,6 +54,8 @@ class Component {
   bool get hasChildren => _children?.isNotEmpty ?? false;
   ComponentSet? _children;
 
+  var _mountCompleter = Completer<void>();
```
spydon, line 57: We'll have to somehow add this without initializing it in the component, maybe with a mixin? Because we don't want more objects than necessary initialized inside component since it affects performance a lot.

erickzanardo: How can a simple instantiation affect the perfomance a lot? 🤔

erickzanardo: I am not against a mixin though, but I just want to understand how come this would affect perfomance that much

spydon: Because you have a lot of components in your game since it is the basic building block, that's why the fields that can in `Component` are lazily initialized.

erickzanardo: Interesting, I am still curious on how this should impact perfomance, I mean we are instantiating a lot of things all the time in flame, like our `toRect` method and so on.

But either way, it will be simple to make this lazy loaded, I will add some changes regarding that

spydon (approved):

packages/flame/lib/src/components/component.dart
st-pasha: Maybe it can live inside the `_LifecycleManager`? That way it can safely disappear once it is no longer needed.

packages/flame/test/components/component_lifecycle_test.dart
```diff
@@ -47,6 +47,47 @@ void main() {
       );
     });
 
+    flameGame.test('component hasMounted completes', (game) async {
```
st-pasha, line 50:
```suggestion
    testWithFlameGame('component hasMounted completes', (game) async {
```

packages/flame/test/components/component_lifecycle_test.dart
```diff
@@ -47,6 +47,47 @@ void main() {
...
+      final component = _MyComponent();
+      await game.add(component);
+      final hasMounted = component.hasMounted;
```
st-pasha, line 53: I presume it is not possible to await for this future, since it cannot complete without the game loop running its course -- hence this would become an infinite cycle.

Perhaps it would be useful to have at least some test where we await `component.hasMounted`, a test case that would mimic the intended use case for this feature.

packages/flame/lib/src/components/component.dart
spydon: That was the suggestion @alestiago had too,I think it sounds like a good idea.

packages/flame/lib/src/components/component.dart
```diff
@@ -203,6 +206,16 @@ class Component {
   /// ```
   void onMount() {}
 
+  /// A future that will complete once the component is mounted on its parent
+  ///
+  /// Note: Calling this method after the component has already mounted will
```
alestiago, line 211: @erickzanardo I wonder if it could be possible to avoid this being a never-ending future when it has already mounted.

If I were using this API and read `hasMounted` I'll expect to receive a boolean value from it. I wonder if `whenMounted` or `onMounted` is a better name for it. This leads me to think that perhaps passing a callback to these is a more intuitive API (reminds me of hooks).  Something as `whenMounted(() => print('hello'))`.

In the case that the `Component` has already been mounted I think I'll expect the callback to never be reached.

wdyt?

spydon: I don't know which way I prefer. 🤔 
For collision detection we have just a nullable field where you can set the wanted callback, that could be an option too.

packages/flame/test/components/component_lifecycle_test.dart
erickzanardo: All of this file is using an instance, better keep with the flow

packages/flame/lib/src/components/component.dart
erickzanardo: Actually, this is very easy to do, I can just check if it is already mounted, and then just return a `Future.value` instead of the future of the completer, I am gonna change this

packages/flame/test/components/component_lifecycle_test.dart
erickzanardo: I double checked and the future does completes, otherwise the test would fail. I changed to expectLater to be sure, but even the original code was working.

packages/flame/lib/src/components/component.dart
erickzanardo: I tried adding it to `_LifecycleManager`, but I didn't like very much. When I did, the logic was half on that class and half on the `Component`, for now, I think we should leave it on the Component, wdyt?

spydon: Interesting, what logic had to stay in the `Component`?

packages/flame/lib/src/components/component.dart
```diff
@@ -203,6 +206,17 @@ class Component {
   /// ```
   void onMount() {}
 
+  /// A future that will complete once the component is mounted on its parent
+  Future<void> get hasMounted {
+    if (isMounted) {
+      return Future.value();
```
spydon, line 212: Nice solution!

packages/flame/test/components/component_lifecycle_test.dart
st-pasha: But can you do `await component.hasMounted;`? I mean, isn't that how this future is supposed to be used?

erickzanardo: Yeah, you can await it, the `completes` matcher makes sure that it finishes.

I could change my test to make use of the `await` keyword, but would make the test uglier IMO

packages/flame/lib/src/components/component.dart
```diff
@@ -403,6 +418,7 @@ class Component {
     if (existingChild || _state == LifecycleState.removed) {
       onGameResize(findGame()!.canvasSize);
     }
+    _mountCompleter?.complete();
```
st-pasha, line 421: Set it to null here?

packages/flame/lib/src/components/component.dart
erickzanardo: The getter for the hasMounted now have to use the isMounted, which lives on the component, that was the bigger logic.

packages/flame/lib/src/components/component.dart
```diff
@@ -203,6 +206,17 @@ class Component {
   /// ```
   void onMount() {}
 
+  /// A future that will complete once the component is mounted on its parent
+  Future<void> get hasMounted {
```
alestiago, line 210: I don't know why, but `hasMounted` doesn't convince me as the method name. If I ever read this I'll expect to get a boolean value from it not a `Future<void>`. @erickzanardo suggested `mounted` which I somewhat prefer over `hasMounted`.

In addition, I have a question on why do we think that using a `Completer` is better than registering callbacks (for example, `whenMounted(callback)`?

erickzanardo: I like the `mounted` name! @spydon what do you think?

About the future vs callback, I think it have a nicer API, and actually, by using a future, we can let the user use a callback if they prefer, by doing `component.mounted.wheComplete(fn)`, so we can have the best of both worlds.

packages/flame/lib/src/components/component.dart
st-pasha: If we simply replace `_mountCompleter` -> `lifecycleManager._mountCompleter`, then nothing needs to be moved anywhere. 

In general, the job of the `LifecycleManager` class is to handle the temporary state that needs to be kept briefly while the component goes through lifecycle transitions; so `mountCompleter` seem to fit perfectly there.

erickzanardo: > If we simply replace _mountCompleter -> lifecycleManager._mountCompleter, then nothing needs to be moved anywhere.

We are using the isMounted method to check if we return the completer future or not, if we move the completer to the lifecycle manager I believe the isMounted should go there too, otherwise we will end up with logic spread across these two classes.

I don't mind moving isMounted there, but then we will end up with two "proxy" methods forwarding methods from the life cycle manager to the component, not sure what is best.

packages/flame/lib/src/components/component.dart
alestiago: @erickzanardo true. Missed that API! 

I think it's nice if we have it as `mounted`. I love how it reads:

`component.mounted.whenComplete`
and 
`await component.mounted`

packages/flame/test/components/component_lifecycle_test.dart
st-pasha: The point is, right now we have no tests to check whether this future is actually awaitable.

erickzanardo: How come? This test is doing exactly that 🤔

Issue #1429, opened 2022-03-09
Allow prefix of the Images instance on the Game to be customisable


erickzanardo:
# Problem to solve

Flame `Images` class has a default prefix to `assets/images`, which can be changed upon instantiation.

`Game` bundles an `Images` instance, but it is directly instantiated upon declaration, which makes it impossible to customise the prefix of that instance.

So users that want to use a different prefix can't use the bundled `images` instance on the `Game` and have to create and manually manage another instance.

# Proposal
We should provide a way for that prefix to be changed.

I believe that a simple solution would be to add an attribute to game, called `imagesPrefix = 'assets/images`, which users could override.

That is a simple solution, which will be retro compatible, but also open to other ideas 

# More information

wolfenrain:
As discussed in our group chat, I am for having it more customizable but not sure if adding another "floating" attribute is the way to go. I currently do not have an alternative tho.

I'll noodle on this as well and see if I can find a "cleaner" solution that is backwards compatible, but it wont be the end of the world for me if we go with this route ;-).

st-pasha:
What if we simply make `Images.prefix` a settable (non-final) property? Then a user would be able to write at the beginning of the game something like
```dart
void main() {
  final game = MyGame();
  game.images.prefix = 'assets/pictures/';
  runApp(GameWidget(game: game));
}
```

erickzanardo:
Yeah, we raised that on other forum, but @wolfenrain was worried that this could be more prone to user error, any more context you want to bring about it @wolfenrain ?

st-pasha:
My thinking was that it would be a setter, which would do basic checks on the validity of the passed value.

st-pasha:
I guess this can be closed now.

PR #1433, opened 2022-03-10
feat: allowing changing of the images prefix and allowing empty prefixes

erickzanardo:
# Description

Solves #1429 

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/lib/src/assets/images.dart
+++ b/packages/flame/lib/src/assets/images.dart
@@ -11,16 +11,20 @@ import '../flame.dart';
 
 class Images {
   Images({String prefix = 'assets/images/'})
-      : assert(prefix.endsWith('/')),
+      : assert(prefix.isEmpty || prefix.endsWith('/'), _assertMessage),
         _prefix = prefix;
 
   String _prefix;
   final Map<String, _ImageAssetLoader> _loadedFiles = {};
 
+  static const _assertMessage = 'Prefix must be empty or end with a "/"';
+
   set prefix(String value) {
-    assert(value.endsWith('/'));
+    assert(prefix.isEmpty || value.endsWith('/'), _assertMessage);
     _prefix = value;
   }
+
+  String get prefix => _prefix;
 
   /// Adds an [image] into the cache under the key [name].
   void add(String name, Image image) {
--- a/packages/flame/lib/src/assets/images.dart
+++ b/packages/flame/lib/src/assets/images.dart
@@ -10,11 +10,17 @@ import 'package:flutter/services.dart';
 import '../flame.dart';
 
 class Images {
-  Images({this.prefix = 'assets/images/'})
-      : assert(prefix.isEmpty || prefix.endsWith('/'));
+  Images({String prefix = 'assets/images/'})
+      : assert(prefix.endsWith('/')),
+        _prefix = prefix;
 
-  final String prefix;
+  String _prefix;
   final Map<String, _ImageAssetLoader> _loadedFiles = {};
+
+  set prefix(String value) {
+    assert(value.endsWith('/'));
+    _prefix = value;
+  }
 
   /// Adds an [image] into the cache under the key [name].
   void add(String name, Image image) {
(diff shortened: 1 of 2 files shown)


packages/flame/lib/src/assets/images.dart
```diff
@@ -10,12 +10,18 @@ import 'package:flutter/services.dart';
 import '../flame.dart';
 
 class Images {
-  Images({this.prefix = 'assets/images/'})
-      : assert(prefix.isEmpty || prefix.endsWith('/'));
+  Images({String prefix = 'assets/images/'})
+      : assert(prefix.endsWith('/')),
```
erickzanardo, line 14: Note: empty prefixes should be allowed, that way we can let users use the full path if they want.

This is useful when using `flutter_gen` for example, which generates type safe paths which always include the whole path.

packages/flame/lib/src/assets/images.dart
```diff
@@ -10,12 +10,18 @@ import 'package:flutter/services.dart';
...
+        _prefix = prefix;
 
-  final String prefix;
+  String _prefix;
   final Map<String, _ImageAssetLoader> _loadedFiles = {};
 
+  set prefix(String value) {
+    assert(value.endsWith('/'));
```
spydon, line 21: An assert message here maybe?

spydon: Or we just add `/` if it doesn't exist?

spydon (approved):

packages/flame/lib/src/assets/images.dart
erickzanardo: could be, I just kept the previous behaviour, but I don't think automatically adding the `/` could be considered a breaking change right? Because if anyone would be passing without, they would be receiving an error already

packages/flame/lib/src/assets/images.dart
luanpotter: > empty prefixes should be allowed

you say that, yet you are removing the check that was allowing empty prefixes...

packages/flame/lib/src/assets/images.dart
```diff
@@ -10,12 +10,18 @@ import 'package:flutter/services.dart';
 import '../flame.dart';
 
 class Images {
-  Images({this.prefix = 'assets/images/'})
-      : assert(prefix.isEmpty || prefix.endsWith('/'));
+  Images({String prefix = 'assets/images/'})
+      : assert(prefix.endsWith('/')),
+        _prefix = prefix;
```
luanpotter, line 15: if you use the setter here you can have the assert on only one place?

packages/flame/lib/src/assets/images.dart
luanpotter: should you clear the cache here? otherwise you would get weird inconsistencies for the existing keys when they invalidate

packages/flame/lib/src/assets/images.dart
```diff
@@ -10,12 +10,18 @@ import 'package:flutter/services.dart';
...
 
-  final String prefix;
+  String _prefix;
```
luanpotter, line 17: can we have a getter too? just for completeness

packages/flame/lib/src/assets/images.dart
erickzanardo: True! I totally got this the wrong way around, good catch

packages/flame/test/assets/images_test.dart
```diff
@@ -37,5 +37,11 @@ void main() {
         images.length,
       );
     });
+
+    test('throws when setting an invalid prefix', () {
+      final images = Images();
+
+      expect(() => images.prefix = 'adasd', throwsAssertionError);
```
luanpotter, line 44: we really need a simple method to check the assertion message, otherwise checks like these can be really brittle

packages/flame/lib/src/assets/images.dart
erickzanardo: unfortunately dart don't allow that :(

packages/flame/lib/src/assets/images.dart
erickzanardo: > should you clear the cache here? otherwise you would get weird inconsistencies for the existing keys when they invalidate

Good question, the pro of not clearing the cache is that if the user changes the prefix back again, the cache is already there, not sure what would be best in this case

packages/flame/test/assets/images_test.dart
erickzanardo: We have, but I am stupid and forgot to use it, changed

PR #1439, opened 2022-03-10
fix: Don't use debug rendering by default in BodyComponent

spydon:
# Description

Since debug rendering now renders on top of everything (since ages), `BodyComponent` shouldn't hijack the debug render method for its normal rendering.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame_forge2d/lib/body_component.dart
+++ b/packages/flame_forge2d/lib/body_component.dart
@@ -14,18 +14,17 @@ import 'sprite_body_component.dart';
 /// it is a good idea to turn on [debugMode] for it so that the bodies can be
 /// seen
 abstract class BodyComponent<T extends Forge2DGame> extends Component
-    with HasGameRef<T> {
+    with HasGameRef<T>, HasPaint {
   static const defaultColor = Color.fromARGB(255, 255, 255, 255);
   late Body body;
-  late Paint paint;
 
-  /// [debugMode] is true by default for body component since otherwise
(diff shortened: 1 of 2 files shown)


packages/flame_forge2d/lib/body_component.dart
```diff
@@ -14,18 +14,17 @@ import 'sprite_body_component.dart';
 /// it is a good idea to turn on [debugMode] for it so that the bodies can be
 /// seen
 abstract class BodyComponent<T extends Forge2DGame> extends Component
-    with HasGameRef<T> {
+    with HasGameRef<T>, HasPaint {
```
erickzanardo, line 17: Excellent! I was going to suggest that but I forgot to raise the issue, glad you did it here!

erickzanardo (commented):
Overall LGTM 🚀, but there is a different issue now, children of body component are rendered upside down.

To reproduce this, try adding a `TextComponent` inside a body component.

st-pasha:
> but there is a different issue now, children of body component are rendered upside down.

Is this problem caused by this PR, or was it there before?

spydon:
> > but there is a different issue now, children of body component are rendered upside down.
> 
> Is this problem caused by this PR, or was it there before?

If it was added directly as a child to the `BodyComponent` I think the problem was already there before, I think it should be fixed now though (this whole mess really needs a refactoring though).

wolfenrain (approved):

erickzanardo (approved):

Issue #1456, opened 2022-03-16
Include `findNestedChildren` and `findNestedChild`


alestiago:
# Problem to solve
Currently one needs to define their custom logic to find if a component has a child in its subtrees that matches the given criteria. Particularly useful for tests where you don't want to make subcomponents public.

# Proposal
I think something, as follows, might be helpful:

<details>

<summary> Code example </summary>

```dart
extension ComponentX on Component {
  T findNestedChild<T extends Component>({
    bool Function(T)? condition,
  }) {
    T? nestedChild;
    propagateToChildren<T>((child) {
      final foundChild = (condition ?? (_) => true)(child);
      if (foundChild) {
        nestedChild = child;
      }

      return !foundChild;
    });

    if (nestedChild == null) {
      throw Exception('No child of type $T found.');
    } else {
      return nestedChild!;
    }
  }

  List<T> findNestedChildren<T extends Component>({
    bool Function(T)? condition,
  }) {
    final nestedChildren = <T>[];
    propagateToChildren<T>((child) {
      final foundChild = (condition ?? (_) => true)(child);
      if (foundChild) {
        nestedChildren.add(child);
      }

      return true;
    });

    return nestedChildren;
  }
}
```

The above is an `Extension` ideally, this would be within `Component` and with documentation that warns that this might be a  very expensive method. The above is just an example, function signatures can be modified to adapt to what's more convenient. For example, perhaps a `ComponentSet` is more desired than a `List`?
</details>

# More information

I'm open for better names, suggestions and discussion.
I'm interested in working on a PR for this.

st-pasha:
In Dart, iterables are lazy. That is, `descendants()` by itself will do almost no work -- just create an `Iterable` object and return it. An Iterable is an object that _knows_ how to retrieve its elements one-by-one, but it won't retrieve those elements unless explicitly asked. Then `descendants().firstWhere(condition)` will ask for those elements, one by one, and will stop as soon as it finds the first one that satisfies the condition. So there won't be any unnecessary computations involved. Similarly, `descendants().where(condition).toList()` will iterate over all the components only once, for each one testing whether it satisfies the condition, and if yes placing it into the final list.

Dan-Crane:
Oh! it very good proposal. I wanted to do the same thing this weekend, but you beat me to it) thank you❤️

```dart
Iterable<Component> get descendants sync* {
  if (!hasChildren) return;

  for (var element in children) {
    if (element.hasChildren) {
      yield element;
[...]

PR #1461, opened 2022-03-17
feat: Added method that returned descendants

Dan-Crane:
# Description
This PR makes it possible to get the descendants of a component

## Checklist

## Breaking Change

## Related Issues

https://github.com/flame-engine/flame/issues/1456#issue-1171350725

### DIFF
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -331,12 +331,14 @@ class Component {
     }
   }
 
-  /// An iterator that performs a deep search on all [children].
+  /// Recursively enumerates all nested [children].
   ///
-  /// This can be an expensive operation if a large nesting tree is used.
+  /// The search is depth-first, hence descendants are in postorder. In other
+  /// words, it explores the first child completely before visiting the next
+  /// sibling.
   ///
-  /// In order to filter [descendants] it is usually convenient to use 
-  /// [Iterable] lazy methods, such as [Iterable.where].
+  /// In order to filter descendants it is usually convenient to use [Iterable]
+  /// lazy methods, such as [Iterable.where].
   Iterable<Component> descendants({bool includeSelf = false}) sync* {
     if (includeSelf) {
       yield this;
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -331,7 +331,13 @@ class Component {
     }
   }
 
-  /// An iterator that returns all added components.
+  /// An iterator that performs a deep search on all [children].
+  ///
+  /// This can be an expensive operation if a large nesting tree is used.
+  ///
+  /// In order to filter [descendants]
+  /// it is usually convenient to use [Iterable] lazy methods,
+  /// such as [Iterable.where].
   Iterable<Component> descendants({bool includeSelf = false}) sync* {
     if (includeSelf) {
       yield this;
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -335,9 +335,8 @@ class Component {
   ///
   /// This can be an expensive operation if a large nesting tree is used.
   ///
-  /// In order to filter [descendants]
-  /// it is usually convenient to use [Iterable] lazy methods,
-  /// such as [Iterable.where].
+  /// In order to filter [descendants] it is usually convenient to use 
+  /// [Iterable] lazy methods, such as [Iterable.where].
   Iterable<Component> descendants({bool includeSelf = false}) sync* {
     if (includeSelf) {
       yield this;
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -346,7 +346,7 @@ class Component {
     if (!hasChildren) {
       return;
     }
-    
+
     for (final component in children) {
       yield component;
       if (component.hasChildren) {
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -331,6 +331,22 @@ class Component {
     }
   }
 
+  /// An iterator that returns all added components.
+  Iterable<Component> descendants({bool includeSelf = false}) sync* {
+    if (includeSelf) {
+      yield this;
+    }
+    if (!hasChildren) {
+      return;
+    }
+    for (final component in children) {
+      yield component;
+      if (component.hasChildren) {
+        yield* component.descendants();
+      }
+    }
+  }
+
   //#region Add/remove components
 
   /// Schedules [component] to be added as a child to this component.
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -344,6 +344,7 @@ class Component {
     if (!hasChildren) {
       return;
     }
+    
     for (final component in children) {
       yield component;
       if (component.hasChildren) {
--- a/packages/flame/test/components/component_test.dart
+++ b/packages/flame/test/components/component_test.dart
@@ -207,7 +207,7 @@ void main() {
     });
     testWithFlameGame(
       'length of descendants must be 0 '
-      'when it is called before the add method',
+      'when it is called before being loaded',
       (game) async {
         final component = Component()..add(Component()..add(Component()));
 
--- a/packages/flame/test/components/component_test.dart
+++ b/packages/flame/test/components/component_test.dart
@@ -228,8 +228,12 @@ void main() {
         (game) async {
           final component = Component()..add(Component()..add(Component()));
           await game.ensureAdd(component);
+          final descendants = game.descendants(includeSelf: true);
 
-          expect(game.descendants(includeSelf: true).length, 4);
+          expect(descendants.length, 4);
+          for (final component in descendants) {
+            expect(component.findGame() != null, true);
+          }
         },
       );
 
--- a/packages/flame/test/components/component_test.dart
+++ b/packages/flame/test/components/component_test.dart
@@ -41,6 +41,12 @@ class _ParentOnPrepareComponent extends _OnPrepareComponent {
   }
 }
 
+class _OrderCheckingComponent extends Component {
+  final int order;
+
+  _OrderCheckingComponent(this.order);
+}
+
 void main() {
   final prepareGame = FlameTester(() => _PrepareGame());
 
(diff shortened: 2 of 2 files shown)


packages/flame/test/components/component_test.dart
```diff
@@ -200,6 +200,56 @@ void main() {
...
+      (game) async {
+        final component = Component()..add(Component()..add(Component()));
+        await game.ensureAdd(component);
+
+        expect(game.descendants().length, 3);
+      },
+    );
+
+    flameGame.test(
+      'length must be equal to the number of added components including itself',
```
alestiago, line 216: Maybe this test name that follows "given-when-then" pattern? Same applies for other test names.
```suggestion
      'length must be equal to the number of added components '
      'when including itself',
```

packages/flame/lib/src/components/component.dart
```diff
@@ -319,6 +319,22 @@ class Component {
     }
   }
 
+  /// An iterator that returns all added components.
+  Iterable<Component> descendants({bool includeSelf = false}) sync* {
```
alestiago, line 323: I was thinking, for convenience only, having it generic as:
```
 Iterable<T> descendants<T extends Component>({bool includeSelf = false}) sync* 
```
But one could use `whereType`, although this is more verbose.

@st-pasha @Dan-Crane wdyt?

packages/flame/test/components/component_test.dart
```diff
@@ -200,6 +200,56 @@ void main() {
       );
     });
   });
+
+  group('Component descendants', () {
```
alestiago, line 204:
```suggestion
  group('descendants', () {
```

Should already be under the `group('Component')`.

packages/flame/test/components/component_test.dart
```diff
@@ -200,6 +200,56 @@ void main() {
...
+        expect(game.descendants().length, 0);
+      },
+    );
+
+    flameGame.test(
+      'length should not change '
+      'if [hasPendingLifecycleEvents] is true after adding',
+      (game) async {
+        final component = Component()..add(Component()..add(Component()));
+        await game.add(component);
```
alestiago, line 241:
```suggestion
        await game.ensureAdd(component);
```

For testing, I prefer `ensureAdd` over `add`, for consistency. Same applies for other.

packages/flame/test/components/component_test.dart
```diff
@@ -200,6 +200,56 @@ void main() {
...
+      (game) async {
+        final component = Component()..add(Component()..add(Component()));
+        await game.ensureAdd(component);
+
+        expect(game.descendants(includeSelf: true).length, 4);
+      },
+    );
+
+    flameGame.test(
+      'length must be 0 if [hasPendingLifecycleEvents] is true',
```
alestiago, line 226:
```suggestion
      'length must be 0 when [hasPendingLifecycleEvents] is true',
```

packages/flame/test/components/component_test.dart
```diff
@@ -200,6 +200,56 @@ void main() {
...
+        game.add(component);
+
+        expect(game.hasPendingLifecycleEvents, true);
+        expect(game.descendants().length, 0);
+      },
+    );
+
+    flameGame.test(
+      'length should not change '
+      'if [hasPendingLifecycleEvents] is true after adding',
```
alestiago, line 238:
```suggestion
      'when [hasPendingLifecycleEvents] is true after adding',
```

alestiago (commented):
Thanks for contributing! Left some minor comments and questions.

I didn't start working on the PR, however would be nice if next time you let us know in the issue that you are interested in working on the PR, to avoid having multiple people resolving (assigned) to the same issue. I'm happy to stick with these changes.

packages/flame/lib/src/components/component.dart
spydon: Better to have it consistent with `ancestors`

alestiago, line 323: I think some documentation on this would be helpful:

```suggestion
  ///
  /// In order to filter [descendants] it is usually convenient to use [Iterable]
  /// lazy methods, such as [Iterable.where].
  Iterable<Component> descendants({bool includeSelf = false}) sync* {
```

packages/flame/lib/src/components/component.dart
```diff
@@ -319,6 +319,22 @@ class Component {
     }
   }
 
+  /// An iterator that returns all added components.
```
alestiago, line 322: I think this can be improved. Here is a suggestion, not sure about it, but something along the lines could be nice.

```suggestion
  /// Deeply searches all [children].
```

packages/flame/test/components/component_test.dart
alestiago, line 204: @Dan-Crane can we add a test where `descendants` is called before being added?

alestiago (commented):
Forgot to mention some things, hope it's helpful.

packages/flame/test/components/component_test.dart
Dan-Crane: Im embarrassed, but can you show an example? I dont know English well🙈

packages/flame/lib/src/components/component.dart
Dan-Crane: You're right

packages/flame/lib/src/components/component.dart
Dan-Crane: Totally agree with you)

packages/flame/test/components/component_test.dart
Dan-Crane: yes, it will be returned 0 elements. I will add

packages/flame/test/components/component_test.dart
alestiago: This is very nitpicky, but I like to keep the test name consistent. On the testing limbo there are some conventions people usually go for. For example, the [given-when-then](https://en.wikipedia.org/wiki/Given-When-Then#:~:text=Given%2DWhen%2DThen%20\(GWT,words%20given%2C%20when%20and%20then) pattern, or Arrange Act Assert. This refers to how we order the things inside our tests.

However, personally, I also like it to have a similar logic when naming the tests. So it's usually "x happens when this occurs". I included suggestions below where I show this. Sometimes just modifying "if" for "when" is sufficient.

Just to reiterate, this is nitpicky and is a personal preference.

packages/flame/test/components/component_test.dart
st-pasha: This group needs to be moved inside the "Component" group, so that every top-level group name corresponds to the name of the class being tested.

packages/flame/test/components/component_test.dart
```diff
@@ -200,6 +200,56 @@ void main() {
...
+      (game) async {
+        final component = Component()..add(Component()..add(Component()));
+        await game.add(component);
+        game.update(0);
```
st-pasha, line 242:
```suggestion
        await game.ready();
```

packages/flame/test/components/component_test.dart
st-pasha: `game.add()` is perfectly fine here, and it's closer to the "real" code since there is no `ensureAdd()` in production.

packages/flame/test/components/component_test.dart
```diff
@@ -200,6 +200,56 @@ void main() {
       );
     });
   });
+
+  group('Component descendants', () {
+    flameGame.test(
```
st-pasha, line 205: Please prefer `testWithFlameGame()` (see #1358 for reasons), which has exactly the same signature. We're slowly refactoring all test code to stop using `flameGame.test()`.

st-pasha (commented):
Please add a test for the order of returned descendants. For example, suppose the game has children `[A, B, C]`, and `B` has children `[D, E]`, and `E` has single child `F` -- verify that the descendants are produced in the order `[A, B, D, E, F, C]`.

packages/flame/test/components/component_test.dart
alestiago: @st-pasha is correct. That's exactly what I meant, sorry for the ambiguity.

packages/flame/test/components/component_test.dart
alestiago: Just out of curiosity, when would you `ensureAdd` over `add`?

packages/flame/lib/src/components/component.dart
Dan-Crane: @st-pasha in the old version of `ancestors()` it was:
```dart
Iterable<Component> ancestors() sync* {
  var current = parent;
  while (current != null) {
    yield current;
    current = current.parent;
  }
}
```
Should I stick to it or stick to new version with generic type?

st-pasha: I prefer the non-template version, because there are already simple and well-known primitives to filter an iterable based on type. So, `ancestors().whereType<T>()` would work perfectly well instead of the templated version. On the other hand, if filtering by type is not needed, then there is no penalty for doing unnecessary work.

Dan-Crane: leave the current implementation for `descendants()`? Also i can remake the ancestor method in this PR and write tests for it

packages/flame/lib/src/components/component.dart
Dan-Crane: done

packages/flame/lib/src/components/component.dart
Dan-Crane: done

st-pasha: When done please resolve the comments, so that's easier to review.

packages/flame/test/components/component_test.dart
Dan-Crane: done

packages/flame/test/components/component_test.dart
Dan-Crane: done

packages/flame/test/components/component_test.dart
Dan-Crane: done

packages/flame/test/components/component_test.dart
Dan-Crane: done

packages/flame/test/components/component_test.dart
Dan-Crane: done

packages/flame/test/components/component_test.dart
Dan-Crane: What do you think of this test group structure?
```dart
group('ClassName.method -', () {
  testWithFlameGame(
    'testName',
    (game) async {
      // Setup

      // Action

      // Result
    },
  );
});
```

Output will be: 
`ClassName.method - testName`
This will visually show which method is currently being tested.

packages/flame/test/components/component_test.dart
Dan-Crane: Done. Please check me😊

packages/flame/test/components/component_test.dart
alestiago: I prefer having:

```
group('ClassName', () {
  group('method1', () {});
  group('method2', () {});
});
```

Notice how this way we don't repeat the "ClassName".

alestiago (commented):
Looks good! Just left a couple of comments regarding the documentation and testing.

packages/flame/lib/src/components/component.dart
```diff
@@ -319,6 +319,28 @@ class Component {
     }
   }
 
+  /// An iterator that performs a deep search on all [children].
```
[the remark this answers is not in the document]
st-pasha: I feel "deep search" is a bit ambiguous: would it search for a particular child and then return it? 
```suggestion
/// Enumerates all children, grand-children, etc. of the current component, in postorder.
```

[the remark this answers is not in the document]
alestiago: I agree that "deep search" can be seen as ambiguous. Your suggestion is nice but I don't like "etc". What about:

```suggestion
  /// Performs a postorder traversal on all [children].
```

There's no need to specify "of the current component" since referring to the `[children]` variable?
In addition, we already know from the signature that it "Enumerates" since it's an `Iterable`?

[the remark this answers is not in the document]
st-pasha: "Child components" sounds like they are immediate children only. Generally, the first line should give a short summary of what the method does, and then there could be a more elaborate description in the subsequent paragraphs.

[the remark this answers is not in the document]
alestiago: Indeed, I agree with that since it is detailed in the Effective Dart. I'm running out of ideas. Maybe we can do something like:

1.
```suggestion
  /// Enumerates all nested [children].
  ///
  /// The search is depth-first-search, hence [descendants] are in postorder.
```

2.
```suggestion
  /// Enumerates all [Component]s in the subtree.
  ///
  /// The search is depth-first-search, hence [descendants] are in postorder.
```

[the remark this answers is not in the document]
Dan-Crane: Maybe?
```dart
/// Enumerates all nested [children].
///
/// The idea is that we move from parent to child until we reach a child without descendants.
/// If we reach such a child we go back (to the point of branching or divergence of paths)
/// and choose another path. This is the case until all the descendants have been traversed.
```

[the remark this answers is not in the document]
st-pasha: I like (1), though there is probably no point in linking `[descendants]`.

Given that the target audience may not be familiar with CS terminology, it would probably be helpful to also add an explanation like @Dan-Crane suggests (but the CS definition must also be present, so that the functionality can be easily understood by experienced programmers as well).

packages/flame/test/components/component_test.dart
```diff
@@ -199,6 +205,111 @@ void main() {
...
+      'length of descendants must be 0 '
+      'when it is called before the add method',
+      (game) async {
+        final component = Component()..add(Component()..add(Component()));
+
+        expect(game.descendants().length, 0);
+
+        await game.ensureAdd(component);
+
+        expect(game.descendants().length, 3);
```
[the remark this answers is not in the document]
Dan-Crane: You mean that the test above is similar to this one, therefore this test is meaningless?
If i change the name to?
`should return the correct length of the descendants after the onLoaded method finishes`

packages/flame/lib/src/components/component.dart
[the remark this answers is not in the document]
Dan-Crane: @st-pasha can you tell me what `CS` means?🙈

packages/flame/test/components/component_test.dart
Dan-Crane: done

packages/flame/test/components/component_test.dart
```diff
@@ -199,6 +205,111 @@ void main() {
...
+        //        ComponentE(         -- 3
+        //          children: [
+        //            ComponentF(),   -- 4
+        //          ]
+        //        ),
+        //      ]
+        //    ),
+        //    ComponentC(),           -- 5
+        //  ],
+        // );
```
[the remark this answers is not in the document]
Dan-Crane: done

packages/flame/test/components/component_test.dart
```diff
@@ -199,6 +205,111 @@ void main() {
...
+        await game.ready();
+
+        final result = game
+            .descendants()
+            .whereType<_OrderCheckingComponent>()
+            .map((e) => e.order)
+            .toList();
+
+        expect(result, expectedOrder);
```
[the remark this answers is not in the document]
Dan-Crane: done

packages/flame/lib/src/components/component.dart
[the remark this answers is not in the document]
alestiago: @Dan-Crane good question! CS stands for Computer Science.

packages/flame/test/components/component_test.dart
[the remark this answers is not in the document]
alestiago: I think this test is great!

I think we can leave it as it is. Great job @Dan-Crane !

I marked this as nit because from the test name (which I like) I was not expecting to see the last `expect` but I do find value on the last `expect`, since it's verifying that there are no aide effects on calling it again after being called before loaded.

packages/flame/lib/src/components/component.dart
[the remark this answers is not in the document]
luanpotter: how about `Recursively enumerates...` (that's the word that in my mind immediately clarifies that it is multi-level)

packages/flame/test/components/component_test.dart
```diff
@@ -199,6 +199,103 @@ void main() {
...
+      );
+
+      testWithFlameGame(
+        'length must be equal to the number of added components including '
+        'itself when the passed parameter includeSelf is true',
+        (game) async {
+          final component = Component()..add(Component()..add(Component()));
+          await game.ensureAdd(component);
+
+          expect(game.descendants(includeSelf: true).length, 4);
```
luanpotter, line 232: how about one more assert:
```suggestion
          final descendants = game.descendants(includeSelf: true);
          expect(descendants.length, 4);
          // expect descendants contains game
```

luanpotter (approved):
LGTM, nice use of generator functions!

packages/flame/lib/src/components/component.dart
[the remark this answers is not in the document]
alestiago: I love how `Recursively enumerates all [children]` sounds!

wolfenrain (approved):
Apart from the whole doc comment discussion, LGTM! Love that you used a generator function for this!

packages/flame/test/components/component_test.dart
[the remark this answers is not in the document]
Dan-Crane: I am not offended if i wrote bad code/comment, speak up) it will help me improve)

yes, this is a test that everything went well

packages/flame/test/components/component_test.dart
Dan-Crane: done

packages/flame/lib/src/components/component.dart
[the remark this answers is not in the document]
Dan-Crane: done

alestiago (approved):

packages/flame/lib/src/components/component.dart
spydon: Agree with @alestiago

spydon (approved):

PR #1463, opened 2022-03-17
feat: Automatic Scaling for Isometric tiles and Negative indexed tile access

Sid911:
# Description
- Added Support for Automatic Scaling for isometric tile. This ensures that Any art which has a different isometric view with longer height than width and even the opposite scales automatically. And there is no problem rendering them side by side and also modified helper functions.

- Added support for negative indexing in `IsometricTileMapComponent` which starts from  -2 in bottom left of the spritesheet's image then move right and overflows to upper rows with the help of `propSize` which returns `tilesheet.srcSize` so user can basically access tiles from negative indexes.

before scaling :
image

after : 
flutter_07

> Use case for props can be where things are taller than normal images
flutter_08

## Breaking Change


doc/flame/components.md
```diff
@@ -788,6 +789,14 @@ This is an example of how a quarter-length map looks like:
 
 ![An example of a isometric map with selector](../images/isometric.png)
 
+The component Also supports negative indexing starting from -2 *(as -1 is reserved for empty tile)* which returns sprite from bottom Left to bottom right using negative Indexing and wraps to upper rows on overflow of tileset's image width.
```
spydon, line 792:
```suggestion
The component supports negative indexing starting from -2 *(as -1 is reserved for empty tile)* 
which returns the sprite from the bottom left to the bottom right using negative indexing and
wraps to the upper rows on overflow of the tileset's image width. 
```

doc/flame/components.md
```diff
@@ -788,6 +789,14 @@ This is an example of how a quarter-length map looks like:
 
 ![An example of a isometric map with selector](../images/isometric.png)
 
+The component Also supports negative indexing starting from -2 *(as -1 is reserved for empty tile)* which returns sprite from bottom Left to bottom right using negative Indexing and wraps to upper rows on overflow of tileset's image width. 
+
+Use `propSize` Property to give a custom Size to negative Indexed elements *(This can be used to add props and art as Tiles which rise higher than most tiles)*.
```
spydon, line 794:
```suggestion
Use the `propSize` property to give a custom size to negative indexed elements 
*(This can be used to add props and art such as tiles which rise higher than most tiles)*.
```

doc/flame/components.md
```diff
@@ -788,6 +789,14 @@ This is an example of how a quarter-length map looks like:
...
+
+An Example of rendering Props as tiles is something like those mirrors some of which are higher than normal tiles
```
spydon, line 796:
```suggestion
The following is an example of rendering props as tiles. In this example it is specifically the
mirrors, some of which are higher than the normal tiles.
```

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -52,11 +52,18 @@ class IsometricTileMapComponent extends PositionComponent {
   /// Note: this must be measured in the destination space.
   double? tileHeight;
 
+  /// propTile's Max Size. This size is used to define negative indexed elements.
+  /// Note : Currently this only accounts for variable height, width is disregarded
+  /// and [tileset.srcSize.x] is used instead.
+  ///
+  Vector2? propSize;
```
spydon, line 59:
```suggestion
  /// propTile's max size. This size is used to define negative indexed elements.
  /// Note : Currently this only accounts for variable height, width is disregarded
  /// and [tileset.srcSize.x] is used instead.
  Vector2? propSize;
```

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -75,20 +82,36 @@ class IsometricTileMapComponent extends PositionComponent {
   /// This is the size the tiles will be drawn (either original or overwritten).
   Vector2 get effectiveTileSize => destTileSize ?? tileset.srcSize;
 
-  /// This is the vertical height of each block; by default it's half the tile
-  /// size.
+  /// Gets the current Scaling factor for the isometric View
```
spydon, line 85:
```suggestion
  /// The current scaling factor for the isometric view.
```

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -75,20 +82,36 @@ class IsometricTileMapComponent extends PositionComponent {
...
+  double get scalingFactor => effectiveTileSize.y / effectiveTileSize.x;
+
+  /// This is the vertical height of each block; by default it's half the tile size.
   double get effectiveTileHeight => tileHeight ?? (effectiveTileSize.y / 2);
 
+  /// gets the effective propSize. PropSize should be assigned if prop dimension are different from srcSize of tilset
```
spydon, line 91:
```suggestion
  /// The effective [propSize]. [propSize] should be assigned if prop dimensions are different from the [tileset.srcSize].
```

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -75,20 +82,36 @@ class IsometricTileMapComponent extends PositionComponent {
...
+  Vector2 get pSize => propSize ?? effectiveTileSize;
```
spydon, line 92: This should have a more explanatory name

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -75,20 +82,36 @@ class IsometricTileMapComponent extends PositionComponent {
...
         final element = matrix[i][j];
         if (element != -1) {
-          final sprite = tileset.getSpriteById(element);
-          final p = getBlockRenderPositionInts(j, i);
-          sprite.render(c, position: p, size: size);
+          final Sprite currentSprite;
+          if (element < -1) {
+            final yRowShift = ((element + 2).abs() / (tileset.image.width / pSize.x).floor()).floor() + 1;
+            final xColShift = (element + 2).abs() % (tileset.image.width / pSize.x).floor();
+            final sourcePosition = Vector2(
```
spydon, line 105: We shouldn't create a new `Vector2` object in every tick here, you can create a private vector outside of the render method that you use instead, for example `_renderSourcePosition` and then you use it in here with `_renderSourcePosition.setValues`.

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -75,20 +82,36 @@ class IsometricTileMapComponent extends PositionComponent {
...
+              pSize.x * xColShift,
+              tileset.image.height - pSize.y * yRowShift,
+            );
+            currentSprite = Sprite(tileset.image, srcPosition: sourcePosition, srcSize: pSize);
```
spydon, line 109: You'll have to do something similar for the `Sprite` object here (maybe `_renderSprite`).

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -105,32 +128,40 @@ class IsometricTileMapComponent extends PositionComponent {
...
     final pos = Vector2(i.toDouble(), j.toDouble())..multiply(halfTile);
     return cartToIso(pos) - halfTile;
   }
 
+  /// Same as [getBlockRenderPositionInts] but is used for props ie. elements with <= -2 values
+  ///
+  /// This accounts for difference in size of the propSize.
+  Vector2 getVariableRenderPositionInts(int i, int j) {
+    final initialPosition = getBlockRenderPositionInts(i, j);
+    initialPosition.sub(Vector2(pSize.x - effectiveTileSize.x, pSize.y - effectiveTileSize.y));
```
spydon, line 141: Use `..x -=` and `..y -=`, or `.setValues(initialPosition.x - ..., initialPosition.y - ...)`

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -105,32 +128,40 @@ class IsometricTileMapComponent extends PositionComponent {
...
   Vector2 getBlockRenderPositionInts(int i, int j) {
-    final halfTile = effectiveTileSize / 2;
+    final halfTile = Vector2(effectiveTileSize.x / 2, (effectiveTileSize.y / 2) / scalingFactor);
     final pos = Vector2(i.toDouble(), j.toDouble())..multiply(halfTile);
     return cartToIso(pos) - halfTile;
   }
 
+  /// Same as [getBlockRenderPositionInts] but is used for props ie. elements with <= -2 values
+  ///
+  /// This accounts for difference in size of the propSize.
```
spydon, line 138:
```suggestion
  /// This accounts for difference in size of the [propSize].
```

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -105,32 +128,40 @@ class IsometricTileMapComponent extends PositionComponent {
 
   /// Same as getBlockRenderPosition but the arguments are exploded as integers.
   Vector2 getBlockRenderPositionInts(int i, int j) {
-    final halfTile = effectiveTileSize / 2;
+    final halfTile = Vector2(effectiveTileSize.x / 2, (effectiveTileSize.y / 2) / scalingFactor);
     final pos = Vector2(i.toDouble(), j.toDouble())..multiply(halfTile);
     return cartToIso(pos) - halfTile;
```
spydon, line 133: This should preferably only create one new `Vector2` too, but since you didn't create more objects here than what was already created you don't have to do that if you don't want to.

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -105,32 +128,40 @@ class IsometricTileMapComponent extends PositionComponent {
...
+
   /// Get the position of the center of the surface of the isometric tile in
   /// the cartesian coordinate space.
   ///
   /// This is the opposite of [getBlock].
   Vector2 getBlockCenterPosition(Block block) {
     final tile = effectiveTileSize;
-    return getBlockRenderPosition(block) +
-        Vector2(tile.x / 2, tile.y - effectiveTileHeight - tile.y / 4);
+    return getBlockRenderPosition(block) + Vector2(tile.x / 2, tile.y - effectiveTileHeight - tile.y / 4);
```
spydon, line 151: Use `..x +=` and `..y +=` instead of creating a new vector

spydon: Ah, this was like that before already, so optional to fix :)

spydon (commented):
Good job! Mostly some comments about not creating new objects in methods called from the render-loop.

st-pasha:
@Sid911 Can you please explain why negative indexing is necessary? Counting from -2 sounds very confusing to me...

Sid911:
@st-pasha I guess you could understand why variable prop Size can be useful because props which usually would have to be rendered separately and would not benefit from being in a grid but negative indexing would be awesome for this in artistic sense. For example in this sprite sheet
image
The blocks can be placed above and the props can be placed below .. because of variable heights it wouldn't make sense to place props after blocks. 

In perfect world I would like the matrix for the isometricTileMap to start from 1 and 0 be the empty place and negative be the for access to well negative indexing. And for the spritesheet's id function to have support for negative indexing with an option for custom height and width of the sprite this way we could basically retrieve any kind of sprite from a `imaginary` grid of passed in custom `height` and `width` or Spritesheet's given size.

But that would mean a breaking change. I guess @spydon can reply if that should be done instead of this *(as the 'perfect world)* can be used later on too... most of the code translates anyway. But I was hesitant of bringing in any breaking changes.

Sid911:
ahm... I definitely didn't hard reset my fork because of some issues which ended up closing this (its 4 am I should sleep). I will send another pull request to follow I have the changes locally.

st-pasha:
Thanks, @Sid911 for the detailed explanation. Though, I must admit, I still don't quite follow :(. Here's what I've got so far:

IsometricTileMap is a rectangular piece of land, where each small plot (called "tile") is represented by one of the images in the `tileset`. The images in the tileset are referred by their indices, so really an `IsometricTileMap` is just a 2D array of integer indices into the tileset spritesheet.

There is also a special index `-1` which means there is no land at all at that point in the map.

Now, this PR adds a notion of props, which I understand to be special objects that are on top of the land: foliage, clutter, structures, items, etc. What makes those objects different from tiles is that they have irregular shape, and therefore need special rendering.

What I don't understand though is how these props coexist with the regular tilemap: if an entry in the 2D array is a single integer, then that integer is either a tile or a prop, so when a prop is drawn, there will be no ground at that place? Or did I get everything wrong?

Sid911:
@st-pasha you got it right about them I guess the picture above in pr doesen't explain well. This will be a lot useful in conjunction with multilayered isometricTileMap rendering where we render multiple tilmaps above each otherwith just `-1 * effectileTileSize.y` (ie. whatever the tilesize is going to be which will be set in srcSize or destSize) offset like in this picture 
flutter_12
enabling things like shadow to be added in higher layers in order to make them seem like they had shadow already. For 

example, in this pic there are three layers going on atm, this is the base layer has light blocks
image

The props here and the darker blocks in layer 2 
image

And shadow images as props in layer 3 
image

I understood why you this seems kinda wrong for a single layer isometricTileMap. For single layers propTiles doesn't makes much sense unless I add an offset parameter for render distance to allow things like below pic where propTiles have their base already made
image

st-pasha:
I see, so there will be multiple `IsometricTileMap`s: one for the ground, then another objects on top (or maybe more than one).

In such a case, the layer with props could be a special class "props" class, derived from IsometricTileMap, with special rendering that's needed for props. And then the tileset indices won't have to be negative.

In fact, the "props" layer doesn't even have to follow the same 2D-array approach as the IsometricTileMap, given that:
 - on some pieces of the ground there won't be any props;
 - in some places there will be multiple props;
 - props don't separate into layers neatly -- for example, consider your image 3 above: the "shadows" layer -- if you add another prop in front of the building, it should obscure the shadows, so would it become the fourth layer? What if you add more objects?
 - props may need to be moved (doors open/close, objects picked up or dropped), or maybe they move by themself (NPCs or the player), and jumping between layers would be problematic.

Sid911:
> I see, so there will be multiple IsometricTileMaps: one for the ground, then another objects on top

Not specifically as you can see in layer 2 blocks and props can be added in same isometric Component. Here props means a static object which cannot be moved like bushes or fence or things like that.

> props don't separate into layers neatly -- for example, consider your image 3 above: the "shadows" layer -- if you add another prop in front of the building, it should obscure the shadows, so would it become the fourth layer? What if you add more objects?

buildings, fences and other static props can have their own shadow as you can see for mirror and the small block. The shadows as props are just for higher placed blocks themselves, so they give a sense of height not for giving shadow to other props themselves.

> props may need to be moved (doors open/close, objects picked up or dropped), or maybe they move by themself (NPCs or the player), and jumping between layers would be problematic.

again anything rendered by isometricTileMap cannot be moved, so this is just for static props, NPS etc. should be handled later on anyway because they can be moved so they need to handle priority changes as the move in the map because they need to be obscured.

>  on some pieces of the ground there won't be any props;

That is true, but for static props usually a map contains a lot of them not even including like higher placed blocks in the above example which will be in same layer, so layered tilemap rendering seemed best option. 

I built this first with a puzzle game in mind which doesn't really have any other sprite rendering but upon creating a pull request I did think how else this could be implemented which suits best for every style possible. For my original game I have a different logic as I don't need to worry about breaking changes. I hope you get the point of this, I do agree layered rendering adds tons of complexity behind it but I also agree that if you want to make game look populated there is no escaping it at the end.

st-pasha:
> The max layer a shadow should be required is if you have massive column of blocks not a huge prop (already would require tons of layers stacked up) and you want to shadow something in same level? which would mean you are either trying to create huge height difference at that point it would be faster to use priority method for like column of shadow. This doesn't make any sense to me.

Consider picture "layer 3" that you posted above. The block in the top right corner has a shadow on it. You say that the block is in layer 2, while the shadow is in layer 3. Now, imagine there is another block directly south (i.e. diagonally) from that dark grey block. This new block will partially obscure the shadow, so it means it must be in layer 4. Now, this new block may itself has a shadow -- layer 5. Now I add another block to the south, and it's layer 6 already. And so on -- and this is all with a perfectly reasonable and relatively flat map layout.

>  NPCs etc. should be handled later on anyway because they can be moved so they need to handle priority changes as the move in the map because they need to be obscured.

And how do you suggest this to be implemented? If an NPC needs to be partially obscured by a block, but will itself obscure other props, then how would this be rendered? Forget even about shadows and other complications, imagine there's a single layer of props -- how would we make an NPC appear in-between those props?

Sid911:
> Consider picture "layer 3" that you posted above. The block in the top right corner has a shadow on it. You say that the block is in layer 2, while the shadow is in layer 3. Now, imagine there is another block directly south (i.e. diagonally) from that dark grey block. This new block will partially obscure the shadow, so it means it must be in layer 4. Now, this new block may itself has a shadow -- layer 5. Now I add another block to the south, and it's layer 6 already. And so on -- and this is all with a perfectly reasonable and relatively flat map layout.

Not really just make a new type of shadow for that region, so all the blocks will be in 2nd layer and shadows on 3rd... there aren't many ways a cube of same size can obscure another cube of same size in isometric view. Its pretty easy to make anyways takes like 30 sec for even someone like me to add to tile sheet.

image.

> And how do you suggest this to be implemented? If an NPC needs to be partially obscured by a block, but will itself obscure other props, then how would this be rendered? Forget even about shadows and other complications, imagine there's a single layer of props -- how would we make an NPC appear in-between those props

It is true I haven't given much thought into how a playable character would work. But first thing that comes to mind (for new users) is you would already be using some kind of per cycle collision detection if you are adding an playable character which checks which block player is on using its position and using the position to check surrounding blocks in the relevant layers isometricTileMap, so just change the priority to less when player move to behind to non -1 element and change it back to higher when it comes to front. And we could optimize this by only checking this all if player position is changing.

off Course that wouldn't exactly work for tight trenches and player where a block/prop will be behind and front of player like you said. For best method that you could do is, hack into the render loop of the isometric Map and render it that way so we wouldn't need priority just render in a way player is obscured inside the for loop.

PR #1464, opened 2022-03-18
feat: Move effects can now be applied to camera's Viewport and Viewfinder

st-pasha:
# Description

- Added mixin `EffectTarget<T>` to be used by effects -- this is similar to `ComponentEffect<T>`, except that the target doesn't have to be a component;
- Added abstract interface `PositionProvider` -- any class that implements this interface can be a valid target of a move effect;
- Classes `PositionComponent`, `Viewfinder` and `Viewport` now implement the `PositionProvider` interface;
- Small refactor in the `Effect` class: the `lastProgress` property is now defined in the `Effect` class itself, not in the derived classes (since it is almost universally useful);

Note: in subsequent PRs I'm planning to expand this approach to other effects as well: scale effects, rotation effects, etc. The top-level documentation will also be updated then.

## Checklist

## Breaking Change

## Related Issues


packages/flame/lib/src/effects/provider_interfaces.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:vector_math/vector_math_64.dart';
+
+/// Interface for a component that can be affected by move effects.
+abstract class PositionProvider {
```
spydon, line 4: Will there be one interface for each part of the transform matrix? It will be very bloated with like `implements PositionProvider, ScaleProvoder, AngleProvider` everywhere where it is needed? Or will there be a mixin or that implements all of them that can be used?

packages/flame/lib/src/components/position_component.dart
```diff
@@ -402,4 +403,14 @@ class PositionComponent extends Component {
     size.setValues(rect.width, rect.height);
     topLeftPosition = rect.topLeft.toVector2();
   }
+
+  //#region Effect interfaces
```
spydon, line 407: There will be a lot of effect specific code in here if we continue down this path.
if `PositionProvider` "looks" for the `position` field instead at least these overrides shouldn't be necessary.

st-pasha: Alright, I'll make the interface refer to the `position` field instead.

spydon (approved):

erickzanardo (approved):

wolfenrain (approved):

PR #1465, opened 2022-03-18
feat: Possibility to mark gesture events as handled

spydon:
# Description

`HasTappables`, `HasHoverables` and `HasDraggables` now returns a bool in their methods so that the user can know whether their components marked the event as handled.

## Checklist

## Breaking Change

- [x] Nope, not anymore

## Related Issues


doc/flame/inputs/gesture-input.md
```diff
@@ -282,6 +282,25 @@ class MyGame extends FlameGame with HasTappables {
...
+```dart
+class MyGame extends FlameGame with HasTappables {
+  @override
+  bool onTapDown(int pointerId, TapDownInfo info) {
+    final handled = !super.onTapDown(info);
+    if(handled) {
+      // Do something if a child handled the event
+    }
+    return handled;
+  }
```
st-pasha, line 300: Compare with the following API:
```dart
void onTapDown(TapDownEvent event) {
  super.onTapDown(event);
  if (event.handled) {
    // do something
  }
}
```

st-pasha (commented):
Could you please describe which part of this PR constitutes a breaking change? And how would the users need to adjust their code in order to accommodate this change?

doc/flame/inputs/gesture-input.md
spydon: I agree that is nicer, I don't think we wanted mutable events though (ping @ErickZanardo). But I think it would be good to go that way anyways, because then this wouldn't be a breaking change.

st-pasha: We have plenty of mutable vectors everywhere, not sure why mutable events are a problem then...

spydon: That's not the same thing though, because you always change and get the `Vector2` from a known object, here "something unknown" further up in the chain will modify the value. But as I said, I think it's still worth it.

packages/flame/test/game/mixins/has_tappables_test.dart
```diff
@@ -0,0 +1,107 @@
...
+    final handled = !super.onTapUp(pointerId, info);
+    if (handled) {
+      handledOnTapUp++;
+    }
+    return true;
+  }
+
+  @override
+  bool onTapCancel(int pointerId) {
+    final handled = !super.onTapCancel(pointerId);
```
erickzanardo, line 34: I wonder if we can think on an alternative to always have to check super, can't the engine check if this was handled already on the top of the component tree so the method ins't called and we don't need to always check?

doc/flame/inputs/gesture-input.md
erickzanardo: Yeah, not a big fan of mutating events, I mean a event should be a snapshot of something that have already happened, and you can't change the past (at least AFAIK :P)

But I wonder if we couldn't think on an alternative to having to check if the event was already handled in other to do our logic, maybe the engine could check that and only call the method if it wasn't handled yet?

spydon: Possibly, but that closes down some options for the users, like handling events differently depending on whether it was handled or not, and it would still be a breaking change. The `Has*` game mixins should have a bit more control than the component mixins I think.

Having a mutable `handled` flag on the event isn't really changing the past imho, since we know the flow of these events. If the user isn't passing the event somewhere else, and then reusing it later, but that would be a really strange design... And if they look at that event later for some weird reason and the flag has changed, then it is actually true that the event was handled.

Isn't it better with a flag than a breaking change at least?

erickzanardo: What if we had a method instead of a flag then? Like a `stopPropagation` method, so internally we could still make the event be immutable and we would have the same scenario of the attribute.

spydon: This is not for the propagation stopping or not, this is for knowing whether an event was handled or not in the `Has*` methods.

erickzanardo: I think what is confusing me, is all that if you did on the return value of the parent method right? But looking deeper now I bet you did that because with a return type of bool we had to use must call super....

If we had the mutating flag, we wouldn't need to force the user to check that right? Because that is my biggest concern, forcing the user to always write that, which will pollute the code for the most cases.

If this assumption is correct, I think I can live with the mutating variable as its benefits will overgrow

spydon: I'll change the PR and then we can compare the commits to see which way we like better? :)

erickzanardo: Sounds good!

examples/lib/stories/collision_detection/circles_example.dart
```diff
@@ -71,6 +71,7 @@ class MyCollidable extends PositionComponent
   @override
   void onCollisionEnd(PositionComponent other) {
     super.onCollisionEnd(other);
+    print('on collision end');
```
erickzanardo, line 74:
```suggestion
```

packages/flame/lib/src/gestures/events.dart
```diff
@@ -219,3 +221,7 @@ class ScaleUpdateInfo extends PositionInfo<ScaleUpdateDetails> {
     ScaleUpdateDetails raw,
   ) : super(game, raw.focalPoint, raw);
 }
+
+mixin _HandledField {
```
erickzanardo, line 225: That is cheating :P

erickzanardo (approved):

st-pasha (commented):
@spydon API details aside, I guess the big question is this: How do we even know that an event was handled?

Say, `HasDraggables.onDragStart` returns the result of propagating `handleDragStart()`. However, the `handleDragStart` returns true/false indicating whether or not the event ought to continue being propagated to other components. Which is not the same as "being handled".

packages/flame/lib/src/gestures/events.dart
spydon: I thought you said that we'd try out to see how it looked like with a mutable boolean? Or how is it cheating? :smile:

erickzanardo: I am just messing with you, I liked the approach to create a mixin so you don't need to alter all the base classes

spydon:
> @spydon API details aside, I guess the big question is this: How do we even know that an event was handled?
> 
> Say, `HasDraggables.onDragStart` returns the result of propagating `handleDragStart()`. However, the `handleDragStart` returns true/false indicating whether or not the event ought to continue being propagated to other components. Which is not the same as "being handled".

That's true, maybe we should separate those two things. Then you'd just set the `handled` flag in your methods instead then, and the booleans would just be for propagation. I still don't like having the mutable events, but I don't see a better option for passing that information through the flow.

packages/flame_forge2d/example/analysis_options.yaml
```diff
@@ -0,0 +1 @@
+include: package:flame_lint/analysis_options.yaml
```
luanpotter, line 1: we should have a linter check to enforce this on all packages!

packages/flame/lib/src/gestures/events.dart
luanpotter: why can't this be added only once to `BaseInfo`?

packages/flame/lib/src/components/mixins/keyboard_handler.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:flutter/services.dart';
+
+import '../../../components.dart';
+import '../../../input.dart';
+
+/// A [Component] mixin to add keyboard handling capability to components.
+/// Must be used in components that can only be added to games that are mixed
+/// with [HasKeyboardHandlerComponents].
+mixin KeyboardHandler on Component {
+  bool onKeyEvent(RawKeyEvent event, Set<LogicalKeyboardKey> keysPressed) {
```
luanpotter, line 10: some dartdocs on what the bool mean can come in handy

packages/flame/example/lib/main.dart
```diff
@@ -28,10 +28,8 @@ class MyGame extends FlameGame with DoubleTapDetector, HasTappables {
   @override
   void onTapUp(int id, TapUpInfo info) {
     super.onTapUp(id, info);
-    final touchPoint = info.eventPosition.game;
-    final handled = children.any((c) => c.shouldRemove);
-
-    if (!handled) {
+    if (!info.handled) {
```
luanpotter, line 31: nice!

doc/flame/inputs/gesture-input.md
```diff
@@ -382,6 +434,32 @@ The provided event info is from the mouse move that triggered the action (enteri
...
+  bool onHoverEnter(PointerHoverInfo info) {
+    info.handled = true;
+    return true;
+  }
+}
+
+class MyGame extends FlameGame with HasHoverables {
+  @override
+  void onHoverEnter(PointerHoverInfo info) {
+    if(info.handled) {
```
luanpotter, line 456:
```suggestion
    if (info.handled) {
```

doc/flame/inputs/gesture-input.md
```diff
@@ -356,6 +382,32 @@ class MyGame extends FlameGame with HasDraggables {
...
+ bool onDragStart(DragStartInfo info) {
+   info.handled = true;
+   return true;
+ }
+}
+
+class MyGame extends FlameGame with HasDraggables {
+  @override
+  void onDragStart(int pointerId, DragStartInfo info) {
+    if(info.handled) {
```
luanpotter, line 404:
```suggestion
    if (info.handled) {
```

packages/flame_forge2d/example/analysis_options.yaml
spydon: Haha, is that even possible? Lintception

luanpotter: well via a bash script. not using the flutter linter :P

PR #1468, opened 2022-03-18
feat: Automatic Isometric Grid scaling

Sid911:
# Description
- Added Support for Automatic Scaling for isometric tile. This ensures that Any art which has a different isometric view with longer height than width and even the opposite scales automatically. And there is no problem rendering them side by side and also modified helper functions.

## Checklist

before scaling :
image

after : 
flutter_07

## Breaking Change


Sid911:
Sorry for the fork hard reset which closed last PR. @st-pasha pls continue about your comment here https://github.com/flame-engine/flame/pull/1463#issuecomment-1072841199

Sid911:
In perfect world I would like the matrix for the isometricTileMap to start from 1 and 0 be the empty place and negative be the for access to well negative indexing. And for the spritesheet's getSpriteFromID function *(I don't remember the name)* to have support for negative indexing with an option for custom height and width of the sprite this way we could basically retrieve any kind of sprite from a `imaginary` grid of passed in custom `height` and `width` or Spritesheet's given size.

But that would mean a breaking change. I guess maintainers can reply if that should be done instead of this PR *(as the 'perfect world can be used later on too with many more uses)*... most of the code translates anyway and can be repourposed. But I was hesitant of bringing in any breaking changes.

doc/flame/components.md
```diff
@@ -788,6 +789,18 @@ This is an example of how a quarter-length map looks like:
 
 ![An example of a isometric map with selector](../images/isometric.png)
 
+The component supports negative indexing starting from -2 *(as -1 is reserved for empty tile)*
```
luanpotter, line 792: we should consider changing the empty tile to null now that we have NS in Dart. then you can start on -1.
or maybe have another field that defines how to interpret the index (then we could have even more options)
thoughts @spydon @erickzanardo @wolfenrain @renancaraujo

spydon: One of those options sounds good, with the second one we wouldn't need to have these negative overflow indexes that are used in this PR right?

Sid911: by 2nd option do mean passing a comparison function returning bool to the component in order to determine which indices to use propSize for?

doc/flame/components.md
```diff
@@ -788,6 +789,18 @@ This is an example of how a quarter-length map looks like:
...
+which returns the sprite from the bottom left to the bottom right using negative indexing and
+wraps to the upper rows on overflow of the tileset's image width. 
+
+Use the `propSize` property to give a custom size to negative indexed elements 
+*(This can be used to add props and art such as tiles which rise higher than most tiles)*.
+
+The following is an example of rendering props as tiles. In this example it is specifically the
+mirrors, some of which are higher than the normal tiles.
+
+![An example use case of props](../images/isometricPropsThumb.png)
```
wolfenrain, line 802: It would maybe be better if we had an example for it and reference that image here so people can also see the code behind the visualization

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -52,18 +52,25 @@ class IsometricTileMapComponent extends PositionComponent {
   /// Note: this must be measured in the destination space.
   double? tileHeight;
 
+  /// propTile's max size. This size is used to define negative indexed elements.
+  /// Note : Currently this only accounts for variable height, width is disregarded
```
wolfenrain, line 56:
```suggestion
  ///
  /// Note: Currently this only accounts for variable height, width is disregarded
```

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -75,20 +82,53 @@ class IsometricTileMapComponent extends PositionComponent {
...
+  /// The current scaling factor for the isometric view.
+  double get scalingFactor => effectiveTileSize.y / effectiveTileSize.x;
+
+  /// This is the vertical height of each block; by default it's half the tile size.
   double get effectiveTileHeight => tileHeight ?? (effectiveTileSize.y / 2);
 
+  /// gets the effective propSize. PropSize should be assigned if prop dimension are different from srcSize of tilset
+  Vector2 get propsMaxSize => propSize ?? effectiveTileSize;
+
+  // render specific variables
```
wolfenrain, line 94: Is this like a region? If so we should use a region comment

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -75,20 +82,53 @@ class IsometricTileMapComponent extends PositionComponent {
...
+
+  /// sourceRenderPosition of the sprite on the image
+  final Vector2 _renderSourcePosition = Vector2.zero();
+
+  /// renderSprite
```
wolfenrain, line 99: Private variables dont really have to be documented, especially if it is just the name itsef ;-) no extra value in that

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -75,20 +82,53 @@ class IsometricTileMapComponent extends PositionComponent {
   /// This is the size the tiles will be drawn (either original or overwritten).
   Vector2 get effectiveTileSize => destTileSize ?? tileset.srcSize;
 
-  /// This is the vertical height of each block; by default it's half the tile
-  /// size.
+  /// The current scaling factor for the isometric view.
```
wolfenrain, line 85:
```suggestion
  /// Returns the current scaling factor for the isometric view.
```

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -75,20 +82,53 @@ class IsometricTileMapComponent extends PositionComponent {
...
+  double get scalingFactor => effectiveTileSize.y / effectiveTileSize.x;
+
+  /// This is the vertical height of each block; by default it's half the tile size.
```
wolfenrain, line 88:
```suggestion
  /// Returns the vertical height for each block.
  ///
  /// By default it's half the tile size.
```

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -75,20 +82,53 @@ class IsometricTileMapComponent extends PositionComponent {
...
   double get effectiveTileHeight => tileHeight ?? (effectiveTileSize.y / 2);
 
+  /// gets the effective propSize. PropSize should be assigned if prop dimension are different from srcSize of tilset
```
wolfenrain, line 91:
```suggestion
  /// Returns the effective prop size. 
  ///
  /// Note: Assign [propSize] if the prop dimension are different from the `srcSize` of the tileset.
```

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -75,20 +82,53 @@ class IsometricTileMapComponent extends PositionComponent {
...
+  Vector2 get propsMaxSize => propSize ?? effectiveTileSize;
+
+  // render specific variables
+
+  /// sourceRenderPosition of the sprite on the image
```
wolfenrain, line 96:
```suggestion
  /// Represents the render position of the sprite on the source image.
```

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -105,11 +145,22 @@ class IsometricTileMapComponent extends PositionComponent {
 
   /// Same as getBlockRenderPosition but the arguments are exploded as integers.
```
wolfenrain, line 146:
```suggestion
  /// Same as [getBlockRenderPosition] but the arguments are exploded as integers.
```

Know it is outside ur PR, but noticed that u did change it for the new method

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -105,11 +145,22 @@ class IsometricTileMapComponent extends PositionComponent {
...
   Vector2 getBlockRenderPositionInts(int i, int j) {
-    final halfTile = effectiveTileSize / 2;
+    final halfTile = Vector2(
+        effectiveTileSize.x / 2, (effectiveTileSize.y / 2) / scalingFactor);
     final pos = Vector2(i.toDouble(), j.toDouble())..multiply(halfTile);
     return cartToIso(pos) - halfTile;
   }
 
+  /// Same as [getBlockRenderPositionInts] but is used for props ie. elements with <= -2 values
```
wolfenrain, line 154:
```suggestion
  /// Same as [getBlockRenderPositionInts] but it is used for props ie. elements with <= -2 values
```

No idea what you mean with `ie` so maybe we should not use abbrevations?

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -52,18 +52,25 @@ class IsometricTileMapComponent extends PositionComponent {
   /// Note: this must be measured in the destination space.
   double? tileHeight;
 
+  /// propTile's max size. This size is used to define negative indexed elements.
```
wolfenrain, line 55:
```suggestion
  ///  The [propSize] is used to define negative indexed elements.
```

wolfenrain (changes requested):

doc/flame/components.md
Sid911: Yeah, but the source code for the image larger than what a typical example would be. I could add a link to the project which currently wouldn't work but will work as I will set the project public soon anyways. Should I do that or make a smaller example in the examples for flame?

packages/flame/lib/src/components/isometric_tile_map_component.dart
Sid911: I just like to categorize variables ... 🤦🏽‍♂️ that just leaked into this code, I will remove it as this is a different style compared to rest.

packages/flame/lib/src/components/isometric_tile_map_component.dart
Sid911: yup This just slipped through.

packages/flame/lib/src/components/isometric_tile_map_component.dart
Sid911: by ie. I meant 'in other words' as `prop` itself might not be known to the person reading this. I will change it to full sentence instead.

st-pasha:
@Sid911 Given that props do not appear in the same layer as the regular "ground" tiles, can we simply have a boolean flag on a `IsometricTileMapComponent` indicating whether it has regular tiles or prop tiles? That way you won't have to use negative indexing for props.

doc/flame/components.md
spydon: How long is the wxanple currently? We have considered putting more full-blown examples into the example page anyways.

Sid911:
@st-pasha I am aware that negative indexing is little weird but it is a standard practice in python etc as accessing from behind is kinda intuitive. Adding a bool would mean you wouldn't be able to render normal blocks in same components. I am considering to implement as suggested in https://github.com/flame-engine/flame/pull/1468#discussion_r830501179

That way null can be used instead of -1 and we pass a function as an optional field for giving users the way to change what values of the tilesheet to use propSize for or not. This way negative indexing can start from -1 and normal indexes can start from 0 by default but can be changed by user. 
While doing that I can also have a render offset for props so that users can create uneven sprites and handle this ontheir own. Also dynamic width while we are at it.

@spydon @luanpotter @wolfenrain would this be a good option? Although this is a  breaking change.

doc/flame/components.md
Sid911: The current game implementation of the levelLoader page as we are seeing here is about 250-300 lines but that includes things like camera movement, tap, state management etc. not just rendering. 

I am thinking of adding just a simple example for the flame examples, while we are at it.

st-pasha:
@Sid911 I'm fully aware about Python's negative indexing, the question though whether it's the right paradigm for this particular case.

AFAIU, you want to have negative indexes because in your spritesheet all props are located at the back of the image, and all regular tiles in the front. Of course, this is a perfectly reasonable organization. But Flame is a generic library, used by thousands of people, so the question is: would this organization also be reasonable for everybody else? If another user wants to have props interleaved with tiles in the spritesheet (perhaps they want to organize assets by geography, or by material type, etc) -- would they need to write "prop index -99" in a spritesheet of 100 images? What if you want to later add more tiles or props to the spritesheet -- would you have to re-code the indices of all props?

> Adding a bool would mean you wouldn't be able to render normal blocks in same components. 

I believe you mentioned in an earlier comment that the intention is to not have tiles and props in the same layer, since that would imply the prop is hanging in the "void", since there won't be any ground underneath.

spydon:
Do we have any good options for avoiding the negative indexing? I'm not a big fan of it either tbh.

st-pasha:
> Do we have any good options for avoiding the negative indexing? I'm not a big fan of it either tbh.

AFAIU, all objects in the layer will either be regular tile pieces, or props, but they won't intermix. Which means a simple boolean flag at the IsometricTileMapComponent should be sufficient.

A perhaps better alternative is to skip the boolean flag and create a new class specifically for props (this class would be derived from IsometricTileMap). It appears the new class would need to simply override the `render()` method.

Sid911:
@spydon @st-pasha so what should be done instead of this? 
I am currently working on many projects atm... I guess I can reduce this pr to the automatic scaling instead of props... So, we can have that at least, then we can worry about implementing the ideal props behavior later? Maybe discussing about it in an issue thread?

spydon:
> @spydon @st-pasha so what should be done instead of this? 
> I am currently working on many projects atm... I guess I can reduce this pr to the automatic scaling instead of props... So, we can have that at least, then we can worry about implementing the ideal props behavior later? Maybe discussing about it in an issue thread?

Sounds like a good idea!

Sid911:
Cool I will update this pr and make an issue for the support of props and variable sized rendering as soon as I get time.

Sid911:
Hi, sorry for the delay in this PR. Usually I am very active but due to many reasons I had been deadlocked by my schedule. Also, for some reason melos Analyze is not passing, please let me know if there is a problem.

packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -75,20 +76,28 @@ class IsometricTileMapComponent extends PositionComponent {
   /// This is the size the tiles will be drawn (either original or overwritten).
   Vector2 get effectiveTileSize => destTileSize ?? tileset.srcSize;
 
-  /// This is the vertical height of each block; by default it's half the tile
-  /// size.
+  /// The current scaling factor for the isometric view.
+  double get scalingFactor => effectiveTileSize.y / effectiveTileSize.x;
+
+  /// This is the vertical height of each block; by default it's half the tile size.
```
spydon, line 82:
```suggestion
  /// This is the vertical height of each block; by default it's half the
  /// tile size.
```

spydon (approved):

wolfenrain (approved):

PR #1469, opened 2022-03-19
fix: `debugMode` should be inherited from parent when mounted

spydon:
# Description

`debugMode` wasn't propagating properly from the parent component to it's descendants.
If you for example added a component `a` to a component `b` first, and then added `b` to `c` which had `debugMode = true`, `a` would not have `debugMode = true`. But if `a` was added to `b` after it was added to `c`, then `a` would have `debugMode = true`.

This also fixes the `debugMode` for `HitboxShape`s.

## Checklist

## Breaking Change

## Related Issues


wolfenrain (approved):
LGTM

luanpotter (approved):
LGTM

PR #1476, opened 2022-03-21
feat: adding blueprint

erickzanardo:
# Description

Adds a new way of grouping flame components that will be added directly on the game level and does not introduces a component who only purposes is to group other components.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/doc/flame/components.md
+++ b/doc/flame/components.md
@@ -154,6 +154,48 @@ void update(double dt) {
 ```
 
 
+### Blueprints
+
+Flame also offers a virtual approach for grouping components which allows the developer to keep
+their components grouped "code wise", but the components (or other things, like a
+`ContactCallback` for Forge2d Games) are individually added to the game and there isn't a parent
+component.
+
+This is useful when you have a complex feature that will require different components, but they
+all need to be on the same rendering level of different faatures, a simple example for this is
+a top down view game, we would like to play with the rendering order of things so we can give the
+player the impression of depth, so if the player character goes behind a house, its roofs is
+rendered "above" the player, to be able to achieve that we would need to split our building
+component into more components, for example: `HouseBase`, `HouseSecondFloor` and `HouseRoof`.
+
+So to avoid leaving these components "floating" around, we could create a `Blueprint` to group
+them:
+
+```dart
+class House extends Blueprint {
+  @override
+  void build() {
+    addAll([
+      HouseBase(),
+      HouseSecondFloor(),
+      HouseRoof(),
+    ]);
+  }
+}
+```
+
+And then just add it to the game, quite similar to how simple components would be added:
+
+```dart
+class MyGame extends FlameGame {
+  @override
+  Future<void> onLoad() async {
+    addFromBlueprint(House());
+  }
+}
+```
+
+
 ### PositionType
 If you want to create a HUD (Head-up display) or another component that isn't positioned in relation
 to the game coordinates, you can change the `PositionType` of the component.
--- a/packages/flame/lib/src/components/blueprint.dart
+++ b/packages/flame/lib/src/components/blueprint.dart
@@ -1,0 +1,50 @@
+import 'package:flutter/foundation.dart';
+
+import '../../components.dart';
+import '../game/flame_game.dart';
+
+const _attachedErrorMessage = "Can't add to attached Blueprints";
+
+// so we can experiment with the idea, but this is a
+// potential upstream change on Flame.
+
+/// A [Blueprint] is a virtual way of grouping [Component]s
+/// that are related, but they need to be added directly on
+/// the [FlameGame] level.
+abstract class Blueprint {
+  final List<Component> _components = [];
+  bool _isAttached = false;
+
+  /// Called before the the [Component]s managed
+  /// by this blueprint is added to the [FlameGame]
+  void build();
+
+  /// Attach the [Component]s built on [build] to the [game]
+  /// instance
+  @mustCallSuper
+  Future<void> attach(FlameGame game) async {
+    build();
+    await game.addAll(_components);
+    _isAttached = true;
+  }
+
+  /// Adds a list of [Component]s to this blueprint.
+  void addAll(List<Component> components) {
+    assert(!_isAttached, _attachedErrorMessage);
+    _components.addAll(components);
+  }
+
+  /// Adds a single [Component] to this blueprint.
+  void add(Component component) {
+    assert(!_isAttached, _attachedErrorMessage);
+    _components.add(component);
+  }
+
+  /// Returns a copy of the components built by this blueprint
+  List<Component> get components => List.unmodifiable(_components);
+
+  /// [bool] that indicates if this [Blueprint] instance is already
+  /// attached to a [FlameGame], attached [Blueprint]s can't have
+  /// more additions to it.
+  bool get isAttached => _isAttached;
+}
--- a/packages/flame/lib/components.dart
+++ b/packages/flame/lib/components.dart
@@ -2,6 +2,7 @@
 export 'src/anchor.dart';
 export 'src/collisions/has_collision_detection.dart';
 export 'src/collisions/hitboxes/screen_hitbox.dart';
+export 'src/components/blueprint.dart';
 export 'src/components/component.dart';
 export 'src/components/component_set.dart';
 export 'src/components/custom_painter_component.dart';
--- a/packages/flame/lib/src/game/flame_game.dart
+++ b/packages/flame/lib/src/game/flame_game.dart
@@ -2,6 +2,7 @@ import 'dart:ui';
 
 import 'package:meta/meta.dart';
 
+import '../components/blueprint.dart';
 import '../components/component.dart';
 import '../extensions/vector2.dart';
 import 'camera/camera.dart';
@@ -150,4 +151,10 @@ class FlameGame extends Component with Game {
 
   @override
   Projector get projector => camera.combinedProjector;
+
+  /// Shortcut to attach a [Blueprint] instance to this game
+  /// equivalent to `MyBluepinrt().attach(game)`
+  Future<void> addFromBlueprint(Blueprint blueprint) async {
+    await blueprint.attach(this);
+  }
 }
--- a/packages/flame/test/components/blueprint_test.dart
+++ b/packages/flame/test/components/blueprint_test.dart
@@ -1,0 +1,45 @@
+import 'package:flame/components.dart';
+import 'package:flame/game.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mocktail/mocktail.dart';
+
+class MockGame extends Mock implements FlameGame {}
+
+class MyBlueprint extends Blueprint {
+  @override
+  void build() {
+    add(Component());
+    addAll([Component(), Component()]);
+  }
+}
+
+void main() {
+  group('Blueprint', () {
+    test('components can be added to it', () {
+      final blueprint = MyBlueprint()..build();
+
+      expect(blueprint.components.length, equals(3));
+    });
+
+    test('adds the components to a game on attach', () {
+      final mockGame = MockGame();
+      when(() => mockGame.addAll(any())).thenAnswer((_) async {});
+      MyBlueprint().attach(mockGame);
+
+      verify(() => mockGame.addAll(any())).called(1);
+    });
+
+    test(
+      'throws assertion error when adding to an already attached blueprint',
+      () async {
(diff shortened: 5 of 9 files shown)


packages/flame/lib/src/components/blueprint.dart
```diff
@@ -0,0 +1,50 @@
+import 'package:flutter/foundation.dart';
+
+import '../../components.dart';
+import '../game/flame_game.dart';
+
+const _attachedErrorMessage = "Can't add to attached Blueprints";
+
+// so we can experiment with the idea, but this is a
+// potential upstream change on Flame.
+
```
erickzanardo, line 10:
```suggestion
```

packages/flame/lib/src/components/blueprint.dart
```diff
@@ -0,0 +1,47 @@
...
+import '../../components.dart';
+import '../game/flame_game.dart';
+
+const _attachedErrorMessage = "Can't add to attached Blueprints";
+
+/// A [Blueprint] is a virtual way of grouping [Component]s
+/// that are related, but they need to be added directly on
+/// the [FlameGame] level.
+abstract class Blueprint {
```
spydon, line 11: Could this be a component with `HasGameRef`? But instead of having children, when you add components to it it adds them to the game instance?

spydon (commented):
Maybe this should be added to the experimental directory, because it should be removed as soon as we figure out a global priority system right?

packages/flame/lib/src/components/blueprint.dart
erickzanardo: Yes, this was my first idea, but it bothered me to game a component added to the game tree, that would not be doing anything else once it have loaded.

erickzanardo:
> Maybe this should be added to the experimental directory, because it should be removed as soon as we figure out a global priority system right?

I think we should keep this even when the global priority system is figured out. The rendering order is one of the things that this solves, but I think it is not the only, if you look on the forge2d blueprint, it also groups callbacks, so we can use this a thing that helps us organise our code too

packages/flame/lib/src/components/blueprint.dart
spydon: I think it is better to keep it as a `Component` than to add another `add` method to `FlameGame` when we have just managed to remove all of the other ones from the components (like `addHitbox` and `addEffect`).
It could be useful to put logic in its `update` method that is related to the group too, for example logic relating to the whole house (in the example) and not only specific parts of the house.

erickzanardo: Idk, it feels super weird to me to have a component that its purpose is to call `gameRef.add`, also, it could be confusing IMO, imagine we create `House` component instead that adds the components to the game, if you remove the `House` component, it would not remove "its" children.

By having a group that is not inside the game tree we make a statement that this is a virtual grouping, and we could even add more methods to it to help managing this group, like a remove method for example, that would call `game.remove` for all its components.

erickzanardo:
> I think that it is better to group callbacks with the help of mixins to be able to be more dynamic in the long run (not having to add every method that we want to group to one class), see for example how the [CollisionPassthrough](https://github.com/flame-engine/flame/blob/main/packages/flame/lib/src/collisions/collision_passthrough.dart) works, that one also basically does this, but still supports FCS.

This is an interesting approach, but it requires the user to write a mixin and do some specific logic, with this approach I am proposing, we just enable the user to group Callbacks with their components, so instead of adding the component, then adding the callback, you can only add the blueprint and it will add both things.

If in the future more things should be added for that feature, only the blueprint needs to be changed, basically creating a single entrypoint for that feature.

doc/flame/components.md
```diff
@@ -154,6 +154,48 @@ void update(double dt) {
 ```
 
 
+### Blueprints
+
+Flame also offers a virtual approach for grouping components which allows the developer to keep
+their components grouped "code wise", but the components (or other things, like a
+`ContactCallback` for Forge2d Games) are individually added to the game and there isn't a parent
+component.
```
luanpotter, line 162: avoid some repetition:
```suggestion
Flame also offers a virtual approach for grouping components that allows the developer to keep them together "code wise", but each component (or other things, like a `ContactCallback` for Forge2d Games) are individually added to the game without a parent.
```

luanpotter: arent' `ContactCallback` components?? your impl seems to take only Components and nothing else

erickzanardo: > arent' ContactCallback components?? your impl seems to take only Components and nothing else
Look on the `Forge2dBlueprint` implementation 😉

doc/flame/components.md
```diff
@@ -154,6 +154,48 @@ void update(double dt) {
...
+
+This is useful when you have a complex feature that will require different components, but they
+all need to be on the same rendering level of different faatures, a simple example for this is
+a top down view game, we would like to play with the rendering order of things so we can give the
+player the impression of depth, so if the player character goes behind a house, its roofs is
+rendered "above" the player, to be able to achieve that we would need to split our building
+component into more components, for example: `HouseBase`, `HouseSecondFloor` and `HouseRoof`.
```
luanpotter, line 169:
```suggestion
This is useful when you have several multi-part components that each need to control the rendering level of its parts globally. A simple example would be a top-down game, in which in order to give the impression of depth, when player character goes behind a house, its roof is rendered "above" the player. In this case, you could still "group" all the building components (`HouseBase`, `HouseSecondFloor`, `HouseRoof`) with a Blueprint instead of a parent component, while they are still all siblings to the Player component.
```

doc/flame/components.md
```diff
@@ -154,6 +154,48 @@ void update(double dt) {
...
+
+So to avoid leaving these components "floating" around, we could create a `Blueprint` to group
```
luanpotter, line 171:
```suggestion
So, to avoid leaving these components "floating" around, we could create a `Blueprint` to group them, but still keeping them as siblings of the Player:
```

packages/flame/lib/src/components/blueprint.dart
erickzanardo: I don't think it would be intuitive to do this by a component, the user would have to create a class based on `Component`, add HasGameRef for it, etc, etc. We could provide a component for it maybe, like a `BlueprintComponent` that implements this concept using a component and maybe this would be more straight forward, but this would still have a children attribute that wouldn't be used and etc, that is what kind makes me think that using a component to solve this is a overkill.

luanpotter: also if it's a component you can use Blueprints on any level of the tree, for example you might want to have [Player, HouseTop, HouseBottom] as sibling but within the Stage component, not the root. or camera component, etc

erickzanardo:
>  It also becomes problematic on the blueprint class if you only want some callback types, but not all of the ones that are implemented.

I don't see how this can be problematic, we would only add the callbacks that a related to the Blueprint, A PlayerBlueprint would only add player related callbacks and etc.

packages/flame/lib/src/components/blueprint.dart
spydon: > We could provide a component for it maybe, like a `BlueprintComponent` that implements this concept using a component and maybe this would be more straight forward

Yes! That is my suggestion, I think it would improve this solution a lot. :)

erickzanardo: > Yes! That is my suggestion, I think it would improve this solution a lot. :)

Don't you think this could be hard to track? We would have to override everything related to children on this component so it would access the virtual components instead of its own components.

spydon: > Don't you think this could be hard to track? We would have to override everything related to children on this component so it would access the virtual components instead of its own components.

I think that it is only `add`, `addAll`, `remove` and `removeAll` that we need to override.
They will probably stay the same for some time now (I hope), feels like the `Component` API has stabilized quite well.

spydon: Only `add` and `remove`.*
(Since the others just proxy the calls to `add` and `remove`).

st-pasha (commented):
Why is this called a "Blueprint"?
And what is a "virtual group", and how is it different from a normal group?
How is components lifecycle going to work?
What if the game has non-trivial component structure at the root so that adding things to the root of the game is not feasible?

erickzanardo:
> Why is this called a "Blueprint"?

Because it is not a real component, just one that will "build" others

> And what is a "virtual group", and how is it different from a normal group?

Same answer as above

> How is components lifecycle going to work?

That is what we are going to test as discussed [here](https://github.com/flame-engine/flame/pull/1476#discussion_r831304503)

> What if the game has non-trivial component structure at the root so that adding things to the root of the game is not feasible?

Then you probably shouldn't use this on your feature.

erickzanardo:
@spydon @luanpotter I've put some more thought on this and I would like to suggest the following:

In the end, even though this feature is more than that, the main reason this was needed is because the problem I described regarding the rendering order, all the other groupings can be done inside a Component.

Either way, I think it is nice to have a built-in feature to "non-physically" group components, like the current implementation of this PR proposes.

As far as I got from our discussion here, the biggest concern about the current implementation, is to add a new "add" method to the FlameGame, which was something that we got rid of and I agree that we should keep like that.

So I would like to propose a "middle ground" way.

The `addFromBlueprint` method was supposed to be just a shortcut for the `Blueprint.attach`, let's get rid of it, the user would have to use that `attach` method, that way the `FlameGame` doesn't even knows what a `Blueprint` is.

Lets move this code to the experimental package, so we can make it clear to the user that we are experimenting with it and it could be removed or changed on next versions.

wdyt?

st-pasha:
> The addFromBlueprint method was supposed to be just a shortcut for the Blueprint.attach, let's get rid of it, the user would have to use that attach method, that way the FlameGame doesn't even knows what a Blueprint is.

:+1:

> Lets move this code to the experimental package, so we can make it clear to the user that we are experimenting with it and it could be removed or changed on next versions.

:+1:

> > Why is this called a "Blueprint"?
>
> Because it is not a real component, just one that will "build" others

I still find the name "Blueprint" to be ill-fitting here. Surely, it builds other components like you say, but then it _holds onto them_. It would be weird if you had a blueprint to build a building, then you built that building, then you live in that building but still call it a "blueprint".

I notice that at different places you describe the class as "virtual approach for grouping components", or "feature to 'non-physically' group components". Maybe these descriptions can be used to create a better name? Say, `VirtualGroup`, `VirtualComponent`, `GhostComponent`, `ShadowComponent`, or something like that.

erickzanardo:
> I still find the name "Blueprint" to be ill-fitting here. Surely, it builds other components like you say, but then it holds onto them. It would be weird if you had a blueprint to build a building, then you built that building, then you live in that building but still call it a "blueprint".

I mean, it does holds the components, but that is just how I implemented it, I could refactor it a little bit so it becomes "stateless", I guess it would make more sense to be honest, like the `build` method could return a `BlueprintPlan`, or `BlueprintResult`, etc, and the components would live there, the attach method would call the build method, get the `BlueprintPlan` a add the elements to the game received as parameters. Thoughts?

st-pasha:
This would depend on what functionality you want from this class. If it's a bit of like a "prefab" in Unity, i.e. a spawner of components, then `Blueprint` is a reasonable name. But if you want these components to stay grouped (virtually) in the future and not "float around", then we'd need a new name. A possible reason to live them grouped would be to be able to do some operations for the entire group. Consider `Blueprint house`:
 - remove the group `house.components.forEach((c) => c.removeFromParent)`;
 - find an item in the group `house.components.whereType<Door>().first.open()`;
 - move the group as a whole `house.forEach((c) => c.position += Vector2(100, 0))`.

erickzanardo:
Unity's prefab is pretty much what I have in mind

spydon:
> In the end, even though this feature is more than that, the main reason this was needed is because the problem I described regarding the rendering order, all the other groupings can be done inside a Component.

Wouldn't you want access to the lifecycle methods to be able to act on the lifecycle events with the whole group that you would get if this was a `Component`? 
 
> The `addFromBlueprint` method was supposed to be just a shortcut for the `Blueprint.attach`, let's get rid of it, the user would have to use that `attach` method, that way the `FlameGame` doesn't even knows what a `Blueprint` is.

Sgtm!
 
> Lets move this code to the experimental package, so we can make it clear to the user that we are experimenting with it and it could be removed or changed on next versions.

👍

st-pasha:
Is this is indeed akin to a prefab in Unity, then the API should allow to create multiple "instances" over time. Maybe instead of `attach()` call it `spawn()`?

PR #1483, opened 2022-03-23
fix: Component.propagateToChildren() order of traversal

st-pasha:
# Description

* `propagateToChildren` is supposed to visit components in the order opposite of how they were rendered. However, there was a bug that when called with `includeSelf=true`, the root of the tree will be visited first instead of last.
* Also, there was a bug where, in case of a game-in-game structure, the children of the nested game would be traversed twice.
* Added flag `bool reversed = false` to the `descendants()` method.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -331,25 +331,34 @@ class Component {
 
   /// Recursively enumerates all nested [children].
   ///
-  /// The search is depth-first, hence descendants are in postorder. In other
-  /// words, it explores the first child completely before visiting the next
-  /// sibling.
+  /// The search is depth-first in preorder. In other words, it explores the
+  /// first child completely before visiting the next sibling, and the root
+  /// component is visited before its children.
   ///
-  /// In order to filter descendants it is usually convenient to use [Iterable]
-  /// lazy methods, such as [Iterable.where].
-  Iterable<Component> descendants({bool includeSelf = false}) sync* {
-    if (includeSelf) {
+  /// This ordering of descendants is considered standard in Flame: it is the
+  /// same order in which the components will normally be updated and rendered
+  /// on every game cycle. The optional parameter [reversed] allows iterating
+  /// through the same set of descendants in reverse order.
+  ///
+  /// The [Iterable] produced by this method is "lazy", which means it will only
+  /// traverse the component tree when required. This allows efficient chaining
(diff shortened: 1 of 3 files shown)


spydon (commented):
> However, there was a bug that when called with includeSelf=true, the root of the tree will be visited first instead of last.

Isn't it supposed to be visited first though? Since it is rendered and updated before its children?

packages/flame/lib/src/components/component.dart
```diff
@@ -331,26 +331,35 @@ class Component {
...
-  /// words, it explores the first child completely before visiting the next
-  /// sibling.
-  ///
-  /// In order to filter descendants it is usually convenient to use [Iterable]
-  /// lazy methods, such as [Iterable.where].
-  Iterable<Component> descendants({bool includeSelf = false}) sync* {
-    if (includeSelf) {
+  /// The search is depth-first in preorder. In other words, it explores the
+  /// first child completely before visiting the next sibling, and the root
+  /// component is visited before its children.
```
spydon, line 336: This sounds like BFS to me, not DFS?

packages/flame/lib/src/components/component.dart
```diff
@@ -331,26 +331,35 @@ class Component {
...
+  /// on every game cycle. The optional parameter [reversed] allows iterating
+  /// through the same set of descendants in reverse order.
+  ///
+  /// The [Iterable] produced by this method is "lazy", which means it will only
+  /// traverse the component tree when required. This allows efficient chaining
+  /// of various iterable methods, such as filtering, early stopping, folding,
+  /// and so on -- see the documentation of the [Iterable] class for details.
+  Iterable<Component> descendants({
+    bool includeSelf = false,
+    bool reversed = false,
```
spydon, line 349: Maybe we should have this flag on ancestors too?

spydon (approved):

erickzanardo (approved):

packages/flame/lib/src/components/component.dart
st-pasha: I don't think there is a way to implement this for ancestors more efficiently than `ancestors().toList().reversed`.

spydon: True, since we don't have access to the root that it is attached to. Let's skip that then.

PR #1485, opened 2022-03-24
feat: adding Component.lastChild and Component.firstChild

erickzanardo:
# Description

Adds a simple new method on `Component` to enable easy and safe way to recover the first child of a given type, basically a convenience method for `children.whereType<T>.first`.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -45,6 +45,9 @@ class Component {
   Component? _parent;
 
   /// Returns the first child that matches the given type [T].
+  ///
+  /// As opposed to `children.whereType<T>().first`, this method returns null
+  /// instead of a [StateError] when no matching children are found.
   T? firstChild<T extends Component>() {
     final match = children.whereType<T>();
 
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -49,7 +49,7 @@ class Component {
     final match = children.whereType<T>();
 
     final it = match.iterator;
-    return it.moveNext()? it.current : null;
+    return it.moveNext() ? it.current : null;
   }
 
   /// The children of the current component.
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -44,6 +44,17 @@ class Component {
   Component? get parent => _parent;
   Component? _parent;
 
+  /// Returns the first child that matches the given type [T].
+  T? child<T extends Component>() {
+     final match = children.whereType<T>();
+
+     if (match.isNotEmpty) {
+       return match.first;
+     }
+
+     return match.first;
+  }
+
   /// The children of the current component.
   ///
   /// This getter will automatically create the [ComponentSet] container within
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -49,9 +49,7 @@ class Component {
   /// As opposed to `children.whereType<T>().first`, this method returns null
   /// instead of a [StateError] when no matching children are found.
   T? firstChild<T extends Component>() {
-    final match = children.whereType<T>();
-
-    final it = match.iterator;
+    final it = children.whereType<T>().iterator;
     return it.moveNext() ? it.current : null;
   }
 
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -55,13 +55,8 @@ class Component {
 
   /// Returns the last child that matches the given type [T].
   T? lastChild<T extends Component>() {
-    final match = children.whereType<T>();
-
-    if (match.isNotEmpty) {
-      return match.last;
-    }
-
-    return null;
+    final it = children.reversed().whereType<T>().iterator;
+    return it.moveNext() ? it.current : null;
   }
 
   /// The children of the current component.
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -48,11 +48,8 @@ class Component {
   T? child<T extends Component>() {
     final match = children.whereType<T>();
 
-    if (match.isNotEmpty) {
-      return match.first;
-    }
-
-    return match.first;
+    final it = match.iterator;
+    return it.moveNext()? it.current : null;
   }
 
   /// The children of the current component.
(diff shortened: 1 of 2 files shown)


packages/flame/lib/src/components/component.dart
```diff
@@ -44,6 +44,17 @@ class Component {
...
 
+  /// Returns the first child that matches the given type [T].
+  T? child<T extends Component>() {
+    final match = children.whereType<T>();
+
+    if (match.isNotEmpty) {
+      return match.first;
+    }
+
+    return match.first;
```
st-pasha, line 55: Shouldn't this be `return null`?

Anyways, a more efficient implementation would be
```dart
final it = match.iterator;
return it.moveNext()? it.current : null;
```
(this avoids running the iterator twice).

packages/flame/test/components/component_test.dart
```diff
@@ -353,10 +353,33 @@ void main() {
...
+          expect(nonExistentChild, isNull);
+        },
+      );
     });
   });
 }
 
+class ComponentA extends Component {}
+
+class ComponentB extends Component {}
```
spydon, line 381: Are these two needed? Can't you just create new components in the test directly?

packages/flame/lib/src/components/component.dart
```diff
@@ -44,6 +44,17 @@ class Component {
   Component? get parent => _parent;
   Component? _parent;
 
+  /// Returns the first child that matches the given type [T].
+  T? child<T extends Component>() {
```
spydon, line 48: I think the name is a bit unclear, but I think the method will be very useful!

packages/flame/test/components/component_test.dart
spydon: Ofc they are needed, I should stop reviewing things right after I have woken up in the morning. 😅

packages/flame/lib/src/components/component.dart
erickzanardo: Any suggestions? My first idea was `get`, but I thought it was too generic, `child` I think kind of goes well since with have an `children`.

My other ideas were:
 - findChild
 - getChild
 - fetchChild
 - firstChild

packages/flame/lib/src/components/component.dart
erickzanardo, line 55: Good idea.

Note to self: never to PRs half sleep
```suggestion
  final it = match.iterator;
  return it.moveNext()? it.current : null;
```

packages/flame/lib/src/components/component.dart
erickzanardo: Yeah, I need a simple and safe way to recover an specific child of a component, doing `children.whereType<Bla>().first`, even though is not that long, starts to became repetitive after a couple of places that you write it and I feel that it isn't very much safe as it will throw and error is it can't find the child, so by returning null it makes that a little bit easier too as it is explicit that the child may not exist

spydon: Maybe `firstChild` could be a good name?

erickzanardo: I like `firstChild` over `findChild` because it kind makes it more clear what this method does

spydon: Lol, `firstChild` was even in your list of suggestions, missed that 😅

alestiago:
Personally, I'm okay with delegating this functionality to the collection. In other words, I'm happy with `.children.whereType<T>.first`.

I see how can this be very verbose. Personally, I think another great option would be `component.children.first<T>`. Thoughts?

In addition, if we have a `firstChild` I'll also expect a `lastChild`. And since this is now looking just like how one manipulated the HTML DOM we could potentially also have `nextSibling`. Thoughts?

st-pasha:
I believe this PR is more about getting the _single_ child of type `T`. 

Except that we don't want to create unnecessary exceptions, so if there aren't any `T` children, the method returns `null` (as opposed to standard `.first` which throws an exception), and if there are multiple `T` children, the method returns the first one (as opposed to standard `.single` which throws an exception).

alestiago:
Of course! I don't see the need of making the PR any larger with `lastChild` and `nextSibling`. Was just dropping some ideas and my two cents. Someone landing on Flame with some DOM background might expect those methods if `firstChild` is a thing. That doesn't mean we have to include them though.

I think a brief section of:
> Except that we don't want to create unnecessary exceptions, so if there aren't any T children, the method returns null (as opposed to standard .first which throws an exception), and if there are multiple T children, the method returns the first one (as opposed to standard .single which throws an exception).

Could be used to enhance the documentation since I bet someone that first looks at this method might possibly wonder this. For example, my future old self (👴) coming back in a couple of years' time without having looked at the Dart language and Flame in quite a while.

packages/flame/lib/src/components/component.dart
```diff
@@ -44,6 +44,14 @@ class Component {
   Component? get parent => _parent;
   Component? _parent;
 
+  /// Returns the first child that matches the given type [T].
```
alestiago, line 47:
```suggestion
  /// Returns the first child that matches the given type [T].
  ///
  /// As opposed to [children.whereType<T>().first] this method returns null
  /// instead of a [StateError] when no matching children are found.
```

erickzanardo:
renamed to `firstChild` and added `lastLast` for symmetry

packages/flame/lib/src/components/component.dart
```diff
@@ -44,6 +44,25 @@ class Component {
...
+  T? firstChild<T extends Component>() {
+    final match = children.whereType<T>();
+
+    final it = match.iterator;
+    return it.moveNext() ? it.current : null;
+  }
+
+  /// Returns the last child that matches the given type [T].
+  T? lastChild<T extends Component>() {
+    final match = children.whereType<T>();
```
spydon, line 57: You can do `children.reversed` here to not have to go through the full list (since reversed is already cached).

spydon (approved):

packages/flame/lib/src/components/component.dart
```diff
@@ -44,6 +44,25 @@ class Component {
   Component? get parent => _parent;
   Component? _parent;
 
+  /// Returns the first child that matches the given type [T].
+  T? firstChild<T extends Component>() {
+    final match = children.whereType<T>();
+
+    final it = match.iterator;
```
st-pasha, line 51:
```suggestion
    final it = children.whereType<T>().iterator;
```

Issue #1488, opened 2022-03-24
ParticleComponent should be a PositionComponent


erickzanardo:
# What could be improved
ParticleComponent if a plan Component, which makes it hard to position when there are camera transformations on the game.

# Why should this be improved
So particle components could easily used alongside the camera system

# Any risks?
This will kind of a minor breaking change.

# More information

spydon:
A risk that I see is that the `PositionComponent` might be a bit heavier component than the `ParticleComponent`, on the other hand `PositionComponent` has gained a lot of optimizations lately. It would be interesting to see if it would make a difference spawning a large amount of  `PositionComponent`s compared to `ParticleComponent`s. If it doesn't make a difference, I think this sounds like a great idea!

erickzanardo:
Cool, I am gonna work on this soon then!

st-pasha:
@spydon `ParticleComponent` is a wrapper around `Particle` (which is not a component), so a single `ParticleComponent` may hold hundreds if not thousands of smaller particles inside. Which means there is little performance loss in making `ParticleComponent` slightly heavier, since this weight will be applied only once to the whole particle system.

Perhaps the name `ParticleComponent` is a bit confusing? Because it seems like there could be one component for every particle, but that's not the case.

spydon:
Right had completely forgotten it was like that, it was a looong time since I looked at the particle system, and I think there barely has been any changes to it for almost two years. Maybe just `ParticlesComponent` then? Feels like most names describing what it does would become very long.

st-pasha:
hmm, how about `ParticleSystem`?

erickzanardo:
I like ParticlesComponent, but it will be a big breaking change instead of a small one, will that be ok?

st-pasha:
We could make a rename, and leave the old name as a deprecated typedef. That way it won't be a breaking change.

erickzanardo:
Yeah, makes sense, I wouldn't remove the `Component` suffix though, So I think we could choose between

 - `ParticlesComponent`
 - `ParticlesSystemComponent`

Which one looks better? Or any other options?

spydon:
The last one would be `ParticleSystemComponent` in that case I think (without the plural s on `Particle`).

alestiago:
I feel like "System" is such a big word. Personally, who knows what a "System" means in what context! 

[...]

PR #1489, opened 2022-03-25
feat: adding ParticleSystemComponent

erickzanardo:
# Description

Adds a new component to handle particles into FCS and deprecates the old one

Solves #1488 

## Checklist

## Breaking Change

### DIFF
--- a/doc/flame/rendering/particles.md
+++ b/doc/flame/rendering/particles.md
@@ -1,21 +1,21 @@
 # Particles
 
 Flame offers a basic, yet robust and extendable particle system. The core concept of this system is
-the `Particle` class, which is very similar in its behavior to the `ParticleComponent`.
+the `Particle` class, which is very similar in its behavior to the `ParticleSystemComponent`.
 
 The most basic usage of a `Particle` with `FlameGame` would look as following:
 
 ```dart
 import 'package:flame/components.dart';
 
-// ... 
+// ...
 
 game.add(
-  // Wrapping a Particle with ParticleComponent
+  // Wrapping a Particle with ParticleSystemComponent
   // which maps Component lifecycle hooks to Particle ones
   // and embeds a trigger for removing the component.
-  ParticleComponent(
-    CircleParticle(),
+  ParticleSystemComponent(
+    particle: CircleParticle(),
   ),
 );
 ```
@@ -40,15 +40,15 @@ Random rnd = Random();
 Vector2 randomVector2() => (Vector2.random(rnd) - Vector2.random(rnd)) * 200;
 
 // Composition.
-// 
-// Defining a particle effect as a set of nested behaviors from top to bottom, one within another: 
-// ParticleComponent 
-//   > ComposedParticle 
-//     > AcceleratedParticle 
+//
+// Defining a particle effect as a set of nested behaviors from top to bottom, one within another:
+// ParticleSystemComponent
+//   > ComposedParticle
+//     > AcceleratedParticle
 //       > CircleParticle
 game.add(
-  ParticleComponent(
-    Particle.generate(
+  ParticleSystemComponent(
+    particle: Particle.generate(
       count: 10,
       generator: (i) => AcceleratedParticle(
         acceleration: randomVector2(),
@@ -61,12 +61,12 @@ game.add(
 );
 
 // Chaining.
-// 
+//
 // Expresses the same behavior as above, but with a more fluent API.
 // Only Particles with SingleChildParticle mixin can be used as chainable behaviors.
 game.add(
-  ParticleComponent(
-    Particle.generate(
+  ParticleSystemComponent(
+    particle: Particle.generate(
       count: 10,
       generator: (i) => pt.CircleParticle(paint: Paint()..color = Colors.red)
     )
@@ -74,19 +74,19 @@ game.add(
 );
 
 // Computed Particle.
-// 
+//
 // All the behaviors are defined explicitly. Offers greater flexibility
 // compared to built-in behaviors.
 game.add(
-  ParticleComponent(
-      Particle.generate(
+  ParticleSystemComponent(
+      particle: Particle.generate(
         count: 10,
         generator: (i) {
           Vector2 position = Vector2.zero();
           Vector2 speed = Vector2.zero();
           final acceleration = randomVector2();
           final paint = Paint()..color = Colors.red;
-          
+
           return ComputedParticle(
             renderer: (canvas, _) {
               speed += acceleration;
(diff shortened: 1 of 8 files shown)


erickzanardo:
cc @alestiago

packages/flame/lib/src/components/particle_system_component.dart
```diff
@@ -0,0 +1,53 @@
...
+import '../../components.dart';
+import '../../particles.dart';
+
+/// {@template particle_system_component}
+/// A [PositionComponent] that renders a [Particle] at the designated
+/// position, scaled to have the designated size and rotated to the specified
+/// angle.
+/// {endtempalte}
+class ParticleSystemComponent extends PositionComponent {
+  Particle? particle;
```
st-pasha, line 12: Why is this nullable?

packages/flame/lib/src/components/particle_component.dart
```diff
@@ -7,6 +7,7 @@ import 'component.dart';
 /// to a [Component] tree. Could be added either to FlameGame
 /// or an implementation of [Component].
 /// Proxies [Component] lifecycle hooks to nested [Particle].
+@Deprecated('Will be removed after v1.1, use ParticleSystemComponent instead')
 class ParticleComponent extends Component {
```
st-pasha, line 11: This could have simply been 
```
class ParticleSystemComponent extends PositionComponent {
```

st-pasha:
Title: ~~adding ParticleSystemComponent~~ ParticleSystemComponent now inherits from PositionComponent

spydon:
> Title: ~~adding ParticleSystemComponent~~ ParticleSystemComponent now inherits from PositionComponent

That sounds like there existed a component with that name previously though. Maybe just change "now" to "that"?

packages/flame/lib/src/components/particle_component.dart
spydon: Then we couldn't have deprecated it, it would have been a breaking change right? And as you said previously, the name was confusing.

doc/flame/rendering/particles.md
```diff
@@ -1,21 +1,21 @@
...
 
 game.add(
-  // Wrapping a Particle with ParticleComponent
+  // Wrapping a Particle with ParticleSystemComponent
   // which maps Component lifecycle hooks to Particle ones
   // and embeds a trigger for removing the component.
-  ParticleComponent(
-    CircleParticle(),
+  ParticleSystemComponent(
+    particle: CircleParticle(),
```
spydon, line 18: This carries a bit of the confusing part in the other component, maybe this could be called `particles`?

spydon (approved):

packages/flame/lib/src/components/particle_component.dart
st-pasha: `typedef ParticleComponent = ParticleSystemComponent;` fixes it. I think we did this multiple times before, didn't we?

spydon: It would work if it is only a name change, but the changes in the class will be breaking right?

erickzanardo: And eventually ParticleComponent will be removed só better to properly warn the user about that

packages/flame/lib/src/components/particle_system_component.dart
erickzanardo: So it can be loaded later, removed, etc. it is the same what we do on other components

packages/flame/lib/src/components/particle_component.dart
spydon: > And eventually ParticleComponent will be removed só better to properly warn the user about that

That wouldn't be a problem when having the deprecated typedef though.

erickzanardo: That is true. But either way the classes are not equivalent anymore. The new one is more in line with the rest of our components, so typedef don't work for this situation

doc/flame/rendering/particles.md
erickzanardo: I wonder if that won't cause another type of confusion, `particles` would give me the idea that this would be a list, and since how our particle system is implemented, a `Particle` instance can hold multiple particles, so maybe naming this `particle` is fine? Wdyt?

spydon: True, `particleType` or something maybe?

erickzanardo: `particleType` sounds a little bit weird, I would expect it be an enum or something, but this is the actually value.

Maybe `value`? It maybe a little too generic but maybe better?

spydon: True, what about `particleGenerator` then? 😄

erickzanardo: That crossed my mind, but that will in turn gives me the idea that this would be a function hahahaha

Also that is a particle accelerator right? :P

spydon: > Also that is a particle accelerator right? :P

The days that it is lucky it is both. 😄

Hmm... `particleSpawner`, `spawn` or something like that then?

erickzanardo: `particleSpawner` I liked the most so far

erickzanardo: generator do bring me the strong idea that this would be a function even more because `Particle` class has a generate factory that receives a named parameter called generator.

So spawmer would be better here IMO, leaving unnamed would make this not follow the standards of the other components, so my vote goes to spawmer or particleSpawmer

st-pasha: but it doesn't spawn anything

erickzanardo: it does 🤔 , it is that class that spawn the particles

st-pasha: Umm, it is a `Particle` object; usually a `CompositeParticle` so that it has many other particles inside. But it is a fully-made particle system.

Of course, it is possible to create a `Particle` that would self-mutate over time and spawn other particles inside, but we don't have any such functionality implemented yet.

erickzanardo: >  it is a Particle object

Yeah, that is why it was originally named `particle`, maybe we should just stick to that...

st-pasha: > it was originally named particle, maybe we should just stick to that...

I would prefer `particle` over `particleSpawner` or `particleGenerator`. However, maybe we could revisit the rule that all arguments must be named, and allow an exception for the case when the meaning of the argument is unambiguous from its type. So that we could write `GameWidget(game)` instead of `GameWidget(game: game)`, and `SpriteComponent(sprite)` instead of `SpriteComponent(sprite: sprite)`, and `ParticleSystemComponent(particle)` instead of `ParticleSystemComponent(particle: particle)`.

alestiago (approved):

doc/flame/rendering/particles.md
spydon: Yeah, sounds like the best option!

spydon (approved):

Issue #1452, opened 2022-03-14
Add Game.backgroundDecoration property


st-pasha:
Based on discussion in #1448, we would like to have the `backgroundColor` property of a `Game` to be dynamically changeable. One problem with that, however, is that right now `backgroundColor` is a method, not a property, and I don't think there is a way to convert a method into a property without making a breaking change.

So, instead of changing `backgroundColor`, I suggest to add a coexisting property `backgroundDecoration`, while deprecating `backgroundColor` method. The `backgroundDecoration` property will work like this:
- by default it will return null;
- if it returns an object, that object will be used as the `decoration:` property of the widget;
- if it returns null, then we will call `backgroundColor()` and construct a decoration box out of that color;
- the `backgroundColor()` method will be deprecated.

Note that `backgroundDecoration` allows more complicated backgrounds than simple color. For example, you can implement gradient colors through it.

spydon:
How can it be dynamically changeable? Since both `color` and `decoration` are set on the widget and we can't refresh the widget?
I think if you'd want to change it mid-game it is better that the user does it at a widget level with state management than from inside of the game.

st-pasha:
You can refresh the widget, since it's stateful.

st-pasha:
> I don't think we want to do that though, the state of the widget should lay on the Flutter side, not inside of the game.

The state of the widget belongs to the `GameWidget` class, and `Game` has a means of communicating with its widget. There is nothing new here, we're already doing that.

> And if you initialize your game in the widget then your game will restart when you change the background color.

The `game` is a `final` property inside the `GameWidget`, which means it is impossible to "restart" it from within the `GameWidget`. Not that we wanted to.

wolfenrain:
Weird idea, but can we not just do a `canvas.drawRect(Vector2.zero() & size, Paint()..color = backgroundColor())` (with some caching etc) at the start of our render pipeline? I mean even Flutter eventually uses some form of canvas drawing to draw the background color of a widget, why should we let a widget do that for us if we can provide it "easily" ourselves?

spydon:
> The state of the widget belongs to the GameWidget class, and Game has a means of communicating with its widget. There is nothing new here, we're already doing that.

You're right, saw that we are handling the `mouseCursor` in the same way as this would be done.
Then I think this is a better idea (if we'd want this feature) than to draw a rectangle on the canvas manually.

st-pasha:
[...]

PR #1498, opened 2022-03-28
fix!: Game.mouseCursor and Game.overlays can now be safely set during onLoad

### LINKED ISSUE #1493: Error when setting Game.mouseCursor during onLoad
```dart
  @override
  Future<void> onLoad() async {
    mouseCursor.value = MouseCursor.uncontrolled;
  }
}

void main() {
  runApp(GameWidget(game: MyGame()));
}
```
produces exception
```
The following assertion was thrown while dispatching notifications for ValueNotifier<MouseCursor?>:
setState() or markNeedsBuild() called during build.
```

st-pasha:
# Description

Refactor state management between the `GameWidget` and a `Game` instance. Now instead of having multiple properties each having its own state notifications, the `Game` exposes a unified interface to signal that the internal state has changed. That interface makes sure that changing Game's properties during `onLoad()` is safe.

The `mouseCursor` property is now a plain property with a setter, not a `ChangeNotifier`. Consequently, instead of writing `game.mouseCursor.value = ...`, one needs to write `game.mouseCursor = ...` now.

## Checklist

## Breaking Change

## Related Issues

Closes #1493
WIP for #1452


packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -210,34 +165,17 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
...
-
-  void addOverlaysListener() {
-    widget.game.overlays.addListener(onChangeActiveOverlays);
-    initialActiveOverlays = widget.game.overlays.value;
-  }
-
-  void removeOverlaysListener(T game) {
-    game.overlays.removeListener(onChangeActiveOverlays);
+  void onGameStateChange() {
+    setState(() {});
```
spydon, line 178: Hmm... Is this really kosher?

st-pasha: What `setState()` does is, removing all the asserts:
```dart
fn();
_element!.markNeedsBuild();
```
so passing a no-op function seems fine to me, considering that all we want to do is to schedule a rebuild of the `GameWidget`.

erickzanardo (changes requested):
I am not sure about doing a refactor and breaking change due to an issue that could have a simple fix.

This part of the code is kind sensitive as there are many edge cases, so any refactor here should be done with a lot of care and a lot of testing.

If I understood correctly, to fix the original issue, we just need to check the widget's [mounted](https://api.flutter.dev/flutter/widgets/State/mounted.html) attribute before calling a `setState`.

st-pasha:
> I am not sure about doing a refactor and breaking change due to an issue that could have a simple fix.

The refactor is useful in itself, it's just that we don't usually create issues for refactors. What this refactor does is that it notices that there are 2 (3) `Game` properties that communicate with the `GameWidget`, and it would be less code, and less bugs, to do this through a common mechanism.

Current line count for the PR is +56/-102, so it means this approach saves at least 50 lines.

As for the breaking change, the only thing that is breaking here is that instead of `game.mouseCursor.value` we now say `game.mouseCursor`. Which is arguably an improvement. If not, this particular part of the PR can be undone -- but I don't think there is a way to effect the change to `mouseCursor` through the normal deprecation process.

> This part of the code is kind sensitive as there are many edge cases, so any refactor here should be done with a lot of care and a lot of testing.

Agree. Perhaps I should add some more tests here? Do you have any recommendations on which specific scenarios to test?
So far, this PR only relied on the existing tests, plus I checked that the examples are working as expected.

erickzanardo:
I do think the API proposal is an improvement! Not needing to to `.value` definitely makes it better to read, write, etc.

But the new API will:
 - Be a breaking change
 - Require more tests

Meanwhile this bug is already affecting the users, and its fix should be easy, so my proposal here is:
 - Lets fix the bug with the current source, without generating a breaking change
 - Lets write more tests for these features
 - Lets release a patch with the fix

Then in a separate issue/PR we can refactor. Ideally we will want to deprecate the current way so users have a warning before having their code broke.

st-pasha:
@erickzanardo I've just tried your suggestion about checking `GameWidget.mounted` like this:
```dart
  void onChangeMouseCursor() {
    if (mounted) {
      setState(() {
        _mouseCursor = widget.game.mouseCursor.value;
      });
    }
  }
```
However, this didn't work -- the error remains.

I believe the problem is that `.mounted` is very broad: it's set to true before `initState()`, and then to false after `dispose()`. Which means that during `build()` this property is true, so `setState()` is still called, and then an exception is thrown because `setState()` is not allowed to be called during `build()`.

So, ultimately, we need some reliable mechanism to stop any calls to `setState()` during `build()`, and then if they do happen, schedule a new build right after the current build finishes.

The approach taken in this PR is probably not very reliable either. Maybe we should add some explicit flags like `isBuilding` into the game widget, WDYT?

packages/flame/lib/src/game/mixins/game.dart
```diff
@@ -269,72 +270,78 @@ mixin Game {
...
   VoidCallback? resumeEngineFn;
 
-  /// A property that stores an [ActiveOverlaysNotifier]
+  /// A property that stores an [_ActiveOverlays]
   ///
-  /// This is useful to render widgets above a game, like a pause menu for
-  /// example.
-  /// Overlays visible or hidden via [overlays].add or [overlays].remove,
-  /// respectively.
+  /// This is useful to render widgets above a game, such as a pause menu.
```
spydon, line 275:
```suggestion
  /// This is useful to render widgets on top of a game, such as a pause menu.
```

spydon (approved):
I think this solution looks pretty solid, it probably even solves some bugs that existed before with the user changing things during a build right?

wolfenrain (approved):

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -151,55 +142,61 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
...
+  /// raised during the build.
+  ///
+  /// This is needed because our build function invokes user code, which in turn
+  /// may change some of the [Game]'s properties which would require the
+  /// [GameWidget] to be rebuilt. However, Flutter doesn't allow widgets to be
+  /// marked dirty while they are building. So, this method is needed to avoid
+  /// such a limitation and ensure that the user code can set [Game]'s
+  /// properties freely, and that they will be propagated to the [GameWidget]
+  /// at the earliest opportunity.
+  Widget _protectedBuild(Widget Function() build) {
```
erickzanardo, line 164: This function concerns me.

A `Widget.build` method should be pure, so it shouldn't trigger any side effects.

I am not sure I completely understand why it is needed, before we did everything on the `initState` can't we keep doing that?

erickzanardo: Yeah, I get it, but I don't like the idea of those side effects, maybe we should completely disallow user to do those modifications on build time of the game widget, and provide initial properties for setting the initial values (which is something I believe we already do?). And then just throw meaningful messages when the user tries to change something that they shouldn't, like we do with the `hasLayout` method.

st-pasha: I understand that this makes the code logic slightly more complicated and "less pure". So it's a detriment to us, developers. At the same time, from the user's perspective, the ability to set up their `Game` from within the class itself -- is a very welcome one. Setting these same properties from `GameWidget` is not as convenient, since it breaks encapsulation of that logic. 

So, it really comes down as a choice between developer convenience and user convenience.

Another difficulty with "completely disallowing user to do these modifications during build time" is that _build time_ is not a well-defined concept within Flame. If it was just the case that the `GameWidget` is built once and for all, then it might be feasible.  But the widget is constantly rebuilt, either due to changes in system environment, or because of changes in the surrounding widget tree, or because of changes coming internally from the `Game`. So how do we explain all these complexities to the user? Do we even understand them ourselves?

The solution in this PR ensures that the widget works regardless of the external or internal sequence of events.

erickzanardo: > The solution in this PR ensures that the widget works regardless of the external or internal sequence of events.

It does, but the solution relies on build side effects, which is the problem for me.

I wonder if we couldn't queue the events when the game is not attached yet, and the. Trigger them on the game attachment

st-pasha: > I wonder if we couldn't queue the events when the game is not attached yet, and then trigger them on the game attachment

This is kinda what this PR is doing, isn't it? The events are "queued" while the game is being built, and then re-triggered at the end of the build. 
Handling them at the end of attachment wouldn't work because a build can trigger multiple times during the course of the game, and we want those builds to be protected too.

> It does, but the solution relies on build side effects, which is the problem for me.

The only real "side effect" here is that a build can schedule a future rebuild if there were modifications to `Game` instance during that build. I'm not sure how this can be considered a "problem" if the alternatives -- either to do nothing and have the changes not being reflected in the UI, or to let the game crash (the current behavior) -- seem much worse.

erickzanardo: Any changes to a state from inside the build is a side effect, that is highly discouraged, [this](https://codewithandrea.com/articles/side-effects-flutter/) simplifies the problem but it can give a rough idea.

If the only way to solve the issue is by the way it is implemented now, I would rather keep the user informing the initial mouse cursor and overlays on the game widget and the user can only change those after the game is attached.

But I will spend some time taking a look on this and testing a few ideas.

erickzanardo: > It seems like this could be a situation were the rules can be bent.

It may be indeed! But I want to explore a little bit more before we go on that path, I will report back here.

erickzanardo: Okay, I did some research, and it seems that having things like this on more complicated scenarios is not that bad, provider for example does a bunch of stuff in the same idea that were implemented here. So I think I am okay with going with this approach.

erickzanardo (approved):

PR #1501, opened 2022-03-29
feat: flame tests can now generate golden tests

erickzanardo:
# Description

The current implementation of the `widgetTest` had a conflict when `expectLater` (which is used for golden tests) because it has an internal `runAsync` method. So to solve this, the `verify` method was split into `setUp` when the components should be added to the game and `verify` where the actual assertions should go.

## Checklist

## Breaking Change

## Related Issues


erickzanardo:
cc @felangel

packages/flame_test/example/pubspec.yaml
```diff
@@ -12,7 +12,8 @@ dependencies:
     sdk: flutter
   flame: ^1.1.0
 dev_dependencies:
-  flame_test: ^1.2.0
+  flame_test:
+    path: ../
```
wolfenrain, line 16: Hmm I think you need to run a `melos clean` right?

packages/flame_test/lib/src/flame_test.dart
```diff
@@ -116,10 +120,26 @@ class GameTester<T extends Game> {
...
+  /// which runs inside the Flutter test environment.
+  ///
+  /// Use [setUp] closure to prepare your game instance (e.g. add components to
+  /// it)
+  /// Use [verify] closure to make verifications/assertions.
+  @isTest
+  void testGameWidget(
+    String description, {
+    WidgetSetupFunction<T>? setUp,
+    WidgetVerifyFunction<T>? verify,
```
wolfenrain, line 141: The verify should be required right?

packages/flame_test/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ version: 1.2.0
 homepage: https://github.com/flame-engine/flame/tree/main/packages/flame_test
 
 environment:
-  sdk: ">=2.14.0 <3.0.0"
+  sdk: ">=2.15.0 <3.0.0"
```
wolfenrain, line 7: Is this constraint upgrade needed for something? Because flame itself still has the `2.14` constraint and if for some reason someone is on `2.14` they can no longer test their games easily.

packages/flame_test/test/golden_test.dart
```diff
@@ -0,0 +1,30 @@
+import 'package:flame/components.dart';
+import 'package:flame/game.dart';
+import 'package:flame_test/flame_test.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('Golden tests', () {
```
wolfenrain, line 8: Woot! This is awesome to be able to do! I never saw a game engine that had the ability to easily write golden tests as well.

wolfenrain (approved):
LGTM, left a few nits

packages/flame_test/pubspec.yaml
erickzanardo: Yep, so we can use constructors tear off, should we increase on flame too (on a separate PR probably)?

packages/flame_test/example/pubspec.yaml
erickzanardo: I think that melos don't handle examples :/

packages/flame_test/lib/src/flame_test.dart
erickzanardo: We didn't required it on the previous API, so I kept it the same way, not sure if we should require on this new API but usually I don't testing libraries to require this, I think `bloc_test` also don't require it.

packages/flame/test/game/mixins/has_draggables_test.dart
```diff
@@ -79,21 +79,25 @@ void main() {
...
         expect(game.handledOnDragUpdate > 0, isTrue);
         expect(game.handledOnDragCancel, 0);
       },
     );
 
-    withDraggables.widgetTest(
+    withDraggables.testGameWidget(
       'drag outside of component is not registered as handled',
-      (game, tester) async {
+      setUp: (game, _) async {
```
alestiago, line 97: I love how this is similar to bloc_test 😉

alestiago (approved):
Love this addition!

Just wondering if the `setUp` or `verify` fails where/how can I pick up the errors?

felangel (approved):

st-pasha (commented):
Golden files are great; but there should be documentation for how to actually create these tests, and how to create/update the golden files.

packages/flame_test/pubspec.yaml
st-pasha: Why not raise to 2.16 while we're at it? 
An important caveat though is that raising the SDK version is a **big** breaking change, so at the very least it has to be in a standalone PR so that it would be prominently visible in the changelog.

wolfenrain: Agreed, would keep it at 2.14 for now and then open a PR with the upgrade changes where we can discuss them more in depth.

packages/flame_test/lib/src/flame_test.dart
```diff
@@ -116,10 +120,26 @@ class GameTester<T extends Game> {
...
+  }
+
+  /// Creates a [Game] specific test case with given [description]
+  /// which runs inside the Flutter test environment.
+  ///
+  /// Use [setUp] closure to prepare your game instance (e.g. add components to
+  /// it)
+  /// Use [verify] closure to make verifications/assertions.
+  @isTest
+  void testGameWidget(
```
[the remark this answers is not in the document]
erickzanardo: Not really, just tested and I can still filter a single test:

```
flutter test --name "renders correctly"
00:04 +1: All tests passed!
```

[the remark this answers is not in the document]
st-pasha: Screen Shot 2022-03-30 at 10 02 37 AM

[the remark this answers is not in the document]
erickzanardo: Screenshot 2022-03-30 at 15 01 50

Probably an issue with your IDE/Plugin/etc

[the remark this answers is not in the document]
st-pasha: That's Android Studio, one of the two [most popular IDEs](https://blog.logrocket.com/best-ides-flutter-2022/) for Flutter. I don't have exact statistics, but it probably has around 50% market share.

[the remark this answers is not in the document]
wolfenrain: > That's Android Studio, one of the two [most popular IDEs](https://blog.logrocket.com/best-ides-flutter-2022/) for Flutter. I don't have exact statistics, but it probably has around 50% market share.

It is still a problem with the Android Studio Flutter plugin in then, because this should be expected behavior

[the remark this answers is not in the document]
st-pasha: The description of the `@isTest` annotation says:
> /// Used to annotate a test framework function that runs a single test.
///
/// Tools, such as IDEs, can show invocations of such function in a file
/// structure view to help the user navigating in large test files.
///
/// The first parameter of the function must be the description of the test.

There is no mention of this being applicable to methods; so it is even not clear that Android Studio's behavior qualifies as a bug.

But regardless of whether it's a bug or a missing feature, it is unlikely to be fixed soon. In the meanwhile, close to half of all Flame users will be affected by this problem.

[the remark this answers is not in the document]
spydon: > close to half of all Flame users will be affected by this problem.

I doubt that half of all Flame users run our tests, I wish it was true though. :wink:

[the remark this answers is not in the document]
st-pasha: `flame_test` is a public library though; so any user who writes tests for their own game can be affected. I concede though that it is probably less common than we'd like.

Still, this is a separate problem, one that we can address at the documentation level. For example, if all our tutorials included discussions on how to add tests, or if we had a separate tutorial dedicated just to writing tests, then there would be much more uptake.

PR #1508, opened 2022-04-01
fix: Bug with anchor parameter in Sprite.render()

st-pasha:
# Description

The `anchor` parameter in `Sprite.render()` was applied incorrectly.

## Checklist

## Breaking Change


spydon (approved):
Interesting that this hasn't been noticed earlier, must have been like this for a long time.

packages/flame/test/sprite_test.dart
```diff
@@ -0,0 +1,50 @@
...
+
+  @override
+  Future<void> onLoad() async {
+    sprite = Sprite(await loadFlameImage());
+  }
+
+  @override
+  void render(Canvas canvas) {
+    canvas.drawRect(
+      Rect.fromLTRB(0, 0, size.x, size.y),
```
luanpotter, line 41: I think there was an ext function to convert size to rect, not sure though

packages/flame/test/_assets/images.dart
```diff
@@ -0,0 +1,377 @@
+import 'dart:ui';
+
+import 'package:flame/assets.dart';
+
+const flameImageEncoded = '''
```
luanpotter, line 5: cant we use image files on tests?

luanpotter (approved):
LGTM, nice catch

packages/flame/test/sprite_test.dart
```diff
@@ -0,0 +1,50 @@
...
+    FlameTester(() => FlameGame()).testGameWidget(
+      'Render with anchor',
+      setUp: (game, tester) async {
+        game.add(MyComponent()..position = Vector2.all(25));
+        await game.ready();
+      },
+      verify: (game, tester) async {
+        await expectLater(
+          find.byGame<FlameGame>(),
+          matchesGoldenFile('_goldens/sprite_test_1.png'),
```
erickzanardo, line 22: So we keep consistent with the other folders we created on flame_test.

```suggestion
          matchesGoldenFile('goldens/sprite_test_1.png'),
```

packages/flame/test/sprite_test.dart
```diff
@@ -0,0 +1,50 @@
...
+import 'package:flame/game.dart';
+import 'package:flame_test/flame_test.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import '_assets/images.dart';
+
+void main() {
+  group('Sprite', () {
+    FlameTester(() => FlameGame()).testGameWidget(
```
erickzanardo, line 13:
```suggestion
    FlameTester(FlameGame.new).testGameWidget(
```

packages/flame/test/_assets/images.dart
st-pasha: It should be possible but I don't know how. Any suggestions?

packages/flame/test/sprite_test.dart
st-pasha: I wanted to keep some distinction between the "regular" directories that match the structure of the `lib/src/` folder, and "utility" directories that host additional content that's needed for tests only.

In `flame_test` we don't have any folder structure right now, so no inconsistency introduced.

packages/flame/test/sprite_test.dart
st-pasha: This requires Dart 2.15+, and we're still on 2.14.

packages/flame/test/sprite_test.dart
erickzanardo: golden is a very specific name already. So it will be distinct. The inconsistency is about the name of the folder containing the snapshots, not about the structure though, so IMO the inconsistency is introduced yes.

I would rather keep our golden folders without the `_`.

packages/flame/test/sprite_test.dart
st-pasha: You're right, it's `size.toRect()`

packages/flame/test/sprite_test.dart
st-pasha: Sorry, I don't follow your argument -- you're suggesting that the name with underscore is inconsistent with some other use of golden snapshots, but AFAIK the only other place that contains a golden file has no folder at all.

The reason I wanted an underscore is in order to keep a distinction between folders with tests and other auxiliary directories. For example, `_assets` folder to keep assets that may be needed for tests. This cannot be called `assets`, because we already have a test directory with such a name. Also, in the future we can add `_components` to contain certain components that are reused across multiple tests. That one also has to have a prefix in order not to confuse it with the regular `components`. Thus, the name `_goldens` is consistent to these directories, and removing the underscore would introduce inconsistency.

erickzanardo: Just went to confirm and the `flame_test` test folder don't have the folder, the place where I introduced the folder was the test on the example: https://github.com/flame-engine/flame/tree/main/packages/flame_test/example/test/goldens, I will revisit flame_test golden location next time I change something there.

My point is that we don't need the `_` to make a distinction, the name `golden` is super specific already, adding the _ just introduces an unnecessary character.

@flame-engine/flame-colab , what do you think name the folder? `_goldens` or `goldens`? My vote goes to `goldens`

spydon: I'd vote for `goldens`, but I'll admit I didn't read through the discussion super carefully.

packages/flame/test/sprite_test.dart
spydon: Let's uuupgrade! 🤩

st-pasha: I'm all for it, but has to be in a separate PR though

packages/flame/test/sprite_test.dart
luanpotter: as long as it's the same on all places both work for me :)

erickzanardo: @st-pasha we talked internally and settled on `goldens`, with that done this will LGTM!

wolfenrain (approved):

packages/flame/test/sprite_test.dart
st-pasha: That would mean we would have to move all current tests into the `tests/lib` folder, so as to avoid name clash between `assets` that contain test assets, and `assets` that contain tests for `lib/src/assets/*.dart`.
Maybe we should leave such renames to a separate PR, in order to have a better changelog record?

spydon: I don't follow, what is it that would clash? Isn't it a simple rename of `_goldens` to `goldens` that is missing?

spydon: Ah, now I see the clash.

spydon: Maybe we could have a directory containing both `goldens` and `assets`?

st-pasha: Hmm, I don't think this is the best solution. Surely it has its pros, but there are a few cons as well:
 - The directory is hard to find if you don't know already about it. That is, for a new developer it would be non-obvious to search for golden files inside `content/`;
 - It's not really consistent with `flame_test/example`;
 - It's +6 extra characters! This whole thread started because a single underscore character was deemed too excessive.

spydon: That makes sense, but it also sounds weird to rename the `_assets` folder in another PR, since it was introduced in this PR. Tbh I don't mind the underscore now when going through the options. I'll ask if the ones voting for a non-underscore solution have another name suggestion for the `_assets` directory.

packages/flame/test/_assets/images.dart
spydon: We are using images for goldens at work, I'll look up how we do it there.

packages/flame/test/sprite_test.dart
spydon: In melos they use a directory called `test_assets` maybe that could be an option if people feel strongly against the underscores.

st-pasha: We're already using an underscore for `doc/_sphinx`, so it looks totally fine to me :)

packages/flame/test/_assets/images.dart
st-pasha: Maybe this will work: https://stackoverflow.com/questions/45780255 ?

packages/flame/test/sprite_test.dart
spydon: image

So if you rename `_assets` to `_resources` we can get this merged! :)

st-pasha: done

PR #1524, opened 2022-04-08
refactor: Added classes MoveByEffect and MoveToEffect

st-pasha:
# Description

The classes `MoveByEffect` and `MoveToEffect` are both derived from abstract `MoveEffect`. This allows cleaner separation of functionality; and also makes it easier to check which effects are applied to the target. For example, you can find and cancel a `MoveToEffect` without affecting an existing `MoveByEffect`, or the other way around.

Note: `MoveEffect.by()` and `MoveEffect.to()` factory constructors remain in place, though we might consider deprecating them in the future.

## Checklist

## Breaking Change


spydon (approved):
I quite liked the `to` and `by` factories as an API, but might just be because I've gotten used to them. I do agree that the code gets cleaner like this.
I guess you'll do the same update for all the other effects?

st-pasha:
Yeah, `MoveEffect.to()` and `MoveEffect.by()` do sounds nice. Though `ScaleEffect.by` is already very questionable :)

> I guess you'll do the same update for all the other effects?

I'll get to other effects eventually too, they need to be converted to use `<Property>Provider` interfaces anyways.

erickzanardo (approved):

PR #1506, opened 2022-03-30
feat!: Flip gravity in flame_forge2d to be able to mix Forge2D and Flame components

spydon:
# Description

Since the coordinate system in Forge2D is flipped we have had a lot of problems mixing Flame and Forge2D components in a `Forge2DGame`, but with this change the gravity of the Y-axis is flipped so a gravity of `Vector2(10, 0)` would pull your bodies downwards instead of upwards as it was previously.

## Checklist

- [x] The title of my PR starts with a [Conventional Commit] prefix (`fix:`, `feat:`, `docs:` etc).
- [x] I have read the [Contributor Guide] and followed the process outlined for submitting PRs.
- [x] I have updated/added tests for ALL new/updated/fixed functionality.
- [x] I have updated/added relevant documentation in `docs` and added dartdoc comments with `///`.
- [x] I have updated/added relevant examples in `examples`.

## Breaking Change

Since it flips the gravity this is a breaking change.

- [x] Yes, this is a breaking change.

## Related Issues

Closes #1484 
Closes #1402
Closes #1132 
Closes #1395

[issue database]: https://github.com/flame-engine/flame/issues
[Contributor Guide]: https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md
[Flame Style Guide]: https://github.com/flame-engine/flame/blob/main/STYLEGUIDE.md
[Conventional Commit]: https://conventionalcommits.org

PR #1525, opened 2022-04-09
feat: Added children parameter to Component constructor

### LINKED ISSUE #1522: Improve components with a declarative API
# What could be improved

Flame already favors composability, Components can have children, their children can have their own and so on.

To put Flame more in line with Flutter APIs, we could a `children` attribute on all components in order to allow a more declarative way of developing composable components, for example:

```dart
add(
  PositionComponent(
    children: [
       SpriteComponent(sprite: playerSprite),
       KeyboardController(),
       ....
    ],
  ),
);
```

# Why should this be improved
It will allow better declarative programming and bring Flame even closer to the "Flutter way"

# Any risks?
Idk how complicated this would be to implement, on a naive idea, I think that just by receiving a list o Component on the constructor and adding then all on the component default `onLoad` will me enough, but we need to test and check if this will bring any issues with the Flame lifecycle.

# More information

st-pasha:
# Description

This PR allows to pass the list of children as a parameter directly into the `Component`'s constructor. This is an alternative to standard `add()` or `addAll()` methods.

## Checklist

## Breaking Change

## Related Issues

Resolves #1522


erickzanardo:
I think there a few components missing, SpriteAnimationComponent for example.

Also we probably should do the same on the bridge packages

spydon (commented):
Missing docs, and I thought maybe Erick wanted to implement this (and might have already started) since he opened the issue so always good to discuss on the issue who's going to implementing it before doing it.
Other than that it looks good!

st-pasha:
> I think there are a few components missing, SpriteAnimationComponent for example.

Right, so my thinking here is that we probably don't want to add `children` to those components which do not have children usually. For example, `Effect`s are components, but it is not common to add other components to effect components.

Since I'm not exactly sure which components usually do or do not have children, I'll make a list and then you let me know which ones should get the parameter.

spydon:
> Since I'm not exactly sure which components usually do or do not have children, I'll make a list and then you let me know which ones should get the parameter.

I would say at least all `Component`s deriving from `PositionComponent`.
`BodyComponent`in flame_forge2d should also have it, but #1506 should be merged first.
There are probably some more too?

st-pasha:
- `CircleHitbox`
- `CompositeHitbox`
- `PolygonHitbox`
- `RectangleHitbox`
- `ScreenHitbox`
- `ButtonComponent`
- `HudButtonComponent`
- `HudMarginComponent`
- `JoystickCompoenent`
- `SpriteButtonComponent`
- `CustomPainterComponent`
- `IsometricTileMapComponent`
- `NineTileBoxComponent`
- `ParallaxComponent`
- `ParticleSystemComponent`
- `SpriteAnimationComponent`
- `SpriteAnimationGroupComponent`
- `SpriteBatchComponent`
- `TextBoxComponent`
- `TextComponent`
- `TimerComponent`
- `CircleComponent`
- `PolygonComponent`
- `RectangleComponent`
- `ShapeComponent`
- flame_flare:
  - `FlareActorComponent`
- flame_forge2d:
  - `BodyComponent`
- flame_rive:
  - `RiveComponent`
- flame_svg:
  - `SvgComponent`
- flame_tiled:
  - `TiledComponent`

spydon:
The hitboxes and the TimerComponent doesn't need the field I'd say.

erickzanardo:
Flame flare is kind obsolete because flame rive, so I would say that it can be omitted too

spydon:
The `CompositeHitbox` would be nice to have the children field on btw, maybe with a tighter type than `Component` to indicate that hitboxes should be added to it.

doc/flame/components.md
```diff
@@ -122,6 +123,36 @@ class GameOverPanel extends PositionComponent with HasGameRef<MyGame> {
...
+        ],
+      ),
+    );
+  }
+}
+```
+
+The two approaches can be combined freely: the children specified within the
+constructor will be added first, and then any additional child components
+after.
```
alestiago, line 154: Shall we also clarify, if the added children via `children` are accesible (or not) from an lifecycle method such as `onLoad` by reading `children`?

packages/flame/test/components/component_test.dart
```diff
@@ -45,6 +45,36 @@ void main() {
...
+      expect(game.children.first.children.elementAt(0), isA<ComponentA>());
+      expect(game.children.first.children.elementAt(1), isA<ComponentB>());
+    });
+
+    testWithFlameGame('children in constructor and onLoad', (game) async {
+      final component = TwoChildrenComponent(
+        children: [ComponentA(), ComponentB()],
+      );
+      game.add(component);
+      await game.ready();
```
alestiago, line 67: ensureAdd?

alestiago (commented):
Looks good! I left some minor comments. But overall this is quite useful.

I wonder if it would be valuable to also ad tests for each subclass of Component to check if children is correctly set.

doc/flame/components.md
st-pasha: Added clarification.

packages/flame/test/components/component_test.dart
st-pasha: `game.ready()` does the same, without the need to use a special extension.

erickzanardo: You do write one line leas though

spydon (approved):

erickzanardo (approved):

spydon:
After checking again I think that `SpriteBatchComponent` and `ParticleSystemComponent` shouldn't have this argument either.
After removing those I think we can merge this. :)

st-pasha:
> After checking again I think that SpriteBatchComponent and ParticleSystemComponent shouldn't have this argument either.
After removing those I think we can merge this. :)

Done.

Issue #1274, opened 2021-12-26
Clipping mask


mancj:
Is there a possibility to create a clipping mask for sprites? 
I mean something similar to [this](https://www.google.com/search?q=clipping+mask+sprite&tbm=isch&ved=2ahUKEwihy9TN1YL1AhXPPOwKHWXgBzcQ2-cCegQIABAC&oq=clipping+mask+sprite&gs_lcp=ChJtb2JpbGUtZ3dzLXdpei1pbWcQAzoFCAAQgAQ6BAgAEB46BAgAEBM6BAgAEBg6BAgeEApQ9QdYzxpghx5oAHAAeACAAXaIAeAGkgEDMi42mAEAoAEBwAEB&sclient=mobile-gws-wiz-img&ei=ZwHJYeH7PM_5sAflwJ-4Aw&bih=655&biw=360&client=ms-android-samsung-ss&prmd=ivsxn&hl=ru#imgrc=TGj4KWVUEFjATM&imgdii=A6QEmxD3vbmrcM)

spydon:
You can clip the canvas that the sprite is rendered on by overriding `render` in `SpriteComponent`, would that achieve what you want?

erickzanardo:
Closing this due inactivity, feel free to add a comment here and we can reopen the issue.

PR #1528, opened 2022-04-10
feat: Adding classes for raw geometric shapes

### LINKED ISSUE #1521: Refactor geomerty shape classes
Currently, some important `geometry` classes exist only as components: `CircleComponent`, `PolygonComponent`, `RectangleComponent`. I propose to separate their geometrical logic into separate `Shape` classes: `Circle`, `Polygon`, `Rectangle`, allowing these to be reused for various purposes.

Thus, for example, a `CircleComponent` would contain a `Circle` shape, plus some rendering logic to display that shape. At the same time, `Circle` would be responsible for purely geometric functionality, such as determining intersections, whether a point is inside or outside, etc.

st-pasha:
# Description

Added classes for the following shapes:
   - [Circle]
   - [Rectangle]
   - [RoundedRectangle]
   - [Polygon]

The shapes themselves do not provide any end-user functionality. Instead, they are intended as building blocks for several other systems, including ShapeComponent, Hitbox, GestureHitbox, Viewport, ClipComponent, and so on. These use cases still need to be built.

## Checklist

## Breaking Change

## Related Issues

Closes #1521


packages/flame/lib/src/geometry/shapes/circle.dart
```diff
@@ -0,0 +1,64 @@
+import 'dart:ui';
+
+import 'package:meta/meta.dart';
+
+import '../../extensions/vector2.dart';
+import '../../game/transform2d.dart';
+import 'shape.dart';
+
+/// The circle with a given [center] and [radius].
+@immutable
```
spydon, line 10: If the `CircleComponent` should be backed by this later it needs to be mutable right?

packages/flame/lib/src/geometry/shapes/line_segment.dart
```diff
@@ -0,0 +1,78 @@
...
+  @override
+  double get perimeter => (to - from).length;
+
+  @override
+  Aabb2 get aabb {
+    final minX = min(from.x, to.x);
+    final maxX = max(from.x, to.x);
+    final minY = min(from.y, to.y);
+    final maxY = max(from.y, to.y);
+    return Aabb2.minMax(Vector2(minX, minY), Vector2(maxX, maxY));
```
spydon, line 40: It should be possible to create this `Aabb2` without creating two new vectors I think?

packages/flame/lib/src/geometry/shapes/rectangle.dart
```diff
@@ -0,0 +1,83 @@
+import 'dart:math';
+import 'dart:ui';
+
+import 'package:meta/meta.dart';
+import 'package:vector_math/vector_math_64.dart';
+
+import '../../game/transform2d.dart';
+import 'shape.dart';
+
+/// An axis-aligned rectangle.
```
spydon, line 10: If the `RectangleComponent` should be backed by this later it needs an angle

spydon (approved):
Nice, some comments if the plan is to back the shape components directly by these

packages/flame/lib/src/geometry/shapes/line_segment.dart
st-pasha: I don't see how, `Aabb2` has only the following constructors:
- `Aabb2.zero()`
- `Aabb2.copy(Aabb2 other)`
- `Aabb2.minMax(Vector2 min, Vector2 max)`
- `Aabb2.centerAndHalfExtents(Vector2 center, Vector2 halfExtents)`
- `Aabb2.fromBuffer(ByteBuffer buffer, int offset)`

packages/flame/lib/src/geometry/shapes/rectangle.dart
st-pasha: An axis-aligned rectangle is a much more simple shape (so its logic is easier to write and execute). For a rotated rectangle, a Polygon shape would probably be more appropriate.

packages/flame/lib/src/geometry/shapes/line_segment.dart
spydon: Alright starbge decision of them to not include a cobstructor from primitives, this wont be a concern when I've created the vector2 pool anyways.

packages/flame/lib/src/geometry/shapes/rectangle.dart
spydon: It'll be a bit confusing when the shape components don't map to the shape classes with the same names, but I get your point.

packages/flame/lib/src/geometry/shapes/circle.dart
st-pasha: That's one of the things that needs to be decided.
If we want the objects to have `operator==` and `hashCode`, then the object has to be declared as `@immutable` (we have an analyzer rule for that). This wouldn't necessarily preclude the shape from being usable in many places -- just that we would need to create new objects on each modification.

Alternatively, we can make all shapes mutable, and then maybe instead of `operator==` declare a simple function `.equals(Shape other)`.

Immutable shapes are a bit easier to work with -- no need to check the validity of the shape on each mutation, certain properties that are expensive to calculate can be easily computed once and memorized, etc.

packages/flame/lib/src/geometry/shapes/rectangle.dart
st-pasha: Ok, perhaps we are overcomplicating things here. You can define a generic `ShapeComponent` as
```dart
class ShapeComponent extends PositionComponent {
  ShapeComponent(this.shape, this.paint, ...) : super(...);

  final Shape shape;
  final Paint paint;

  @override
  void render(Canvas canvas) {
    canvas.drawPath(shape.toPath(), paint);
  }
}
```
and the `shape` could be anything (including `Rectangle`), and you'd still be able to move/rotate/squeeze it via the PositionComponent's facilities.

spydon: Remember that these should be the base for the hitboxes too (or they are currently at least), and I don't think we want an API where you always have to wrap a shape with a hitbox component to get a hitbox, instead of having the arguments directly on the hitbox.

spydon: And also, I think we should avoid doing a breaking change again here since we have just refactored this API.

st-pasha:
The main idea is that a geometric `Shape` is an object that could be used in a variety of applications: 
 - hitboxes for collision detection;
 - gesture hitboxes ("tapboxes") for tap detection;
 - clip mask for a `ClipComponent` (cf #1274);
 - custom-shaped viewport;
 - world bound for camera;
 - `ShapeComponent`s, i.e. components made of simple geometric shapes;
 - etc.

Regarding `Hitbox` components, the biggest challenge is figuring out the intersection API. The fact that `Rectangle` is not rotatable, however, is not a problem -- we can simply make it that `Hitbox.rectangle(angle=...)` would produce a hitbox whose internal shape is a `Polygon` instead of a `Rectangle`. We would have to do for `Circle`s too: any non-uniform scaling turns it into an ellipse, and rotation/shearing would either require a special `RotatedEllipse` shape, or a more generic `Polycurve`.

packages/flame/lib/src/geometry/shapes/circle.dart
spydon: I would love to have them immutable, but I don't think that would work performance wise in practice, since the hitboxes and shape components need to be able to change over time and we don't want to create shape objects in every tick when they are transforming.

packages/flame/lib/src/geometry/shapes/rectangle.dart
st-pasha: > Remember that these should be the base for the hitboxes too (or they are currently at least)

A `ShapeComponent` is a lightweight component which draws a geometrical shape using a simple paint. It shouldn't be the base for hitboxes. The only reason a Hitbox is based on a `ShapeComponent` right now is that shapes don't exist as standalone reusable entities, instead they are "baked in" into the `ShapeComponent`. 

Which is exactly what this PR is trying to address: create `Shape` as a separate abstract entity, and then let various components use those shapes as they need.

st-pasha:
When it comes to Hitboxes, I think the following design considerations apply:
- Hitboxes are based on shapes, in the sense that a `Hitbox` requires a `Shape` to support its functionality.
- In the majority of cases, a hitbox is static _within its component_.
- When a component moves around the world, the **projection** of that component's shape onto the world coordinate plane is moving.
- More often than not, the projection is a simple translation. That is, simple movement of a component is more common than rotations, scalings, and skews.
- For all shapes, a translation is a very simple operation that can be done in-place.
- More generic transforms are trickier, for some shapes they even result in a different shape being produced. For example, a circle may turn into a rotated ellipse, a rectangle into a parallelogram. Clearly, such transforms _cannot_ be done in-place.

Given these considerations, I think we should aim for the following architecture for the collision system:
- Collision system is attached to the world;
- For every hitbox mounted to that world, it is possible to construct the `Transform2D` matrix that projects the hitbox onto the world plane;
- Collision system maintains the list of projections from every hitbox shape onto the world plane. Here "maintains" means it keeps track of which components have moved during each tick, and either moves the corresponding projection (if it was a simple translation), or re-projects the hitbox if it was a more complicated movement.
- The collision detection algorithm runs on the set of projected shapes.

In view of this design, the `Shape` class has to support the following operations:
- `projection(Transform2D)`;
- `move(Vector2)` -- in-place translation;
- `intersection(Shape)`.

The intersection functionality deserves more discussion. Clearly, an intersection between two shapes is again a shape. However, computing that shape may not be always practical (for example, a circle intersecting a circle is a lune). On the other hand, often we don't want to know the exact shape of an intersection. For example, just knowing that the shapes did intersect might be sufficient. In other applications you might want just a single representative point of an intersection, or projection of that point onto the shape's boundary. In the current implementation we return the set of extreme points of the intersection shape.

spydon:
> Given these considerations, I think we should aim for the following architecture for the collision system:
> 
>     * Collision system is attached to the world;

It kind of is already, currently the `world` _is_ the FlameGame, since the `world` concept is still in experimental.

>     * For every hitbox mounted to that world, it is possible to construct the `Transform2D` matrix that projects the hitbox onto the world plane;

Agree that this would be a much better solution than the current one that takes the `absolute*` fields from the parent, since that doesn't work for all cases as discussed previously.

>     * Collision system maintains the list of projections from every hitbox shape onto the world plane. Here "maintains" means it keeps track of which components have moved during each tick, and either moves the corresponding projection (if it was a simple translation), or re-projects the hitbox if it was a more complicated movement.

Why couldn't the hitbox itself take care of that, like it does today?

>     * The collision detection algorithm runs on the set of projected shapes.

When using projection, instead of mutability, we'd create one new temporary object per hitbox per tick (possibly only the transformed ones, but could still be very many).
 
> In the current implementation we return the set of extreme points of the intersection shape.

I think that should be enough, anything more advanced will lead to heavier computation without much practical gain (might even make the API more complex).

st-pasha:
>> * Collision system is attached to the world;
>
> It kind of is already, currently the world is the FlameGame, since the world concept is still in experimental.

I use the "world" here in a loose sense: a `FlameGame` is a world, a `World` is a world, any other component can be the world too, it just needs to declare itself an owner of a `CollisionSystem` (which itself is either a mixin or a component). The only requirement for this "world" is that component tree mounted to it would have a way to propagate the transform matrices. So that if the "world" contains component A which contains B, then there should be a way for a hitbox attached to B to find its projection matrix onto the root world. Basically, the only requirement is that A and B implement the extended `Projector` interface.

> Why couldn't the hitbox itself take care of that, like it does today?

The `Shape` class knows how to project itself, using the provided `Transform2D` object. The `Hitbox` then can be responsible for finding that transform matrix. As mentioned above, this can be done by traversing up the tree until you reach the `CollisionSystem` root. However, it's possible that traversing _down_ the tree will be more efficient. 

For example, suppose you have tree `A -> B -> [C, D, E]`. Traversing down from A, you compute transform matrix AB once, and pass it to each C, D, and E. Traversing up from C, D, E, you need to recompute the entire transform ABC, ABD, ABE.

Also, it would be good if we only recomputed those transforms that actually need recomputing. So, having some kind of central authority to keep track of which component moved would be beneficial here.

> When using projection, instead of mutability, we'd create one new temporary object per hitbox per tick

We could have an API where `project()` takes an optional `target` parameter, and will attempt to modify that target in-place when it is given, and assuming the target has a suitable type. So, something like this:
```dart
Shape project(Transform2D transform, [Shape? target]);
```

>> In the current implementation we return the set of extreme points of the intersection shape.
>
> I think that should be enough, anything more advanced will lead to heavier computation without much practical gain (might even make the API more complex).

I'm not suggesting that we perform more advanced calculations here, if anything, I'd rather we did less.

But we could always return a special `Intersection` object that _knows_ how to perform these advanced calculations if the user needs them. Similarly, the extreme points are also non-trivial to compute, so they can be calculated by the `Intersection` object by user request.

The question only what is the minimum amount of calculation that we need to perform in order to know that two shapes did in fact intersect. AFAIU, there are algorithms that can tell that without computing the intersection points.

So what I'm suggesting here is that:
 - `Shape`'s `intersection()` method returned an `Intersection?` object;
 - the `Intersection` object is able to calculate `extremes()` on request;
 - the collision callback system for now returns `Intersection.extremes()`, but we slowly transition towards returning the `Intersection` object itself.

spydon:
> We could have an API where `project()` takes an optional `target` parameter, and will attempt to modify that target in-place when it is given, and assuming the target has a suitable type. So, something like this:
> 
> ```dart
> Shape project(Transform2D transform, [Shape? target]);
> ```

Sounds good.
 
> I'm not suggesting that we perform more advanced calculations here, if anything, I'd rather we did less.
> 
> But we could always return a special `Intersection` object that _knows_ how to perform these advanced calculations if the user needs them. Similarly, the extreme points are also non-trivial to compute, so they can be calculated by the `Intersection` object by user request.

Gotcha, that makes sense.

>     * the collision callback system for now returns `Intersection.extremes()`, but we slowly transition towards returning the `Intersection` object itself.

As long as that isn't too near in the future so that we don't have a breaking change again on this API already.

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
```diff
@@ -0,0 +1,219 @@
...
+      if (crossProduct <= 0) {
+        nExteriorAngles++;
+      }
+    });
+    if (nInteriorAngles < nExteriorAngles) {
+      _reverseVertices();
+      _initializeEdges();
+      nInteriorAngles = nExteriorAngles;
+    }
+    _convex = nInteriorAngles == n;
```
spydon, line 81: Is the "orientation" really correct term for determining whether a shape is convex or not?

st-pasha: The primary purpose of this method is to verify that the vertices are listed in the counter-clockwise order, and if not reverse the list in-place (which puts it into the CCW order). This is needed because otherwise the `containsPoint()` method won't work correctly. As a by-product, the method also detects whether the shape is convex or not.

spydon: Hmm, I see the usefulness of setting `_convex` here, but it is a side effect of what you'd expect this method to do.

st-pasha: Not sure what you're suggesting? Add a comment clarifying the purpose of the method?

spydon: I don't have any suggestion yet, maybe tomorrow. I just don't like methods with side effects.

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
```diff
@@ -0,0 +1,219 @@
...
+    } else {
+      _convex = convex;
+    }
+  }
+
+  /// The vertices (corners) of the polygon.
+  ///
+  /// The user should treat this list as read-only and not attempt to modify
+  /// either the list itself or individual points.
+  List<Vector2> get vertices => _vertices;
```
spydon, line 39: This could return an `UnmodifiableList` maybe?

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
```diff
@@ -0,0 +1,219 @@
...
+    var previousVertex = _vertices.last;
+    _edges = _vertices.map((Vector2 vertex) {
+      final edge = vertex - previousVertex;
+      previousVertex = vertex;
+      return edge;
+    }).toList(growable: false);
+  }
+
+  /// Number of vertices/edges in the polygon.
+  int get n => _vertices.length;
```
spydon, line 59: This should have a more descriptive name

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
spydon: Didn't have time to look properly at this today either, but I think maybe `check` is the wrong wording since it actually updates the vertices, so maybe a name that reflects that?

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
st-pasha: I was basing this on the fact that a [polygon](https://en.wikipedia.org/wiki/Polygon) is also called an **n-gon**, so the property _n_ of a polygon is actually the standard terminology.

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
st-pasha: I don't think I like `UnmodifiableList`, it looks rather inefficient.

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
spydon: Sounds like something that would be more known to matematicians, but maybe not programmers? I haven't seen any geometrical game api that exposes it as n at least, but maybe you have some examples?

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
st-pasha: Ok, I renamed it into `_ensureProperOrientation()`.

spydon: Much better!

packages/flame/lib/src/experimental/geometry/shapes/rounded_rectangle.dart
```diff
@@ -0,0 +1,182 @@
...
+  double get bottom => _bottom;
+  double get radius => _radius;
+  double get width => _right - _left;
+  double get height => _bottom - _top;
+
+  @override
+  bool get isConvex => true;
+
+  @override
+  double get perimeter => (tau - 8) * _radius + 2 * (width + height);
```
spydon, line 97: Why is `tao - 8` used here?

packages/flame/lib/src/experimental/geometry/shapes/shape.dart
```diff
@@ -0,0 +1,86 @@
...
+  /// example, a closed shape can be filled with a paint.
+  bool get isClosed => true;
+
+  /// True if the shape is convex, i.e. a line segment connecting any two points
+  /// of the shape would lie completely within the shape.
+  bool get isConvex;
+
+  /// The length of the shape's boundary. For some more complicated shapes this
+  /// can be computed approximately.
+  double get perimeter;
```
spydon, line 30: Curius of what this is used for

packages/flame/lib/src/experimental/geometry/shapes/shape.dart
```diff
@@ -0,0 +1,86 @@
...
+
+/// Base class for various 2D geometric primitives defined on a Cartesian
+/// coordinate plane.
+///
+/// Implementations include:
+///   - [Circle]
+///   - [Polygon]
+///   - [Rectangle]
+///   - [RoundedRectangle]
+abstract class Shape {
```
spydon, line 19: Maybe we discussed this earlier but I forgot, shouldn't the shapes have an angle?

packages/flame/lib/src/experimental/geometry/shapes/rounded_rectangle.dart
st-pasha: Imagine you cut off the 4 rounded corners. Since each corner is a quarter of a circle, you can make a full circle out of them, with perimeter tau*r. The remaining sides each is shorter by 2r now. So their perimeter is (width-2r + height-2r)*2, which is (width+height)*2 - 8r. Adding back the circle, you have (tau-8)r.

packages/flame/lib/src/experimental/geometry/shapes/shape.dart
st-pasha: Say, if you want to move a component along the border of the shape, you may want to know the length of the border in order to calculate the timing.
It may not be very useful, but it doesn't impose any extra cost, and it's a fundamental geometric property of a shape.

packages/flame/lib/src/experimental/geometry/shapes/shape.dart
st-pasha: The project() method can apply an arbitrary transform to an existing shape, including a rotation. The method also has the target= parameter to do the transformation in-place.

packages/flame/lib/src/experimental/geometry/shapes/rounded_rectangle.dart
spydon: Makes sense, thanks for the explanation!

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
spydon: I think something like `numberOfSides` would be clearer (albeit much longer), that is what they use in the text on the wikipedia page too.

st-pasha: That's even longer than `edges.length` though. Maybe just remove it?

spydon: True, I think `edges.length` is nice and clear

st-pasha: ok, i removed `.n`.

spydon (approved):

wolfenrain (approved):
LGTM! This will hopefully bring some interesting new aspects to flame!

PR #1531, opened 2022-04-11
feat: Added flame_test function closeToAabb()

st-pasha:
# Description

A convenience function for testing that an Aabb is close to the value that you'd expect.

## Checklist

## Breaking Change

### DIFF
--- a/packages/flame_test/lib/flame_test.dart
+++ b/packages/flame_test/lib/flame_test.dart
@@ -1,3 +1,4 @@
+export 'src/close_to_aabb.dart' show closeToAabb;
 export 'src/close_to_vector.dart';
 export 'src/expect_double.dart';
 export 'src/expect_vector2.dart';
--- a/packages/flame_test/lib/src/close_to_aabb.dart
+++ b/packages/flame_test/lib/src/close_to_aabb.dart
@@ -1,0 +1,60 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:vector_math/vector_math_64.dart';
+
+/// Returns a matcher which checks if the argument is an axis-aligned bounding
+/// box sufficiently close (within distance [epsilon]) to [expected]. Example
+/// of usage:
+/// ```dart
+/// expect(
+///   component.aabb,
+///   closeToAabb(Aabb2.minMax(Vector2.zero(), Vector2.all(1)), 1e-5),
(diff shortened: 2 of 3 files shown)


packages/flame_test/lib/flame_test.dart
```diff
@@ -1,3 +1,4 @@
+export 'src/close_to_aabb.dart' show closeToAabb;
```
spydon, line 1: Do you need `show` here when `closeToAabb` is the only public method?

spydon (approved):
Nice with the mismatch description!

erickzanardo (approved):

packages/flame_test/lib/flame_test.dart
st-pasha: Not _needed_ per se, but it's a protection against possible future modifications accidentally leaking into the global namespace.

spydon: Or the other way around, that something that should be exported won't be exported when we publish, since the developer could assume that everything in the file was exported since it is in almost all other barrel files. I don't think we should have an exception here. 

I see that we have it on the effects and experimental files too, we should probably settle for either using `show` or using `@internal` to have it clearer. I think `@internal` is nicer, since then you see it directly in the affected file.

st-pasha: > something that should be exported won't be exported when we publish

Protection against this scenario is that the test files use the library's public interface. So when you add tests for the new functionality, you'd notice that it's not properly exported.

The concern about accidental leaking of API is not theoretical -- for example see https://pub.dev/documentation/flame/latest/game/game-library.html -- apparently we're exposing some classes like `Sphere` or `Vector4` that we should never have.

Another advantage of using `show` for every `export` is that you can see from the file the totality of all classes/functions available from that library. Otherwise you'd have to go into each and every file to check what's there. This would be especially helpful for libraries like `flame_test` where some files export multiple functions.

> I think @internal is nicer, since then you see it directly in the affected file.

Agree, this is nice too. But there is no harm in doing both: have the `@internal` annotation, and the `show` statement on export.

spydon: Alright, don't like that we have the inconsistency, but that's not this PRs fault, merging this.

Issue #1532, opened 2022-04-11
Error when add and remove component while game engine paused


Hwan-seok:
# Current bug behaviour

Add the component to the game after pausing engine and remove it before engine resume causes assertion fails.

# Expected behaviour
Prevent failing assertions.

# Steps to reproduce
```
import 'package:flame/components.dart';
import 'package:flame/game.dart';
import 'package:flame_test/flame_test.dart';

class MyGame extends FlameGame {}

final tester = FlameTester(MyGame.new);

void main() {
  tester.test('test', (game) async {
    game.pauseEngine();

    final component = Component();
    await game.add(component);
    game.remove(component);

    game.resumeEngine();
    game.update(0);
  });
}

```
# Flutter doctor output

```
✓] Flutter (Channel stable, 2.10.4, on macOS 12.1 21C52 darwin-arm, locale ko-KR)
    • Flutter version 2.10.4 at /Users/lafity101/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision c860cba910 (2 weeks ago), 2022-03-25 00:23:12 -0500
    • Engine revision 57d3bac3dd
    • Dart version 2.16.2
    • DevTools version 2.9.2

[✓] Android toolchain - develop for Android devices (Android SDK version 32.0.0)
    • Android SDK at /Users/lafity101/Library/Android/sdk
    • Platform android-32, build-tools 32.0.0
    • Java binary at: /Applications/Android
      Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7249189)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 13.3)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • CocoaPods version 1.11.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2020.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7249189)

[✓] IntelliJ IDEA Ultimate Edition (version 2021.3)
[...]

Issue #1549, opened 2022-04-14
game frozen with assertion failed


GuiTom:
# Current bug behaviour
 report 'Assertion failed at line 501 pos 12 of src/component/component.dart:"_state == LifecycleState.loaded || _state == LifecycleState.removed?"' ,and game frozen

# Expected behaviour
   It shound not asseet fail

# Steps to reproduce
    I do not know how to reproduce,but it occurs frequently

# Flutter doctor output

```
[✓] Flutter (Channel unknown, 2.8.1, on macOS 11.4 20F71 darwin-x64, locale zh-Hans-CN)
    • Flutter version 2.8.1 at /Users/chenchao/flutter_2.8.1
    • Upstream repository unknown
    • Framework revision 77d935af4d (4 months ago), 2021-12-16 08:37:33 -0800
    • Engine revision 890a5fca2e
    • Dart version 2.15.1
    • Pub download mirror https://pub.flutter-io.cn
    • Flutter download mirror https://storage.flutter-io.cn

[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
    • Android SDK at /Users/chenchao/Library/Android/sdk
    • Platform android-31, build-tools 31.0.0
    • ANDROID_SDK_ROOT = /Users/chenchao/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7281165)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 13.2.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • CocoaPods version 1.11.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2020.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.10+0-b96-7281165)

[✓] VS Code (version 1.66.1)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension can be installed from:
[...]

PR #1551, opened 2022-04-15
fix: Components in uninitialized state can now be safely removed

st-pasha:
# Description

Correctly handle removal of components that are still in uninitialized state (which can happen when a component is added to a detached parent). Added assertion checks to guard against component tree corruption when removing a component from the wrong parent.

## Checklist

## Breaking Change

## Related Issues

WIP for #1549

### DIFF
--- a/packages/flame/test/components/component_test.dart
+++ b/packages/flame/test/components/component_test.dart
@@ -172,6 +172,75 @@ void main() {
       },
     );
 
+    testWithFlameGame(
+      'remove a component before it was ever added',
+      (game) async {
+        expect(
+          () => game.remove(Component()),
+          failsAssert(
+            "Trying to remove a component that doesn't belong to any parent",
+          ),
+        );
+      },
+    );
+
+    testWithFlameGame(
+      'remove a component from a wrong parent',
+      (game) async {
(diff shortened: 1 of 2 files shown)


spydon (approved):
Very nice, seems like many have been troubled by this

Hwan-seok:
Is it also fixes #1532 ?
If it is, then it should as a part of the related issue.

wolfenrain (approved):

packages/flame/test/components/component_test.dart
```diff
@@ -472,3 +541,13 @@ class TwoChildrenComponent extends Component {
...
   }
 }
+
+class _ComponentWithOnLoad extends Component {
+  int onLoadCalledCount = 0;
+
+  @override
+  Future<void> onLoad() async {
+    onLoadCalledCount++;
+    await null;
```
erickzanardo, line 551: I don't think this is needed, just by marking the method as async should be enough
```suggestion
```

erickzanardo (approved):

packages/flame/test/components/component_test.dart
st-pasha: Yep, makes sense.

Issue #1543, opened 2022-04-13
Add on complete callback to SpriteAnimationWidget


erickzanardo:
# Problem to solve
there is no easy way to know if an animation is finished when using it through `SpriteAnimationWidget` since it don't exposes a callback and internally it overrides any set callback on `animation.onComplete`

# Proposal
Provide a callback on the widget constructor so the user can register the callback.

# More information

alestiago:
I need to have a closer look but perhaps having a Controller would be closer to the ideal solution. Perhaps we can reuse the controller.

So for example we could expose the `playing` and listen when this value changes to `false` from `true`.

Then, a setter for this could be used to, for example if we want the animation to be paused. 

Anyways, this is just me naively dropping some ideas. I need to have a closer look at the actual code 👨‍💻

spydon:
@alestiago I think it is important that it should be similar to the rest of our APIs, so passing in a callback I think would be the optimal solution. :)

spydon:
> @alestiago I think it is important that it should be similar to the rest of our APIs, so passing in a callback I think would be the optimal solution. :)

Forgot that this is the widget, I guess it should be as Fluttery as possible instead then.

alestiago:
I'll be soon raising a PR so that we can discuss how the API of this Widget should be.

erickzanardo:
> @erickzanardo does #1564 solve this issue?

Humm, I think it offers an alternative, but a callback on the widget would be more straight forward and better for users.

alestiago:
Doesn't `animation.completed.whenComplete(callback)` allows using a callback.

erickzanardo:
> Doesn't `animation.completed.whenComplete(callback)` allow using a callback?

Yes, but that means that you need to hold an `animation` instance, which will probably require the use of a stateful widget.

SpriteAnimationWidget has some helper constructors that loads and manages the `animation` instance for you. So a callback received by the `SpriteAnimationWidget` would be the best.

We could even use the `animation.completed` internally to hook that callback that we would receive

PR #1552, opened 2022-04-16
feat!: implemented SpriteAnimationController

alestiago:
# Description

Refactors `SpriteAnimationController` API to be more Flutter friendly. Supporting new features, for example, listening to when the animation completes.

## Checklist

## Breaking Change

- [X] Yes, this is a breaking change.

## Related Issues

#1543 

#1492 

---

 ⚠️  This is currently under development.


packages/flame/lib/src/widgets/animation_widget.dart
```diff
@@ -2,169 +2,62 @@ import 'dart:math';
...
-    setState(() {
-      widget.animation.reset();
-      _lastUpdated = DateTime.now().millisecond.toDouble();
-      _controller?.repeat(
-        // Approximately 60 fps
-        period: const Duration(milliseconds: 16),
-      );
-    });
+  }) : super(vsync: vsync) {
+    duration = Duration(seconds: animation.totalDuration().ceil());
```
alestiago, line 47: The `ceil` makes the total duration to be approximated. I still need to define an algorithm to calculate the `totalDuration` as `Duration`, without losing precision.

packages/flame/lib/src/widgets/animation_widget.dart
```diff
@@ -2,169 +2,62 @@ import 'dart:math';
...
-    }
-  }
-
-  void _initAnimation() {
-    setState(() {
-      widget.animation.reset();
-      _lastUpdated = DateTime.now().millisecond.toDouble();
-      _controller?.repeat(
-        // Approximately 60 fps
-        period: const Duration(milliseconds: 16),
```
alestiago, line 130: @spydon this logic seemed kind of weird. It has now been deleted and the user has full control over the period. I need to look into it, however, do you have any additional context to this? Is there any side effects if for example, the period is increased/decreased?

spydon: It wasn't me who implemented this, but this is so that it will be running at 60fps, kind of imitating a simple game loop.
Not sure if it is needed.

packages/flame/lib/src/widgets/animation_widget.dart
```diff
@@ -2,169 +2,62 @@ import 'dart:math';
...
-    return Container(
-      child: CustomPaint(
-        painter: SpritePainter(
-          widget.animation.getSprite(),
-          widget.anchor,
-        ),
-      ),
-    );
+    final now = DateTime.now().millisecond.toDouble();
+    final dt = max(0, (now - (_lastUpdated ?? 0)) / 1000).toDouble();
```
spydon, line 59:
```suggestion
    final dt = max<double>(0, (now - (_lastUpdated ?? 0)) / 1000);
```

packages/flame/lib/src/widgets/animation_widget.dart
```diff
@@ -2,169 +2,62 @@ import 'dart:math';
...
-  Widget build(BuildContext ctx) {
-    return Container(
-      child: CustomPaint(
-        painter: SpritePainter(
-          widget.animation.getSprite(),
-          widget.anchor,
-        ),
-      ),
-    );
+    final now = DateTime.now().millisecond.toDouble();
```
spydon, line 58: Can't the vsync ticker take care of this?

st-pasha, line 58: `DateTime.now().millisecond` returns the "millisecond" part of the datetime, a number between 0 and 999. Which means there will be animation jank every second.

packages/flame/lib/src/widgets/animation_widget.dart
alestiago: The user can now specify the period manually, so I currently don't see any harm in this

packages/flame/lib/src/widgets/animation_widget.dart
alestiago: I copied it from the previous logic, so I think this issue was still in the previous code. I will look into this.

packages/flame/lib/src/widgets/animation_widget.dart
spydon: It doesn't have a default now right and the user must specify it explicitly? I think a default would be good since many users won't know that this is to reflect 60/120 fps.

erickzanardo (commented):
I have a few concerns about this.

When this widget was initially implemented, we considered the possibility of it receiving a controller, but we went with the current way because that would make API less friendly, the user would need to create and manage a controller of its own, that means that they would need to have a stateful widget and so on.

But a SpriteAnimation is a pretty straight forward thing, so a path more like how an [AnimatedController](https://api.flutter.dev/flutter/widgets/AnimatedContainer-class.html) is, made more sense here.

Right now, what we need to solve here is that the oncomplete callback on the `SpriteAnimation` is overriden by the `SpriteAnimationWidget`. IMO we should solve that instead of refactoring everything here. My suggestion to solve that is to change the on complete on `SpriteAnimation` to be able to hold several callbacks instead of only one by implementing a `registerCompleteListerner` and `removeCompleteListener`.

Then the widget would register and remove its callback as needed.

We could make some fallbacks for the current API on the `SpriteAnimation` to avoid a breaking change.

alestiago:
>  When this widget was initially implemented, we considered the possibility of it receiving a controller, but we went with the current way because that would make API less friendly, the user would need to create and manage a controller of its own, that means that they would need to have a stateful widget and so on.

Whilst it is true that the "API" is less friendly because of the controller. I prefer this API over the previous one. Personally, for me and from what I've seen from other developers using the API it is much much easier to pick up for Flutter developers (since it is just how animations are handled in Flutter). This is a Widget so I feel strongly about making the API as closer to Flutter as possible. If the issue is regarding the boilerplate code regarded to making Flutter animations, then  I'll say the problem falls to Flutter.

In addition, this exposes much more power and control to the developer. For example, with this pattern the period can be easily adjusted or even better, the controller could be possibly shared between widgets if the developer wishes to. In addition by exposing the functionality Flutter offers it is less likely that new functionality will have to be constantly added to this Widget. 

> Right now, what we need to solve here is that the oncomplete callback on the SpriteAnimation is overriden by the SpriteAnimationWidget. IMO we should solve that instead of refactoring everything here. My suggestion to solve that is to change the on complete on SpriteAnimation to be able to hold several callbacks instead of only one by implementing a registerCompleteListerner and removeCompleteListener.

This is already possible with ` _controller.addStatusListener((status) { })` and  ` _controller.addListener((status) { })`. 

It is true that this solves way more than just #1543 , let say this PR solving #1543 is a side-effect. Since the aim of this PR is to bring the widget closer to the Flutter world.

erickzanardo:
> Whilst it is true that the "API" is less friendly because of the controller. I prefer this API over the previous one. Personally, for me and from what I've seen from other developers using the API it is much much easier to pick up for Flutter developers (since it is just how animations are handled in Flutter). This is a Widget so I feel strongly about making the API as closer to Flutter as possible. If the issue is regarding the boilerplate code regarded to making Flutter controlled animations, then I'll say the problem falls to Flutter.

I disagree that requiring the user to passa a controller makes it more Flutter way, like I referenced the `AnimatedContainer`, Flutter offers animated widgets that have their controller internally.

It does add more power, but IMO the boilerplate is not worth it, all the animation information is already stored on the `SpriteAnimation` instance, so the power added is not that much and not worth requiring the user to write their own controller.

> If we go down your solution, I feel will be reinventing the wheel since this logic you've described is the same for Listenable? AndAnimationController is already a Listenable.

I don't think we are, because that API can be used on the Flame side as well. Right now we can't have multiple on complete callbacks on animations.

PR #1561, opened 2022-04-21
feat: Added FollowBehavior and ability for the new Camera to follow a component

st-pasha:
# Description

- Created `FollowBehavior` which can be attached to any component and cause that component to follow the given target;
- Added `CameraComponent` methods `follow()`, `stopFollowing()`, and `moveTo()` which are similar to the existing `Camera` methods.

Still needed: tests, docs and examples.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/lib/src/experimental/camera_component.dart
+++ b/packages/flame/lib/src/experimental/camera_component.dart
@@ -9,7 +9,6 @@ import '../effects/controllers/effect_controller.dart';
 import '../effects/move_effect.dart';
 import '../effects/move_to_effect.dart';
 import '../effects/provider_interfaces.dart';
-import '../effects/rotate_effect.dart';
 import 'follow_behavior.dart';
 import 'max_viewport.dart';
 import 'viewfinder.dart';
--- a/packages/flame/lib/src/experimental/camera_component.dart
+++ b/packages/flame/lib/src/experimental/camera_component.dart
@@ -1,8 +1,11 @@
 import 'dart:ui';
 
 import 'package:meta/meta.dart';
+import 'package:vector_math/vector_math_64.dart';
 
 import '../components/component.dart';
+import '../effects/provider_interfaces.dart';
+import 'follow_behavior.dart';
 import 'max_viewport.dart';
 import 'viewfinder.dart';
 import 'viewport.dart';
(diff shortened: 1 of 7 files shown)


packages/flame/lib/src/experimental/camera_component.dart
```diff
@@ -115,4 +118,51 @@ class CameraComponent extends Component {
...
+  /// first.
+  ///
+  /// Parameters [maxSpeed], [horizontalOnly] and [verticalOnly] have the same
+  /// meaning as in the [FollowBehavior.new] constructor.
+  ///
+  /// If [snap] is true, then the viewfinder's starting position will be set to
+  /// the target's current location. If [snap] is false, then the viewfinder
+  /// will move from its current position to the target's position at the given
+  /// speed.
+  void follow(
```
spydon, line 135: It would be good to have a `followVector2` too which can be used by the `BodyComponent` for example, since that one can't have a `PositionProvider` since the `body.position` shouldn't be changed.

st-pasha: My thinking here is that if `target` is a `PositionProvider`, then it can really be tied to anything: to another component, to a particle system, to a mouse cursor, to a geometric center of a swarm of components, to an abstract calculation, etc.
The fact that `body.position` shouldn't change is really not an obstacle, since we can declare a no-op setter in this case.

st-pasha: I created the `APositionProvider` class to make this process easier.

spydon: That is quite a strange name. :sweat_smile: 
Is this class really necessary? We can just implement the `PositionProvider` on `BodyComponent` and do a no-op + assert in the setter just like you said.

st-pasha: It's helpful if you want to create a new `PositionProvider` on the fly. Say,
```dart
camera.follow(APositionProvider(getValue: () => (player1.position + player2.position)/2));
```
or
```dart
camera.follow(APositionProvider(getValue: () => player.position + player.velocity*0.5));
```

spydon: I think we can start doing basic implementations of the providers if we see a real need for it, I think it will confuse people more than it will help them currently (and it is still very simple to create a new class that implements the provider with the information that they would want if they know what they are doing).

st-pasha: Why would it confuse people if they are not even required to use this class? It's just a handy helper. If somebody thinks this helper is not really helpful, then they can use whatever other solution they prefer.

spydon: I think it is mostly the name that I don't like. Maybe the old school java naming convention would be better here, `PositionProviderImpl`?

packages/flame/lib/src/experimental/follow_behavior.dart
```diff
@@ -0,0 +1,75 @@
...
+/// either [PositionComponent]s, or camera's [Viewfinder]/[Viewport], or any
+/// other objects, including custom implementations.
+///
+/// The [maxSpeed] parameter controls the maximum speed with which the [owner]
+/// is allowed to move as it pursues the [target]. By default, the max speed is
+/// infinite, allowing the owner to stay on top of the target all the time.
+///
+/// The flags [horizontalOnly]/[verticalOnly] allow constraining the [owner]'s
+/// movement to the horizontal/vertical directions respectively.
+class FollowBehavior extends Component {
```
spydon, line 19: You planned to add constraints to within which bounds the camera would move if I remember correctly, are you intending to put those bounds on this class or directly on the viewfinder, or maybe both?

st-pasha: My plan was to have a separate component for that. Something like `ConstrainedPositionBehavior`. You would add it after the `FollowBehavior`, and it would check whether the position stays within the specified bounds (given by a Shape). If not, the camera would be moved towards the shape's boundary.

packages/flame/lib/src/experimental/camera_component.dart
st-pasha: ok, renamed

spydon (approved):

wolfenrain (approved):
LGTM!

Issue #1558, opened 2022-04-20
Bug in sample code: packages / flame_forge2d / example / lib / mouse_joint_sample.dart


JCzz:
I think I found a bug, the code should be:

      @override
      bool onDragUpdate(int pointerId, DragUpdateInfo details) {
        final mouseJointDef = MouseJointDef()
          ..maxForce = 3000 * ball.body.mass * 10
          ..dampingRatio = 0.1
          ..frequencyHz = 5
          ..target.setFrom(ball.body.position)
          ..collideConnected = false
          ..bodyA = groundBody
          ..bodyB = ball.body;
    
        if (mouseJoint == null) {
          mouseJoint = MouseJoint(mouseJointDef);
          world.createJoint(mouseJoint!);
        }
        mouseJoint?.setTarget(details.eventPosition.game);
        return false;
      }

And not

      @override
      bool onDragUpdate(int pointerId, DragUpdateInfo details) {
        final mouseJointDef = MouseJointDef()
          ..maxForce = 3000 * ball.body.mass * 10
          ..dampingRatio = 0.1
          ..frequencyHz = 5
          ..target.setFrom(ball.body.position)
          ..collideConnected = false
          ..bodyA = groundBody
          ..bodyB = ball.body;
    
        mouseJoint ??= MouseJoint(mouseJointDef);
        world.createJoint(mouseJoint!);
    
        mouseJoint?.setTarget(details.eventPosition.game);
        return false;
      }

See this for elaboration: https://stackoverflow.com/questions/71928016/mousejoint-gets-less-and-less-reactive/71938577#71938577

spydon:
Makes sense, so you want to open a PR for this fix?

JCzz:
Yes, created: https://github.com/flame-engine/flame/pull/1562

PR #1562, opened 2022-04-21
fix: MouseJoint gets less and less reactive(#1558)

JCzz:
# Description
Found a functional bug in Forge2d samplecode: `mouse_joint_sample.dart`

## Breaking Change
- [ x] No, this is *not* a breaking change.

https://stackoverflow.com/questions/71928016/mousejoint-gets-less-and-less-reactive/71938577#71938577


spydon (approved):
Thanks for your contribution!

PR #1564, opened 2022-04-22
feat: included `completed` to `SpriteAnimation`

alestiago:
# Description

Adds `completed` future that completes whenever the animation completes.

Similar to how we have `loaded` and `mounted` on `Component`.

Reasoning:
- The `onCompleted` can be overridden whereas with `completed` you can add multiple "callbacks"
(that overriding issue is what raised the issue on the widget in the first place)
- In some scenarios, the API `completed` is easier to test and read over `onCompleted`

## Checklist

- [X]  I have updated/added tests for ALL new/updated/fixed functionality.
- [X] I have updated/added relevant examples in `examples`.

## Breaking Change

## Related Issues

#1543 

### DIFF
--- a/doc/flame/components.md
+++ b/doc/flame/components.md
@@ -367,7 +367,7 @@ animation object needs to be ticked to move the frames.
 To listen when the animation is done (when it reaches its last frame and is not looping) you can
 use `animation.completed`.
 
-For example, you can:
+Example:
 
 ```dart
 await animation.completed; 
--- a/doc/flame/components.md
+++ b/doc/flame/components.md
@@ -367,6 +367,17 @@ animation object needs to be ticked to move the frames.
 To listen when the animation is done (when it reaches its last frame and is not looping) you can
 use `animation.completed`.
 
+For example, you can:
+
+```dart
+await animation.completed; 
+doSomething();
+
+// or alternatively
+
+animation.completed.whenCompleted(doSomething);
+```
+
 
 ## SpriteAnimationGroup
 
--- a/doc/flame/components.md
+++ b/doc/flame/components.md
@@ -364,7 +364,7 @@ this.player = SpriteAnimationComponent.fromFrameData(
 If you are not using `FlameGame`, don't forget this component needs to be updated, because the
 animation object needs to be ticked to move the frames.
 
-To listen when the animation is done (when it reaches its last frame and is not looping) you can
+To listen when the animation is done (when it reaches the last frame and is not looping) you can
 use `animation.completed`.
 
 Example:
--- a/doc/flame/components.md
+++ b/doc/flame/components.md
@@ -375,7 +375,7 @@ doSomething();
 
 // or alternatively
 
-animation.completed.whenCompleted(doSomething);
+animation.completed.whenComplete(doSomething);
 ```
 
 
--- a/doc/flame/components.md
+++ b/doc/flame/components.md
@@ -364,6 +364,9 @@ this.player = SpriteAnimationComponent.fromFrameData(
 If you are not using `FlameGame`, don't forget this component needs to be updated, because the
 animation object needs to be ticked to move the frames.
 
+To listen when the animation is done (when it reaches its last frame and is not looping) you can
+use `animation.completed`.
+
 
 ## SpriteAnimationGroup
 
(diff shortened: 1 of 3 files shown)


packages/flame/lib/src/sprite_animation.dart
```diff
@@ -243,6 +246,20 @@ class SpriteAnimation {
   /// still image).
   bool get isSingleFrame => frames.length == 1;
 
+  /// A future that will complete when the animation completes.
+  ///
+  /// An animation is considered to be completed if it reaches its [isLastFrame]
+  /// and is not [loop]ing.
+  Future<void> get completed {
```
wolfenrain, line 253: When I read this property name I would assume it returns a `bool` instead of a `Future<void>`

Maybe a more descriptive name would fit? Something`whenCompleted`? What do you think @st-pasha @spydon

spydon: This follows the same pattern as `mounted` etc where we also had this discussion, so I think this is fine.
A bool should have `is` or `has` prepended to it in most cases.

alestiago: I'm open to renaming but this follows exactly the same logic that `loaded` and `mounted` in `Component` have.

doc/flame/components.md
```diff
@@ -364,6 +364,9 @@ this.player = SpriteAnimationComponent.fromFrameData(
 If you are not using `FlameGame`, don't forget this component needs to be updated, because the
 animation object needs to be ticked to move the frames.
 
+To listen when the animation is done (when it reaches its last frame and is not looping) you can
+use `animation.completed`.
```
spydon, line 368: An example of how to use it would be good here

packages/flame/test/sprite_animation_test.dart
```diff
@@ -55,6 +58,71 @@ void main() {
...
+        );
+
+        animation.update(1);
+
+        expectLater(animation.completed, completes);
+      },
+    );
+
+    test(
+      "completed doesn't completes when animation is yet to complete",
```
spydon, line 92:
```suggestion
      "completed doesn't complete when the animation is yet to complete",
```

packages/flame/lib/src/sprite_animation.dart
wolfenrain: Aight, that clears it up for me!

packages/flame/test/sprite_animation_test.dart
```diff
@@ -55,6 +58,71 @@ void main() {
...
+        loop: false,
+      );
+
+      expectLater(animation.completed, completes);
+
+      animation.update(1);
+    });
+
+    test(
+      'completed completes when animation has alrady completed',
```
spydon, line 76:
```suggestion
      'completed completes when the animation has already completed',
```

packages/flame/test/sprite_animation_test.dart
```diff
@@ -55,6 +58,71 @@ void main() {
...
+      () async {
+        final sprite = MockSprite();
+        final animation = SpriteAnimation.spriteList([sprite], stepTime: 1);
+
+        expectLater(animation.completed, doesNotComplete);
+      },
+    );
+
+    test(
+      "completed doesn't completes when animation is looping and on last frame",
```
spydon, line 116:
```suggestion
      "completed doesn't complete when animation is looping and on last frame",
```

doc/flame/components.md
```diff
@@ -364,6 +364,20 @@ this.player = SpriteAnimationComponent.fromFrameData(
...
+
+For example, you can:
```
spydon, line 370:
```suggestion
Example:
```
A sentence can't end with "you can" :)

alestiago: Good catch, I was thinking "For example, you can do as follows:" but forgot to actually write it down haha.

doc/flame/components.md
```diff
@@ -364,6 +364,20 @@ this.player = SpriteAnimationComponent.fromFrameData(
 If you are not using `FlameGame`, don't forget this component needs to be updated, because the
 animation object needs to be ticked to move the frames.
 
+To listen when the animation is done (when it reaches its last frame and is not looping) you can
```
wolfenrain, line 367:
```suggestion
To listen when the animation is done (when it reaches the last frame and is not looping) you can
```

wolfenrain (approved):

spydon (approved):

PR #1566, opened 2022-04-22
feat: add `ParentIsA` to force parent child relations

wolfenrain:
# Description

The `HasParent` mixin allows the developer to ensure that their component will always have a given parent type. It can't be added to a parent of a different type.

Therefore it can provide a strongly typed parent.

## Checklist

## Breaking Change

## Related Issues


packages/flame/lib/src/game/mixins/has_parent.dart
```diff
@@ -0,0 +1,13 @@
...
+/// A mixin that ensures a parent is of the given type [T].
+mixin HasParent<T extends Component> on Component {
+  @override
+  T get parent => super.parent! as T;
+
+  @override
+  void onMount() {
+    assert(super.parent is T, 'Parent must be of type ${T.toString()}');
+    super.onMount();
+  }
```
alestiago, line 12:
```suggestion
  @override
  Future<void> addToParent(Component parent) async {
    assert(
      parent is T,
      'Parent must be of type ${T.toString()}.',
    );
    await super.addToParent(parent);
  }
```

st-pasha: 1. `'$T'` is simpler than `${T.toString()}`
2. `addToParent()` is not async

st-pasha:
Possible names:
- `ParentIsA<T>`
- `ParentHasType<T>`

The closest equivalent in Dart is the `isA<T>()` matcher in `flutter_test`.

packages/flame/lib/src/game/mixins/has_parent.dart
alestiago: Agree with 1, just copied the same assert message and didn't pay attention. Now thinking about it, both approaches will display an invalid string if toString is overrided.

With 2 that's true. But wouldn't the computation be stopped before in addToParent than in onMount?

wolfenrain:
> Possible names:
> 
> * `ParentIsA<T>`
> * `ParentHasType<T>`
> 
> The closest equivalent in Dart is the `isA<T>()` matcher in `flutter_test`.

The `ParentIsA<T>` is best suited! It explains it and tells the developer that it will enforce something as well.

packages/flame/lib/src/game/mixins/has_parent.dart
wolfenrain: Good call for doing this on the addToParent! Will test if that works as expected but I am quite sure it will

Also the `toString` is a Github copilot "mistake", will turn that into `$T`. The toString can't be overridden by the user as it is a generic Type so I don't think that is a concern for us.

wolfenrain: Reworked it to the `addToParent`, works as expected.

alestiago:
I fully agree with @wolfenrain, my vote goes for `ParentIsA`.

My only concern is how it deviates from the standard `HasFoo` going on.

doc/flame/components.md
```diff
@@ -158,6 +158,24 @@ available eventually: after they are loaded and mounted. We can only assure
...
+#### Ensuring a component has a given parent
+
+When a component requires to be added to a specific parent type the 
+`HasParent` mixin can be used to enforce a strongly typed parent.
+
+Example:
+
+```dart
+class MyComponent extends Component with HasParent<MyParentComponent> {
+  Future<void> onLoad() async {
```
st-pasha, line 170: Add `@override`

doc/flame/components.md
```diff
@@ -158,6 +158,24 @@ available eventually: after they are loaded and mounted. We can only assure
...
+    // parent is of type MyParentComponent
+    print(parent.myValue);
+  }
+}
+```
+
+If `MyComponent` is added to a component that is not `MyParentComponent` 
+an assertion error will be thrown stating such.
```
st-pasha, line 178:
```suggestion
If you try to add `MyComponent` to a parent that is not `MyParentComponent`, 
an assertion error will be thrown.
```

packages/flame/test/game/mixins/has_parent_test.dart
```diff
@@ -0,0 +1,32 @@
...
+
+      expect(component.parent, isA<ParentComponent>());
+    });
+
+    testWithFlameGame('throws assertion error when the wrong parent is used',
+        (game) async {
+      final parent = DifferentComponent();
+      final component = TestComponent();
+
+      expect(() => parent.add(component), throwsAssertionError);
```
st-pasha, line 29:
```suggestion
      expect(() => parent.add(component), failsAssert('Parent must be of type ParentComponent'),);
```

packages/flame/test/game/mixins/has_parent_test.dart
```diff
@@ -0,0 +1,32 @@
...
+void main() {
+  group('HasParent', () {
+    testWithFlameGame('successfully sets the parent link', (game) async {
+      final parent = ParentComponent();
+      final component = TestComponent();
+
+      await parent.add(component);
+      await game.add(parent);
+
+      expect(component.parent, isA<ParentComponent>());
```
st-pasha, line 21: This would have succeeded even if there was no `HasParent` mixin.
Perhaps we could add some `ParentComponent { int foo; }` property, and then check that `component.parent.foo` can be read without an explicit type-cast?

doc/flame/components.md
```diff
@@ -158,6 +158,24 @@ available eventually: after they are loaded and mounted. We can only assure
 that they will appear in the children list in the same order as they were
 scheduled for addition.
 
+#### Ensuring a component has a given parent
```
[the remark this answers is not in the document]
wolfenrain: I did that as a fourth as I didnt find a good place to put this, and this chapter talks about children order and parent relations so I thought this might be a nice fit.

packages/flame/test/game/mixins/has_parent_test.dart
wolfenrain: The problem with that is, we can't. It would fail before that, and if it didn't fail it will be correct anyway. It wouldn't test anything really. This whole test doesn't really test anything except that there was no error from the `add`

spydon (approved):
`ParentIsA` sounds good to me too!

st-pasha:
I'm worried about overriding `addToParent` though -- perhaps the previous approach with `onMount` is better?

This is because there are several ways to add a component to a given parent, and `addToParent` is only one of them. There is also `add()` and `addAll()` which currently defer to `addToParent`, but this is merely an implementation detail which may change at some point in the future. Then, there is `changeParent()` which is currently independent of `addToParent`, which means it would circumvent the check in `ParentIsA<T>`.

wolfenrain:
> I'm worried about overriding `addToParent` though -- perhaps the previous approach with `onMount` is better?
> 
> This is because there are several ways to add a component to a given parent, and `addToParent` is only one of them. There is also `add()` and `addAll()` which currently defer to `addToParent`, but this is merely an implementation detail which may change at some point in the future. Then, there is `changeParent()` which is currently independent of `addToParent`, which means it would circumvent the check in `ParentIsA<T>`.

You bring up a good point here, it is indeed more susceptible to breaking changes (even tho we should catch these bugs). But the onMount has the downside that the user will only know it was "wrong" after the component gets mounted and not when you are already adding it to a parent. 

I personally think that the onMount is indeed a safer bet to guarantee that it won't break when we rework our adding logic.

packages/flame/test/game/mixins/has_parent_test.dart
```diff
@@ -0,0 +1,35 @@
...
+      final parent = DifferentComponent();
+      final component = TestComponent();
+
+      expect(
+        () => parent.add(component),
+        failsAssert('Parent must be of type ParentComponent'),
+      );
+    });
+  });
+}
```
alestiago, line 35: @wolfenrain 

Can we add a test that "throws assertion error when changed to the wrong parent"?

Something like:
1. Adds a component to a valid parent
2. Calls `changeParent` to an invalid parent
3. Expect an error

packages/flame/lib/src/game/mixins/has_parent.dart
```diff
@@ -0,0 +1,16 @@
...
+mixin HasParent<T extends Component> on Component {
+  @override
+  T get parent => super.parent! as T;
+
+  @override
+  @mustCallSuper
+  Future<void>? addToParent(Component parent) {
+    assert(parent is T, 'Parent must be of type $T');
+    return super.addToParent(parent);
+  }
```
alestiago, line 15:
```suggestion
  @override
  @mustCallSuper
  @override
 Future<void>? addToParent(covariant T parent) {
    return super.addToParent(parent);
  }
```

packages/flame/test/game/mixins/has_parent_test.dart
wolfenrain: as I reworked it to `onMount` for now we can come back to this if we decide to change the way we check it.

doc/flame/components.md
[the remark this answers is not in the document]
spydon: I think this would be fine as a third level header too, if you look at the section after this one it is also third level and I think they fit on the same level.

packages/flame/test/game/mixins/has_parent_test.dart
alestiago: I think no matter how we implement it, it will be nice to have this tested. 

Personally, adding this test does increase my confidence on the codebase.

spydon (approved):

PR #1571, opened 2022-04-25
feat!: Size effects will now work only on components implementing SizeProvider

st-pasha:
# Description

Similar to all other effects, the `SizeEffect` now targets components implementing `SizeProvider`.

Currently the following classes implement `SizeProvider`:
 - `CircleComponent`,
 - `Viewport`,
 - `NineTileBoxComponent`,
 - `SpriteComponent`,
 - `SpriteGroupComponent`,
 - `SpriteAnimationComponent`,
 - `SpriteAnimationGroupComponent`.

Unlike with other effects, `PositionComponent` does NOT implement `SizeProvider`. This is because the components usually don't pay attention to their `size` when rendering, and thus changing their `size` would not create any visual difference, which users assume is a bug. Now this class of bugs will disappear: `SizeEffect` will simply throw an error saying that the target is not a `SizeProvider`.

Breaking change:
 - If users have their own components that they target with `SizeEffect`, they would need to add `implements SizeProvider` to the declaration of their component.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/examples/lib/stories/effects/size_effect_example.dart
+++ b/examples/lib/stories/effects/size_effect_example.dart
@@ -1,8 +1,8 @@
 import 'dart:ui';
 
+import 'package:flame/components.dart';
 import 'package:flame/effects.dart';
 import 'package:flame/game.dart';
-import 'package:flame/geometry.dart';
 import 'package:flame/input.dart';
 import 'package:flame/palette.dart';
 import 'package:flutter/animation.dart';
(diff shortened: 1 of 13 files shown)


spydon:
Many common PositionComponents does take size into consideration when rendering as far as I'm aware, so this will be a breaking change for all those ones. Either all the ones that SizeEffect currently works on need to have SizeProvider, or we need to make all PositionComponents take size into consideration.

st-pasha:
Yep. If a component allows its size to be dynamically updated, it marks itself with `implements SizeProvider`. Otherwise, `SizeEffect` will complain about being attached to a component that's not resizable.

examples/lib/stories/effects/size_effect_example.dart
```diff
@@ -15,35 +15,29 @@ class SizeEffectExample extends FlameGame with TapDetector {
     down, depending on its current state.
   ''';
 
-  late RectangleComponent square;
+  late Component shape;
   bool grow = true;
 
   @override
   Future<void> onLoad() async {
-    square = RectangleComponent.square(
```
spydon, line 23: I wondered how this one could have worked previously, the answer was simple; it didn't. :sweat_smile:

spydon, line 23: I wondered how this one could have worked previously, the answer was simple; it didn't. :sweat_smile:

spydon (approved):

wolfenrain (approved):

PR #1572, opened 2022-04-25
feat: Add range constructor on SpriteAnimationData

oloshe:
When using sprite animation, you can control the range of animation display by passing in the two parameters of start and end

### DIFF
--- a/packages/flame/lib/src/sprite_animation.dart
+++ b/packages/flame/lib/src/sprite_animation.dart
@@ -70,6 +70,37 @@ class SpriteAnimationData {
     });
   }
 
+  /// Specifies the range of the sprite grid.
+  ///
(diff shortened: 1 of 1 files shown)


packages/flame/lib/src/sprite_animation.dart
```diff
@@ -70,6 +70,37 @@ class SpriteAnimationData {
     });
   }
 
+  /// Specifies the range of the sprite grid.
+  ///
+  /// Make sure your sprites are placedleft-to-right and top-to-bottom
```
erickzanardo, line 75:
```suggestion
  /// Make sure your sprites are placed left-to-right and top-to-bottom
```

erickzanardo (approved):
Good addition! Thanks

spydon (approved):

PR #1578, opened 2022-04-26
fix: correctly calculating frame length

oloshe:
Sorry I didn't double check this error in yesterday's pr due to my oversight, because the index starts from 0, and the calculated length should be incremented by 1。


spydon (approved):
And none of us caught it in code review either 😅

oloshe:
> 

Yes, boundary check is so difficult to detect with the naked eye, I found this problem in actual use today

st-pasha:
Shouldn't there be a test for this? Seems like a very easy thing to miss in subsequent refactors.

spydon:
> I agree with that, but I don't seem to find any tests about `SpriteAnimationData`

Feel free to add a file for it, some old classes are missing test files.

PR #1587, opened 2022-04-28
feat: Added the onLongTapDown event

st-pasha:
# Description

- Event `onLongTapDown` added to mixins:
  - `Tappable`
  - `HasTappables`
  - `MultiTouchTapDetector` (it existed here before, but was never invoked, so it was essentially broken)
- The mixins listed above collected into a single `events/` folder, since their functionality is tightly coupled. I feel that putting all functionality related to events handling into a single place would make code maintenance easier.
  - This is not a breaking change: all the classes are still exported as before;
  - We'd probably want to do similar consolidation with other events (such as drag events) in subsequent PRs.
- Added documentation to all aforementioned mixins.
- Due to the fact that `HasTappables` and `MultiTouchTapDetector` are very similar in their structure, created a common interface `MultiTapListener` that both of them implement.
  - This common interface allows removing code duplication at the `GameWidget` level in file `gestures.dart`. 

## Checklist

## Breaking Change


packages/flame/lib/src/events/component_mixins/tappable.dart
```diff
@@ -1,21 +1,24 @@
...
 
+/// Mixin that can be added to any [Component] allowing it to receive tap
+/// events.
+///
+/// When using this mixin, also add [HasTappables] to your game, which handles
+/// propagation of tap events from the root game to individual components.
+///
+/// See [MultiTapGestureRecognizer] for the description of each individual
+/// event.
 mixin Tappable on Component {
```
spydon, line 16: I think that the structure might be clearer with the mixins for components lives in `src/componentes/mixins/`, possibly with a sub directory for events in there. than having the `components_mixins` sub directory here. If we decide to change, I think that the structure should be changed for all the files at once, in a separate PR, so that we don't have some classes related to this in the `events` directory and some on other places in the intermediate time.

spydon (commented):
The code itself looks good! But I don't think the directory structure should change in this PR.

packages/flame/lib/src/events/component_mixins/tappable.dart
st-pasha: Yeah, makes sense. I'll move the files back for now, and then we'll discuss the restructuring in a different PR.

spydon (approved):

packages/flame/lib/src/events/interfaces/multi_tap_listener.dart
```diff
@@ -0,0 +1,33 @@
...
+import '../../game/mixins/has_tappables.dart';
+import '../../game/mixins/multi_touch_tap_detector.dart';
+
+/// Interface that must be implemented by a game in order for it to be eligible
+/// to receive events from a [MultiTapGestureRecognizer].
+///
+/// Instead of implementing this class directly consider using one of the
+/// prebuilt mixins:
+///  - [HasTappables] for a `FlameGame`
+///  - [MultiTouchTapDetector] for a custom `Game`
```
luanpotter, line 12: maybe I am misremembering, lmk if this comment makes sense, but can't you use MultiTouchTapDetector on `FlameGames` if you are not using tappables and want to listen to multi-touch events?

packages/flame/lib/src/game/mixins/has_tappables.dart
```diff
@@ -1,10 +1,25 @@
...
-import '../../../game.dart';
+import '../../components/mixins/tappable.dart';
+import '../../events/interfaces/multi_tap_listener.dart';
 import '../../gestures/events.dart';
+import '../flame_game.dart';
+import 'multi_touch_tap_detector.dart';
 
-mixin HasTappables on FlameGame {
+/// Mixin that can be added to a [FlameGame] allowing it (and the components
+/// attached to the game) to receive tap events.
```
luanpotter, line 11:
```suggestion
/// attached to it) to receive tap events.
```

luanpotter (approved):
Nice!

packages/flame/lib/src/events/interfaces/multi_tap_listener.dart
st-pasha: Since FlameGame is a Game, you can absolutely use MultiTouchTapDetector on it. The HasTappables mixing is just a more concrete implementation, which ensures the event propagates down into the component tree.

PR #1594, opened 2022-04-30
feat: adding KeyboardListenerComponent

erickzanardo:
# Description

Adds a simple component helper to allow users to handle keyboard events directly by a built in component.

## Checklist

## Breaking Change

## Related Issues


examples/lib/stories/input/keyboard_component_example.dart
```diff
@@ -0,0 +1,73 @@
...
+    ember = Ember(position: size / 2, size: Vector2.all(100));
+    add(ember);
+
+    add(
+      KeyboardComponent(
+        keyUp: {
+          LogicalKeyboardKey.keyA: _stopX,
+          LogicalKeyboardKey.keyD: _stopX,
+          LogicalKeyboardKey.keyW: _stopY,
+          LogicalKeyboardKey.keyS: _stopY,
```
luanpotter, line 35: on key up you need to check what other keys are pressed, otherwise this will cause non-ideal behavior.

the more correct impl IMO is:
* user holds A -> goes left
* user holds D -> goes right
* users releases D -> as A is pressed, goes left (not stop, which is the current impl)

I think this a perfect place to have a correct example so we can share with people when they do the naive solution

luanpotter: example from [ghost v. wonderland](https://github.com/bluefireteam/ghost-vs-wonderland/blob/master/game/lib/game.dart#L63)

erickzanardo: Sounds good, will look on ghost game

st-pasha:
At some point in the not-too-distant future I'd like to create a component that draws a keyboard on a screen. Like a built-in system virtual keyboard, only fully styleable so that one can make it look harmonious with the rest of the game. I think this could be a great addition to any game that needs some kind of text input. 

And for such a component the most natural name would be `KeyboardComponent` -- since it's a component that draws a keyboard.

So I was wondering if we could consider renaming the component in this PR in order to avoid name conflict. For example, how about `KeyboardNavigator` or `KeyboardNavigationComponent`?

erickzanardo:
I don mind renaming, but I don think Navigation or Navigator is a good definition for what the component does.

An alternative for me would be KeyboardListenerComponent

spydon:
> An alternative for me would be KeyboardListenerComponent

That's a pretty nice and descriptive name.

packages/flame/lib/src/components/input/keyboard_component.dart
```diff
@@ -0,0 +1,38 @@
...
+import '../../../game.dart';
+import '../../game/mixins/keyboard.dart';
+
+/// The signature for a key handle function
+typedef KeyHandlerCallback = bool Function();
+
+/// {@template keyboard_component}
+/// A [Component] that receives keyboard input and executes registered methods.
+/// This component is based on [KeyboardHandler], which requires the [FlameGame]
+/// which it is used to be mixed with [HasKeyboardHandlerComponents]
```
spydon:
```suggestion
/// which is used to be mixed with [HasKeyboardHandlerComponents].
```

examples/lib/stories/input/keyboard_component_example.dart
erickzanardo: done!

doc/flame/inputs/keyboard-input.md
```diff
@@ -91,6 +91,31 @@ that triggered the callback in the first place. The second is a set of the curre
 The returned value should be `true` to allow the continuous propagation of the key event among other
 components. To not allow any other component to receive the event, return `false`.
 
+Flame also provides a default implementation called `KeyboardListenerComponent` which can be used to handle
+keyboard events which can be composed with a `FlameGame` or any other `Component`:
+
+For example, imagining a `PositionComponent` which has methods to move on the X and Y axis,
+the following code could be used to bind those methods to key events:
```
spydon, line 98:
```suggestion
Flame also provides a default implementation called `KeyboardListenerComponent` which can be used to handle
keyboard events.  It can be used as a mixin with a `FlameGame` or any other `Component`, or just extended:

For example, imagine a `PositionComponent` which has methods to move on the X and Y axis,
then the following code could be used to bind those methods to key events:
```

examples/lib/stories/input/keyboard_listener_component_example.dart
```diff
@@ -0,0 +1,131 @@
...
+                false,
+                LogicalKeyboardKey.keyW,
+                keys,
+              ),
+          LogicalKeyboardKey.keyS: (keys) => _handleKey(
+                false,
+                LogicalKeyboardKey.keyS,
+                keys,
+              ),
+        },
```
luanpotter, line 51: can be a followup, but I wonder if we can provide a less verbose API where you just list the keys we want to listen on an array and bind 1 `_handleKey` function taking 3 parameters

luanpotter (approved):

doc/flame/inputs/keyboard-input.md
```diff
@@ -91,6 +91,32 @@ that triggered the callback in the first place. The second is a set of the curre
 The returned value should be `true` to allow the continuous propagation of the key event among other
 components. To not allow any other component to receive the event, return `false`.
 
+Flame also provides a default implementation called `KeyboardListenerComponent` which can be used to handle
+keyboard events. Like any other component, it can be added as a child to a `FlameGame` or another
+`Component`:
```
spydon, line 96:
```suggestion
Flame also provides a default implementation called `KeyboardListenerComponent` which can be used to
handle keyboard events. Like any other component, it can be added as a child to a `FlameGame` or
another `Component`:
```

packages/flame/lib/src/components/input/keyboard_listener_component.dart
```diff
@@ -0,0 +1,38 @@
...
+import 'package:flame/src/game/mixins/keyboard.dart';
+import 'package:flutter/services.dart';
+
+/// The signature for a key handle function
+typedef KeyHandlerCallback = bool Function(Set<LogicalKeyboardKey>);
+
+/// {@template keyboard_listener_component}
+/// A [Component] that receives keyboard input and executes registered methods.
+/// This component is based on [KeyboardHandler], which requires the [FlameGame]
+/// which is is used to be mixed with [HasKeyboardHandlerComponents].
```
spydon, line 12:
```suggestion
/// which is used to be mixed with [HasKeyboardHandlerComponents].
```

examples/lib/stories/input/keyboard_listener_component_example.dart
```diff
@@ -0,0 +1,131 @@
...
+import 'package:flame/palette.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+
+class KeyboardListenerComponentExample extends FlameGame
+    with HasKeyboardHandlerComponents {
+  static const String description = '''
+    Similar to the default Keyboard example, but shows a different implementation approach,
+    which uses Flame's KeyboardListenerComponent to handle input.
+    Usage: Use A S D W to steer Ember.
```
spydon, line 14:
```suggestion
    Similar to the default Keyboard example, but shows a different
    implementation approach, which uses Flame's
    KeyboardListenerComponent to handle input.
    Usage: Use A S D W to steer Ember.
```

spydon (approved):

PR #1595, opened 2022-05-01
feat: Add `FpsComponent` and `FpsTextComponent`

spydon:
# Description

This PR adds two components:
- `FpsComponent` which calculates the current FPS within a window.
- `FpsTextComponent` which renders the current FPS on the screen using a `TextComponent`.

Will add docs and tests if you think it looks like a good idea.

## Checklist

## Breaking Change

## Related Issues

Related to #1582 

### DIFF
--- a/packages/flame/lib/src/components/fps_component.dart
+++ b/packages/flame/lib/src/components/fps_component.dart
@@ -12,22 +12,25 @@ class FPSComponent extends Component {
 
   /// The number of game ticks over which the fps measure will be averaged.
   final int windowSize;
-  /// The queue of the recent game tick durations. The length of this queue will not exceed [windowSize].
-  final Queue<double> _window = Queue();
-  /// The sum of all values in the [_window] queue.
+
+  /// The queue of the recent game tick durations.
+  /// The length of this queue will not exceed [windowSize].
+  final Queue<double> window = Queue();
+
+  /// The sum of all values in the [window] queue.
   double _sum = 0;
 
   @override
   void update(double dt) {
-    _window.addLast(dt);
+    window.addLast(dt);
     _sum += dt;
-    if (_window.length > windowSize) {
-      _sum -= _window.removeFirst();
+    if (window.length > windowSize) {
+      _sum -= window.removeFirst();
     }
   }
 
   /// Get the current average FPS over the last [windowSize] frames.
   double get fps {
-    return _window.isEmpty ? 0 : _window.length / _sum;
+    return window.isEmpty ? 0 : window.length / _sum;
   }
 }
--- a/packages/flame/lib/src/components/fps_component.dart
+++ b/packages/flame/lib/src/components/fps_component.dart
@@ -2,11 +2,11 @@ import 'dart:collection';
 
 import '../../components.dart';
 
-/// The [FPSComponent] is a non-visual component which you can get the current
+/// The [FpsComponent] is a non-visual component which you can get the current
 /// fps of the game with by calling [fps], once the component has been added to
 /// the component tree.
-class FPSComponent extends Component {
-  FPSComponent({
+class FpsComponent extends Component {
+  FpsComponent({
     this.windowSize = 60,
   });
 
--- a/packages/flame/lib/src/components/fps_component.dart
+++ b/packages/flame/lib/src/components/fps_component.dart
@@ -16,15 +16,16 @@ class FPSComponent extends Component {
 
   @override
   void update(double dt) {
-    _window.add(dt);
-    _sum += dt;
-    if (_window.length >= windowSize) {
+    final frameFps = 1 / dt;
+    _window.add(frameFps);
+    _sum += frameFps;
+    if (_window.length > windowSize) {
       _sum -= _window.removeFirst();
     }
   }
 
   /// Get the current average FPS over the last [windowSize] frames.
   double get fps {
-    return 1 / (_sum / _window.length);
+    return _window.isEmpty ? 0 : _sum / _window.length;
   }
 }
--- a/packages/flame/lib/src/components/fps_component.dart
+++ b/packages/flame/lib/src/components/fps_component.dart
@@ -10,6 +10,7 @@ class FPSComponent extends Component {
     this.windowSize = 60,
   });
 
+  /// The number of game ticks over which the fps measure will be averaged.
   final int windowSize;
   final Queue<double> _window = Queue();
   /// The sum of all values in the [_window] queue.
--- a/packages/flame/lib/src/components/fps_component.dart
+++ b/packages/flame/lib/src/components/fps_component.dart
@@ -12,6 +12,7 @@ class FPSComponent extends Component {
 
   /// The number of game ticks over which the fps measure will be averaged.
   final int windowSize;
+  /// The queue of the recent game tick durations. The length of this queue will not exceed [windowSize].
   final Queue<double> _window = Queue();
   /// The sum of all values in the [_window] queue.
   double _sum = 0;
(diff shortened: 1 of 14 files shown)


luanpotter (approved):
LGTM

st-pasha:
FYI: Dart style guide [recommends](https://dart.dev/guides/language/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words) capitalizing acronyms like words: `FpsComponent`, `FpsTextComponent`.

packages/flame/lib/src/components/fps_component.dart
```diff
@@ -0,0 +1,31 @@
...
+    _sum += frameFps;
+    if (_window.length > windowSize) {
+      _sum -= _window.removeFirst();
+    }
+  }
+
+  /// Get the current average FPS over the last [windowSize] frames.
+  double get fps {
+    return _window.isEmpty ? 0 : _sum / _window.length;
+  }
```
st-pasha, line 30: I recommend averaging times `dt`, not the `fps` values. Consider, for example, what would happen if there is a single `dt` that is very high, or very low.

```suggestion
  void update(double dt) {
    _window.addLast(dt);
    _sum += dt;
    if (_window.length > windowSize) {
      _sum -= _window.removeFirst();
    }
  }

  /// Get the current average FPS over the last [windowSize] frames.
  double get fps {
    return _window.isEmpty ? 0 : _window.length / _sum;
  }
```

packages/flame/lib/src/components/fps_text_component.dart
```diff
@@ -0,0 +1,40 @@
...
+        );
+
+  final int windowSize;
+  final int decimalPlaces;
+  late final FPSComponent fpsComponent;
+
+  @override
+  Future<void> onLoad() async {
+    positionType = PositionType.game;
+    add(fpsComponent = FPSComponent(windowSize: windowSize));
```
st-pasha, line 33: Suggestion: move these lines into the constructor, then you wouldn't have to store the `windowSize` variable.

packages/flame/lib/src/components/fps_text_component.dart
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  @override
+  void update(double dt) {
+    text = '${fpsComponent.fps.toStringAsFixed(decimalPlaces)}FPS';
```
st-pasha, line 38:
```suggestion
    text = '${fpsComponent.fps.toStringAsFixed(decimalPlaces)} FPS';
```

packages/flame/test/components/fps_component_test.dart
```diff
@@ -0,0 +1,50 @@
+import 'package:flame/components.dart';
+import 'package:flame_test/flame_test.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('FPSComponent', () {
+    testWithFlameGame('reports correct FPS for 1 frames', (game) async {
+      final fpsComponent = FPSComponent();
+      await game.ensureAdd(fpsComponent);
+      game.update(1 / 60);
```
st-pasha, line 10:
```suggestion
      expect(fpsComponent.fps, 0);
      game.update(1 / 60);
```

packages/flame/test/components/fps_component_test.dart
```diff
@@ -0,0 +1,50 @@
...
+      expect(fpsComponent.fps, 60.0);
+    });
+
+    testWithFlameGame('reports correct FPS with varying dt', (game) async {
+      const windowSize = 40;
+      final fpsComponent = FPSComponent(windowSize: windowSize);
+      await game.ensureAdd(fpsComponent);
+      for (var i = 0; i < 1.5 * windowSize; i++) {
+        // Alternating between 50 and 100 FPS
+        final dt = 1 / (100 / (1 + i % 2));
```
st-pasha, line 43:
```suggestion
        final dt = i.isEven? 1/100 : 1/50;
```

packages/flame/test/components/fps_component_test.dart
```diff
@@ -0,0 +1,50 @@
...
+        game.update(dt);
+      }
+
+      expect(fpsComponent.fps, 75.0);
```
st-pasha, line 47: This should be around 67.
If one frame takes 0.01s, and next 0.02s, then 2 frames take 0.03s, and 66 frames take 0.99s. Add one more 0.01s frame -- and there were 67 frames in total within the last 1.00 second.

packages/flame/lib/src/components/fps_component.dart
spydon: I did that from the beginning, but the rounding errors become quite big.

```
So say that we do 4 frames, 2 of them has the dt for 40fps and 2 for 60fps:
40fps = 1/40 = 0.025dt
60fps = 1/60 = 0.016...dt
(0.025+0.025+0.016+0.016)/4=0.0208... (average dt)
1/0.0208=48fps (dt to fps)
(Should be 50)
```

st-pasha: The average of 40fps and 60fps is `2/(1/40 + 1/60) = 48 fps`.
"FPS" literally means "frames per second", i.e. how many frames there were (on average) in one second.

Imagine you have one frame rendering in 0.999s, and another frame rendering in 0.001s. Would you say it's 2 frames per second; or is it (1.001 + 1000)/2 ≈ 500 fps?

spydon: That's true, late Sunday evenings clearly doesn't produce the best code. 😄

packages/flame/lib/src/components/fps_component.dart
```diff
@@ -0,0 +1,36 @@
...
+  FPSComponent({
+    this.windowSize = 60,
+  });
+
+  /// The number of game ticks over which the fps measure will be averaged.
+  final int windowSize;
+
+  /// The queue of the recent game tick durations.
+  /// The length of this queue will not exceed [windowSize].
+  final Queue<double> window = Queue();
```
wolfenrain, line 18: This currently is writeable from the outside, while I love to expose the values, can we somehow make sure people cant add their own values to it?

wolfenrain (approved):
LGTM, one small nit.

packages/flame/lib/src/components/fps_component.dart
spydon: I did that intentionally actually, so that people can calculate other things on the window. Since this should be as efficient as possible I don't want to wrap it in an `UnmodifiableList`, but I should write a comment about that it shouldn't be modfied manually. I think the pinned message in telegram applies here. 😄

packages/flame/lib/src/components/fps_component.dart
```diff
@@ -0,0 +1,36 @@
...
+/// The [FPSComponent] is a non-visual component which you can get the current
+/// fps of the game with by calling [fps], once the component has been added to
+/// the component tree.
+class FPSComponent extends Component {
+  FPSComponent({
+    this.windowSize = 60,
+  });
+
+  /// The number of game ticks over which the fps measure will be averaged.
+  final int windowSize;
```
erickzanardo, line 14: This name is a little bit confusing for me, at first I thought this had to do with the game window. Maybe we could call it `expectedFps` or something like that?

erickzanardo (commented):
Overall LGTM! Just one minor comment

packages/flame/lib/src/components/fps_component.dart
spydon: I updated the dartdoc for the field

erickzanardo (approved):

aytunch:
Hi, I can not use `FpsTextComponent` with Flame 1.1.1
`add(FpsTextComponent(position: Vector2(0, size.y - 24)));`
Is this not merged already? Do I need to do some manual import for it to work?

spydon:
Hi, it is merged, but not released. We'll have a release in the coming days that ibcludes it.

PR #1597, opened 2022-05-01
fix: Use a fixed update rate for `Forge2DGame`

spydon:
# Description

I think this is how it is supposed to be, to avoid overflowing values in the calculations if there is a lag.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame_forge2d/lib/forge2d_game.dart
+++ b/packages/flame_forge2d/lib/forge2d_game.dart
@@ -21,10 +21,12 @@ class Forge2DGame extends FlameGame {
 
   final World world;
 
+  static const _worldStepTime = 1 / 60;
+
   @override
   void update(double dt) {
     super.update(dt);
-    world.stepDt(dt);
+    world.stepDt(_worldStepTime);
   }
 
   Vector2 worldToScreen(Vector2 position) {


wolfenrain (approved):
It obviously looks good to me, but do we have any sources to confirm that the world step time should be constant?

It makes sense from a frame rendering vs world physics side of view but would love to read more about it

st-pasha:
Some devices run at 100 or 120 fps, so this fixed fps would obviously create problems on such devices.
On the other hand, some games may have such a complicated environment that they cannot render within 60fps budget - for these games the time will now flow slower than expected.

st-pasha:
Some devices run at 100 or 120 fps, so this fixed fps would obviously create problems on such devices.
On the other hand, some games may have such a complicated environment that they cannot render within 60fps budget - for these games the time will now flow slower than expected.

alestiago:
@spydon Are you suggesting having `worldStepTime` be optionally specified by the developer?

spydon:
> So I'm not really sure what to recommend here, since I don't fully understand which problem is this trying to solve. Maybe @spydon can elaborate a little on the nature of the bug that he's trying to fix?

I think the solution here isn't optimal, I read up on it a bit more this morning. I think that the problem is that if `dt` becomes too large, the physics engine overflows int in some calculations and produces weird results, but I'm not certain this is the case so I'll do some more testing.
What one could do instead is use some max value of `dt`, and as long as it is under that value it just uses the normal `dt`.

alestiago:
@spydon I have reached what you've mentioned:

> I think that the problem is that if dt becomes too large, the physics engine overflows int in some calculations and produces weird results.

More specifically a motorised revolute joint starts randomly spinning and surpassing the limits. When debugging it one of our hypotheses is regarding the one you mention here.

st-pasha:
I like the `min(dt, maxDt)` approach, because it means the game runs normally under normal circumstances, but then if there is a single spike in latency due to system, it would just effectively pause the game for that period.

This could be in fact a useful functionality for any game, not just Forge2D game -- so perhaps it can be added at a GameLoop level? As long as there is a way to opt out of it, just in case.

erickzanardo:
What if we come with a threshold that we know that it will not bug the physics the engine, and if dt is bigger than that, we call the update of the forge2d world the X amount of times that the threshold fits on it?

That way we avoid the overflowing and we keep both worlds synched

st-pasha:
If the bottleneck is physics simulation, then we better just drop the fps and update more rarely (with larger dt). 

On the other hand, there is a possibility of a big spike in dt for one frame only. I'm not sure whether that would actually happen, but there could be things like daytime saving clock change, time zone change, the system updating its clock because it synchronized with the internet, or the user changing their time manually. In these cases, we don't want to run update() with dt that is 1000 or 10000000 times larger than usual, nor do we want to repeat update() a million times; and we definitely don't want to run with a negative dt.

spydon:
> If the bottleneck is physics simulation, then we better just drop the fps and update more rarely (with larger dt). 

The problem is that you can't have too large `dt`s in the Forge2D without causing problems.

> On the other hand, there is a possibility of a big spike in dt for one frame only. I'm not sure whether that would actually happen, but there could be things like daytime saving clock change, time zone change, the system updating its clock because it synchronized with the internet, or the user changing their time manually. In these cases, we don't want to run update() with dt that is 1000 or 10000000 times larger than usual, nor do we want to repeat update() a million times; and we definitely don't want to run with a negative dt.

It's based on the ticks of a timer, not `date.now-lastDate` right?

Anyhow, I think in the first solution we'll just cap the `dt` for the physics simulation (with an option to turn the cap off) and let the other update be like it is now.

st-pasha:
> It's based on the ticks of a timer

The documentation for `Ticker` doesn't mention what clock is being used, and I couldn't find it in the code either...

spydon:
This produces more weirdness than it solves unfortunately so I'll close it.

PR #1601, opened 2022-05-03
fix: Component can now be removed in any lifecycle stage

### LINKED ISSUE #1532: Error when add and remove component while game engine paused

### LINKED ISSUE #1549: game frozen with assertion failed

st-pasha:
# Description

- It should now be possible to remove a component from any possible lifecycle stage.
- The `Component._state` variable converted from an enum into a bitfield in order to be able to better track the state.
- Removed restriction that components can't be added to a game from the constructor.

## Checklist

- [x] The title of my PR starts with a [Conventional Commit] prefix.
- [x] I have updated/added relevant documentation in dartdoc comments.

## Breaking Change

## Related Issues

Closes #1549 
Closes #1532


packages/flame/lib/src/components/component.dart
```diff
@@ -765,28 +760,28 @@ class _LifecycleManager {
     return !(_children.isEmpty && _removals.isEmpty && _adoption.isEmpty);
   }
 
+  /// Attempt to resolve pending events in all lifecycle event queues.
+  ///
+  /// This method must be periodically invoked by the game engine, in order to
```
spydon, line 765: Periodically -> each tick?

spydon (approved):
Looks good! Checks are failing but I'm sure you're aware.

packages/flame/lib/src/components/component.dart
```diff
@@ -133,6 +157,8 @@ class Component {
     removeFromParent();
   }
 
+  //#region Debug rendering
```
alestiago, line 160: What IDE are you using @st-pasha ?

packages/flame/lib/src/components/component.dart
```diff
@@ -37,21 +36,46 @@ class Component {
...
+  ///     [onLoad] method, and can be checked via the [isLoaded] getter. Once
+  ///     set, this bit is never turned off.
+  ///
+  /// [_mounted]: this flag is set when the component becomes properly mounted
+  ///     to the component tree, and then turned off when the component is
+  ///     removed from the tree.
+  ///
+  /// [_removing]: this bit indicates that the component is scheduled for
+  ///     removal at the earliest possible opportunity, and then cleared when
+  ///     the component is actually removed.
```
alestiago, line 62: This is an awesome documentation! Great job!

Any chance we can add an example of how this changes when a component is added, changed parent, and then removed?

alestiago (commented):
Nice work!

I still need to have a better dive. But overall I'm impressed on the quality of the documentation.

Having said this, I'm concerned about using bit operations. My concern is that usually Flutter developers do not perform these type of operations, hence this code can be seen as slightly more difficult to comprehend.

In addition, personally I think that it requires more thought to read bit operations since I calculate them mentally, over just reading if something is equal to something else. 

I'm thinking how can we make this code more readable, especially for beginners. 

Perhaps we can have an extension on int and have methods like `is(int value)` that checks that `this && value != 0` (mask). It would be nice to keep the performance whilst not sacrificing so much readability. Or keep the enum but have more states and map them to bits with an extension (until enhanced enums).

Also, I feel that there are many combinations that _state can have.  With no guarantee that an invalid state can be reached. Making it more error prone and difficult to maintain. Perhaps we can have a setter that, for example, checks that the state is not loaded and removed simultaneously. A possible implementation is to have the  but assertion that checks _state is always a valid state.

packages/flame/test/components/component_lifecycle_test.dart
```diff
@@ -4,25 +4,36 @@ import 'package:flame_test/flame_test.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_test/flutter_test.dart';
 
-class _MyComponent extends Component {
+class LifecycleComponent extends Component {
```
alestiago, line 7: Why public? Shouldn't this be self contained in this test file?

st-pasha: This component is also used in the `component_test.dart`. Technically, the two test files should be combined (since we decided on a 1 test file per source file policy), but we haven't gotten to that yet. Once the files are combined, this component can be made private once again.

packages/flame/lib/src/components/component.dart
st-pasha: Android Studio -- I believe Erick taught me this trick some long time ago. Does this annotation not work in any other IDE?

spydon:
> Having said this, I'm concerned about using bit operations. My concern is that usually Flutter developers do not perform these type of operations, hence this code can be seen as slightly more difficult to comprehend.

They are only internal, I think this is a perfect use for bitfield and bit operations.
(And if someone wants to do changes to this behavior, it really isn't any rocket science to read up on how it works)

> I'm thinking how can we make this code more readable, especially for beginners.

Beginners really shouldn't be fiddling around in this part of the code. :sweat_smile: 

> Also, I feel that there are many combinations that _state can have. With no guarantee that an invalid state can be reached. Making it more error prone and difficult to maintain. Perhaps we can have a setter that, for example, checks that the state is not loaded and removed simultaneously. A possible implementation is to have the but assertion that checks _state is always a valid state.

That's a good point, but I think there are sufficient tests (and if there aren't they should be added) that ensures that a faulty state can't be reached, note again that the user isn't manipulating this field freely, but the internal state machine changes it.

packages/flame/lib/src/components/component.dart
```diff
@@ -493,22 +517,26 @@ class Component {
...
   }
 
-  Future<void>? _load() {
+  Future<void>? _startLoading() {
+    assert(_state == _initial);
     assert(_parent != null);
-    assert(_state == LifecycleState.uninitialized);
-    _state = LifecycleState.loading;
+    assert(_parent!.findGame() != null);
+    assert(_parent!.findGame()!.hasLayout);
```
erickzanardo, line 524: We are starting to have a quite some assertions here, would it be good to have a message to give more context about them?

erickzanardo (approved):
I must confess that bit operations scares me a little bit (HA, get it?).

Not that I can't understand them, but they are not so human friendly, so if possible I think we could add some internal comments on the lines where the operations take place.

But even with that said, this is a nice clean up, so LGTM!!

packages/flame/lib/src/components/component.dart
st-pasha: I'm trying to distinguish between assertions that trigger when the user does something wrong -- these assertions will have proper messages, -- versus the assertions that can only trigger if we did something wrong, i.e. if there is a bug in our code. The latter kind of assertions are not possible to trigger, at least we don't know how they can be possibly triggered (if we did, we'd be able to fix that). So their function is to (1) state the pre-/post-conditions of a method (for example, at a start of this method we believe that the parent is set, and that the game is laid out), and (2) guard against possible problems during future refactors.

packages/flame/lib/src/components/component.dart
st-pasha: Currently we do call this method on every game tick.

But nothing stops us from calling it more or less often. For example, someone might decide to call this twice per tick - once at the beginning of update and once after the render. Or, conversely, you could decide to call this less often if the fps becomes too low.

Seems like "periodically" fits quite well to these.

st-pasha:
Thanks for the feedback, I see there's some desire for more comments about how this all fits together, so i'll try to add those later tonight.

alestiago:
@spydon I don't need any completer to be removed. Sorry for the confusion I meant currently we have a "loaded" completer, a "mounted" completer, but we don't have a "removed" completer.

st-pasha:
> I was wondering if after this change we could add it.

We can, absolutely

packages/flame/lib/src/components/component.dart
st-pasha: Added more detailed explanation of how state evolves over time.

packages/flame/lib/src/components/component.dart
```diff
@@ -435,38 +550,30 @@ class Component {
...
-  /// mounted.
+  /// Removes a component from the component tree.
+  ///
+  /// This will call [onRemove] for the component and its children, but only if
+  /// there was an [onMount] call previously, i.e. when removing a component
+  /// that was properly mounted.
+  ///
+  /// A component can be removed even before it finishes mounting, however such
+  /// component cannot be added back into the tree until it at least finishes
+  /// loading.
```
spydon, line 576: Couldn't we place it straight into the loading list if it is only halfway through loading when it is re-added?

spydon (approved):
Excellent docs for `_state`, probably the longest one we have for a field anywhere in the whole code base.

packages/flame/lib/src/components/component.dart
st-pasha: I'm worried about the case when the user is accessing `parent` from `onLoad`, and then suddenly that parent would change or become null in the middle of their code.
Of course, accessing `parent` from `onLoad` is a kind of an anti-pattern, but we have been allowing it so far, so not sure how to handle this.

We could, for example, make it an error to access `parent` while the component `isLoading` -- but that would be a breaking change, and has to be done in a separate PR, after careful consideration.

spydon: Right, I think that could be a good change so that people move such logic to `onMount` instead.
But as you said, outside of the scope of this PR then.

PR #1602, opened 2022-05-03
docs: Added the T-Rex example

spydon:
# Description

Updated the "clone" of the classic Chrome offline T-Rex game from here:
https://github.com/bluefireteam/trex-flame (shout out to @renancaraujo)
and moved it into the examples.

As @st-pasha mentioned the hamburger menu became invisible since it's white, so I had to add a small margin to the game.
image

## Checklist

## Breaking Change

## Related Issues


renancaraujo (approved):
Approved!

st-pasha:
Getting this error:
```
> flutter --no-color pub get
Running "flutter pub get" in examples...                        
The current Flutter SDK version is 2.8.0.

Because examples depends on google_fonts >=2.3.1 which requires Flutter SDK version >=2.10.0-0, version solving failed.
pub get failed (1; Because examples depends on google_fonts >=2.3.1 which requires Flutter SDK version >=2.10.0-0, version solving failed.)
```

The environment lists min Flutter version 2.5.0, so we should either upgrade the Flutter version, or downgrade the google_fonts requirement.

st-pasha:
Another thing I noticed when running this example (i.e. `examples/lib/main.dart`), is that it makes the Dashbook disappear, i.e. no other examples will be accessible anymore. I'm not sure how it happens, but this is definitely something that we'd want fixed.

spydon:
> Another thing I noticed when running this example (i.e. `examples/lib/main.dart`), is that it makes the Dashbook disappear, i.e. no other examples will be accessible anymore. I'm not sure how it happens, but this is definitely something that we'd want fixed.

Ah, it doesn't disappear, but the hamburger menu is white so you can't see it.
Will look into that, and there is a problem when reloading the game too that I'll fix.

st-pasha:
Not sure how "clean" we want this implementation to be, but there are still some defects that we could fix:
 - The dino starts running immediately when the game loads. A better solution would be to wait for the first user interaction - tap or pressing a space bar.
 - The grey rectangle under the menu button isn't always positioned properly. For example, when the dashbook side menu opens, the button disappears but the rectangle doesn't. Likewise, if I select a device frame, the rectangle also becomes misplaced.
   - A possible solution here is to just always keep the side menu open.
 - When the side menu is open, the game draws on top of it.
 - Dino is not able to duck, and there are no pterodactyls.
 - The original game (chrome://dino) awards points for distance traveled, not for obstacles jumped.
 - The font for score is out-of-style with the game. The spritesheet actually contains images for digits needed to display the score, but looks like we're not using those.
 - The high score is not tracked.

ALso, the TRex is now the first example shown on the examples page. Previously all examples were organized alphabetically with "Animations" coming first -- it was probably not a very good organizational principle, but I just wanted to check that putting TRex first is a deliberate choice, not an accident.

spydon:
> Not sure how "clean" we want this implementation to be, but there are still some defects that we could fix:

>     * Dino is not able to duck, and there are no pterodactyls.
>     * The original game (chrome://dino) awards points for distance traveled, not for obstacles jumped.
>     * The high score is not tracked.

I just updated and moved the T-Rex example that we had over here:
https://github.com/bluefireteam/trex-flame
If anyone wants to make it more complete they are of course free to.

>     * The grey rectangle under the menu button isn't always positioned properly. For example, when the dashbook side menu opens, the button disappears but the rectangle doesn't. Likewise, if I select a device frame, the rectangle also becomes misplaced.

Good catch, missed that one.

>       * A possible solution here is to just always keep the side menu open.

I think that has been discussed previously and if I remember correctly we didn't want to do that.
Maybe it is time to reconsider having it open all the time if there is enough space, but on cellphones for example you wouldn't want it open all the time. But I don't think the problem exist on cellphones, because if I remember correctly the example isn't run when the menu is open there.

I don't think that it is currently possible to know whether the menu is open from within the game, so I'm not sure how to solve this.
Any ideas?

>     * When the side menu is open, the game draws on top of it.

Fixed.

>     * The font for score is out-of-style with the game. The spritesheet actually contains images for digits needed to display the score, but looks like we're not using those.

Changed the font to another one.

> ALso, the TRex is now the first example shown on the examples page. Previously all examples were organized alphabetically with "Animations" coming first -- it was probably not a very good organizational principle, but I just wanted to check that putting TRex first is a deliberate choice, not an accident.

It was deliberate, there will be more small games coming in here and I figured that it's a good thing to showcase these before the features to draw people's attention.

wolfenrain (approved):

PR #1605, opened 2022-05-05
feat: World bounds for a CameraComponent

### LINKED ISSUE #812: Сircle camera movement boundaries
# What could be improved
Add a circle bounding area to the camera

# Why should this be improved
The circular map is popular in 2D top view games.

# More information
or to do so to simplify this work

_reference_
` If set, this determines boundaries for the camera movement.  The camera will never move such that a region outside the world boundaries is shown, meaning it will stop following when the object gets close to the  edges. Changing this value can immediately snap the camera if it is a wrong position, but other than that it's just prevent movement so should not add any non-smooth movement.`
`Rect? worldBounds;`

st-pasha:
# Description

- Add `BoundedPositionBehavior` which could restrict the movement of another component.
- Use that behavior to implement camera's "world bounds". The shape of this world bound can be arbitrary.

## Checklist

## Breaking Change

## Related Issues

Closes #812 


packages/flame/lib/src/experimental/camera_component.dart
```diff
@@ -197,4 +199,24 @@ class CameraComponent extends Component {
...
   }
+
+  /// Sets or clears the world bound for the camera's viewfinder.
+  ///
+  /// The bound is a [Shape], given in the world coordinates. The viewfinder's
+  /// position will be restricted to always remain inside this region. Note that
+  /// if you want the camera to never see the empty space outside of the world's
+  /// rendering area, then you should set up the bound to be smaller than the
+  /// size of the world.
+  void setBound(Shape? bound) {
```
spydon, line 210: Shouldn't it be "bounds"?

spydon (approved):

packages/flame/lib/src/experimental/camera_component.dart
st-pasha: renamed

packages/flame/lib/src/experimental/bounded_position_behavior.dart
```diff
@@ -0,0 +1,92 @@
...
+  /// Saved position from the last game tick.
+  final Vector2 _previousPosition;
+
+  @override
+  void onMount() {
+    if (_target == null) {
+      assert(
+        parent is PositionProvider,
+        'Can only apply this behavior to a PositionProvider',
+      );
```
alestiago, line 61: Should we use the `ParentIsA` mixin?

luanpotter (approved):

packages/flame/lib/src/experimental/bounded_position_behavior.dart
st-pasha: The requirement that `parent` must be a `PositionProvider` only applies in case when `_target == null`. If, on the other hand, target was specified explicitly in the constructor, then the parent can be any.

PR #1610, opened 2022-05-07
fix: Fix tile flips when using canvas.drawAtlas

nathanaelneveux:
# Description

This PR makes changes to SpriteBatch to deal with flipped tiles. Specifically since RSTransform and therefore drawAtlas isn't able to handle flips this PR feeds an additional flipped version of the atlas to the SpriteBatch. Since this could create an additional image per flip type (H + V), flips have been simplified to reduce that to just a horizontally flipped image. All vertical flips (+rotations) can be represented as horizontal flips +180º.

The SimpleFlips class has been updated to represent this even simpler representation of angles and flips. **This is the only breaking change in this PR.**

## Checklist

## Breaking Change

- [x] Yes, this contains a breaking change.

## Related Issues

Fixes issue listed at top of package README


packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -178,13 +179,21 @@ class RenderableTiledMap {
           final src = ts.computeDrawRect(t).toRect();
           final flips = SimpleFlips.fromFlips(tile.flips);
           final size = destTileSize;
+          final scale = size.x / src.width;
+          final anchor = Vector2(src.width / 2, src.height / 2);
+          final offset = Vector2((tx + .5) * size.x, (ty + .5) * size.y)
```
spydon, line 184: Are there cases when `_renderLayer` will run each tick?
If that is the case you'd want to have two private `Vector2` objects outside of this method that you can re-use with `setFrom` or `setValues`, so that you don't have to create new `Vector2` objects here each tick.

packages/flame_tiled/lib/src/simple_flips.dart
```diff
@@ -4,65 +4,82 @@ import 'package:tiled/tiled.dart';
 /// Tiled represents all flips and rotation using three possible flips:
 /// horizontal, vertical and diagonal.
 /// This class converts that representation to a simpler one, that uses one
-/// angle (with pi/2 steps) and two flips (H or V).
+/// angle (with pi/2 steps) and one flip (Horizontal). All Vertical flips are
+/// represented as Horizontal flips + 180º.
```
spydon, line 8:
```suggestion
/// angle (with pi/2 steps) and one flip (horizontal). All vertical flips are
/// represented as horizontal flips + 180º.
```

spydon (approved):
Good job! Have you seen that there is a merge conflict?

packages/flame_tiled/lib/src/renderable_tile_map.dart
nathanaelneveux: `_renderLayer` shouldn't be called per tick unless there are some excessive calls to `refreshCache`. Perhaps a scenario with lots of layer visibility changes? But either way a large map would produce a lot of `Vector2` object during load. There isn't really a need for me to store these calculations in a `Vector2` instead of just a pair of doubles so I think I'll just switch to that.

luanpotter (approved):
Nice!

PR #1611, opened 2022-05-08
feat!: Added anchor for the Viewport

st-pasha:
# Description

- The `Viewport` now has its own anchor;
- The coordinate system of the `Viewport` now always has the origin in the top-left corner of its bounding box (previously it was in the center) -- similar to the local coordinate system of a `PositionComponent`.

## Checklist

## Breaking Change

## Related Issues


spydon (approved):
Very nice, I think `topLeft` makes more sense to be the default too (noticed when I created this DartPad game).

packages/flame/lib/src/experimental/fixed_size_viewport.dart
```diff
@@ -26,13 +26,12 @@ class FixedSizeViewport extends Viewport {
 
   @override
   bool containsLocalPoint(Vector2 point) {
-    return point.x.abs() <= size.x / 2 && point.y.abs() <= size.y / 2;
+    final x = point.x, y = point.y;
+    return x >= 0 && x <= size.x && y >= 0 && y <= size.y;
```
luanpotter, line 30: can we just use https://github.com/flame-engine/flame/blob/main/packages/flame/lib/src/extensions/rect.dart#L28

packages/flame/lib/src/experimental/camera_component.dart
```diff
@@ -108,7 +111,10 @@ class CameraComponent extends Component {
 
   @override
   Iterable<ComponentPointPair> componentsAtPoint(Vector2 point) sync* {
-    final viewportPoint = point - viewport.position;
+    final viewportPoint = Vector2(
+      point.x - viewport.position.x + viewport.anchor.x * viewport.size.x,
+      point.y - viewport.position.y + viewport.anchor.y * viewport.size.y,
+    );
```
luanpotter, line 117: is this the same as: `point - anchor.translate(viewport.position, viewport.size)` ?

packages/flame/lib/src/experimental/camera_component.dart
```diff
@@ -86,7 +86,10 @@ class CameraComponent extends Component {
   @override
   void renderTree(Canvas canvas) {
     canvas.save();
-    canvas.translate(viewport.position.x, viewport.position.y);
+    canvas.translate(
+      viewport.position.x - viewport.anchor.x * viewport.size.x,
+      viewport.position.y - viewport.anchor.y * viewport.size.y,
+    );
```
luanpotter, line 92: is this the same as `canvas.translateVector(anchor.translate(viewport.position, viewport.size)`?

we could even have a function to combine an anchor and a viewport that calls `anchor.translate`

packages/flame/lib/src/experimental/fixed_size_viewport.dart
st-pasha: That would require creating a temporary `Rect` object, and then converting a point into an `Offset`. Plus, the Rect's right and bottom edges are exclusive, so that would be a slightly different behavior.

packages/flame/lib/src/experimental/camera_component.dart
st-pasha: Could be, but I don't really understand what is `anchor.translate()`, and when it can be used.

wolfenrain (approved):

erickzanardo (approved):

PR #1613, opened 2022-05-08
refactor: Component's lifecycle futures moved into LifecycleManager

### LINKED ISSUE #1426: Group logic related to lifecycle of components on the _LifecycleManager
# What could be improved

Right now, some life cycle related logic lives directly on the Component class, like its state, if it is mounted, its completers and so on.

It would make better maintainable code if this logic would be center on the `_LifecycleManager` and the component would just expose those methods through getters.

# Why should this be improved
Easier to maintain code.

# Any risks?
None come to mind, this is all internal stuff, and we have a good amount of tests ensuring the correct behaviour so a refactor should be safe.

# More information

st-pasha:
# Description

Small refactor (which was proposed a long time ago) that moves `Component`'s properties `._loadCompleter` and `._mountCompleter` into the `_LifecycleManager` class without changing the user-facing API. The purpose of the move is to de-clutter the `Component`'s namespace.

## Checklist

## Breaking Change

## Related Issues

Closes #1426


spydon (approved):

erickzanardo (changes requested):
Tbh, I don't see much value on this change, we just moved the attributes to the `LifecycleManager` but the logic is still on `Component`.

I would rather keep them together unless we can figure it out a way to keep both things together.

luanpotter (approved):

st-pasha:
The idea here is that both `loadCompleter` and `mountCompleter` are futures that are (1) rarely needed by the user, (2) even when used, are only valid for a very brief period of time - few milliseconds at the beginning of the component's lifecycle. By moving these into the `LifecycleManager` class, we ensure that:
 - the size of the `Component` class is now smaller;
 - debugging is now easier because there are fewer properties in the `Component` class in a debug window.

Given that `Component` is the base class for ALL other components, I believe it's best to keep the class as lean as possible.

erickzanardo:
> the size of the Component class is now smaller;

That isn't true, they completers were already optional, so the footprint don't change, unless you are mentioned size as in line numbers, but if so, that difference is minimal and not worth having the attribute on a place and the logic in another.

> debugging is now easier because there are fewer properties in the Component class in a debug window.

Same as I mentioned above, IMO that difference is minimal and not worth having the attribute on a place and the logic in another.

st-pasha:
> That isn't true, they completers were already optional, so the footprint don't change

Even if a variable is `null`, it still takes up some space inside the class, since that `null` pointer needs to be stored somewhere.

> not worth having the attribute on a place and the logic in another.

The `LifecycleManager` is a `Component`'s appendix; it's specifically designed to hold optional/temporary parts of the `Component` that can be discarded once there is no need for them anymore. This is exactly what `loadCompleter` and `mountCompleter` are. For these fields, the `LifecycleManager` is a more natural place than the main `Component`.

erickzanardo:
> For these fields, the LifecycleManager is a more natural place than the main Component.

This is agree! But for me, none of these are worth the drawback of having the fields on a class, while they are managed on a different place.

If we can find a way to move the logic of `loaded` and `mounted` to the `LifecycleManager`, while keep the API without a breaking change, I am onboard. Otherwise I prefer to keep the field and the logic together.

st-pasha:
> If we can find a way to move the logic of loaded and mounted to the LifecycleManager, while keep the API without a breaking change, I am onboard. Otherwise I prefer to keep the field and the logic together.

Hmm, I see your point. I'll try to think of something that can be done along these lines.

renancaraujo (approved):

PR #1518, opened 2022-04-07
feat: Added componentsAtPoint() iterable

st-pasha:
# Description

- Added method `Component.componentsAtPoint()` which returns an iterable of descendant components intersecting with the given point, together with the points of intersection. This method can be used, for example, for event handling.
- Method `Component.containsLocalPoint()`, similar to `Component.containsPoint` but takes a point in local coordinate space.

## Checklist

- [x] The title of my PR starts with a [Conventional Commit] prefix (`fix:`, `feat:`, `docs:` etc).
- [x] I have read the [Contributor Guide] and followed the process outlined for submitting PRs.
- [x] I have updated/added tests for ALL new/updated/fixed functionality.
- [x] I have updated/added relevant documentation in `docs` and added dartdoc comments with `///`.
- [x] I have updated/added relevant examples in `examples`.

## Breaking Change

- [-] Yes, this is a breaking change.
- [x] No, this is *not* a breaking change.

## Related Issues

WIP for #1326
WIP for #1511 

[issue database]: https://github.com/flame-engine/flame/issues
[Contributor Guide]: https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md
[Flame Style Guide]: https://github.com/flame-engine/flame/blob/main/STYLEGUIDE.md
[Conventional Commit]: https://conventionalcommits.org

PR #1615, opened 2022-05-10
feat: method `componentsAtPoint` now reports the "stacktrace" of points

### LINKED ISSUE #1592: Improve API of componentsAtPoint()
# What could be improved

It's been 4 days since `componentsAtPoint` was introduced, but I think there's a way to make it better. Well, [we already talked](https://github.com/flame-engine/flame/pull/1518#discussion_r844946743) about the API not being ideal, but couldn't come up with a better solution.

So, here's a proposed solution:
```dart
Iterable<Component> componentsAtPoint(Vector2 point, [NestedLocation? nestedLocation]);
```
Here the `nestedLocation` argument, if given, will be used to store the local coordinates of `point` for the component being returned. In fact, it will also store the entire stack of points for every ancestor of the component being returned. For example, if you want to know not only the local coordinates of the point, but also the corresponding coordinates in the parent coordinate space, or in the world coordinate space, then that information will be available.

# Any risks?

Technically this is a breaking behavior, but since the method was just introduced, and there hasn't been a release yet, the risk to users is very low.

# More information

Are you interested in working on a PR for this? Yes.

st-pasha:
# Description

The method `componentsAtPoint` now has an optional parameter `nestedPoints`, which will record the "stacktrace" of points up to and including the currently returned component. This allows for better flexibility in understanding the point's position relative to its parent and the world. This also allows the method `componentsAtPoint` to return an iterable of `Component`s instead of an iterable of `ComponentPointPair`s.

## Checklist

## Breaking Change

  - There is a change in the API of `componentsAtPoint()`. However, since the method itself was only introduced 2 weeks ago (#1518), and there hasn't been a Flame release since then, the change can be considered non-breaking.

## Related Issues

Closes #1592


packages/flame/lib/src/components/component.dart
```diff
@@ -642,9 +641,13 @@ class Component {
...
-  /// The return value is an [Iterable] of `(component, point)` pairs, which
-  /// gives not only the components themselves, but also the points of
-  /// intersection, in their respective local coordinates.
+  /// The return value is an [Iterable] of components. If the [nestedPoints]
+  /// parameter is given, then it will also report the points of intersection
+  /// for each component in its local coordinate space. Specifically, the last
+  /// element in the list is the point in the coordinate space of the returned
+  /// component, the element before the last is in that component's parent's
+  /// coordinate space, and so on. The [nestedPoints] list must be growable and
+  /// modifiable.
```
erickzanardo, line 650: > list must be growable and modifiable

Can't we clone the list to avoid this type of side effect? Or does this gives a huge impact on perfomance?

If so, we could at least clone at on the first call?

st-pasha: It is an output parameter, so the entire idea is that the list will be modified by the function so that the user would then be able to look at it.

wolfenrain (commented):
Love that this removes the `ComponentPointPair` class, but this means it is a breaking change right?

spydon (approved):
Very happy that the `ComponentPointPair`, good job!

wolfenrain (approved):

PR #1620, opened 2022-05-13
feat: Aligned text in the TextBoxComponent

### LINKED ISSUE #1088: `TextBoxComponent` should have the possiblity to align text within itself
# Problem to solve
Right now the text within a `TextBoxComponent` can only be left aligned.

# Proposal
I propose that we add a `textAlign` field in `TextBoxConfig`. The type of the field could either be `TextAlign` or an `Anchor`.

# More information

st-pasha:
# Description

- Added option `align` in the `TextBoxComponent` which controls the alignment of text.
- Added option for the `TextBoxComponent` to have a fixed size (before the only mode was for the textbox to automatically expand/shrink to fit the text).

## Checklist

## Breaking Change

## Related Issues

Closes #1088


doc/flame/rendering/text.md
```diff
@@ -93,25 +93,32 @@ class MyGame extends FlameGame {
...
+  MyTextBox(String text)
+    : super(text: text, textRenderer: tiny, boxConfig: TextBoxConfig(timePerChar: 0.05));
 
   @override
-  void drawBackground(Canvas c) {
+  void render(Canvas canvas) {
     Rect rect = Rect.fromLTWH(0, 0, width, height);
-    c.drawRect(rect, Paint()..color = Color(0xFFFF00FF));
-    c.drawRect(
+    canvas.drawRect(rect, Paint()..color = Color(0xFFFF00FF));
```
spydon, line 115: We shouldn't encourage our users to create `Paint` objects in the render loop (even though it was done before this PR).

examples/lib/stories/rendering/text_example.dart
```diff
@@ -72,21 +76,29 @@ final _shaded = TextPaint(
...
+            margins: EdgeInsets.all(margins ?? 25),
           ),
         );
 
   @override
-  void drawBackground(Canvas c) {
+  void render(Canvas canvas) {
     final rect = Rect.fromLTWH(0, 0, width, height);
-    c.drawRect(rect, Paint()..color = Colors.white10);
+    canvas.drawRect(rect, Paint()..color = Colors.white10);
```
spydon, line 101: Same here

packages/flame/lib/src/components/text_box_component.dart
```diff
@@ -66,26 +66,52 @@ class TextBoxComponent<T extends TextRenderer> extends TextComponent {
...
+  /// example, setting this property to `Align.center` will make the text
+  /// centered inside its box. Similarly, `Align.bottomRight` will render the
+  /// text that's aligned to the right and to the bottom of the box.
+  ///
+  /// Custom alignment anchors are supported too. For example, if this property
+  /// is set to `Anchor(0.1, 0)`, then the text would be positioned such that
+  /// its every line will have 10% of whitespace on the left, and 90% on the
+  /// right. You can use an `AnchorEffect` to make the text gradually transition
+  /// between different alignment values.
+  Anchor align;
```
spydon, line 106: Will the "text alignment" always be left aligned even though `center` is sent in here? I think that could be confusing, this feels like it is more like an `anchor` and and that `align` should be text alignment?

spydon: Aah, or maybe it aligns every row separately, that could work I guess!

spydon (approved):
Lgtm

doc/flame/rendering/text.md
st-pasha: fixed

st-pasha:
@spydon Please take a look at the failing test, I suspect it might be due to differences between Flutter versions on my machine and in the CI pipeline.

spydon:
> @spydon Please take a look at the failing test, I suspect it might be due to differences between Flutter versions on my machine and in the CI pipeline.

I would guess that it is due to the classic font rendering problem when using goldens:
https://pub.dev/packages/alchemist#about-platform-tests-vs-ci-tests

spydon:
> Golden tests involving texts are usually a pitfall because text rendering is influenced by the OS where the test is run, so if we generate a snapshot on linux and then try to run the test on MacOs or windows, there are high possibility that the test will fail.
> 
> So IMO we shouldn't use golden tests for these situations.

We can use them, but render boxes instead where the text would have been.
Alchemist for example has a mode for this.
(There are also a few fonts that are specialized for golden testing, but that's more work.)

erickzanardo:
> Alchemist for example has a mode for this.
Interesting, would be cool to test this.

spydon:
I tried Flutter 3.0.0 and 2.10.5 locally and they generated identical images (same hash), but they are different to the one generated by the CI. It seems like there is one pixel difference in the boxes, I just downloaded the image generated by the CI and replaced the other file with it and ran the test again to get the diffs.
text_box_component_test_1_isolatedDiff
text_box_component_test_1_maskedDiff
text_box_component_test_1_testImage
text_box_component_test_1_masterImage

st-pasha:
Thanks @spydon -- looks like some anti-aliasing artifacts on the edges of the lines.

A workaround that we can try here is to switch to the SpriteFontRenderer once #1634 is merged.

spydon:
Maybe we can try Alchemist for handling the goldens instead?

spydon:
I tried to use Alchemist for running these tests, but I get a problem with the delayed future:

```
Pending timers:
Timer (duration: 0:00:00.010000, periodic: false), created:
#0      new FakeTimer._ (package:fake_async/fake_async.dart:308:62)
#1      FakeAsync._createTimer (package:fake_async/fake_async.dart:252:27)
#2      FakeAsync.run.<anonymous closure> (package:fake_async/fake_async.dart:185:19)
#6      TextBoxComponent.redraw (package:flame/src/components/text_box_component.dart:290:14)
#7      TextBoxComponent.update (package:flame/src/components/text_box_component.dart:300:7)
#8      Component.updateTree (package:flame/src/components/component.dart:459:5)
#9      FlameGame.updateTree.<anonymous closure> (package:flame/src/game/flame_game.dart:80:31)
#10     IterableMixin.forEach (dart:collection/iterable.dart:45:35)
#11     FlameGame.updateTree (package:flame/src/game/flame_game.dart:80:14)
#12     FlameGame.update (package:flame/src/game/flame_game.dart:69:7)
#13     GameRenderBox.gameLoopCallback (package:flame/src/game/game_render_box.dart:53:10)
#14     GameLoop._tick (package:flame/src/game/game_loop.dart:46:13)
#15     Ticker._tick (package:flutter/src/scheduler/ticker.dart:238:12)
#16     SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1146:15)
#17     SchedulerBinding.handleBeginFrame.<anonymous closure> (package:flutter/src/scheduler/binding.dart:1059:11)
#18     _LinkedHashMapMixin.forEach (dart:collection-patch/compact_hash.dart:614:13)
#19     SchedulerBinding.handleBeginFrame (package:flutter/src/scheduler/binding.dart:1057:17)
#20     AutomatedTestWidgetsFlutterBinding.scheduleWarmUpFrame (package:flutter_test/src/binding.dart:1127:5)
#21     runApp (package:flutter/src/widgets/binding.dart:1046:7)
#22     TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:840:7)
<asynchronous suspension>
(elided 3 frames from dart:async)

══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following assertion was thrown running a test:
A Timer is still pending even after the widget tree was disposed.
'package:flutter_test/src/binding.dart':
Failed assertion: line 1291 pos 12: '!timersPending'

#2      AutomatedTestWidgetsFlutterBinding._verifyInvariants (package:flutter_test/src/binding.dart:1291:12)
#3      TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:848:7)
<asynchronous suspension>
(elided 2 frames from class _AssertionError)

The test description was:
  Alignment options (variant: Linux)
```

st-pasha:
I hope that after merging #1634 we can switch to sprite fonts here, and there wouldn't be a problem anymore.

spydon:
> I hope that after merging #1634 we can switch to sprite fonts here, and there wouldn't be a problem anymore.

Feels a bit like cheating though, since normal fonts aren't tested then (and this will most likely not be the last time we have this problem in the CI).

st-pasha:
I looked at Alchemist documentation, and they say they use Ahem font when rendering in CI, which replaces all glyphs with blocks. However, Flutter itself uses Ahem font for rendering goldens, yet somehow we see a 1-pixel shift. I'm not sure using Alchemist would solve this...

There's also the `golden_toolkit` package, and the approach they took is to bundle Roboto font together with the tests, and load it directly from there: https://github.com/eBay/flutter_glove_box/blob/9e881929543e1ed634f854897ded903d4bc2f55a/packages/golden_toolkit/lib/src/font_loader.dart I wonder if we can maybe replicate that somehow?

spydon:
I just noticed that all positions have .5 pixels, maybe this wouldn't cause anti-aliasing if the positions were in .0 instead?

spydon:
@st-pasha Let's put `skip: true` on the test and open an issue about fixing it, so that this feature can be included in v1.2.0

st-pasha:
So, it looks like the issue is caused by different ways that the fonts are smoothed on different operating systems. I made these zoomed-in screenshots, which show that both OSes perform smoothing, but in a different way:
Screen Shot 2022-06-03 at 10 54 24 AM
Screen Shot 2022-06-03 at 10 55 05 AM

I've tried to render the text using a paint with antialiasing turned off - but it didn't make any difference, looks like macOS simply doesn't support switching off antialiasing for fonts (except maybe as a global system option, but that would be too disruptive for day-to-day life)...

In view of this, I believe the best we can do is to not rely on system font rendering altogether and use sprite fonts instead (at least for golden files); or wait until such time as Flutter team figures how to make golden tests match if the difference between images is up to anti-aliasing effects.

spydon:
> So, it looks like the issue is caused by different ways that the fonts are smoothed on different operating systems. I made these zoomed-in screenshots, which show that both OSes perform smoothing, but in a different way

That is the classic Flutter issue, didn't we post that one here? Otherwise I know @erickzanardo has the link to it somewhere, because he has a workaround for it in that issue.
Alchemist should solve that issue.

spydon:
Here is the issue (Erick digged it up):
https://github.com/flutter/flutter/issues/56383
(apparently the suggested workaround doesn't work all the time, so we should avoid that)

st-pasha:
> because he has a workaround for it in that issue.

The suggested "workaround" - to change the font smoothing at system global level - would only work on a computer that is dedicated solely to running golden tests and to nothing else. Because in a day-to-day life font smoothing is a very much desirable feature, there is a reason companies invest money in improving it.

Of course, ideally, Apple would have an API for rendering a specific string without any antialiasing. But apparently they don't have this, and it's not clear that they ever will (even if they do, we don't want tests that run on some version of OS but not on other, and not everyone can upgrade their OS easily).

In view of this, there may be no other option than manual sprite-based fonts. If Alchemist has a work-around, then this is what they must be doing.

spydon:
> The suggested "workaround" - to change the font smoothing at system global level - would only work on a computer that is dedicated solely to running golden tests and to nothing else. Because in a day-to-day life font smoothing is a very much desirable feature, there is a reason companies invest money in improving it.

Yeah, sorry I was unclear, I didn't mean that we should use that work around, it was just how I remembered the issue. :)

> If Alchemist has a work-around, then this is what they must be doing.

Indeed, let's investigate that post v1.2.0.

erickzanardo:
> The suggested "workaround" - to change the font smoothing at system global level - would only work on a computer that is dedicated solely to running golden tests and to nothing else. Because in a day-to-day life font smoothing is a very much desirable feature, there is a reason companies invest money in improving it.

Yeah, that workaround was something that used to work, but since some time isn't accurate anymore. There is only one workaround that I think is to be trusted. Running golden tests on the same environment as the CI will run them. On a multi platform world, docker would be an alternative for that, but that would render a very bad developer experience IMO.

> In view of this, there may be no other option than manual sprite-based fonts. If Alchemist has a work-around, then this is what they must be doing.

I am a little bit skeptical about this still, giving the nature of the issue. But I would love to see this in action and would love even more to be proven wrong. Lets try to plan for doing this, even if as a POC for now.

packages/flame_test/lib/src/test_golden.dart
```diff
@@ -29,7 +29,11 @@ void testGolden(
   PrepareGameFunction testBody, {
   required String goldenFile,
   FlameGame? game,
+  bool skip = false,
 }) {
+  if (skip) {
+    return;
+  }
   testWidgets(testName, (tester) async {
```
spydon, line 37: You could just pass in skip here instead

st-pasha: ok

erickzanardo (approved):

PR #1625, opened 2022-05-14
refactor: Move to package imports

spydon:
# Description

Moves from relative to package imports.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/test/collisions/collision_callback_benchmark.dart
+++ b/packages/flame/test/collisions/collision_callback_benchmark.dart
@@ -5,7 +5,7 @@ import 'package:flame/components.dart';
 import 'package:flame_test/flame_test.dart';
 import 'package:test/test.dart';
 
-import 'collision_test_helpers.dart';
+import '/home/spydon/repos/flame/packages/flame/test/collisions/collision_test_helpers.dart';
 
 class _TestBlock extends PositionComponent with CollisionCallbacks {
   final Vector2 velocity;
--- a/packages/flame/test/collisions/collision_callback_benchmark.dart
+++ b/packages/flame/test/collisions/collision_callback_benchmark.dart
@@ -5,7 +5,7 @@ import 'package:flame/components.dart';
 import 'package:flame_test/flame_test.dart';
 import 'package:test/test.dart';
 
-import '/home/spydon/repos/flame/packages/flame/test/collisions/collision_test_helpers.dart';
+import 'collision_test_helpers.dart';
 
 class _TestBlock extends PositionComponent with CollisionCallbacks {
   final Vector2 velocity;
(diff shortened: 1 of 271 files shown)


st-pasha:
This seems like it would be a big breaking change for any project that uses flame_lint.

spydon:
> This seems like it would be a big breaking change for any project that uses flame_lint.

Is it a breaking change when it is only a warning? 🤔

erickzanardo:
> Is it a breaking change when it is only a warning? 🤔


I agree, this is a dev dependency too, so nothing to worry IMO

st-pasha:
> Is it a breaking change when it is only a warning?

When we put a `@Deprecated` annotation on a method or class, it creates warnings to whoever uses that method or class. And we don't consider deprecation to be a breaking change. So, it makes sense to say that warnings are not breaking change.

On the other hand, this change would cause hundreds if not thousands warnings at once, so one could argue that the sheer quantity of warnings causes this to become a "breaking change".

On an unrelated note, what is the reason for the switch? It seems to me that the vast majority of import statements are now longer. Also, the Dart itself seems to prefer relative imports (https://dart.dev/guides/language/effective-dart/usage#prefer-relative-import-paths).

packages/flame/test/collisions/collision_callback_benchmark.dart
```diff
@@ -5,7 +5,7 @@ import 'package:flame/components.dart';
 import 'package:flame_test/flame_test.dart';
 import 'package:test/test.dart';
 
-import 'collision_test_helpers.dart';
+import '/home/spydon/repos/flame/packages/flame/test/collisions/collision_test_helpers.dart';
```
st-pasha, line 8: This doesn't seem right.

st-pasha: (also same in a few files below this one)

spydon: Good catch, will fix them.

spydon:
> When we put a `@Deprecated` annotation on a method or class, it creates warnings to whoever uses that method or class. And we don't consider deprecation to be a breaking change. So, it makes sense to say that warnings are not breaking change.
> On the other hand, this change would cause hundreds if not thousands warnings at once, so one could argue that the sheer quantity of warnings causes this to become a "breaking change".

We will bump `flame_lint` just the same as it would have been a breaking change, but I don't want to put the PR as breaking since it touches all packages.

> On an unrelated note, what is the reason for the switch? It seems to me that the vast majority of import statements are now longer. Also, the Dart itself seems to prefer relative imports (https://dart.dev/guides/language/effective-dart/usage#prefer-relative-import-paths).

Among Flutter packages and applications it seems to be more standard to use only package imports only since it is easier to read when you have barrel files (no need for `../../../../..`.

erickzanardo (approved):

wolfenrain (approved):

PR #1626, opened 2022-05-14
feat!: remove `onTimingsCallback` for Flutter 3.0

wolfenrain:
# Description

This technically speaking is a breaking change, but since we are already planning on deprecating the `FPSCounter` the team has internally decided to turn the `fps()` method into a copy of the functionality from #1595 so that we can remove the method `onTimingsCallback` already so Flutter 3.0 users wont have any warnings. This method is either not used or a really small portion of people are using it, compared to the `FPSCounter`. 

## Checklist

## Breaking Change

- [x] Yes, this is a breaking change, ~~but as we are doing this on purpose I have not added the suffix `!`~~.

## Related Issues

### DIFF
--- a/packages/flame/lib/src/game/game_render_box.dart
+++ b/packages/flame/lib/src/game/game_render_box.dart
@@ -10,10 +10,7 @@ class GameRenderBox extends RenderBox with WidgetsBindingObserver {
   Game game;
   GameLoop? gameLoop;
 
-  GameRenderBox(this.buildContext, this.game) {
-    //ignore: deprecated_member_use_from_same_package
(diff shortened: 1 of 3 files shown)


spydon (approved):

st-pasha:
:+1: to this change - even with the old FpsCounter it was unclear why the method was exposed in the Game's API, since it was possible to subscribe to the timing events from the mixing itself.

> this is a breaking change, but as we are doing this on purpose I have not added the suffix !.

That doesn't sound right though. All breaking changes that we do are deliberate, and are marked with `!`. Of course, there could be some breaking changes that we introduced without noticing, by accident - those will have no `!`.

erickzanardo (approved):

PR #1634, opened 2022-05-17
feat: Added ability to render spritesheet-based fonts

### LINKED ISSUE #1621: Add support for sprite-based text rendering
# Problem to solve

Currently we support rendering text via Flutter's `TextPainter` mechanism, which offers a variety of font and style choices.

However, sometimes a game may want to use a different approach and render text using custom images for every character. This could be either for stylistic purposes, or to reduce the bundle size, or to create special alphabets, or for any other purposes.

One concrete example is the TRex game, where the spritesheet contains characters for "0123456789HI", but displaying the score using those characters is complicated without a dedicated spritesheet-based renderer.

Note that we have anticipated the need for custom text renderers since long time ago, so this is a feature that we always wanted to implement.

## More information

Are you interested in working on a PR for this? 
- [x] yes

st-pasha:
# Description

This PR adds [SpriteFontRenderer], which is a new implementation of [TextRenderer], allowing to render text based on fonts embedded into a spritesheet.

See the T-Rex game example, where the score is now rendered based on a spritesheet font.

(Also added highscore tracking for the TRex game).

## Checklist

## Breaking Change

## Related Issues

Closes #1621


packages/flame/lib/src/text/sprite_font_renderer.dart
```diff
@@ -0,0 +1,132 @@
...
+  });
+
+  final Image source;
+  final double charWidth;
+  final double charHeight;
+  final double scale;
+  double letterSpacing;
+  bool get isMonospace => true;
+
+  Map<int, _GlyphInfo> glyphs = {};
```
erickzanardo, line 41: Can we make this private since it is internal and the value is an internal class?

erickzanardo (commented):
Nice one! I have one suggestion though.

`TextPaint` is an immutable class IIRC, and for consistency I think we should that on this renderer too, so instead of a `addGlyph` method, we could have a new attribute on the constructor to receive the metadata of mapped glyphs. Other than that, this looks great!

wdyt?

spydon (approved):
Good job, glad to see that the T-Rex game is already getting improvements too!
I agree with Erick's suggestions too.

wolfenrain (approved):

st-pasha:
> `TextPaint` is an immutable class IIRC, and for consistency I think we should that on this renderer too, so instead of a `addGlyph` method, we could have a new attribute on the constructor to receive the metadata of mapped glyphs. Other than that, this looks great!
> 
> wdyt?

We could make something like `SpriteFontBuilder` which would have the `addGlyph()` method, and then `.build()` will return an instance of `SpriteFontRenderer` (which would not be constructible directly). 

Need to keep an eye on future ability of create new `SpriteFontRenderer`s out of an existing `SpriteFontRenderer`. For example, create a different-sized variant of the same font.

packages/flame/lib/src/text/sprite_font_renderer.dart
st-pasha: done

erickzanardo:
> Need to keep an eye on future ability of create new SpriteFontRenderers out of an existing SpriteFontRenderer. For example, create a different-sized variant of the same font.

I don't think that making this immutable would disallow that, we can take TextStyle from flutter itself as an example, where it provides a copyWith method, that same approach would work great on this too IMO

packages/flame/lib/src/text/sprite_font_renderer.dart
```diff
@@ -0,0 +1,163 @@
...
+    required Map<int, _GlyphInfo> glyphs,
+    required this.letterSpacing,
+    required this.scale,
+  }) : _glyphs = glyphs;
+
+  final Image source;
+  final double charWidth;
+  final double charHeight;
+  final double scale;
+  double letterSpacing;
```
spydon, line 39: Shouldn't this be final too?

st-pasha: done

packages/flame/lib/src/text/sprite_font_renderer.dart
```diff
@@ -0,0 +1,163 @@
...
+    final glyph = _glyphs[i];
+    if (glyph == null) {
+      throw ArgumentError('No glyph for character "${String.fromCharCode(i)}"');
+    }
+    return glyph;
+  }
+}
+
+/// Helper class for creating [SpriteFontRenderer] instances.
+class SpriteFontBuilder {
```
erickzanardo, line 100: I really think we should avoid the builder pattern, it is something not common on Flutter/Dart and on Flame we don't use that as well.

A solution more in line with, for example, how we made the `SpriteAnimation` constructor would be preferable. Where we have a metadata class that describe the animation and internally it is used to load the frames. Here we could have the same thing, a class for a glyph metadata and `SpriteFontRenderer` will receive a map of it.

That we keep the API more declarative and more in line with the APIs we current have.

st-pasha: Alright, I changed it into `SpriteFontData`.

st-pasha: Alright, I changed it into `SpriteFontData`.

packages/flame/lib/src/text/sprite_font_renderer.dart
```diff
@@ -0,0 +1,144 @@
...
+    final glyph = _data._glyphs[i];
+    if (glyph == null) {
+      throw ArgumentError('No glyph for character "${String.fromCharCode(i)}"');
+    }
+    return glyph;
+  }
+}
+
+/// Helper class for creating [SpriteFontRenderer] instances.
+class SpriteFontData {
```
erickzanardo, line 94: I think this isn't quite the way I suggested, sorry if I wasn't clear on my previous comments, I will add a snippet of how I expect this API to look like:

```dart
final renderer = SpriteFontRenderer({
  SpriteFontData(
    source: image,
    charWidth: 6,
    charHeight: 8,
    glyphs: {
     'a': GlyphData(top: 0, left: 0),
     'b': GlyphData(top: 0, left: 6),
     ...
    },
});
```

That way this will be truly immutable and we don't need to imperatively call an `addGlyph` method.

st-pasha: Alright, implemented this approach

erickzanardo (approved):
LGTM, can't wait to use this.

PR #1637, opened 2022-05-18
fix: performance improvements on `SpriteBatch` APIs

### LINKED ISSUE #1614: Performance is dramatically decreased After #1610
# Current bug behaviour
I was very looked forward to #1605 and I upgraded flame to this. 
But I found the performance Is Very slow after upgrade. (see videos at the end).
Android can check the performance at least but IOS cannot even see the app launches because it shutting down immediately after launched due to memory issue.
So I bisected what commit went wrong.
After some reverts, I found #1610  is the problem.

It is very normal at #1596 (commit hash 60df3b9f60f538fbad7a3d806f5d38262ab6d66c)
But after #1610 (b4ad498fe5488795deb2c2e098fcde357b448bf), It has dramatically slown performance. 

I carefully assume that it has some memory leak and GCs constantly.

# Expected behaviour
Fix the performance issue.

# Flutter doctor output

```

[✓] Flutter (Channel stable, 2.10.5, on macOS 12.1 21C52 darwin-arm, locale ko-KR)
    • Flutter version 2.10.5 at /Users/lafity101/flutter
    • Framework revision 5464c5bac7 (3 weeks ago), 2022-04-18 09:55:37 -0700
    • Engine revision 57d3bac3dd
    • Dart version 2.16.2
    • DevTools version 2.9.2

[✓] Xcode - develop for iOS and macOS (Xcode 13.3.1)
    • CocoaPods version 1.11.2

[✓] VS Code (version 1.67.0)
    • Flutter extension version 3.40.0

    • SM G977N (mobile)           •                • android-arm64  • Android 11 (API 30)
    • iPhone 12mini (mobile) •  • ios            • iOS 14.8 18H17

[✓] HTTP Host Availability
```

- Explained at the beginning

# More information

### Before #1610 
https://user-images.githubusercontent.com/38072762/167560363-d8694ad2-ba53-4834-a860-847f6d0faf74.mp4

### After #1610 - I didn't manipulated video speed
https://user-images.githubusercontent.com/38072762/167560356-07333277-31c1-4165-ad43-e73f4635fc09.mp4

nathanaelneveux:
# Description

Fix for the performance decrease mentioned in #1614 

Some items that need work:
- [ ] I feel like I can be more efficient with object creation here and do the entire operation on the first Uint8List
- [x] After finding decodeImageFromPixels() I believe I can do the whole operation without adding the Bitmap package
- [x] I need to do more testing of the caching of the image. What I've written so far seems straight forward but I'm not totally convinced subsequent calls to retrieve the image from the cache are working. As I was stepping through the debugger testing something else I kept seeing the same image being generated, but I haven't done a lot of testing here yet.

## Checklist

## Breaking Change

## Related Issues

Fixes #1614 


packages/flame/lib/src/sprite_batch.dart
```diff
@@ -180,18 +183,36 @@ class SpriteBatch {
     );
   }
 
+  static const String _generatedAtlasKeySuffix = '#withFlipAttached';
+
   static Future<Image> _generateAtlas(Images? images, String path) async {
```
spydon, line 188: Have you made sure that this one isn't called more than necessary? I don't really understand why the `PictureRectorder` weren't able to handle this usecase... In which cases does it need to be re-generated?

st-pasha (commented):
I don't think there is anything wrong with the current approach that uses `PictureRecorder`. (If there is, we need to have a clear confirmation of that, then we need to file a bug to Flutter team, and use the suggested workaround until that bug is resolved, but also put it into a separate function so that it can be reused through multiple other places where we use PictureRecorder in Flame).

The more probable cause of #1614 is that `SpriteBatch.load()` method attempts to generate atlas on every call - even if the sprite batch won't be using the flipped images later on. And that image generation is not cached. So, if the user has code that does `SpriteBatch.load()` whenever a component is created, and if many such components are created per second, then there will be lots of unnecessary image generation happening. This explanation fits well with the problem described in #1614 as "IOS cannot even see the app launches because it shutting down immediately after launched due to memory issue."

So, I suggest to do the following:
1. Restore the original image generation through PictureRecorder;
2. Make invocation of `_generateAtlas` happen only if the user explicitly requests support for flipped sprites in their SpriteBatch;
3. Cache that generated image;
4. (Optionally) Measure the speed difference between generating an image via PictureRecorder vs via the Bitmap library.

packages/flame/pubspec.yaml
```diff
@@ -14,6 +14,7 @@ dependencies:
   collection: ^1.15.0
   ordered_set: ^5.0.0
   vector_math: ^2.1.1
+  bitmap: ^0.1.2
```
spydon, line 17: I don't think we'd want to pull in another dependency for this.

packages/flame/lib/src/sprite_batch.dart
nathanaelneveux: How I replicated the performance issues mentioned in #1614 was to load a test .tmx file I had in a for loop x30 because I assumed my test map was just not complicated enough to see the issue.
```
for (int i = 0; i < 30; i++) {
      TiledComponent anotherMap =
          await TiledComponent.load('map.tmx', Vector2.all(64));
      add(anotherMap);
```
Running the above code against [`60df3b`](https://github.com/flame-engine/flame/commit/60df3b9f60f538fbad7a3d806f5d38262ab6d66c) the game runs normally.

I set a breakpoint within `_generateAtlas` to test whether this function was being called more than necessary (or worse per game tick) based on st-pasha's [comment](https://github.com/flame-engine/flame/issues/1614#issuecomment-1123076693). Once the spriteBatch has been loaded images are not recreated - `_generateAtlas` is not called again after the initial load, but performance in the game is terrible.

After replacing the `PictureRecorder` code performance was the same as [`60df3b`](https://github.com/flame-engine/flame/commit/60df3b9f60f538fbad7a3d806f5d38262ab6d66c).

I can't actually confirm that `PictureRecorder` is the culprit here since I replaced the whole code block with something new. Something in the current approach is causing a performance problem however.

packages/flame/pubspec.yaml
nathanaelneveux: I was already working on removing that as a dependency so that is reflected in the most recent commit.

I am having trouble with the caching however. It seems that after running `_images.add('$path$_generatedAtlasKeySuffix', atlas)` the `atlas` image is disposed. I've also tried setting `atlas` to the result of `_images.fromCache` after adding it to the cache but this still results in `atlas` being disposed.

nathanaelneveux: I figured out the caching - the problem was because all the layers and images of the `SpriteBatch` are being loaded asynchronously (with `Future.wait`). The image generation ends up being called multiple times before it's ever added to the cache the first time.

I added a variation of `add` and `fromCache` to `Images` that returns `Future<Image>` and with that I am able to get caching working.

nathanaelneveux:
I think we can do better than a flag for the generation of the mirrored images. For one thing `SpriteBatch.load` is buried pretty deep within `RenderableTiledMap` with the main user facing function being `.fromFile()`.

How do you feel about doing it as a "lazy load" where once we get a call to `addTransform` with `flip = true` we call `_generateAtlas`?

spydon:
> How do you feel about doing it as a "lazy load" where once we get a call to `addTransform` with `flip = true` we call `_generateAtlas`?

Will that not potentially impact performance mid-game instead of in `onLoad`?

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -170,29 +171,66 @@ class SpriteBatch {
     RSTransform? defaultTransform,
     Images? images,
     bool useAtlas = true,
+    bool needsFlips = true,
```
st-pasha, line 174: This should be false by default, because we can detect when the user needs to use this option but forgot to set it, but we cannot detect the opposite situation.

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -170,29 +171,66 @@ class SpriteBatch {
...
     canvas.drawImage(image, Offset.zero, _emptyPaint);
     canvas.scale(-1, 1);
-    canvas.drawImage(image, Offset(-image.width * 2, 0), _emptyPaint);
+    canvas.drawImage(image,
+        Offset(-image.width.toDouble(), image.height.toDouble()), _emptyPaint);
 
     final picture = recorder.endRecording();
-    final atlas = picture.toImage(image.width * 2, image.height);
-    return atlas;
+    return picture.toImage(image.width, image.height * 2);
```
st-pasha, line 211:
```suggestion
    return picture.toImageSafe(image.width, image.height * 2);
```

packages/flame/lib/src/sprite_batch.dart
nathanaelneveux: I only switched it to true for testing since it won't currently pass tests with it set to false.

nathanaelneveux:
> > How do you feel about doing it as a "lazy load" where once we get a call to `addTransform` with `flip = true` we call `_generateAtlas`?
> 
> Will that not potentially impact performance mid-game instead of in `onLoad`?

`addTransform` is part of the loading process as well and should happen pretty early on in the lifecycle of this component. You have to add all your transforms before you would do your first render. See how its done in Flame Tiled for [example](https://github.com/flame-engine/flame/blob/843ddc36249272fcb518b44672e1012307dfa1b5/packages/flame_tiled/lib/src/renderable_tile_map.dart#L188).

spydon:
> `addTransform` is part of the loading process as well and should happen pretty early on in the lifecycle of this component. You have to add all your transforms before you would do your first render. See how its done in Flame Tiled for [example](https://github.com/flame-engine/flame/blob/843ddc36249272fcb518b44672e1012307dfa1b5/packages/flame_tiled/lib/src/renderable_tile_map.dart#L188).

Alright, sounds like it makes sense to have it lazily initialized then!

packages/flame/lib/src/sprite_batch.dart
st-pasha: And if we switch to lazy loading, then this property will disappear altogether, right?

nathanaelneveux: That would be the idea - that way it's only loaded as needed but also doesn't require any user action.

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -171,29 +172,66 @@ class SpriteBatch {
...
+  static const String _generatedAtlasKeySuffix = '#withFlipAttached';
+
+  static Future<Image> _atlasFromCache(Images images, String path) async {
+    if (!images.containsKey('$path$_generatedAtlasKeySuffix')) {
+      return images.addFuture(
+        '$path$_generatedAtlasKeySuffix',
+        _generateAtlas1(images, path),
+      );
+    }
+    return images.fromCacheFuture('$path$_generatedAtlasKeySuffix');
```
st-pasha, line 198: Perhaps we should consider combining these into a single method, something like `Images.fetchOrGenerate(key, builder)`

nathanaelneveux: That is nice! - I really wanted something similar to how `.load` works but for images and this sounds like it could be that.

spydon: That sounds good

spydon:
@nathanaelneveux any progress on this one? Anything that you need help with? :)

nathanaelneveux:
So obviously you’ll see in the most recent commit that it’s not quite there yet. It’s still generating the `flippableAtlas` right away. 

I was hoping `late` would cause it to wait until it was used to begin generating which right now would be at `render` but it isn’t waiting. Once I can get `late` working then the next step would be to move the late generation up into the `addTranform` so that it’s still being generated early on in the lifecycle when needed.

I need a reference to `Images` and a reproducible key - which right now I’m using `path` with a suffix. I’m trying to avoid adding those to the constructor because I think if your using the constructor it doesn’t make sense to do lazy initialization? Or maybe I’m over thinking that?

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -172,28 +180,72 @@ class SpriteBatch {
...
+    final _images = images ?? Flame.images;
+    //flippableAtlas = await _atlasFromCache(_images, path);
     return SpriteBatch(
-      await _generateAtlas(images, path),
+      await _images.load(path),
       defaultColor: defaultColor,
       defaultTransform: defaultTransform ?? RSTransform(1, 0, 0, 0),
       defaultBlendMode: defaultBlendMode,
       useAtlas: useAtlas,
+      flippableAtlas: await _images.fetchOrGenerate(
```
spydon, line 191: If you move this to where the field is declared it should be lazily initialized, now it'll be initialized directly.

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -224,13 +276,15 @@ class SpriteBatch {
       color: color ?? defaultColor,
     );
 
+    if (flip && useAtlas) _needsFlips = true;
```
st-pasha, line 279: If you replace `_needsFlips` variable with `_hasFlips`, then this could simply be
```
if (flip && useAtlas && !_hasFlips) {
  _hasFlips = true;
  _generateFlippedAtlas();
}
```
and there's no need for separate `atlas` and `flippedAtlas` variables.

packages/flame/lib/src/cache/images.dart
```diff
@@ -45,6 +45,19 @@ class Images {
...
   }
 
+  /// Returns the image [name] from the cache. If no image with [name] exists
+  /// uses [ImageGenerator] to generate and then add the resulting image into
+  /// the cache.
+  /// The cache will assume the ownership of the resulting image, and will
+  /// properly dispose of it at the end.
+  Future<Image> fetchOrGenerate(
+    String name,
+    Future<Image> Function(Images) ImageGenerator,
```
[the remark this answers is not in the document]
spydon: We must have missed that rule, added it in #1656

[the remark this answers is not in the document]
nathanaelneveux: I don't know if its managed separately than the linter but the autocomplete also capitalized the function name (I remember thinking that was odd)

packages/flame/lib/src/sprite_batch.dart
nathanaelneveux: The only thing missing from this is caching the flipped atlas - this is what I've been stuck on and circling around as well. I think the caching is still important because every layer of a map that is loaded will reload the same set of atlas images each time. So at a minimum caching prevents regenerating for each layer of a map.

The suggestions you made here and @spydon comment as well has given me an idea on how to get the caching working though

nathanaelneveux:
This isn't quite working yet but `late` is working now so generation is happening at the correct time and caching is working.

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -172,28 +182,63 @@ class SpriteBatch {
...
+      _emptyPaint,
+    );
 
     final picture = recorder.endRecording();
-    final atlas = picture.toImageSafe(image.width * 2, image.height);
-    return atlas;
+    return picture.toImageSafe(image.width, image.height * 2);
+  }
+
+  static Future<Image> _generateFlippedAtlas2(
```
spydon, line 221: Why is this a static method? (And the other one)

packages/flame/lib/src/sprite_batch.dart
spydon: Since you're using `fetchOrGenerate` for the flipped atlas, isn't it already being cached?

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -216,21 +261,26 @@ class SpriteBatch {
...
     final batchItem = BatchItem(
       source: source,
       transform: transform ??= defaultTransform ?? RSTransform(1, 0, 0, 0),
       flip: flip,
       color: color ?? defaultColor,
     );
 
+    if (flip && useAtlas && !_hasFlips) {
+      _hasFlips = true;
+      atlas = await _flippableAtlas;
```
st-pasha, line 274: I wouldn't use `await` here (nor declare this method `async`) -- this has potential of introducing some subtle bugs like if a user adds several transforms
```dart
batch.addTransform(...);
batch.addTransform(..., flip: true);
batch.addTransform(...);
```
then the second transform will be added in the end, after the 3rd, because it waits for the new atlas to be generated. This would be problematic for users who want their batches to be in the exact order they specified.

st-pasha:
@nathanaelneveux This is indeed a bit more complicated than it looks at the first sight, but I feel we're already quite close to the finishing line.

So, the `SpriteBatch` has two constructors: the default one `.new`, and async `.load`. We would want the flips to work correctly in both cases. And "correctly" in this case means that the flipped image has to be cached.

For caching, we need two things: (1) the cache itself, and (2) the key. Currently, the `Images` cache is passed as a parameter in one of the constructors only (we can add it as an optional parameter for the second constructor too). But this cache is needed at that later time when we decide we need to make flips, so first thing that we need is:

- [ ] Store `images` argument in `.load` as a parameter `final Images? _imageCache` in the `SpriteBatch` class.
- [ ] Make a getter `Images get imageCache => _imageCache ?? Flame.images;`.

The key is a bit trickier. For the `.load` constructor the path of the image is the key; but the default constructor gets the raw image, without a key. So, we need to work around that:

- [ ] Store `path` argument in `.load` as a parameter `final String? _imageKey`.
- [ ] Create a method for getting the actual image key, something like `String get imageKey => _imageKey ?? imageCache.findKeyForImage(atlas) ?? 'image[${identityHashCode(atlas)}]';`.
- [ ] The `Images.findKeyForImage` method needs to be implemented, but that's fairly straightforward.

Lastly, 

- [ ] have a method that would generate and cache the flipped image:
  ```dart
  Future<void> _makeFlippedAtlas() async {
    _hasFlips = true;
    _atlasReady = false;
    final key = '$imageKey#with-flips';
    atlas = await imageCache.fetchOrGenerate(key, _generateAtlas);
    _atlasReady = true;
  }
  ``` 
  Even though this method is async, you don't await on it in the `addTransform` -- instead in the `render()` method you can replace the check `if (useAtlas)` with `if (useAtlas && _atlasReady)`. This way the renderer will automatically fall back to "regular" rendering while the atlas is being generated.

- [ ] Remove variables `_flippableAtlas` and `flippableAtlasGenerator`.

I hope this would do it, if not please let me know.

spydon:
Do you want help finishing this up @nathanaelneveux? :slightly_smiling_face:

nathanaelneveux:
Yes actually! I just committed what I have so far - it's close but I'm having trouble with some of the parameters.

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -172,28 +198,45 @@ class SpriteBatch {
...
+    atlas = await imageCache.fetchOrGenerate(
+      key,
+      () => _generateFlippedAtlas(atlas),
+    );
+    _atlasReady = true;
+  }
+
+  Future<Image> _generateFlippedAtlas(
+    Image image,
+  ) async {
```
st-pasha, line 226: ~~`async`~~

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -216,21 +259,26 @@ class SpriteBatch {
     RSTransform? transform,
     bool flip = false,
     Color? color,
-  }) {
+  }) async {
```
st-pasha, line 262: ~~`async`~~

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -315,7 +363,7 @@ class SpriteBatch {
   }) {
     paint ??= _emptyPaint;
 
-    if (!useAtlas) {
+    if (!useAtlas || !_atlasReady) {
```
st-pasha, line 366: If you switch the order of then and else blocks here, then the condition becomes simpler: `if (useAtlas && _atlasReady)`

st-pasha (commented):
Looks good overall, can you elaborate please what's the problem with the parameters ?

nathanaelneveux:
Screen Shot 2022-06-01 at 10 04 00 PM
Right now the test renders before `_atlasReady = true` The `Canvas.drawImageRect` method seems to be drawing the flipped images just slightly off (see picture above).

You can test the `Canvas.drawAtlas` method by calling `await Flame.images.ready();` right before rendering in the test.

nathanaelneveux:
Even though this is passing tests this isn't working on my test map - the flips are wrong. I'll have a test that looks at a larger atlas to get the bottom of this and make sure it doesn't happen again.

spydon:
> Even though this is passing tests this isn't working on my test map - the flips are wrong. I'll have a test that looks at a larger atlas to get the bottom of this and make sure it doesn't happen again.

It seems like it is failing both of the tests that you added, even the one not using the atlas?
What error are you experiencing with larger maps?

nathanaelneveux:
> > Even though this is passing tests this isn't working on my test map - the flips are wrong. I'll have a test that looks at a larger atlas to get the bottom of this and make sure it doesn't happen again.
> 
> It seems like it is failing both of the tests that you added, even the one not using the atlas? What error are you experiencing with larger maps?

There was an error introduced when I stacked the flipped image vertically instead of horizontally in [35e08d8](https://github.com/nathanaelneveux/flame/commit/35e08d8e365a8fbb80953758c4b944d55d21480f). `source.top + source.height,` should have been `atlas.height/2 + source.top`. This was still passing tests because the tile map in this test was only 1 tile high - `source.top` = `atlas.height/2` in this instance. I didn't want to introduce a division into `addTransform` so I reverted to the generateAtlas I had before [35e08d8](https://github.com/nathanaelneveux/flame/commit/35e08d8e365a8fbb80953758c4b944d55d21480f). I believe that isn't passing tests now because `atlas.width` is no longer reliable - it may still be the width of the atlas before `_makeFlippedAtlas()` completes.

nathanaelneveux:
Screen Shot 2022-06-02 at 9 17 54 PM

I was wrong about the screenshot I posted yesterday  - the above is what is produced by `Canvas.drawImageRect`. So the `batchItem.source` is also going to need some adjustments.

nathanaelneveux:
Tests are passing now - I know there must be a better way to do the conditional in the BatchItem construction but its too late here for me to think straight

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -129,7 +168,30 @@ class SpriteBatch {
...
+
+  /// When the [_imageKey] isn't specified [imageKey] will return either the key
+  /// for the [atlas] stored in [imageCache] or a key generated from the
+  /// identityHashCode.
+  String get imageKey =>
+      _imageKey ??
+      imageCache.findKeyForImage(atlas) ??
+      'image[${identityHashCode(atlas)}]';
+
+  /// Whether any [BatchItem]s need a flippableAtlas.
```
wolfenrain, line 190:
```suggestion
  /// Whether any [BatchItem]s needs a flippable atlas.
```

wolfenrain (approved):

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -37,6 +39,23 @@ const _defaultScale = 0.0;
...
+    required this.color,
+  })  : matrix = Matrix4(
+          transform.scos, transform.ssin, 0, 0, //
+          -transform.ssin, transform.scos, 0, 0, //
+          0, 0, _defaultScale, 0, //
+          transform.tx, transform.ty, 0, 1, //
+        )
+          ..translate(source.width / 2, source.height / 2)
+          ..rotateY(flip ? pi : 0)
+          ..translate(-source.width / 2, -source.height / 2),
```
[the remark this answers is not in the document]
st-pasha: We could have some extension method to make a `Matrix4` out of `RSTransform`, looks like there isn't one right now.

That `_defaultScale` looks mighty suspicious to me as well, I think we can just remove it without anyone noticing.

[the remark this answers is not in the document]
spydon: Removed it, how come the scale is 0 and not 1? :thinking:

nathanaelneveux:
> Thanks for your very hard and long work on this PR @nathanaelneveux, very appreciated! 🎉
> (Are you on our discord?)

Phew :) that definitely ending up being more than I expected when first tackling this as a hobby project lol. Thank you all for your patience and help - great community you’ve built here

packages/flame/lib/src/sprite_batch.dart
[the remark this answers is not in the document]
nathanaelneveux: This is the difference between the matrix's with the rotateY code and without:
```
WITH FLIPS

[0] 1.0,-0.0,-1.2246467991473532e-16,32.0
[1] -0.0,-1.0,0.0,16.0
[2] -0.0,0.0,0.0,0.0
[3] -0.0,0.0,0.0,1.0

[0] -0.0,1.0,0.0,48.0
[1] 1.0,0.0,-1.2246467991473532e-16,0.0
[2] -0.0,0.0,0.0,0.0
[3] -0.0,0.0,0.0,1.0

[0] -0.0,-1.0,0.0,48.0
[1] -1.0,0.0,1.2246467991473532e-16,32.0
[2] -0.0,0.0,0.0,0.0
[3] -0.0,0.0,0.0,1.0

[0] -1.0,-0.0,1.2246467991473532e-16,64.0
[1] -0.0,1.0,0.0,16.0
[2] -0.0,0.0,0.0,0.0
[3] -0.0,0.0,0.0,1.0

NO FLIPS

[0] -1.0,-0.0,0.0,48.0
[1] 0.0,-1.0,0.0,16.0
[2] 0.0,0.0,0.0,0.0
[3] 0.0,0.0,0.0,1.0

[0] 0.0,1.0,0.0,48.0
[1] -1.0,0.0,0.0,16.0
[2] 0.0,0.0,0.0,0.0
[3] 0.0,0.0,0.0,1.0

[0] 0.0,-1.0,0.0,48.0
[1] 1.0,0.0,0.0,16.0
[2] 0.0,0.0,0.0,0.0
[3] 0.0,0.0,0.0,1.0

[0] 1.0,-0.0,0.0,48.0
[1] 0.0,1.0,0.0,16.0
[2] 0.0,0.0,0.0,0.0
[3] 0.0,0.0,0.0,1.0
```
(`.toString` doesn't look to be in the same order as the Matrix4 initializer)

packages/flame/lib/src/cache/images.dart
```diff
@@ -45,6 +45,19 @@ class Images {
     _assets[name] = _ImageAsset.fromImage(image);
   }
 
+  /// Returns the image [name] from the cache. If no image with [name] exists
+  /// uses [imageGenerator] to generate and then add the resulting image into
+  /// the cache.
+  /// The cache will assume the ownership of the resulting image, and will
+  /// properly dispose of it at the end.
```
alestiago, line 52: Nit: This documentation requires formatting to comply with effective Dart.

I'm using a phone so I cant give a proper formatted example.

However, the first line/s should be as brief and separated by a new line. 

In addition we should avoid being redundant with what the function signature is already telling us. For example, "Returns an image" is already something we now from the signature.

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -9,34 +10,47 @@ import 'package:flame/src/flame.dart';
...
 /// A single item in a SpriteBatch.
 ///
 /// Holds all the important information of a batch item,
 class BatchItem {
+  BatchItem({
+    required this.source,
+    required this.transform,
+    this.flip = false,
+    required this.color,
+  })  : matrix = Matrix4(
```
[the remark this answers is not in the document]
spydon: > why is is not null?

That's true, the matrix should probably be lazily initialized.

> What is this Matrix doing and why are this particular values useful?

It is just another representation of the `RSTransform`.

packages/flame/lib/src/cache/images.dart
spydon: Updated

packages/flame/lib/src/cache/images.dart
```diff
@@ -45,6 +45,19 @@ class Images {
...
   }
 
+  /// If the image with [name] exists in the cache that is returned, otherwise
+  /// the image generated by [imageGenerator] is returned.
+  ///
+  /// If the [imageGenerator] is used, the resulting [Image] is stored with
+  /// [name] in the cache.
+  Future<Image> fetchOrGenerate(
+    String name,
+    Future<Image> Function() imageGenerator,
```
alestiago, line 55: Why is it named "Generator" instead of "Builder"?

packages/flame/lib/src/cache/images.dart
```diff
@@ -45,6 +45,19 @@ class Images {
     _assets[name] = _ImageAsset.fromImage(image);
   }
 
+  /// If the image with [name] exists in the cache that is returned, otherwise
+  /// the image generated by [imageGenerator] is returned.
+  ///
+  /// If the [imageGenerator] is used, the resulting [Image] is stored with
+  /// [name] in the cache.
+  Future<Image> fetchOrGenerate(
```
alestiago, line 53: Is this meant to fetch and if it fails to do so then, it builds an image and also caches it?

packages/flame/lib/src/sprite_batch.dart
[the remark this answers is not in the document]
alestiago: Thanks for taking care of this!

packages/flame/lib/src/cache/images.dart
spydon: Yes, is that not clear from the Dartdocs? 😅

packages/flame/lib/src/cache/images.dart
spydon: I didn't name it, but it's not a builder (if the builder name is used according to the builder pattern).

PR #1643, opened 2022-05-19
fix: Avoid leaks when using PictureRecorders

st-pasha:
# Description

When using `PictureRecorder`, the `Picture` object that it creates needs to be properly disposed after it is no longer used.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/lib/src/extensions/picture_extension.dart
+++ b/packages/flame/lib/src/extensions/picture_extension.dart
@@ -2,7 +2,7 @@ import 'dart:ui';
 
 extension PictureExtension on Picture {
   /// Converts [Picture] into an [Image] and disposes of the picture.
-  Future<Image> asImage(int width, int height) {
+  Future<Image> toImageSafe(int width, int height) {
     return toImage(width, height).then((image) {
       dispose();
       return image;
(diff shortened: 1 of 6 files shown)


packages/flame/lib/src/extensions/picture_extension.dart
```diff
@@ -0,0 +1,11 @@
+import 'dart:ui';
+
+extension PictureExtension on Picture {
+  /// Converts [Picture] into an [Image] and disposes of the picture.
+  Future<Image> asImage(int width, int height) {
```
spydon, line 5: Aah, this is a good solution!

spydon (approved):

spydon:
It would be really nice to be able to have a custom lint rule that checks so that `toImage` isn't used, but I don't think that is possible (or not worth the amount of time it would take to fix that at least).

packages/flame/lib/src/extensions/picture_extension.dart
erickzanardo: I love this solution too! But I have problems with the method name, it is way too similar to `asImage`.

Maybe we could name this as `safeToImage` or something that makes it more clear what it does?

erickzanardo (approved):
LGTM

PR #1645, opened 2022-05-20
feat: add FutureOr support on SpriteButton

oloshe:
It can use Future<Sprite> as its argument


spydon:
I don't see the point of this change, can you please elaborate?

oloshe:
> I don't see the point of this change, can you please elaborate?

Whether it's a Future or not, it ends up being a Future<List<Sprite>>. It can also be passed directly when you need to load the image asynchronously to create a Sprite or a method return Future<Sprite> like Sprite.load. In addition, it will not affect the case of directly passing Sprite as a parameter

spydon:
Ah, now when I looked at the rest of the code it makes sense!

spydon (approved):

erickzanardo (approved):

PR #1648, opened 2022-05-22
feat: allow controlling when a fixture is rendered

alestiago:
# Description

Changes `_renderFixtures` to `renderFixture`. This keeps the consistency with the other public render methods in `BodyComponent`.

Introduces new functionalities:
- Allow listening when a given fixture is rendered 
- Allows modifying the rendering logic for a given (one or more) fixtures. 

For example, avoiding a fixture to be rendered:
```dart
class MyBodyComponent extends BodyComponent {
...
void renderFixture(Canvas canvas, Fixture fixture) {
 final avoidRendering = condition; // Any condition here.
  if (avoidRendering) return;
  super.renderFixture(canvas, fixture);
}
...
}
```

## Checklist

## Breaking Change

## Related Issues


spydon:
I don't think we should rely on userdata for this (preferably we shouldn't rely on userdata for anything, it is a way too fragile pattern).

alestiago:
(cc: @spydon )

This PR is an initial proposal to allow rendering only some Fixtures. Once settled, I'll update tests, examples and docs.

alestiago:
> I don't think we should rely on userdata for this (preferably we shouldn't rely on userdata for anything, it is a way too fragile pattern).

Thanks @spydon for having a look. One doesn't need to rely on `userData`, as pointed out in the description, one can store those fixtures that would be skipped during rendering.

packages/flame_forge2d/lib/body_component.dart
```diff
@@ -74,36 +74,52 @@ abstract class BodyComponent<T extends Forge2DGame> extends Component
...
+  /// to customize how fixtures are rendered. For example, you can filter out
+  /// fixtures that you don't want to render.
+  ///
+  /// **NOTE**: If [renderBody] is false, no fixtures will be rendered. Hence,
+  /// [renderFixture] is not called when [render]ing.
+  void renderFixture(
+    Canvas canvas,
+    Fixture fixture,
+  ) =>
+      _renderFixture(canvas, fixture);
```
st-pasha, line 121: Looks like the `_renderFixture()` method can be inlined here, merging the two.

alestiago: Indeed, that's how I had it initially. I think it fits well.

However, I went with `_renderFixture()` to keep with the file consistency, since `renderChain` has `_renderChain` and so forth. Those methods could be also a single method.

I'm in favour of not using `_renderFixture`. 

Wdyt @spydon ?

spydon: Yeah, it doesn't add anything.

spydon (approved):
Great with some goldens here too :)

packages/flame_forge2d/pubspec.yaml
```diff
@@ -4,17 +4,20 @@ version: 0.11.0
 homepage: https://github.com/flame-engine/flame/tree/main/packages/flame_forge2d
 
 environment:
-  sdk: ">=2.16.0 <3.0.0"
+  sdk: ">=2.17.0 <3.0.0"
```
luanpotter, line 7: should we keep this consistent across projects (ie 16)?
if we want to update we should make a separate PR updating every package

luanpotter (approved):
LGTM with one comment. Nice tests!

packages/flame_forge2d/pubspec.yaml
luanpotter, line 7:
```suggestion
  sdk: ">=2.16.0 <3.0.0"
```

alestiago: For sure! When do we plan to update the SDK? Is there something stopping us from it?

spydon: I'm going to update after the release

spydon: And it is consistent on all other packages, if that was a question :)

Issue #1654, opened 2022-05-25
Add completed callback map on `SpriteAnimationGroupComponent`


oloshe:
Sometimes we use `SpriteAnimationGroupComponent` to control a sprite character to act something, it's convenient and useful. However, when I want to add some logic after some animation completed, it need to do some inconvenient steps, it was like:

```dart
SpriteAnimationGroupComponent<T> comp;

// ...

comp.animation!.onComplete = onCompleted;

// ...

void onCompleted() {
    // do my logic

    // reset onComplete to null
    comp.animation!.onComplete = null;
}
```

# Problem to solve
It may make it easier and convenient to invoke each status's callback or "hook",  such as reset my status or switch to other animation.

# Proposal

The way I solve it is extends it and add a field just like the `removeOnFinish`:

```dart
class SpriteAnimationGroupCallbackComponent<T> extends SpriteAnimationGroupComponent<T> {
  final Map<T, void Function()> onFinish;
  // ...

  @override
  void update(double dt) {
    super.update(dt);
    if (onFinish[current] != null && (animation?.done() ?? false)) {
      onFinish[current]!.call();
    }
  }
}
```

But it's seem to no need to create an new class to do this trivial thing, maybe add it on `SpriteAnimationGroupComponent` is more suitable.

spydon:
Sounds like a good addition to me, do you want to put up a PR for it?

oloshe:
alright, i'll create a pr later

munsterlander:
I am confused on why this is needed.  As a group component, it is just implementing `SpriteAnimation` underneath the `SpriteAnimationComponent` that it wraps.  Those individual animations already have `onStart`, `onFrame`, and `onComplete` callbacks.  As a group component, there is not an end of the group, but an end of the individual animation that is current.  So in your group component - at least this is how I am doing it - you just need:

```
    final testAnimation = testSpriteSheet.createAnimation(row: 0, stepTime: 1, loop: false, to: 8);
    testAnimation.onComplete = (){
    //whatever
    };
    animations = {
      TestAnimationGroupState.testState: testAnimation,
      ....
};
etc...
```

In short, this was the whole reason we added the other callbacks, so why is a wrapper component callback necessary as the group component doesn't really have an end as you can change current to a different animation? 

Again, I could be completely lost which is a common occurrence.

oloshe:
Thanks for @munsterlander 
You answered my doubts. I wrapped so many that I forgot they could be defined individually.

[...]

PR #1655, opened 2022-05-25
feat: add `onFinish` on `SpriteAnimationGroupComponent`

oloshe:
see in #1654


spydon:
Looks good, it would be great with some small tests for this!

oloshe:
Absolutely, I forgot it. I should add it later.

PR #1661, opened 2022-05-27
feat: Implement tap events based on `componentsAtPoint`

### LINKED ISSUE #1326: 🐘 Upgrade events system
This issue records that, which we have known for quite some time now, but didn't dare to bring up: **our events system no longer fits into the Flame engine**.

For the background, the Flame engine is built around the concept of a _component tree_ -- the structure that contains all components that are currently in game. The semantics of this tree is that of ownership: which components "own" which; but we also use the tree in order to traverse it in order to dispatch `update` events, `render` events, `resize` events, pointer (tap/drag) events (tap/drag), etc.

The `render` events are of particular interest here, because pointer events ought to be dispatched to whichever component is rendered at that point of the screen where the event occurred. So, in the "normal" case, the components in the tree are rendered according to their pre-order traversal. But it is also possible that a component overrides their `renderTree()` method, completely altering their rendering pipeline. 

Thus, if an engine allows the components to be rendered out-of-tree-order, it must also allow for the pointer events to be properly dispatched to those rendered components.

I don't know yet how exactly to implement this upgrade, so this issue is a call for proposals. I'll post mine too, if I come up with one. But regardless, this elephant in the room ought to be addressed.

### LINKED ISSUE #1511: Tap gesture not propagating properly with the CameraComponent
# Current bug behaviour
When using the new experimental CameraComponent the tap gesture event doesn't get propagated properly. Tapping on the screen results in incorrect tappable components being called if the camera position is moved. 

# Expected behaviour
The tap event should propagate to the component tapped even after moving the camera.

# Steps to reproduce
1. Create a FlameGame class with HasTappables
2. Extend the world class and include tappable SpriteComponents.
3. Create and add a CameraComponent to the onLoad method of this class.
4. Translate the CameraComponent to another position
5. Try tapping the tappable SpriteComponents in the world -> They don't get called when tapping on them. But you can find a position on the screen where they do get called.

* Flame version: 1.1.0
* Platform affected:  Android, probably iOS too.

# More information
Example code structure:
```
class CustomGame extends FlameGame with HasTappables, HasDraggables{
@override
  Future<void> onLoad() async {
    await super.onLoad();

    final world = CustomWorld();
    await add(world);
    final camera = CameraComponent(world: world);
    await add(camera);

    camera.viewfinder.position = Vector2(200,200);
    camera.viewfinder.zoom = 0.4;
  }
}

class CustomWorld extends World {
  @override
  Future<void> onLoad() async {
     add(CustomSprite(200,200));
     add(CustomSprite(0,0));
     add(CustomSprite(200,400));
    }
  }

}


class CustomSprite extends SpriteComponent with HasGameRef<CustomGame> , Tappable{
   CustomSprite (double x, double y) {
    this.x = x;
    this.y = y;
  }
  @override
  Future<void> onLoad() async {
    await super.onLoad();
    sprite = await gameRef.loadSprite('any-image-here.png');
    size = Vector2(200,200);
  }
  
  @override
  bool onTapDown(TapDownInfo info) {
    print("Tapped");
    return true;
  }

}
```

st-pasha:
# Description

This PR adds a new event system based on the `componentsAtPoint` delivery mechanism. These events allow for a proper support of components that implement `renderTree()` method. The `CameraComponent` is one such component, with more planned in the future.

Additionally, the following improvements compared to the current tap events added:
 - the same component can be tapped with multiple fingers simultaneously;
 - a component that received `onTapDown` is guaranteed to receive `onTapUp` or `onTapCancel` later;
 - a component that moves away from the point of touch will receive `onTapCancel` instead of `onTapUp` (even though the game widget receives `onTapUp` from Flutter).

Due to the fact that the switch from the current event system to the new event system is potentially a significant breaking change, the new event system is introduced as parallel to the existing one. This way we have more time to test the new system before recommending the switch and deprecating the old one; and the switch itself should feel more gradual.

## Checklist

## Breaking Change

## Related Issues

Closes #1326
Closes #1511


packages/flame/lib/src/events/flame_game_mixins/has_tappable_components.dart
```diff
@@ -0,0 +1,203 @@
...
+/// the component tree.
+///
+/// === Usage notes ===
+/// - If your game uses components with [TapCallbacks], then this mixin must be
+///   added to the [FlameGame] in order for [TapCallbacks] to work properly.
+/// - If your game also uses [Tappable] components, then add the
+///   [HasLegacyTappables] mixin as well (instead of `HasTappables`).
+/// - If your game has no tappable components, then do not use this mixin.
+///   Instead, consider `MultiTouchTapDetector`.
+mixin HasTappableComponents on FlameGame implements MultiTapListener {
```
spydon, line 29: This is actually a nicer name than `HasTappables` too, so maybe we should deprecate the other one later instead of renaming this one to that name like we discussed.

spydon (commented):
Looking good so far, well written dartdocs!

erickzanardo (commented):
I need to take a closer look, but I like the changes so far, I have two comments though:

 - This should be exported by the experimental namespace since, well, we are experimenting with it.
 -  I really dislike the `HasLegacyTappables` since `HasTappables` isn't deprecated yet. My suggestion would be something like `HasTappablesProxy` or `HasTappablesBridge` or something like that.

packages/flame/lib/src/events/messages/event.dart
```diff
@@ -0,0 +1,40 @@
...
+  /// This flag is neither set nor read by Flame. Instead, it can be set by the
+  /// user in a component that handles an event, and then read by the user in a
+  /// different component, or at the root Game level.
+  bool handled = false;
+
+  /// If this flag is false (default), the event will be delivered to the first
+  /// component that can handle it. If that component sets this flag to true,
+  /// the event will propagate further down the component tree to other eligible
+  /// components.
+  bool continuePropagation = false;
```
spydon, line 23: Currently this defaults to true, but maybe it's better to default to false.

spydon (approved):
Nice, very keen to see examples running directly in the docs.

packages/flame/lib/src/events/messages/event.dart
st-pasha: I don't understand, the default is false here, and it says so in the doc-comment too.

spydon: Yes, but currently for the `Tappable` it is true.

st-pasha: Ok, so no change's needed for this PR, right?

spydon: No, I think this is good, because I think this is how it usually is in these kind of event systems, that it by default stops after the first "thing" that it hits.

doc/flame/examples/lib/tap_events.dart
```diff
@@ -0,0 +1,125 @@
...
+  }
+}
+
+/// This component is the tappable blue-ish rectangle in the center of the
+/// game. It uses the [TapCallbacks] mixin in order to inform the game that it
+/// wants to receive tap events.
+class TapTarget extends PositionComponent with TapCallbacks {
+  TapTarget() : super(anchor: Anchor.center);
+
+  final Paint _paint = Paint()..color = const Color(0x448BA8FF);
```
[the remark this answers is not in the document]
spydon: Sounds like we should add this rule to the linter, I'll look into that.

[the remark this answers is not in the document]
spydon: Seems like there isn't any rule for it `omit_local_variable_types` is the closest, but that wouldn't apply here.

PR #1696, opened 2022-06-04
feat: Added utility function solveCubic()

st-pasha:
# Description

The function solves a cubic equation, and can be used for a variety of purposes, such as handling the Bezier curves.

## Checklist

## Breaking Change

## Related Issues


spydon (approved):

packages/flame/lib/src/utils/solve_cubic.dart
```diff
@@ -0,0 +1,59 @@
...
+
+/// Solves cubic equation `ax³ + bx² + cx + d == 0`.
+///
+/// Depending on the coefficients, either 1 or 3 real solutions may be returned.
+/// In degenerate cases, when some of the roots of the equation coincide, we
+/// return all such roots without deduplication.
+///
+/// If coefficient [a] is equal to zero, then we solve the equation as a
+/// quadratic one (see [solveQuadratic]).
+List<double> solveCubic(double a, double b, double c, double d) {
```
wolfenrain, line 13: I wonder if this should be in utils or if we should have a dedicated math section?

wolfenrain (approved):
LGTM, left a small comment that I kinda forgot in the previous PR😅

packages/flame/lib/src/utils/solve_cubic.dart
st-pasha: "Math sees itself as a dreamy poet. Science sees it as a supplier of specialized technical equipment. And herein we find one of the greatest paradoxes of human inquiry: These two views, both valid, are hard to reconcile. If math is an equipment supplier, why is its equipment so strangely poetic? And if math is a poet, then why is its poetry so unexpectedly useful?"

spydon: Haha, that can be the top level doc for the math section :smile:
I agree that we can have a math directory and a barrel file for it, I'm guessing we'll have quite a lot of more things going in there later.

spydon: That can be done on a separate PR though since it will move the other file too

PR #1709, opened 2022-06-07
fix: Fix flame package example app

luanpotter:
# Description

Our main example for the flame package was no longer working due to some changes on the input system, namely the need to manually mark events as handled.
I also removed some pausing logic that I do not think belongs on the main example, that can be a separate topic on the dashbook example. This example should be simple and contain key learning topics.

## Checklist

## Breaking Change

### DIFF
--- a/packages/flame/example/lib/main.dart
+++ b/packages/flame/example/lib/main.dart
@@ -14,12 +14,10 @@ void main() {
   );
 }
 
-/// This example simply adds a rotating white square on the screen, if you press
-/// somewhere other than on the existing square another square will be added and
-/// if you press on a square it will be removed.
-class MyGame extends FlameGame with DoubleTapDetector, HasTappables {
-  bool running = true;
-
+/// This example simply adds a rotating white square on the screen.
+/// If you press on a square, it will be removed.
+/// If you press anywhere else, another square will be added.
+class MyGame extends FlameGame with HasTappables {
   @override
   Future<void> onLoad() async {
     add(Square(Vector2(100, 200)));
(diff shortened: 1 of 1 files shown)


packages/flame/example/lib/main.dart
```diff
@@ -80,6 +67,7 @@ class Square extends PositionComponent with Tappable {
   @override
   bool onTapUp(TapUpInfo info) {
     removeFromParent();
+    info.handled = true;
```
spydon, line 70: Why was this needes?

spydon (approved):

renancaraujo (approved):

packages/flame/example/lib/main.dart
luanpotter: because returning true no longer sets it automatically

spydon: But why does it need to be set?

luanpotter: so the check to avoid adding a new square works

spydon: Ah, but just return false instead to stop the propagation, and no check is needed.

packages/flame/example/lib/main.dart
```diff
@@ -80,6 +67,7 @@ class Square extends PositionComponent with Tappable {
...
     return true;
```
spydon, line 71: Here 👆

packages/flame/example/lib/main.dart
luanpotter: let me try

luanpotter: I thought `true` meant handled

luanpotter: doesn't work

spydon: `true` means continue to propagate, but it always ends up on the game it seems like? I wonder if that is a bug...
You can check `info.continuePropagation` on the game, but then it's cleaner to do lile you did before.

spydon: Yeah, it must have always been like that, since that is the entry point, I'll merge this

PR #1715, opened 2022-06-08
feat: Drag events that dispatch using componentsAtPoint

### LINKED ISSUE #1325: Drag events should always dispatch to the component(s) that received `onDragStart`
After `onDragStart`, the system emits a stream of `onDragUpdate` events, culminating in a `onDragEnd` or `onDragCancel`. Currently, each of those follow-up events traverses the game tree again, searching for the component(s) that received the original `onDragStart`. This is not particularly efficient, plus may potentially create bugs when the event is accidentally cancelled by one of the components even if one of the components did not receive it.

The proposed solution is simple: during `onDragStart`, store the component that received this event into a list inside `HasDraggables`, and later dispatch all subsequent events to that list.

### LINKED ISSUE #1326: 🐘 Upgrade events system

### LINKED ISSUE #1677: Implement drag events for the new event system
# What could be improved

Drag events that use `componentsAtPoint` delivery mechanism, and therefore work correctly in the presence of components that override `renderTree()`.

# More information

~~Blocked by #1668~~
~~Blocked by #1661~~
Required for #1325

st-pasha:
# Description

Implement drag events that dispatch using the `componentsAtPoint` mechanism.

## Checklist

## Breaking Change

## Related Issues

Closes #1677
Closes #1326
Closes #1325


spydon (approved):
Lgtm!

wolfenrain (approved):

Issue #1627, opened 2022-05-14
[EPIC] Text rendering improvements


st-pasha:
This issue collects various ideas for improvements of text rendering functionality in Flame. These are of various degrees of difficulty and/or usefulness, and there is no timeline on when we would be able to implement all of these, but still collecting all these ideas in one place is useful in order to be able to design the framework capable of supporting all these various use cases eventually.

- [x] Display text using a custom sprite-sheet based font. For example, a game may use a custom hand-drawn "font" which is prepared as a regular image, and it may not be practical to convert that image into an actual font file. (#1621)
- [x] Mix styles within a single text block.
  For example, you may want to have some part of text displayed as bold, or italic, or in a different color, etc.
- [x] Mix text renderers within a single text block.
  Some texts may need to use pictograms inside, for example `"This quest will give you 8💎"` (except that the pictogram would come from the game's assets and not from the standard emojis). This can be accomplished if a text block would allow mixing text fragments from the standard "TextPaint" renderer, and the spritesheet-based renderer.
- [ ] Automatically expand a marked-up plain text into rich text.
  Once we have the ability to display rich text, we need to also make it convenient, so that for example a string like `"To **boldly** go where no man has gone before"` could be automatically converted into a text fragment with a bold word inside. (#2049)
- [ ] Allow user-customizable text markup.
  As an extension to the previous functionality, some text fragments may need to be expanded into user-defined entities. For example, `"10[GP]"` may render as 10 + an icon for a gold coin.
- [ ] Allow dynamic text effects.
  Sometimes you may need to adjust either the style or placement of each letter individually. For example, each letter may have its own different color, or you may want some words to "shake" (as in Undertale).
- [ ] Allow text fragments to respond to tap events.
  This could allow URLs or URL-like functionality, or even simple tooltips.
- [ ] Render text with "justified" alignment.
- [x] Align different text pieces according to their baseline.
- [ ] In-text exclusion zones.
  For example, in order to insert images inside text, one would need to force the text to flow around that image.
- [ ] Render a string along a curve.

spydon:
I'll close this as stale since Pasha isn't doing any more PRs.

PR #1720, opened 2022-06-11
feat: Added TextFormatter classes

st-pasha:
# Description

* A `TextFormatter` is a class that can take a string of text and convert it into a `TextElement`.
* A `TextElement` is an object that encapsulates a text string prepared for rendering. This object knows its own dimensions (layout), and is able to render itself on a canvas.

A `TextFormatter` is different from the current `TextRenderer` in one crucial respect: whereas `TextRenderer` takes a string of text and draws it onto the canvas directly (performing the layout and measurement internally), the `TextFormatter` creates an object that encapsulates the information necessary for rendering.

Thus, `TextFormatter`'s output can be used to prepare the text once, and then render multiple times. In addition, since each `TextElement` knows its own layout, -- these individual layouts can be collected, manipulated, and arranged into a larger text body necessary for rich text rendering.

In addition, this PR:
- Implements a debug version of the standard text renderer, which draws rectangles instead of text;
- Re-enables the skipped golden test for the TextBoxComponent (hopefully it'll work this time);
- Refactors existing `TextRenderer`s to use the new `TextFormatter`s under the hood.

This PR is a WIP for the Rich Text functionality (#1627).

## Checklist

## Breaking Change

## Related Issues


packages/flame/lib/src/text/common/glyph_info.dart
```diff
@@ -0,0 +1,14 @@
+/// Helper class that stores dimensions of a single glyph for a spritesheet-
+/// based font.
+class GlyphInfo {
```
spydon, line 3: `GlyphDimensions` maybe? It's hard to differentiate `GlyphInfo` and `GlyphData` with their current names.

packages/flame/lib/src/text/common/line_metrics.dart
```diff
@@ -0,0 +1,98 @@
+import 'dart:ui';
+
+import 'package:flame/src/text/common/text_line.dart';
+
+/// [LineMetrics] object contains measurements of a [TextLine].
```
spydon, line 5:
```suggestion
/// The [LineMetrics] object contains measurements of a [TextLine].
```

packages/flame/lib/src/text/formatter_text_renderer.dart
```diff
@@ -0,0 +1,35 @@
+import 'dart:ui';
+
+import 'package:flame/src/anchor.dart';
+import 'package:flame/src/text/formatters/text_formatter.dart';
+import 'package:flame/text.dart';
+import 'package:vector_math/vector_math_64.dart';
+
+/// Helper class that implements a [TextRenderer] using a [TextFormatter].
+class FormatterTextRenderer<T extends TextFormatter> extends TextRenderer {
```
spydon, line 9: `Formatted` maybe?

packages/flame/lib/src/text/sprite_font_renderer.dart
```diff
@@ -14,134 +14,31 @@ import 'package:vector_math/vector_math_64.dart';
...
+  }) : super(
+          SpriteFontTextFormatter(
+            source: source,
+            charWidth: charWidth,
+            charHeight: charHeight,
+            glyphs: glyphs,
+            scale: scale,
+            letterSpacing: letterSpacing,
+          ),
+        );
```
spydon, line 43: Nice!

spydon (commented):
Just docs and an example missing otherwise it looks good!

st-pasha:
> Re-enables the skipped golden test for the TextBoxComponent (hopefully it'll work this time);

It actually did work this time!

packages/flame/lib/src/text/common/glyph_info.dart
st-pasha: Good point, we should probably merge these two somehow.
However, since this PR didn't create these classes -- it merely moved them out of `sprite_font_renderer.dart` file -- we should probably defer this refactoring to a separate PR.

packages/flame/lib/src/text/formatter_text_renderer.dart
st-pasha: The etymology here is that this is a `TextRenderer` based on a `TextFormatter`. So, the name is "TextFormatter TextRenderer", where the first "Text" is elided for brevity.

Note also that this is a mostly internal class -- the only situation where it could be useful to a user is if they tried to implement their own `TextRenderer`, which is rather unlikely to happen.

st-pasha:
> Just docs and an example missing otherwise it looks good!

My thinking was that first we develop the rich text functionality, over the course of several PRs, and only then add the proper documentation. Otherwise we'd likely have to rewrite the documentation when the rest of the pieces are developed.

packages/flame/lib/src/text/common/glyph_info.dart
spydon: Will you remember it? Or should we add an issue to track it with?

spydon (approved):

packages/flame/lib/src/text/common/glyph_info.dart
st-pasha: I think we'll remember. We would have to do some refactoring around those classes anyways, in order to support non-monospace sprite fonts.

packages/flame/lib/src/text/text_paint.dart
```diff
@@ -63,6 +47,7 @@ class TextPaint extends TextRenderer {
   /// However, you probably want to use the [render] method which already
   /// takes the anchor into consideration.
   /// That way, you don't need to perform the math for that yourself.
+  @Deprecated('This method will be removed in v1.4.0')
```
erickzanardo, line 50: Why will this be removed?

st-pasha: You're right, we should postpone such decisions until after the rest of the API is in place. I've removed the deprecation note.

packages/flame/lib/src/text/common/glyph_info.dart
```diff
@@ -0,0 +1,14 @@
...
+  double srcLeft = 0;
```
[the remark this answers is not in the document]
st-pasha: The `rst*` attributes are used in the `rstTransforms` argument for the `canvas.drawRawAtlas` function.

We do need to have proper documentation for them, but I was thinking of leaving that for later, since we still need to merge `GlyphData` and `GlyphInfo` classes, plus there would  be some reorganization of this class to support non-monospace fonts.

PR #1722, opened 2022-06-11
fix: GameWidget will not listen to keyboard events without the KeyboardEvents mixin

### LINKED ISSUE #1710: TextInput not reacting to keyboard

st-pasha:
# Description

Currently, the `GameWidget` uses a `Focus` widget always, even if the game is unable to receive keyboard events.
This PR makes it so that the `Focus` node is only used when the game has `KeyboardEvents` mixin that is used for listening to keyboard events. This is similar to how we use gesture detectors and mouse detectors.

Another advantage here is that now `Focus` is in parallel with the overlay widgets, which means that it won't interfere with the overlay's text input attempts.

## Checklist

## Breaking Change

- [x] No, this is *not* a breaking change (shouldn't be?).

## Related Issues

Closes #1710

### DIFF
--- a/packages/flame/lib/src/game/game_widget/game_widget.dart
+++ b/packages/flame/lib/src/game/game_widget/game_widget.dart
@@ -310,69 +310,73 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
           internalGameWidget,
         );
       }
-
       if (hasMouseDetectors(currentGame)) {
         internalGameWidget = applyMouseDetectors(
           currentGame,
           internalGameWidget,
         );
       }
+      if (currentGame is KeyboardEvents) {
+        internalGameWidget = applyKeyboardDetectors(internalGameWidget);
+      }
 
       final stackedWidgets = [internalGameWidget];
       _addBackground(context, stackedWidgets);
       _addOverlays(context, stackedWidgets);
 
-      // We can use Directionality.maybeOf when that method lands on stable
-      final textDir = widget.textDirection ?? TextDirection.ltr;
-
-      return Focus(
-        focusNode: _focusNode,
-        autofocus: widget.autofocus,
-        onKey: _handleKeyEvent,
-        child: MouseRegion(
-          cursor: currentGame.mouseCursor,
-          child: Directionality(
-            textDirection: textDir,
-            child: Container(
-              color: currentGame.backgroundColor(),
-              child: LayoutBuilder(
-                builder: (_, BoxConstraints constraints) {
-                  return _protectedBuild(() {
-                    final size = constraints.biggest.toVector2();
-                    if (size.isZero()) {
+      return MouseRegion(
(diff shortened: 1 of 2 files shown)


renancaraujo (changes requested):
By defined way back, the default behavior of a game widget is to handle all keyboard events if the game has primary focus. This was defined bacause of the warn sound that is popped by the system (on desktop) when a keyboard event is not handled at all. 
As a default behavior, it can be overridden by the user by setting `Keyboard.handled` on `onKeyEvent` in the game level.
Widgets outside of the GameWidget and overlays that may handle keyboard events, should make sure that they have focus.

st-pasha:
@renancaraujo 

This PR specifically address those `Game`s which do not implement the `KeyboardEvents` interface/mixin, and do not wish to interact with the keyboard.

I understand the scenario that you're describing, where unhandled key presses cause annoying beeping sounds. But if the user faces this problem, they have an easy solution: add the `KeyboardEvents` mixin to the game (whose default behavior is to mark all keys as "handled").

If, on the other hand, the user doesn't want to interfere with the key events, then they have no easy solution right now. They could, again, add the `KeyboardEvents` mixin and mark all keys as "ignored", but this is feels wrong and is very non-intuitive: if you don't want to handle the key events, why would you even think about adding a mixin dedicated to handling key events?

PR #1738, opened 2022-06-17
fix: SpriteAnimationWidget can now be update animation safely

### LINKED ISSUE #1697: SpriteAnimationWidget is causing error when its rebuild
# Current bug behaviour
SpriteAnimationWidget uses the `SingleTickerProviderStateMixin` but its AnimationController can be disposed and recreated on `didUpdateWidget` life cycle.
As `SingleTickerProviderStateMixin` can only create ticker once,  it should be changed to `TickerProviderStateMixin`

```dart

════════ Exception caught by widgets library ═══════════════════════════════════
The following assertion was thrown building BaseFutureBuilder<SpriteAnimation>:
_InternalSpriteAnimationWidgetState is a SingleTickerProviderStateMixin but multiple tickers were created.

A SingleTickerProviderStateMixin can only be used as a TickerProvider once.

If a State is used for multiple AnimationController objects, or if it is passed to other objects and those objects might use it more than one time in total, then instead of mixing in a SingleTickerProviderStateMixin, use a regular TickerProviderStateMixin.

The relevant error-causing widget was
SpriteAnimationWidget
```
# Expected behaviour

Use TickerProviderStateMixin to be able to recreate AnimationController.

# Steps to reproduce
- Minimum reproductions on SingleTickerProviderStateMixin. it causes following.
> _HasAnimationState is a SingleTickerProviderStateMixinn but multiple tickers were created.

```dart
import 'dart:async';
import 'dart:math';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  @override
    return const MaterialApp(
      home: MyHome(),
    );
  }
}

class MyHome extends StatefulWidget {
  const MyHome({Key? key}) : super(key: key);

  @override
  State<MyHome> createState() => _MyHomeState();
}

class _MyHomeState extends State<MyHome> {
  @override
  void initState() {
    Future.delayed(const Duration(seconds: 2)).then((value) => setState(() {}));
    super.initState();
  }

  @override
    return Scaffold(
      body: HasAnimation(
        value: Random().nextInt(100),
      ),
    );
  }
}

class HasAnimation extends StatefulWidget {
  final int value;
  const HasAnimation({Key? key, required this.value}) : super(key: key);

  @override
  State<HasAnimation> createState() => _HasAnimationState();
}

class _HasAnimationState extends State<HasAnimation>
    with SingleTickerProviderStateMixin {
  late AnimationController animationController;

  @override
  void initState() {
    setAnimation();
    super.initState();
  }

  setAnimation() {
    animationController = AnimationController(
      vsync: this,
      duration: const Duration(seconds: 1),
    );
  }

  disposeAnimation() => animationController.dispose();

  @override
  void didUpdateWidget(covariant HasAnimation oldWidget) {
    disposeAnimation();
    setAnimation();
    super.didUpdateWidget(oldWidget);
  }

  @override
    return Container();
  }
}

```

# Flutter doctor output

```
[✓] Flutter (Channel stable, 3.0.1, on macOS 12.1 21C52 darwin-arm, locale ko-KR)
    • Flutter version 3.0.1 at /Users/lafity101/flutter
    • Framework revision fb57da5f94 (2 weeks ago), 2022-05-19 15:50:29 -0700
    • Engine revision caaafc5604
    • Dart version 2.17.1
    • DevTools version 2.12.2

[✓] Xcode - develop for iOS and macOS (Xcode 13.4)
    • CocoaPods version 1.11.3

[✓] VS Code (version 1.67.2)

[✓] Connected device (4 available)
[✓] HTTP Host Availability
```

* Flame version: main

# More information

Are you interested in working on a PR for this? Yes

Hwan-seok:
# Description

Now, `SpriteAnimationWidget` could be execute `didUpdateWidget` safely.
I might think there is a better solution but could not find it.

## Checklist

## Breaking Change

## Related Issues

Fixes #1697 


spydon:
Hmm, can you explain a bit more what this change is supposed to do?
According to [this page](https://api.flutter.dev/flutter/widgets/TickerProviderStateMixin-mixin.html) we should be using the `SingleTickerProviderStateMixin` since there it only one ticker here, but I'm not very familiar with these mixins, so maybe you could explain this change a bit more?

spydon:
Thanks for the very detailed description @Hwan-seok, that makes sense!
Can't we just re-use the animation controller, or is that no possible?

spydon (approved):
Lgtm, good job!

packages/flame/lib/src/widgets/animation_widget.dart
```diff
@@ -140,17 +139,19 @@ class _InternalSpriteAnimationWidgetState
   }
 
   void _setupController() {
-    _controller?.dispose();
+    widget.animation.onComplete = _pauseAnimation;
+    _controller ??= AnimationController(vsync: this)
+      ..removeListener(_onAnimationValueChanged)
+      ..addListener(_onAnimationValueChanged);
```
wolfenrain, line 145: I wonder, if you use the `??=` operator, will it call the `removeListener` and `addListener` only if `_controller` was null? Or will it always call it?

Because I read this code as: If `_controller` is null assign it this object and call these methods on that object

If that is the case then this won't do what you expect it to do right?

packages/flame/lib/src/widgets/animation_widget.dart
```diff
@@ -140,17 +139,19 @@ class _InternalSpriteAnimationWidgetState
   }
 
   void _setupController() {
-    _controller?.dispose();
+    widget.animation.onComplete = _pauseAnimation;
+    _controller ??= AnimationController(vsync: this)
+      ..removeListener(_onAnimationValueChanged)
```
spydon, line 144:
```suggestion
    _controller ??= AnimationController(vsync: this);
    _controller
      ..removeListener(_onAnimationValueChanged)
```
Solving @wolfenrain's concern

packages/flame/lib/src/widgets/animation_widget.dart
Hwan-seok: You are right
It should be as follows.
```dart
(_controller ??= AnimationController(vsync: this))
      ..removeListener(_onAnimationValueChanged)
      ..addListener(_onAnimationValueChanged);
```

I am very confused about this after this review because even though the code has a bug but it works by either test or real build.

And I realized that removing and re-adding the listener doesn't needed.
`widget.animation` in `_onAnimationValueChanged` already referencing the updated animation after the lifecycle event.

Hwan-seok: I removed the `removeListener` 
Thanks for the advice! :)

erickzanardo:
Awesome! Thanks for the contribution and for the careful explanation!

erickzanardo (approved):

PR #1739, opened 2022-06-18
fix: World component can now be queried with `componentsAtPoint`

st-pasha:
# Description

The `componentsAtPoint` function allows us to find a component located at a specific position. Usually it is applied to the top-level Game, but there are instances when it could also be useful to apply this function to the `World` component. This PR enables such functionality.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/lib/src/experimental/world.dart
+++ b/packages/flame/lib/src/experimental/world.dart
@@ -1,6 +1,7 @@
 import 'dart:ui';
 
 import 'package:flame/src/components/component.dart';
+import 'package:flame/src/components/mixins/coordinate_transform.dart';
 import 'package:flame/src/experimental/camera_component.dart';
 import 'package:meta/meta.dart';
 import 'package:vector_math/vector_math_64.dart';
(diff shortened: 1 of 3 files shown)


packages/flame/lib/src/experimental/world.dart
```diff
@@ -25,18 +26,8 @@ class World extends Component {
   bool containsLocalPoint(Vector2 point) => true;
 
   @override
-  Iterable<Component> componentsAtPoint(
-    Vector2 point, [
-    List<Vector2>? nestedPoints,
-  ]) {
-    return const Iterable.empty();
-  }
+  Vector2? localToParent(Vector2 point) => null;
```
spydon, line 29: Isn't this an `UnimplementedException` rather than null?

spydon (approved):

packages/flame/lib/src/experimental/world.dart
st-pasha: The `null` return value for the CoordinateTransform interface indicates that the given point cannot be mapped to the parent coordinate space. Which is exactly the case here: there is no mapping between the device and the world coordinates other than what is provided externally by the camera.

erickzanardo (approved):

PR #1755, opened 2022-06-26
feat: Added RouterComponent 

### LINKED ISSUE #1375: Create Navigator component
A `NavigatorComponent` would fulfill the similar role that Flutter's `Navigator` widget does: it would facilitate switching between multiple independent screens.

A typical game is comprised of multiple UI screens: the start page, the settings page, the level selector page, the world page, the character inventory page, the help page, etc. The `NavigatorComponent`'s job will be to help move from one such page to another.

This is a common functionality that is needed in pretty much every game. So how do games solve this right now?
 - Some games simply remove the old page and mount the new one. This works if the page loading times are negligible, otherwise there will be a gap with no top-level component present.
 - Use Flutter's `Navigator` and have different `GameWidget`s for each game page. This works, but it's not very efficient as it requires that each page is loaded anew on every page switch.

st-pasha:
# Description

This PR adds the `RouterComponent` (see the docs for the description of its functionality).

https://user-images.githubusercontent.com/4231472/175832885-bcd1b821-c263-4d83-a4a6-daedf8e5f79f.mov

## Checklist

## Breaking Change

## Related Issues

~~Prerequisite: #1781~~
Closes #1375


doc/flame/examples/lib/navigator.dart
```diff
@@ -0,0 +1,412 @@
...
+    navigator = Navigator(
+      pages: {
+        'splash': Page(builder: SplashScreen.new),
+        'home': Page(builder: StartPageImpl.new),
+        'level1': LevelPage(builder: Level1PageImpl.new),
+        'level2': LevelPage(builder: Level2PageImpl.new),
+        'pause': Page(builder: PausePageImpl.new, transparent: true),
+      },
+      initialPage: 'splash',
+    )..addToParent(this);
```
spydon, line 21: Maybe clearer to just wrap this in `add` instead, since the assigment can still be done at the same time?

doc/flame/examples/lib/navigator.dart
```diff
@@ -0,0 +1,412 @@
...
+  void onGameResize(Vector2 size) {
+    super.onGameResize(size);
+    _logoOffset = Offset((size.x - _logo.width) / 2, size.y / 3 - _logo.height);
+    _button1.position = Vector2(size.x / 2, _logoOffset.dy + 140);
+    _button2.position = Vector2(size.x / 2, _logoOffset.dy + 200);
+  }
+
+  @override
+  void render(Canvas canvas) {
+    // canvas.drawColor(const Color(0xff063e67), BlendMode.src);
```
spydon, line 92: Should this be removed?

doc/flame/examples/lib/navigator.dart
```diff
@@ -0,0 +1,412 @@
...
+    _logo.paint(canvas, _logoOffset);
```
spydon, line 93: Why not use a `TextComponent`?

doc/flame/examples/lib/navigator.dart
```diff
@@ -0,0 +1,412 @@
...
+    action();
+  }
+
+  @override
+  void onTapCancel(TapCancelEvent event) {
+    scale = Vector2.all(1.0);
+  }
+}
+
+abstract class SimpleButton extends PositionComponent with TapCallbacks {
```
spydon, line 166: Feels like overkill creating new button components in the example, why not use one of the existing button components? I think it takes away focus from the point of this example.

doc/flame/examples/lib/navigator.dart
```diff
@@ -0,0 +1,412 @@
...
+    stopTime();
+    addRenderEffect(
+      PaintRenderEffect()
+        ..addDesaturation(opacity: 0.5)
+        ..addBlur(3.0),
+    );
+  }
+}
+
+class Level1PageImpl extends Component {
```
spydon, line 256: Why the `Impl`?

doc/flame/navigator.md
```diff
@@ -0,0 +1,63 @@
...
+the game. In addition, the "pause" button stops time and applies visual effects to the content of
+the page below it.
+```
+
+
+# Navigator
+
+The **Navigator** is a component whose job is to manage navigation across multiple screens within
+the game. It is similar in spirit to Flutter's [Navigator][Flutter Navigator] class, except that it
+works with Flame components instead of Flutter widgets.
```
spydon, line 17: Maybe an overlay should have the ability to be a page too? I think that would be quite useful.

packages/flame/lib/src/components/navigator.dart
```diff
@@ -0,0 +1,57 @@
+import 'package:collection/collection.dart';
+import 'package:flame/src/components/component.dart';
+import 'package:flame/src/components/page.dart';
+
+class Navigator extends Component {
+  Navigator({
+    required Map<String, Page> pages,
```
spydon, line 7: It would be nice if other types than strings could be used here too, enums for example, then the user can make their code a bit safer from runtime exceptions if they wish.

packages/flame/lib/src/components/navigator.dart
```diff
@@ -0,0 +1,57 @@
...
+      add(page);
+    }
+    _currentPages.remove(page);
+    _currentPages.add(page);
+    _fixPageOrder();
+    activePage?.deactivate();
+    page.activate();
+  }
+
+  void popPage() {
```
spydon, line 32:
```suggestion
  void pop() {
```
If we want to follow the same style as the flutter navigator

packages/flame/lib/src/page_render_effect.dart
```diff
@@ -0,0 +1,35 @@
+import 'dart:ui';
+
+abstract class PageRenderEffect {
+  Canvas preprocessCanvas(Canvas canvas);
+  void postprocessCanvas(Canvas canvas);
+}
+
+class PaintRenderEffect extends PageRenderEffect {
```
spydon, line 8: Maybe this should be called something else than `Effect`, to not confuse it with an effect from the effects system.

spydon (commented):
Looks very good, looking forward to using it!
I think the example can be simplified quite a bit by using already existing Flame components though, to keep it more on point.

doc/flame/examples/lib/navigator.dart
st-pasha: Which one would I use though?
We have `SpriteButtonComponent`, which requires two sprites that I don't have.
We have `ButtonComponent` which uses two `PositionComponent`s, but then those still need to be implemented, so it's pretty much the same amount of code.

Plus, both of those buttons use the `Tappable` interface, which makes it hard to prevent tap events from reaching the button if there is another page on top of it.

packages/flame/lib/src/page_render_effect.dart
st-pasha: Agree, though I'm a bit out of ideas here...

packages/flame/lib/src/components/navigator.dart
st-pasha: Makes sense, however, I wanted to make an additional extension here, the `pageFactory` parameter. This would handle dynamically generated page names, for example `"level/357"`.

My thinking was that if you say `navigator.pushPage("level/357")`, then the navigator would split the name into two parts and then invoke `factories["level"].generatePage("357")` to create the page with the given name. But I don't see how this functionality could work with enums.

doc/flame/examples/lib/navigator.dart
st-pasha: Currently, `Page` is the component whose job is to build the actual contents of the page when it is shown. So, I wanted to make a distinction that `SomethingPage` is a subclass of `Page`, and its content would then be `SomethingPageImpl`, or `SomethingScreen`, or `SomethingComponent`, or `SomethingPageContent`, etc.

The `Page` is an equivalent to Flutter's `Route`. Perhaps if we renamed it into `Route` it would be less confusing?

doc/flame/examples/lib/navigator.dart
st-pasha: yep

doc/flame/examples/lib/navigator.dart
st-pasha: right, I keep forgetting about that syntax

doc/flame/examples/lib/navigator.dart
st-pasha: yep, now changed to TextComponent

doc/flame/navigator.md
st-pasha: Sounds like an interesting idea. Of course, overlays would have limited functionality because they're always displayed on top of the Flame canvas, but it could still be useful to the user.

packages/flame/lib/src/components/navigator.dart
st-pasha: I kinda find the `pushPage() / popPage()` pair more symmetric and balanced than `pushPage() / pop()`.

I don't think it would be confusing to the users: even if they try to write `pop()`, the autocomplete will tell them that it should be `popPage()`.

doc/flame/examples/lib/navigator.dart
st-pasha: Renamed Page into Route now.

doc/flame/navigator.md
st-pasha: I created issue #1762 for this

packages/flame/lib/src/components/navigator.dart
st-pasha: Now renamed into `pushRoute()` / `popRoute()`.

CodeDoctorDE:
Hello,
Is there an option to add a loading screen?
If you have a scene having many objects can be useful to have it.

doc/_sphinx/extensions/flutter_app.py
```diff
@@ -44,13 +44,16 @@ class FlutterAppDirective(SphinxDirective):
...
 
       :show: - a list of one or more run modes, which could include "widget",
-        "popup", and "code". Each of these modes produces a different output:
+        "popup", "code", and "infobox". Each of these modes produces a different
+        output:
           "widget" - an iframe shown directly inside the docs page;
           "popup" - a [Run] button which opens the app to (almost) fullscreen;
           "code" - a [Code] button which opens a popup with the code that was
               compiled.
+          "infobox" - the content will be displayed as an infobox floating on
```
erickzanardo, line 53: This changes of the docs should be extract in a separate PR

erickzanardo:
I realise this is in draft but I want to raise some discussion around this.

I like the idea of navigator inside Flame, but I wonder if this isn't falling a little bit into over engineering.

Flutter already provides a navigation system, it already provides a way to display dialogs, etc. We recommend users that build their UI using flutter, as flutter is a way more powerful tool for building UIs than Flame. So I am failing to see why someone would write dialogs, title screens, stage selection screens with Flame components rather than just going with Flutter.

With the content I am seeing in the docs, I fear that users will be "directed" to think that Flame can't be integrated inside a Flutter app, this is actually something that users always ask on our discord.

With that said, I still think we could have some sort of navigator for Flame as I see clear use cases for it, a fine example is a Pokémon/RPG like game, where you have the over world which is a top down game, and the battle screen, which is a turn based game. We could implement this with Flutter Widgets, but I definitely think it would be nicer and smoother if such transition is done directly on Flame, so a small navigation system, or component stacking system would be good here.

So my thinkings here are this:
 - We can still have a full blown navigation system on Flame, but we need a super thorough documentation that mentions flutter navigation system, how flame is just a widget, how we recommend using Flutter to build pages, widgets, navigation etc.
 - Simplifying this to be a simpler system, Like something that just stack components and allow them to be changed with nice transitions.

IMO the second option is the one I would go

spydon:
> I realise this is in draft but I want to raise some discussion around this.
> 
> I like the idea of navigator inside Flame, but I wonder if this isn't falling a little bit into over engineering.

Interesting, I thought the other way around, that it will be very nice to handle screens and scenes in the same way in Flame no matter if it is a component or an overlay with Flutter components and from the review I've done previously I don't think that it was over engineered and that the user facing API was very simplistic.

> With the content I am seeing in the docs, I fear that users will be "directed" to think that Flame can't be integrated inside a Flutter app, this is actually something that users always ask on our discord.

Is it? They usually ask the other way around right, whether a Flutter widget can be added to Flame?
 
>     * Simplifying this to be a simpler system, Like something that just stack components and allow them to be changed with nice transitions.

What is it that you'd want to remove from the current solution? The possibility to handle overlays from the navigator?

Agree that we should have good docs describing what the difference would be with this compared to using navigation from Flutter.

erickzanardo:
> Interesting, I thought the other way around, that it will be very nice to handle screens and scenes in the same way in Flame no matter if it is a component or an overlay with Flutter components and from the review I've done previously I don't think that it was over engineered and that the user facing API was very simplistic.

The over engineer thing for me is that we have implemented a full blown navigation system for Flame, in a way, it feels to me that we are reinventing the wheel, as this is already ready on Flutter and we are adding it to Flame.

> Is it? They usually ask the other way around right, whether a Flutter widget can be added to Flame?

Yeah, but that is part of the broad subject right? Where we end up explaining that you can't use widgets directly on Flame, but since Flame is a widget in the end, it can be composed into your Flutter App.

> What is it that you'd want to remove from the current solution? The possibility to handle overlays from the navigator?

I think we could have a simpler API actually, like, a StackComponent, where you can Push and pop it, and it would change between the components with nice transitions and such, and let the navigation thing to Flutter itself.

packages/flame/lib/src/page_render_effect.dart
spydon: `Filter` maybe?

erickzanardo:
> Flutter can't handle Flame "scenes", which is a concept that people ask all the time about and I think having scenes consistent with how you handle screens would be very nice for user experience instead of having to handle them separately. 

Yes! That is why on my original comment I said that I like the idea of having this concept, and even gave the example of the pokemon use case, what I don't like is to suggest the user to do UI on Flame instead of Flutter.

> Maybe, I've never seen a question that way around though and I don't think this will increase that confusion, because as soon as they have written GameWidget they should™️ know that they can use it in Flutter. 😄

I disagree, many people still asks about that even after writing their first game. I agree with you that it should be obvious, but I would still thread carefully on this. 

> Isn't that pretty much what this is currently (with the exception that it'll be able to handle overlays too)? Navigator is a component already.

I don't think so, looking at the Navigator component we have a lot of code managing routes, managing the concept of named routes, we also have a class to just represent a route, I am thinking on something more like this:

```dart
SceneManager( // I liked the name concept of Scene that you mentioned 😅 
  children: [
    Overworld(),
  ],
);
```

Then from the `Overworld` component one could (assuming it has the parent mixin): `parent.push(Battle())`, from the Battle, one could just call `parent.pop()`.

For scene transitions, the push and pop methods could receive an effect for example.

I think this API will cover most of the use cases we wold need on Flame and is waaaay more simple.

spydon:
> Yes! That is why on my original comment I said that I like the idea of having this concept, and even gave the example of the pokemon use case, what I don't like is to suggest the user to do UI on Flame instead of Flutter.

Maybe I misunderstood you then, but that only sounds like a documentation issue. :)

> I disagree, many people still asks about that even after writing their first game. I agree with you that it should be obvious, but I would still thread carefully on this.

I don't know what secret communication channel you have with our users, but I haven't seen this on discord at least. 😄 

> I don't think so, looking at the Navigator component we have a lot of code managing routes

It's less than 200 lines (without comments), I don't think that is a lot. 🤷 

> I think this API will cover most of the use cases we wold need on Flame and is waaaay more simple.

Let's skip the overlays discussion for now, since that isn't implemented in this PR anyways. 
But other than that, the suggested solution will only cover the most basic usecase right? What if you want to go directly from a battle to an already initialized inventory screen in the main game for example? The way you navigate to routes shouldn't have to be linear.

erickzanardo:
> Maybe I misunderstood you then, but that only sounds like a documentation issue. :)

I don't think it is only a documentation issue, only the existence of the navigation system on flame could cause some sort of confusion. But I definitely think that documentation is a way to mitigate it (or at least most of it). That is why I added that as an option on my suggestion on my original comment.

> I don't know what secret communication channel you have with our users, but I haven't seen this on discord at least. 😄
(About whether the GameWidget can be used in the Flutter widget tree that is)

I don't have a secrete communication channel 🧐 , the doubts I am mentioning is not specific about GameWidget in the flutter widget, but how Flutter widgets can be used alongside a FlameGame

> It's less than 200 lines (without comments), I don't think that is a lot. 🤷

That isn't a lot, but the whole system is more complicated right? You need to have a route, that route will create the component, etc, I am not saying it is a complicated code to ready, but rather a complicated system. Or even better, I don't even think this is hard to use or hard to understand, I just think that for our use case, it could be even more simpler.

> The way you navigate to routes shouldn't have to be linear.

Hmm, With the way I suggested, you don't need to have things linear, you could add components to the `SceneManager` on whatever order you like, didn't followed you on this one.

spydon:
> Hmm, With the way I suggested, you don't need to have things linear, you could add components to the `SceneManager` on whatever order you like, didn't followed you on this one.

In your suggestion you say that we should remove named routes right? 
Consider the following scenes; `world`, `battle` and `inventory`, where the `inventory` screen can be accessed from any route.
How would you go from the `battle` route to `inventory` when `world` was the last showed scene before `battle`? `inventory` has already been initialized in the usecase and should not have to be re-created.
What I mean is that with only push (without a named push) and pop you can only move linearly, which I guess is why the `Route` concept was introduced here.

erickzanardo:
> What I mean is that with only push (without a named push) and pop you can only move linearly, which I guess is why the Route concept was introduced here.

I only mentioned `pop` and `push` because that was enough for my exemplification. We could have a replace that would be enough for that maybe?

But we could keep the idea of named routes, I am not against that, but I think it could be simpler, and it could be one of the options. I will try to write a small example of this.

spydon:
> I only mentioned `pop` and `push` because that was enough for my exemplification. We could have a replace that would be enough for that maybe?

With replace only you'd have to add the `battle` scene again when you want to go back from the `inventory`.

> But we could keep the idea of named routes, I am not against that, but I think it could be simpler, and it could be one of the options. I will try to write a small example of this.

If we have named routes, then it is pretty much like the implementation in the PR but without the route class? 

I think the route class is a pretty good layer here, because then you can easily define filters and transitions for each route. Without the route class, one would have to pass in all that information when a new component is pushed as a route right?

st-pasha:
Wow, that's a big discussion! I will be able to read it carefully later tonight, but for now here's my general thinking about why I thought the Navigator is needed.

So, let's suppose I'm making a some kind of an RPG game. At the outset, I'm sure there will be the following top-level elements there:
- a loading screen (not sure why, but it seems all games have a loading screen);
- the "main menu" where you can start a new game, or check out settings/credits/etc;
- the character creation screen, where you select the class, give your character a name and choose their look;
- maybe some kind of a cut-scene;
- the main world, where the character is going to kill monster and do quests;
- some inventory UI;
- a dialogue with NPCs UI;
- a trade UI;
- the world map where I can see my current whereabouts;
- etc.

Now, that's already quite a lot to handle, so how can I handle all of it?
Let's say, the world map - how can I implement it? The first instinct is to make an overlay. But overlay displays a widget, and I feel that the map will have a complicated visual that I would really rather custom-render with the Flame's toolset. Ok then, I could make it into a component and then just mount on top of the world when needed. Alright, but now what if my character gets attacked by monsters while looking at the map? I want the world to pause while the map UI is open (and `pauseEngine` is not gonna work, because it will pause everything, including the map). One way to accomplish this is to have the world as a child of another component, and that other component will control the passage of time for the world.

What about all those other screens? Let's take the "main menu" screen. Sure, doing layout in Flutter is very powerful, but the screen is more than just about the layout. I'd want the menu screen to have consistent custom look-and-feel, different from what Flutter's MaterialApp offers. I may need to use my custom buttons, custom text renderers, sprinkle everything with lots of animation and effects. In short, I'd prefer even the main menu to be done with Flame, even if it means that I need to do the layout manually.

Now, I understand that there could be different opinions about this and different preferences. But does not force its way of organization on anybody -- the developers are free to either use it or not use it. We'd definitely want to have good documentation outlining various approaches for organizing the high-level structure of the game, perhaps mentioning the pros and cons of each one.

wolfenrain:
> Alright, while I was thinking about the example, I think I finally managed to think this through better.
> 
> > If we have named routes, then it is pretty much like the implementation in the PR but without the route class?
> 
> True, so here is my thoughts:
> 
> * We could change `Navigator` and `Route` to `SceneManager` and `Scene`. I feel that most of my issues if that how this would be confusing to users on when the user would use Flutter Navigation of the Flame ones. I think that Scene describes better on the flame context, Route reminds me of Page, and that is not really the case on the game.
> * We can keep named routes, I got your point and it makes sense, it is a nice idea to have them in a way that they can be kept alive. But I also like the idea to push "fresh" created components in the stack, so we could have a pushNamed, as well as push.
> * On documentation we should make it clear about the difference between Flame's Scene Manager and Flutter's navigator, also I wouldn't use Title screen and things like that as an example for a scene on the docs, that fits better as a Flutter Page.
> * (Optional but nicely to have): Named scenes could be a generic type instead on string, so people can use enums.
> 
> Thoughts?

Reworking this into a form of scene manager totally makes sense! That is quite often what you see in other game engines as well and in the context of what Pasha just mentioned that makes the most sense for me as well.

st-pasha:
After reading the discussion more carefully, I'll try to summarize the points and suggestions raised:

- **There is a general consensus that this is a useful feature to have**, though few questions / points of contention remain.

- **Are we reinventing the wheel here?**

  Yes, we are, though it's not that hard since we have a first-hand account of how this wheel is supposed to look like :) But really, there's nothing wrong with Flutter's `Navigator`, except that it cannot navigate within the Flame's environment. So, we need our own navigator to do that.

- **Would the users be confused by this new Navigator?**

  There is a certain degree of truth to this. However, there is a second side to the same coin: once you realize that Flame has its own Navigator that is almost the same as the Flutter's Navigator, you also realize that you don't have to learn this new system because the knowledge of the Flutter's Navigator is perfectly transferable here. (Well, at least to people who know Flutter). Which is why, I think, it is good that we follow the Flutter's example here instead of trying to mimic Unity or UnrealEngine.

  Still, a case could be made for renaming `Navigator` -> `NavigatorComponent`, so that the users would be able to easier distinguish the two navigators (esp. if they are used at the same time).

- **What about renaming Route -> Scene?**

  To me, this would feel more confusing. To me, the word "scene" just sounds very BIG. When I hear "scene", I'm thinking along the lines of Scene = Page = Screen = World. However, the Route object is supposed to be much smaller: its job is just to build the corresponding scene/page, and possibly configure the transition affordances or some special effects. There is also the `Scene` object in Flutter, but it's related to canvas rendering.

- **Unnamed routes**

  At first, I didn't think these would be needed since we can just give names to all the routes. But then, thinking of Erick's example, I kinda see their usefulness too. For example, if a character engages in a battle with a monster and we want to bring up the battle UI, then that UI needs to be passed the argument which is the monster properties. One way to accomplish this would be to create the battle route on the fly and pass the `monster` argument to it. But there could be other solutions too, need to check what Flutter is doing in cases like this.

- **Documentation**

  I feel there's a consensus that we ought to have a good documentation about which other navigation systems can be used and how, and also the comparison of their relative advantages and drawbacks.

erickzanardo:
> you don't have to learn this new system because the knowledge of the Flutter's Navigator is perfectly transferable here ... Which is why, I think, it is good that we follow the Flutter's example here instead of trying to mimic Unity or UnrealEngine.

The system will be the same, just with a different name, so knowledge is still applicable, and we can mitigate any possible confusions with a good doc explaining that the scene system is heavily inspired by Flutter's navigation system.

The point of the Scene term is not to mimic Unity or Unreal, the point is that this is a term more suitable for a game engine. Godot also uses this term.

> I'm thinking along the lines of Scene = Page = Screen = World.

The same goes for route imo, Route = Page = Screen = world. But for me, Screen we have one less ting actually, because I don't relate Scene to Page, btw Page is something that I find difficult to relate to game engine at all.

So I still feel that Scene would be a better name for this system. I can see that it can cause some confusions for Flutter devs (which can be mitigated with docs) coming to the engine but on the other hands, game devs coming to Flame from other engines may feel less confused and none of them will ever confuse it with the Flutter's navigation.

st-pasha:
I'm not against the `Scene` term per se, it's a great term. But AFAIK, in all game engine the term **Scene** refers to a large arrangements of game objects. To be more concrete, this is how the Navigator is used in the example in this PR:
```dart
      navigator = Navigator(
        routes: {
          'splash': Route(builder: SplashScreenPage.new),
          'home': Route(builder: StartPage.new),
          'level1': Route(builder: Level1Page.new),
          'level2': Route(builder: Level2Page.new),
          'pause': PauseRoute(),
        },
        initialRoute: 'splash',
      ),
    );
```

So, it's reasonable to say that the start page is a scene, or that a level page is a scene. Thus, we could rename `StartPage` -> `StartScene`, `Level1Page` -> `Level1Scene`, and so on. However, a **Route** is not a scene: it's a descriptor of how to get to the scene.

Note also that I didn't want to store the destination pages (or Scenes) within the navigator, because it would have caused all those pages (scenes) to be instantiated when the Navigator is created, which is potentially costly. So instead we just store a list of routes -- sort of like an address book -- allowing each destination page (scene) to be instantiated only when and if it is actually needed.

Maybe there is a better term than "Route", but I don't think it's the "Scene".

doc/_sphinx/extensions/flutter_app.py
st-pasha: Done: see #1771

packages/flame/lib/src/page_render_effect.dart
st-pasha: Renamed into `Decorator`.

packages/flame/lib/src/components/navigator.dart
st-pasha: Now renamed into `pushNamed` / `pop` :)

doc/flame/examples/lib/navigator.dart
spydon: Fair enough

doc/flame/examples/lib/navigator.dart
spydon: You forgot to rename the components here though, because I guess now the components should be have just the `Page` suffix and not the `PageImpl` one?

st-pasha: They are called `Level1Page` and `Level2Page` now, maybe you're looking at the older version of code?

spydon: I was indeed, must have continued on an old tab. :sweat_smile:

spydon:
> Maybe there is a better term than "Route", but I don't think it's the "Scene".

I think that Route is fine, and I think that we can probably introduce a scene concept later (which basically is just an empty component at the moment).

alestiago:
I think this would be highly beneficial from #1799 . Where the `Navigator` would be an `InheritedComponent`. Similar to how Flutter deals with navigation.

packages/flame/lib/src/components/navigator_component.dart
```diff
@@ -0,0 +1,206 @@
...
+/// Routes that are on the stack are mounted as components. When a route is
+/// popped, it is removed from the stack and unmounted. Routes can be either
+/// transparent or opaque. An opaque route prevents all routes below it from
+/// rendering, and also stops pointer events. In addition, routes are able to
+/// stop or slow down time for the pages that they control, or to apply visual
+/// effects to those pages.
+class NavigatorComponent extends Component {
+  NavigatorComponent({
+    required this.initialRoute,
+    required Map<String, Route> routes,
```
spydon, line 32: What was the verdict of making this `Map<T, Route>` so that enums for example can be used too?

spydon (approved):
I think that it looks good!
I'm thinking whether it is worth having generics on the `NavigatorComponent` so that it can be used in a more safe way than just plain strings. The cons that I see of that though is:
 -  It might be harder for the user to get into using it at first
  - That generic would also spread to the `Route`.

packages/flame/lib/src/components/navigator_component.dart
st-pasha: The advantage of string ids is that they are an open-ended set: the route names can be generated at runtime. For example, if the game includes some kind of downloadable levels, then it needs a way to invoke those levels from the Navigator. For this purpose we have `routeFactories` and `pushRoute()`. It wouldn't be possible to implement route factories with generic `T` (unless we specify that those method work only when `T = string`).

Also, it would be harder to create routes on the fly: you'd have to write `Route<MyNavigatorIds>(...)` instead of `Route(...)`.

At the same time, the benefit that the enums would bring into the Navigator is better type safety: it would be impossible to accidentally misspell the name of a route. However, the same safety could be achieved by simply declaring a number of static strings for each route and using those strings as names. This would work as a "manual enum" without sacrificing any of the other Navigator's functionality.

spydon: The generics can't be omitted since they don't have the same base class, but setting it to `String` would work, but the problem is that it would have to be done everywhere all the time if we don't have an implementation for just strings like`class StringRoute extends Route<String>`.

It would be nice to allow other types than string, I find manual enums a bit dirty. But I also don't want to sacrifice the beginner's user experience in favor of this.

Wdyt of allowing `T` in the base class, but having ready implementations for `String` as key, so that the user doesn't have to specify it everywhere?

erickzanardo: > The generics can't be omitted since they don't have the same base class

Hm, I am not sure about that, I just tried this on dart pad to confirm and it worked nice, `value` is dynamic because no generics was informed:

```dart
class Value<T> {
  T? value;
}

void main() {
  final v = Value();
  v.value = 'A';
  print(v.value);
}
```

Am I missing something obvious?

spydon: Sure, but then it wouldn't be typed at all, the worst of all worlds :smile:

st-pasha:
I think this PR has no action items left, @spydon WDYT?

spydon:
> I think this PR has no action items left, @spydon WDYT?

If I remember correctly, there was a discussion about adding generics to the routes so that enums for example could be used, and then have a default implementation that just handled strings?

st-pasha:
> If I remember correctly, there was a discussion about adding generics to the routes so that enums for example could be used, and then have a default implementation that just handled strings?

My understanding was that because enums/strings have no common super-type, then using generics here would be counter-productive as it will lead for less type-safety (`dynamic` as the inferred type).

An additional problem with generic types is that it wouldn't be able to accommodate some of the existing features like generated route names.

spydon:
> My understanding was that because enums/strings have no common super-type, then using generics here would be counter-productive as it will lead for less type-safety (`dynamic` as the inferred type).
> 
> An additional problem with generic types is that it wouldn't be able to accommodate some of the existing features like generated route names.

Right, that was it, now I remember.

spydon (approved):

spydon:
@st-pasha is there a plan for how to support routes opening overlays? I feel that is pretty important so that users aren't locked in to create menus etc as flame components.

st-pasha:
Yeah, we have issue #1762 for that.

erickzanardo (approved):
I am still not sure about the `Navigator` name, internally me and @spydon discussed a little bit and the `RouterComponent` name could also be a good fit and I think I like it more.

Either way, gonna leave my approval in here.

spydon:
> I am still not sure about the `Navigator` name, internally me and @spydon discussed a little bit and the `RouterComponent` name could also be a good fit and I think I like it more.
> 
> Either way, gonna leave my approval in here.

For me both `RouterComponent` and `NavigatorComponent` are fine, but if you prefer `RouterComponent` maybe we should go with that then so that we can get this merged. :)

PR #1766, opened 2022-06-30
chore: updating dashbook on examples

erickzanardo:
# Description

Updates the Dashbook version on our examples and update  an example to use its new action API.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/examples/pubspec.yaml
+++ b/examples/pubspec.yaml
@@ -10,7 +10,7 @@ environment:
   flutter: ">=3.0.0"
 
 dependencies:
-  dashbook: 0.1.6
+  dashbook: 0.1.8
   flame: ^1.2.0
   flame_audio: ^1.2.0
   flame_forge2d: ^0.11.0


examples/lib/stories/animations/animation_group_example.dart
```diff
@@ -53,21 +60,6 @@ class AnimationGroupExample extends FlameGame with TapDetector {
     add(robot);
   }
 
-  @override
-  void onTapDown(_) {
```
spydon, line 57: Isn't this example better to keep the tappable, since that is more in line with how our users will use this feature?

erickzanardo: I agree and disagree hahaha

I think from a mobile game perspective, indeed a tap is how a running action like this would take place.

But given this is mostly accessed on desktop browser, I think the actions api fits better.

I don't mind reverting the example of you disagree though. I just wanted to leave one story with the new api being used as an example

erickzanardo: > But people won't be using dashbook at least,

Won't they? they see our example from Dashbook, maybe I am not following 😅 

> But there could be buttons that are in the example itself if you want buttons, widgets or not is up to you.

The whole point of the actions API on Dashbook is so you don't need write those buttons manually. But if you don't see value for it on this example, I will just revert the code on the example. Maybe in the future we come up with an example that can take advantage of this new API.

erickzanardo: Reverted the example changes and kept only the dashbook update, take another look please.

spydon (approved):

examples/lib/stories/animations/animation_group_example.dart
spydon: > Won't they? they see our example from Dashbook, maybe I am not following 😅
>
> The whole point of the actions API on Dashbook is so you don't need write those buttons manually. But if you don't see value for it on this example, I will just revert the code on the example. Maybe in the future we come up with an example that can take advantage of this new API.

In their games I mean, so when there is code that is valuable as examples I don't think it should be simplified by dashbook. An example where the dashbook simplifications work very well is with the nine tile widget for example. :)

Issue #1765, opened 2022-06-30
SpriteAnimationWidget test sometimes fails


st-pasha:
# Current bug behaviour
Observed the following error reported in the logs:
```
  ● SpriteAnimationWidget can safely change animation parameter packages/flame/test/widgets/sprite_animation_widget_test.dart:70:5
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following TestFailure was thrown running a test:
Expected: a value not equal to <0>
  Actual: <0.0>
   Which: is not a value not equal to <0>
When the exception was thrown, this was the stack:
#4      main.<anonymous closure>.<anonymous closure> (file:///home/runner/work/flame/flame/packages/flame/test/widgets/sprite_animation_widget_test.dart:93:7)
<asynchronous suspension>
<asynchronous suspension>
(elided one frame from package:stack_trace)
This was caught by the test expectation on the following line:
  file:///home/runner/work/flame/flame/packages/flame/test/widgets/sprite_animation_widget_test.dart line 93
The test description was:
  can safely change animation parameter
════════════════════════════════════════════════════════════════════════════════════════════════════
```

I presume the test is flaky, because this error appeared only once so far.

The error might also be related to PR #1760, given that `SpriteWidget` was modified recently in that PR. Needs further investigation.

st-pasha:
Running this test in a loop shows that it fails approximately 22% of the times.

Hwan-seok:
I did some debugging and I might think the following has some issue.

https://github.com/flame-engine/flame/blob/82f75fcb57c8185a7138ee6ceb9082a418099df8/packages/flame/lib/src/widgets/animation_widget.dart#L148-L154

`now` and `_lastUpdate` has a range of [0,999] and I found that there could be a situation `_lastUpdate` is bigger than `now` when `_lastUpdate` is close to 999 and `now` is close to 0.
![스크린샷 2022-07-01 오전 5 46 13](https://user-images.githubusercontent.com/38072762/176775391-4562051a-75e6-4fd2-a9ea-72e0ebfd8a8b.png)
Then the dt could be 0. so the test will fail.

It indicates that the widget animation was sometime not updated.

I might think the test is fine but `now` and `_lastUpdate` is a little bit need to be changed.
it should be some kind of epoch thing.

PR #1767, opened 2022-06-30
fix: Change AnimationWidget to use epoch time 

Hwan-seok:
# Description
Currently, `AnimationWidget` is using a millisecond of the current second as elapsed time since the last frame.
It is dangerous because when the last frame occurs at end of the second and the current frame at the start of the second.
```
e.g. 
 Last frame: 08:00:00.990
 Current frame: 08:00:01.006
 ```
When the above situation occurs, `dt` became 0 and the SpriteAnimation does not update.

Changed to using elapsed time as subtraction of `microsecondsSinceEpoch` between frames.

## Checklist

## Breaking Change

## Related Issues
Fixes https://github.com/flame-engine/flame/issues/1765

### DIFF
--- a/packages/flame/lib/src/widgets/animation_widget.dart
+++ b/packages/flame/lib/src/widgets/animation_widget.dart
@@ -1,5 +1,4 @@
 import 'dart:async';
-import 'dart:math';
 
 import 'package:flame/src/anchor.dart';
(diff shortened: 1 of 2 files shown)


spydon (approved):
Good catch, that was clearly a bug!

erickzanardo (approved):

PR #1769, opened 2022-07-01
feat: Adding ClipComponent

erickzanardo:
# Description

Adds a new component called `ClipComponent` that clips the canvas area based on its size and shape.

## Checklist

## Breaking Change

## Related Issues


doc/flame/components.md
```diff
@@ -988,6 +988,22 @@ Check the example app
 [custom_painter_component](https://github.com/flame-engine/flame/blob/main/examples/lib/stories/widgets/custom_painter_example.dart)
 for details on how to use it.
 
+## ClipComponent
+
+A `ClipComponent` is a component that will clip the canvas to its size and shape. This means that
+if the any child of a `ClipComponent` is beign rendered outside of its bondaries, the part that is
```
spydon, line 994:
```suggestion
if the component itself or any child of the `ClipComponent` renders outside of the `ClipComponent`'s boundaries, the part that is
```

doc/flame/components.md
```diff
@@ -988,6 +988,22 @@ Check the example app
...
+not inside the area will not be shown.
+
+There are three types of shapes that a `ClipComponent` can take form:
```
spydon, line 997:
```suggestion
There are three types of shapes that a `ClipComponent` can use to define the area that shoud be clipped:
```

doc/flame/components.md
```diff
@@ -988,6 +988,22 @@ Check the example app
...
+
+ - `ClipComponent.rect`: Clips the area in the form a rectangle based on its size.
+ - `ClipComponent.circle`: Clips the area in the form of a circle based on its size.
+ - `ClipComponent.poylon`:  Clips the area in the form of a polygon based on the points received
+on the constructor.
```
spydon, line 1002:
```suggestion
in the constructor.
```

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,92 @@
...
+  polygon,
+}
+
+/// {@template clip_component}
+/// A component that will clip its content.
+/// {@endtemplate}
+class ClipComponent extends PositionComponent {
+  /// {@macro clip_component}
+  ///
+  /// Creates a circle clip based on the components size.
```
spydon, line 17:
```suggestion
  /// Creates a circle clip based on the [size].
```

spydon (approved):
Looks good! Just some small grammar stuff.

st-pasha:
Related issue: #1274

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,92 @@
...
+        _path = Path()..addOval(Rect.fromLTWH(0, 0, size.x, size.y));
+        break;
+      case _ClipType.polygon:
+        // nothing to do here, path is set in the constructor.
+        break;
+    }
+  }
+
+  @override
+  void render(Canvas canvas) => canvas.clipPath(_path);
```
st-pasha, line 91: I think this would affect all subsequent canvas drawing commands, including those at the sibling and higher level.

A more suitable approach would be to override the `renderTree` method, and then wrap the clipping into `canvas.save`/`canvas.restore`.

We might want to have a golden test for this scenario too, just in case.

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,92 @@
...
+}
```
st-pasha, line 92: There are also methods `parentToLocal` and `localToParent` that need to be overridden to return `null` for those points that fall outside of the clipping mask.
This would ensure, for example, that the pointer events would not work on the clipped parts.

st-pasha: Note: `Shape.containsPoint` can be used for this.

packages/flame/lib/src/components/clip_component.dart
erickzanardo: I thought that would be the case too, but as far as I checked, it didn't affected siblings or higher level.

I will double check though, and add a test for this case.

doc/flame/components.md
```diff
@@ -988,6 +988,22 @@ Check the example app
 [custom_painter_component](https://github.com/flame-engine/flame/blob/main/examples/lib/stories/widgets/custom_painter_example.dart)
 for details on how to use it.
 
+## ClipComponent
+
+A `ClipComponent` is a component that will clip the canvas to its size and shape. This means that
+if the component itself or any child of the `ClipComponent` renders outside of the `ClipComponent`'s boundaries, the part that is
```
spydon, line 994: This line became too long ofter my suggestions. :sweat_smile:

examples/lib/stories/components/clip_component_example.dart
```diff
@@ -0,0 +1,81 @@
...
+    descendants().whereType<PositionComponent>().forEach((component) {
+      component.size = component.size * 1.1;
+    });
+  }
+
+  void decreaseSize() {
+    descendants().whereType<PositionComponent>().forEach((component) {
+      component.size = component.size * .9;
+    });
+  }
```
st-pasha, line 51: These functions look unused, right?

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,92 @@
...
+  /// Creates a circle clip based on the [size].
+  ClipComponent.circle({
+    super.position,
+    super.size,
+    super.scale,
+    super.angle,
+    super.anchor,
+    super.children,
+    super.priority,
+  }) : _clipType = _ClipType.circle;
```
alestiago, line 26: Instead of this factory methods I would expect the `ClipComponent` to receive a `Clipper`. The `Clipper` would have the information on how to draw the path to be clipped. 

In this scenario, we would subclass `Clipper` and implement (example names)L`CircleClipper` `PolygonClipper` `RectClipper`.

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,92 @@
...
+    final firstPoint = _points.removeAt(0);
+
+    _path = _points.fold(Path()..moveTo(firstPoint.x, firstPoint.y),
+        (previousValue, value) {
+      return previousValue..lineTo(value.x, value.y);
+    });
+  }
+
+  late Path _path;
+  final _ClipType _clipType;
```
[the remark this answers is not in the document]
alestiago: https://github.com/flame-engine/flame/pull/1769/files#r922651102

packages/flame/lib/src/components/clip_component.dart
st-pasha: Could you please elaborate what would be the advantages of this approach compared to using the existing `Shape` classes?

alestiago: @st-pasha I haven't look at Flame Shape classes in a while. The above is what I would expect if I was coming from the Flutter world.

Do Shape classes allow the user to specify a custom shape?

st-pasha: There are several shape classes available: circle, rectangle, rounded rectangle, and polygon. More can be added in the future.

alestiago: If the user can easily subclass a `ShapeComponent` and have the codebase live on their end, then sounds like a good idea to use `ShapeComponent` subclasses. Is there documentation on how to properly subclass ShapeComponent? For example, documentation that details how to create a StarComponent.

All I could find is: https://docs.flame-engine.org/1.2.0/flame/components.html?highlight=shapecomponent#shapecomponents

examples/lib/stories/components/clip_component_example.dart
erickzanardo: Yeah, good catch, I was going to use them but decided otherwise and forgot to remove them.

doc/flame/components.md
```diff
@@ -988,6 +988,22 @@ Check the example app
...
+
+A `ClipComponent` is a component that will clip the canvas to its size and shape. This means that
+if the component itself or any child of the `ClipComponent` renders outside of the
+`ClipComponent`'s boundaries, the part that is not inside the area will not be shown.
+
+There are three types of shapes that a `ClipComponent` can use to define the area that shoud be clipped:
+
+ - `ClipComponent.rect`: Clips the area in the form a rectangle based on its size.
+ - `ClipComponent.circle`: Clips the area in the form of a circle based on its size.
+ - `ClipComponent.poylon`:  Clips the area in the form of a polygon based on the points received
```
st-pasha, line 1001:
```suggestion
 - `ClipComponent.polygon`:  Clips the area in the form of a polygon based on the points received
```

packages/flame/lib/src/components/clip_component.dart
alestiago: Any chance you can plug the Shape class code URL in here?

Is this Shape class you're referencing from flame?

spydon: I'm on my phone, just search for `shape.dart` and you should find it. (Better to ask these questions on discord so that the PR threads with the discussions don't become polluted)

alestiago: No worries @spydon are you referring to the shape in the [experimental](https://github.com/flame-engine/flame/tree/main/packages/flame/lib/src/experimental/geometry/shapes) folder?

I think we have a naming conflict with `Shape` in forge2d. So perhaps users will not have the best developer experience since they will require to "show" or "hide" or "as" imports when using flame and forge2d.

Anyways, I think using Flame's experimental Shape class is useful however I think it contains more data than is actually needed for the clipping; such as aabb. 

In addition, I don't think it would be simple, from the user POV to subclass shape or implement their own custom clipper. I believe if we go down the `Shape` road I would say the Flutter framework does a better job at exposing a simpler API to clip the canvas.

I hope my two cents help here.

st-pasha: The problem though is that `Path` is very opaque, and cannot be manipulated easily. For example, if you wanted to change the shape dynamically - there isn't any way to do it. With `Shape` you can do it though: change the size of a rectangle, the radius of a rounded rectangle, or positions of individual points of a polygon.

It would be great if there was a reliable way to inspect the structure of a `Path` and modify it (or even convert `Path` into a `Shape`), but there doesn't seem to  be such a way.

alestiago: @st-pasha if the ClipComponent only required a path. Wouldn't the users be able to use the existing experimental shape classes to do all those manipulations and then use `asPath()`?

If instead, the developer wants to provide the `Shape` to later on inspect or manipulate the structure they could subclass `ClipComponent` and have a constructor that requires a `Shape` instead of a Path and then pass `Shape.asPath` to super?

Would the above solve the issue you were describing?

alestiago: Only if it functionally equivalent and provides a good API. If we go all the way with shapes it would be more difficult to provide a custom path. 

We could define a .fromShape in ClipComponent constructor that does exactly what we described above.

renancaraujo (commented):
I like the idea and the implemmentation looks good.

The only thing is that it is hard to trigger a path reconstruction if after load/mount, the user requires to change the shape of the clipped area (eg: update the size of the clipped rectangle given a game constraint)

erickzanardo:
@alestiago @st-pasha @renancaraujo @spydon here is what I am going to do. I will keep the `rect`, `circle` and `polygon` constructors as they are a nice and friendlier API than to just receive a `Shape`.

I will refactor so this uses a shape internally too, so I can easily implement the `containsPoint` method.

I will not include a constructor that receive a Shape yet though, since that class is on the experimental package. But since it will be used internally, it will be super simple to add that constructor once it gets out of the experimental state.

packages/flame/lib/src/components/clip_component.dart
st-pasha: A star is a `Polygon`, so that's easy. But for a more general case we ought to implement a `Polycurve` `Shape` which would be comprised of Bezier curves.
Note that with a Polygon (or Polycurve) you'd be able to manipulate individual nodes of the shape, allowing a significant degree of dynamic flexibility.

st-pasha:
> it will be super simple to add that constructor once it gets out of the experimental state.

I guess we never actually discussed about when to move things out of experimental. Should it be after some future release? After a certain features that are currently lacking are implemented? Any other trigger?

erickzanardo:
> I guess we never actually discussed about when to move things out of experimental. Should it be after some future release? After a certain features that are currently lacking are implemented? Any other trigger?

I think it can vary depending on the complexity of the feature. But we should definitely discuss on some guidelines, maybe it is worth to open a discussion or an issue so we can discuss that.

packages/flame/lib/src/components/clip_component.dart
alestiago: A star was just a symbolic example. Consider instead a round edges star. 

How would a user implement a clip for a rounded edged star? Is there documentation for it? Is it just not supported with this current API? Are there examples the developer can read from to implement a custom shape?

doc/flame/components.md
```diff
@@ -989,6 +989,22 @@ Check the example app
 [custom_painter_component](https://github.com/flame-engine/flame/blob/main/examples/lib/stories/widgets/custom_painter_example.dart)
 for details on how to use it.
 
+## ClipComponent
+
+A `ClipComponent` is a component that will clip the canvas to its size and shape. This means that
+if the component itself or any child of the `ClipComponent` renders outside of the
+`ClipComponent`'s boundaries, the part that is not inside the area will not be shown.
+
+There are three types of shapes that a `ClipComponent` can use to define the area that shoud be clipped:
```
spydon, line 998: Too long line

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,132 @@
...
+import 'package:flame/components.dart';
+import 'package:flame/experimental.dart';
+
+/// {@template clip_component}
+/// A component that will clip its content.
+/// {@endtemplate}
+abstract class ClipComponent extends PositionComponent {
+  /// {@macro clip_component}
+  ///
+  /// Clips the canvas based its shape and its size.
```
spydon, line 12:
```suggestion
  /// Clips the canvas based its shape and size.
```

spydon (approved):
Just some small grammar stuff otherwise lgtm!

spydon (approved):

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,132 @@
...
+
+  @override
+  bool containsLocalPoint(Vector2 point) {
+    return _shape.containsPoint(point);
+  }
+}
+
+/// {@template circle_clip_component}
+///
+/// A component that will clip its content to a circle.
```
alestiago, line 55: endtemplate?

doc/flame/components.md
```diff
@@ -989,6 +989,23 @@ Check the example app
...
+## ClipComponent
+
+A `ClipComponent` is a component that will clip the canvas to its size and shape. This means that
+if the component itself or any child of the `ClipComponent` renders outside of the
+`ClipComponent`'s boundaries, the part that is not inside the area will not be shown.
+
+There are three types of shapes that a `ClipComponent` can use to define the area that shoud be
+clipped:
+
+ - `ClipComponent.rect`: Clips the area in the form a rectangle based on its size.
```
alestiago, line 1001: I think this requires to be updated

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,132 @@
...
+    super.priority,
+  });
+
+  late Path _path;
+  late Shape _shape;
+
+  @override
+  Future<void> onLoad() async {
+    _prepare();
+    size.addListener(_prepare);
```
alestiago, line 29: I need to look into how addListener on Size works. Does it need to be disposed?

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,132 @@
...
+
+  @override
+  Shape _buildShape() {
+    return Rectangle.fromRect(size.toRect());
+  }
+}
+
+/// {@template polygon_clip_component}
+///
+/// A component that will clip its content to a polygon.
```
alestiago, line 101: endtemplate?

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,132 @@
...
+    super.size,
+    super.scale,
+    super.angle,
+    super.anchor,
+    super.children,
+    super.priority,
+  });
+
+  @override
+  Shape _buildShape() {
```
alestiago, line 71: Why is this preferred over having the shape be passed to the ClipComponent constructor?

alestiago (commented):
Looking way better!

packages/flame/lib/src/components/clip_component.dart
erickzanardo: No needed, this is subscribing to its own size, and it should live until the component live, so when the component is removed and garbage collected, this will go along

packages/flame/lib/src/components/clip_component.dart
erickzanardo: We need to recreate the shape based on size of the component, otherwise the user would need to pass the Shape and the Size, and this would be confusing.

alestiago: Would a parameter named shapeBuilder that receives a Size suffice?

packages/flame/lib/src/components/clip_component.dart
alestiago: I see very clever and simple implementation. Love it!

packages/flame/lib/src/components/clip_component.dart
erickzanardo: A shape builder is not a bad idea, but I would think of it as a way of giving more freedom to the user to define their own way of building the shapes. But I would create `.builder` constructor to it and keep the other constructors as requiring the user to define the shape builder always would not be good API IMO.

alestiago: A .builder constructor is very interesting. I like where this ja heading.

Out of curiosity, under what circumstances do you think the shapeBuilder will fail to be a "good" API?

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,132 @@
...
+  /// Clips the canvas based its shape and size.
+  ClipComponent({
+    super.position,
+    super.size,
+    super.scale,
+    super.angle,
+    super.anchor,
+    super.children,
+    super.priority,
+  });
```
st-pasha, line 21: I guess an alternative approach here could be to add the `shapeBuilder` function as a parameter (`Shape Function()`), and then declare 3 factory constructors `ClipComponent.circle`, `ClipComponent.rectangle` and `ClipComponent.polygon` instead of separate components for every shape.

PROs: (1) easier to create clip components with custom shape - no need to declare a new class, (2) easier to document a single component than several components, (3) more flexibility to change the shape at runtime.

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,132 @@
...
+
+  @override
+  Future<void> onLoad() async {
+    _prepare();
+    size.addListener(_prepare);
+  }
+
+  Shape _buildShape();
+
+  void _prepare() {
```
st-pasha, line 34: This could be public, allowing the shape to be modified by an external force (such as an effect or a tap callback).

packages/flame/lib/src/components/clip_component.dart
erickzanardo: Imagine requiring the user to write this every time they want a circle:

```dart
ClipComponent(builder: (size) => ..., size: Vector2.all(50), ...);
```

When they could instead do:

```dart
ClipComponent.circle(size: Vector2.all(50), ...);
```

But with a `builder` factory, we can have the best of both things.

packages/flame/lib/src/components/clip_component.dart
erickzanardo: Yep, this is what I will do following the discussion I am having with @alestiago on a thread above.

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -0,0 +1,132 @@
...
+  late Path _path;
+  late Shape _shape;
+
+  @override
+  Future<void> onLoad() async {
+    _prepare();
+    size.addListener(_prepare);
+  }
+
+  Shape _buildShape();
```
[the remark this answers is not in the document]
erickzanardo: Will do the builder factory instead.

packages/flame/lib/src/components/clip_component.dart
alestiago: The above is exactly what I was referring to and would like. Are we happy with it?

So the API would be:
```dart
ClipComponent(builder: (size) => ..., size: Vector2.all(50), ...);
ClipComponent.circle(size: Vector2.all(50), ...);
ClipComponent.polygon(size: Vector2.all(50), ...);
...
```

erickzanardo: Yep, that is what I will implement.

packages/flame/lib/src/components/clip_component.dart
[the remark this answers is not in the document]
erickzanardo: size has a single instance internally https://github.com/flame-engine/flame/blob/main/packages/flame/lib/src/components/position_component.dart#L178

I don't see the point of exposing this implementation detail to the user, if the size never change, yes the listener is added in vein, but there is no harm in that since this is an internal thing.

packages/flame/lib/src/components/clip_component.dart
st-pasha: Oh, sorry I haven't noticed that thread.

packages/flame/lib/src/components/clip_component.dart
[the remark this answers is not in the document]
st-pasha: ~~in vein~~ in vain.
A simpler approach could be to just override the `size` setter -- this way whenever the user changes the size, the shape gets updated too. This would be slightly less general, in case the user changes the size vector in-place, but on the other hand such practice ought to be discouraged anyway.

packages/flame/lib/src/components/clip_component.dart
alestiago: Looking forward to it! Thanks @erickzanardo for the work put into this!

erickzanardo: Updated! 🚀

packages/flame/lib/src/components/clip_component.dart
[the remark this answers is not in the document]
erickzanardo: I honestly don't think that the listener will impact anything, at most this will allocate something in the memory that may not be used.

But I would argue that this memory footprint will be marginal and this has no risk of leak, and the alternative is weird make it harder for the user, or to "hack" by overriding methods or other ways.

wolfenrain (approved):

Issue #1768, opened 2022-07-01
Separate docs site from docs content


spydon:
# What could be improved
Currently the docs site is built from a Flame version tag.
This means that if we improve the docs site or change its theme, or when we have a bug in it, it will remain like how it was when that tag was published forever when you look at that tag.

My suggestion is that the `_sphinx` directory is moved to the [flame-docs-site](https://github.com/flame-engine/flame-docs-site) repo and that the script for publishing is updated accordingly.

# Why should this be improved
To have consistency on the site and the possibility to fix bugs.

# Any risks?
Older versions could break with new functionality introduced for the site.

# More information

st-pasha:
Hmm, it's still possible to build the older branches using the latest doc theme: 
1. First, check out the latest flame repo
2. Copy its `_sphinx` folder into a temporary folder
3. Check out the branch that is needed
4. Replace its `_sphinx` folder with the one copied in step 2

Surely, this is few extra steps, but those are pretty trivial for a script.

spydon:
https://github.com/flame-engine/flame-docs-site/pull/5/

PR #1770, opened 2022-07-01
feat: Move the _sphinx directory to flame-docs-site

spydon:
# Description

This is an intermediate solution to have the docs site not being tied to Flame tags.
It just moves the `_sphinx` directory to the flame-docs-site repo and then copies it in place every time it generates the docs for a tag. 

https://github.com/flame-engine/flame-docs-site/pull/5

## Checklist

## Breaking Change

## Related Issues

Related to #1768, but we will probably be able to come up with a better solution in the long run.


erickzanardo (approved):

spydon:
> But how would this process work if this directory is moved to another repo?

It will work the same as today, but instead of using the `_sphinx` directory from the checked out tag, it will use the latest `_sphinx` directory for all versions. See this PR: https://github.com/flame-engine/flame-docs-site/pull/5

I think it makes sense to separate the styling and functionality of the docs site from the actual content, because today if we get a bug in the functionality of the docs site that tagged version will always have that bug.

st-pasha:
> It will work the same as today, but instead of using the _sphinx directory from the checked out tag, it will use the latest _sphinx directory for all versions. 

I still don't get it. Currently, in order to develop documentation (and be able to see the results), one would need to:
1. git clone flame repository (once)
2. cd into doc/_sphinx
3. pip install from the requirements file (once)
4. run make html
5. open the generated index.html file in the browser

But if the folder is in another repository, then what? How would the development process look like?

spydon:
> But if the folder is in another repository, then what? How would the development process look like?

We can have a melos target that sets all that up, including pulling down the other repo.
If you have a better idea for how to separate them I'm all ears! Currently we're stuck with our latest released version not having searchable docs and we have to have a way to fix bugs in the functionality of the docs which doesn't only apply to `main` and future versions.

st-pasha:
What do you think about the following:
1. No changes in this repository

Then, in the `flame-docs-site` repository we modify the `publish.sh` script to add the following steps:
```bash
tmp_sphinx_copy = $tmp_flame_src/../sphinx-copy

function main() {
  ...
  git clone https://github.com/flame-engine/flame.git $tmp_flame_src
  cd $tmp_flame_src

  # NEW:
  git checkout main
  cp -rf doc/_sphinx $tmp_sphinx_copy
  
  ...
}

function generate() {
  version=$1
  ...
  cd $tmp_flame_src
  git checkout -f $version
  
  # NEW
  rm -rf doc/_sphinx
  cp -rf $tmp_sphinx_copy doc/_sphinx

  ...
}
```

This way you always get the latest _sphinx when building a particular branch, but the development process in the flame repository is not affected.

spydon:
I guess we could also keep it where it is, but only use the `main` `_sphinx` directory for all tags when we publish, then the current workflow would be kept.

spydon:
> What do you think about the following:
> 
>     1. No changes in this repository
> 
> 
> Then, in the `flame-docs-site` repository we modify the `publish.sh` script to add the following steps:
> 
> ```shell
> tmp_sphinx_copy = $tmp_flame_src/../sphinx-copy
> 
> function main() {
>   ...
>   git clone https://github.com/flame-engine/flame.git $tmp_flame_src
>   cd $tmp_flame_src
> 
>   # NEW:
>   git checkout main
>   cp -rf doc/_sphinx $tmp_sphinx_copy
>   
>   ...
> }
> 
> function generate() {
>   version=$1
>   ...
>   cd $tmp_flame_src
>   git checkout -f $version
>   
>   # NEW
>   rm -rf doc/_sphinx
>   cp -rf $tmp_sphinx_copy doc/_sphinx
> 
>   ...
> }
> ```
> 
> This way you always get the latest _sphinx when building a particular branch, but the development process in the flame repository is not affected.

Haha, you must have posted that just seconds before me. Let's go with that then.

PR #1775, opened 2022-07-02
perf!: Game.images/assets are now same as Flame.images/assets by default

### LINKED ISSUE #1750: Merge game.images and Flame.images
# What could be improved

Multiple objects in Flame take an optional `Images? images` argument, which defaults to `Flame.images` if not provided. At the same time, many `Game.load...` methods default to the `Game.images` cache for storing images. The net result is that a user can quite easily end up in a situation where some image would get loaded twice: once into the game cache, and once into the global `Flame.cache`.

I suggest to simply make `Game.images` refer to the same cache as `Flame.images` by default, and to let the user supply their own cache if necessary.

# Any risks?

Not much, unless the user's code somehow relies on these two caches being distinct. In this case they would have to explicitly request a separate `Images` cache instance for their `Game`.

st-pasha:
# Description

The caches `Game.images` and `Game.assets` now coincide with the static caches `Flame.images` and `Flame.assets` by default.

This aims to avoid the situation where the same image (asset) may be loaded twice into two separate caches without the user noticing.

## Checklist

## Breaking Change

### Migration instructions

For the majority of users this change would not be breaking. However, it can affect those users who have multiple game instances simultaneously, where they use image files with same names but different content.

For such users the solution is to assign a private game instance when creating the game:
```dart
final game = MyGame() 
  ..images = Images()
  ..assets = AssetsCache();
```

## Related Issues

Closes #1750

### DIFF
--- a/packages/flame/lib/src/game/game.dart
+++ b/packages/flame/lib/src/game/game.dart
@@ -1,6 +1,7 @@
 import 'package:flame/cache.dart';
 import 'package:flame/components.dart';
 import 'package:flame/extensions.dart';
+import 'package:flame/src/flame.dart';
 import 'package:flame/src/game/game_render_box.dart';
 import 'package:flame/src/game/projector.dart';
 import 'package:flutter/rendering.dart';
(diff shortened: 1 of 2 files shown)


erickzanardo:
I was going to comment that this could mess with the automatic cache clear code that we used to have a code on Game (or GameWidget not quite sure) but I can't find it anymore 🤔 

But the idea of the game's images and assets attributes was that it would be automatically cleared once the game was removed from the tree, so setting it to the global, could cause some issues?

st-pasha:
> But the idea of the game's images and assets attributes was that it would be automatically cleared once the game was removed from the tree

I'm not sure that's advisable: even if a game is removed from the widget tree, it doesn't mean it's gone for good -- it might be added back at a later time. In fact, it's probably easy to accidentally remove the game from one place in the widget tree and immediately add it in another place when the widget structure changes and the widgets are not keyed properly.

spydon (approved):

wolfenrain (approved):

PR #1785, opened 2022-07-07
feat!: Raycasting and raytracing

spydon:
# Description

This PR implements raytracing and raycasting for the built-in hitboxes.

Depends on #1787 and #1788

Left to do:
 - [x] Raytracing with result (trivial)
 - [x] Raytracing example
 - [x] Docs

## Checklist

## Breaking Change

- [x] Yes, very minor one though

### Migration instructions

If you pass in your own collision detection system to the `HasCollisionDetection` mixin you have to change the signature of that to: `CollisionDetection<ShapeHitbox>`instead of `CollisionDetection<Hitbox>`.

## Related Issues


packages/flame/lib/src/collisions/raycast_result.dart
```diff
@@ -0,0 +1,94 @@
...
+      hitbox: other.hitbox,
+      reflectionRay: other.reflectionRay,
+      normal: other.normal,
+      distance: other.distance,
+      isActive: isActive,
+      isWithin: other.isWithin,
+    );
+  }
+
+  void setWith({
```
wolfenrain, line 64: missing docs pl0x

packages/flame/lib/src/collisions/raycast_result.dart
```diff
@@ -0,0 +1,94 @@
...
+    }
+    if (normal != null) {
+      _normal.setFrom(normal);
+    }
+    _distance = distance ?? double.maxFinite;
+    this.isActive = isActive;
+    this.isWithin = isWithin;
+  }
+
+  RaycastResult<T> clone() {
```
wolfenrain, line 84: nit: explain why this method exists?

packages/flame/lib/src/collisions/standard_collision_detection.dart
```diff
@@ -0,0 +1,103 @@
...
+    }
+    return finalResult;
+  }
+
+  @override
+  Set<RaycastResult<ShapeHitbox>> raytrace(
+    Ray2 ray, {
+    Iterable<RaycastResult<Hitbox<ShapeHitbox>>>? out,
+  }) {
+    // TODO(Spydon): implement raycastAll, rayTrace?
```
wolfenrain, line 100: Should this be turned into an issue? Or are u gonna do that in this PR?

packages/flame/lib/src/extensions/aabb2.dart
```diff
@@ -0,0 +1,3 @@
+import 'package:vector_math/vector_math_64.dart';
+
+extension Aabb2Extension on Aabb2 {}
```
wolfenrain, line 3: 👀  is this still required?

packages/flame/lib/src/geometry/polygon_ray_intersection.dart
```diff
@@ -0,0 +1,72 @@
+import 'package:flame/collisions.dart';
+import 'package:flame/components.dart';
+import 'package:flame/geometry.dart';
+
+/// A [Hitbox] in the shape of a polygon.
+mixin PolygonRayIntersection<T extends ShapeHitbox> on PolygonComponent {
+  late final _temporaryNormal = Vector2.zero();
+
+  RaycastResult<ShapeHitbox>? rayIntersection(
```
wolfenrain, line 9: Some docs would be sweet <3

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,101 @@
+import 'dart:math';
+
+import 'package:flame/geometry.dart';
+import 'package:meta/meta.dart';
+import 'package:vector_math/vector_math_64.dart';
+
+/// A Ray is a ray in the 2d plane.
```
wolfenrain, line 7:
```suggestion
/// A ray that is based in the 2d plane.
```

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,101 @@
...
+  // a division in [intersectsWithAabb2], since a ray will usually be tried
+  // against many bounding boxes it's good to pre-calculate it, which is done
+  // in the direction setter.
+  @visibleForTesting
+  late double directionInvX;
+  @visibleForTesting
+  late double directionInvY;
+
+  // Uses the branchless Ray/Bounding box intersection method by Tavian.
+  // https://tavianator.com/2011/ray_box.html
```
wolfenrain, line 40:
```suggestion
  /// Uses the branchless Ray/Bounding box intersection method by Tavian.
  /// https://tavianator.com/2011/ray_box.html
```

st-pasha (commented):
Perhaps it would make sense to try to split this PR into several smaller ones, so that they are easier to review, and they are better reflected in the change log? For example, the Ray2 class can probably be introduced in a standalone PR; similarly some new methods for the Vector2 class.

packages/flame/lib/src/collisions/broadphase.dart
```diff
@@ -13,6 +14,9 @@ abstract class Broadphase<T extends Hitbox<T>> {
   Broadphase({List<T>? items}) : items = items ?? [];
 
   Set<CollisionProspect<T>> query();
+
+  /// Returns potential [T] hitboxes that the ray can intersect.
+  Set<T> raycast(Ray2 ray);
```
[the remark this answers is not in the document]
spydon: They are cheaper for some operations, but I agree, I don't remember why I made this a set

packages/flame/lib/src/geometry/ray2.dart
spydon: This isn't a dartdoc though, this is a comment for devs changing things here.

packages/flame/lib/src/geometry/ray2.dart
spydon: Fixed in #1788

packages/flame/lib/src/geometry/ray2.dart
wolfenrain: But why not make it a dart doc?

spydon:
> Perhaps it would make sense to try to split this PR into several smaller ones, so that they are easier to review, and they are better reflected in the change log? For example, the Ray2 class can probably be introduced in a standalone PR; similarly some new methods for the Vector2 class.

Yeah, probably a good idea, split it up in #1787 and #1788.

packages/flame/lib/src/collisions/has_collision_detection.dart
```diff
@@ -4,10 +4,11 @@ import 'package:flame/game.dart';
...
 /// collision detection every tick.
 mixin HasCollisionDetection on FlameGame {
-  CollisionDetection<Hitbox> _collisionDetection = StandardCollisionDetection();
-  CollisionDetection<Hitbox> get collisionDetection => _collisionDetection;
+  CollisionDetection<ShapeHitbox> _collisionDetection =
+      StandardCollisionDetection();
+  CollisionDetection<ShapeHitbox> get collisionDetection => _collisionDetection;
 
-  set collisionDetection(CollisionDetection<Hitbox> cd) {
+  set collisionDetection(CollisionDetection<ShapeHitbox> cd) {
```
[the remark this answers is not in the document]
spydon: Yeah I guess, a very minor one though since I doubt anyone has implemented their own collision detection system to pass in, but I can set the PR to breaking.

packages/flame/lib/src/collisions/hitboxes/circle_hitbox.dart
```diff
@@ -32,4 +33,67 @@ class CircleHitbox extends CircleComponent with ShapeHitbox {
...
+  late final _temporaryAbsoluteCenter = Vector2.zero();
+
+  @override
+  RaycastResult<ShapeHitbox>? rayIntersection(
+    Ray2 ray, {
+    RaycastResult<ShapeHitbox>? out,
+  }) {
+    var isWithin = false;
+    _temporaryLineSegment.from.setFrom(ray.origin);
+    // TODO(spydon): absoluteCenter is creating two new vectors
```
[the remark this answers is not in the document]
spydon: Created #1790

st-pasha:
I've tried running the example - and it turned out to be extremely laggy when I move the mouse.
After adding `print('render')` into the render method, and `print('casting')` into the castRays method, I'm seeing the following output:
<details>

```dart
render
render
render
render
render
render
render
render
render
render
render
render
render
render
render
render
render
casting
render
render
render
render
render
casting
casting
render
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
render
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
render
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
casting
```
</details>
My guess here is that there are many mouse-move events being generated, and until each of them is processed, the next render is not scheduled. And since each raycast operation is expensive, they end up clogging the pipeline.

Recommendation: accumulate the results of mouse moves in a variable, and then process the them all at once during the `update()`.

spydon:
> I've tried running the example - and it turned out to be extremely laggy when I move the mouse. After adding `print('render')` into the render method, and `print('casting')` into the castRays method, I'm seeing the following output:
> My guess here is that there are many mouse-move events being generated, and until each of them is processed, the next render is not scheduled. And since each raycast operation is expensive, they end up clogging the pipeline.
> 
> Recommendation: accumulate the results of mouse moves in a variable, and then process the them all at once during the `update()`.

Fixed that! And broke something else, seems like something is wrong with the broadphase now. 😅

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -121,4 +121,7 @@ class Ray2 {
     this.origin.setFrom(origin);
     this.direction = direction;
   }
+
+  @override
+  String toString() => 'origin: $origin, direction: $direction';
```
st-pasha, line 126: This should include `Ray2`.

spydon: Will fix, still debugging some issues with the ray tracing on this one.

st-pasha:
Sometimes this happens (I reduce the number of rays to 20 for clarity):
Screen Shot 2022-08-03 at 7 28 21 PM

It's kinda super-rare, but in case you might now what might be causing it? As if the point of origin was at the same time inside the rectangle and outside the screen at the same time.

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
```diff
@@ -129,6 +130,15 @@ mixin ShapeHitbox on ShapeComponent implements Hitbox<ShapeHitbox> {
...
   }
 
+  /// Returns the reflection of the ray against the closest.
+  ///
+  /// If you are only interested in the intersection point use
+  /// [RaycastResult.intersectionPoint] of the result.
+  RaycastResult<ShapeHitbox>? rayIntersection(
+    Ray2 ray, {
+    RaycastResult<ShapeHitbox>? out,
+  });
```
st-pasha, line 140: It would be good to start moving towards using `Shape` classes more, instead of putting more geometric operations into shape-hitbox components.

packages/flame/lib/src/collisions/sweep.dart
```diff
@@ -1,18 +1,20 @@
...
   Sweep({super.items});
 
-  final List<T> _active = [];
-  final Set<CollisionProspect<T>> _potentials = {};
+  late final List<T> _active = [];
+  late final Set<CollisionProspect<T>> _potentials = {};
+
+  @override
+  void update() {
+    items.sort((a, b) => (a.aabb.min.x - b.aabb.min.x).ceil());
```
st-pasha, line 11: Not sure why this has `.ceil()`?

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -87,7 +101,9 @@ class Ray2 {
     final dot = _v2.dot(_v3);
     final t1 = _v2.cross(_v1) / dot;
     final t2 = _v1.dot(_v3) / dot;
-    if (t1 >= 0 && t2 >= 0 && t2 <= 1) {
+    // Using a small value above 0 just because of rounding errors later that
+    // might cause a ray to go in the wrong direction.
+    if (t1 > 0.00000000001 && t2 >= 0 && t2 <= 1) {
```
st-pasha, line 106: `t1 = 0` corresponds to a ray whose origin lies on the segment. This is considered not intersecting now?

On the other hand, since `t1` is returned from the function, the upstream code can handle that case differently if it needs to.

examples/lib/stories/collision_detection/raycast_example.dart
```diff
@@ -0,0 +1,150 @@
...
+  Ray2? ray;
+  Ray2? reflection;
+  Vector2? origin;
+  Vector2? tapOrigin;
+  bool isOriginCasted = false;
+  bool isTapOriginCasted = false;
+  Paint paint = Paint()..color = Colors.amber.withOpacity(0.2);
+  Paint tapPaint = Paint()..color = Colors.blue.withOpacity(0.2);
+
+  static const numberOfRays = 2000;
```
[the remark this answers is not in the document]
spydon: That was the point, that it should look like a light source since that is a common usecase for it.
But you do have a point, I'll rename this example and create a new one with fewer rays.

examples/lib/stories/collision_detection/raytrace_example.dart
```diff
@@ -0,0 +1,184 @@
...
+    _timePassed += dt;
+    rayPaint.color = _colorTween.transform(0.5 + (sin(_timePassed) / 2))!;
+    if (origin != null && !isOriginCasted) {
+      _ray.origin.setFrom(origin!);
+      _ray.direction
+        ..setValues(-1, 1)
+        ..normalize();
+      const angle = tau / numberOfRays;
+      for (var i = 0; i < numberOfRays; i++) {
+        _ray.direction.rotate(i * angle);
```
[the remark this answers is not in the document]
spydon: I had it like that first, the problem is that the ray's direction can still be modified in place and then you'd have unexpected results. I had it like a `NotifyingVector2` too, but that was too heavy.

examples/lib/stories/collision_detection/raytrace_example.dart
```diff
@@ -0,0 +1,184 @@
...
+        _ray.updateInverses();
+        collisionDetection.raytrace(
+          _ray,
+          maxDepth: 500,
+          out: results,
```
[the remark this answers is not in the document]
spydon: Yes, that's the point, to not create lots of unnecessary objects.

packages/flame/lib/src/collisions/raycast_result.dart
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  /// Whether this result has active results in it.
+  ///
+  /// This is used so that the objects in there can continue to live even when
+  /// there is no result from a ray cast.
+  bool isActive;
+
+  /// Whether the origin of the ray was within the hitbox.
+  bool isWithin;
```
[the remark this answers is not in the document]
spydon: > Perhaps call it isInsideHitbox or something like that, for better clarity.

Good idea, I'll change the name.

>  Also it might be good to make this private (with getter), so that the user wouldn't be able to overwrite it.

With an internal setter? Because these objects needs to be possible to re-use.

packages/flame/lib/src/collisions/sweep.dart
spydon: I don't remember exactly why I used `ceil`, but it needs to be rounded with one of the rounding methods for the sorting to work.

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
spydon: For sure, but the `Shape` classes doesn't have the correct projection yet right?

packages/flame/lib/src/geometry/ray2.dart
spydon: Yes, this fixed some bug. If I remember correctly it was when doing raytracing and doing any step later than the initial one, where origin is used from the intersection of the previous ray that was cast, then the ray will not necessarily continue in the correct direction, since it is counted as intersecting directly.

examples/lib/stories/collision_detection/raytrace_example.dart
```diff
@@ -0,0 +1,184 @@
...
+      _ray.direction
+        ..setValues(-1, 1)
+        ..normalize();
+      const angle = tau / numberOfRays;
+      for (var i = 0; i < numberOfRays; i++) {
+        _ray.direction.rotate(i * angle);
+        _ray.updateInverses();
+        collisionDetection.raytrace(
+          _ray,
+          maxDepth: 500,
```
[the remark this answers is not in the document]
spydon: True, it was a remnant of when I was trying out with many "bounces".

spydon:
> Sometimes this happens (I reduce the number of rays to 20 for clarity): Screen Shot 2022-08-03 at 7 28 21 PM
> 
> It's kinda super-rare, but in case you might now what might be causing it? As if the point of origin was at the same time inside the rectangle and outside the screen at the same time.

Very strange, I can't reproduce this. Can you reproduce this from the same ray origin consistently? If so, what origin and direction do you use? It's weird that it happens to all the rays too...

packages/flame/lib/src/collisions/raycast_result.dart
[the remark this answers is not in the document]
spydon: Changed them to private and only made them changable from `setWith`.

examples/lib/stories/collision_detection/raycast_example.dart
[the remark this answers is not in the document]
st-pasha: Oh is it? I haven't realized.
When we write documentation for this feature, it would be useful to discuss potential ways to apply it.

In this case, in order to create a light-source effect, would you draw these rays on a separate layer, and then apply via a decorator as a transparency mask?

examples/lib/stories/collision_detection/raytrace_example.dart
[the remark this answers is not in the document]
st-pasha: But you'd also have unexpected results if you modify in-place and _do not_ call `updateNormals()`. So, if the user calls `updateNormals`, they already know they are doing something iffy here, and so they might as well use the regular setter instead.

This kinda is a global problem for Flame. Whenever we have an object with some property that is a `Vector2`, then exposing that property to the user may result in the user modifying that vector in-place, which could break some important internal constraints. I can see several solutions here:
1. Create `UnmodifiableVector2` which would be a read-only view on the underlying vector, and return this object. 
2. Return a clone of the underlying vector (perhaps clone into a static "temporary" `Vector2`).
3. Return the underlying vector as-is, but have an implicit agreement with the user that they would not modify that vector. This option is most efficient but least safe.
4. Return a `NotifyingVector2` and register a callback that would update the internal state in case the user changes the vector. This approach is currently used in `PositionComponent`, and it seems to be the most expensive one.

examples/lib/stories/collision_detection/raytrace_example.dart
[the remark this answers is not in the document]
st-pasha: This is a loop with the `numberOfRays` iterations. On each iteration of the loop the `results` array gets overwritten, whereas it should be appended. The array can only be overwritten at the start of the next `update()` cycle.

packages/flame/lib/src/collisions/sweep.dart
st-pasha: Oh, interesting. I checked the docs for the `sort()` function, and seems like the canonical approach here would be
```suggestion
items.sort((a, b) => a.aabb.min.x.compareTo(b.aabb.min.x));
```

packages/flame/lib/src/geometry/ray2.dart
st-pasha: Makes sense, thanks for the explanation.
Still, it would probably be better to have this logic upstream, in case someone else needs ray intersections for other purposes.

examples/lib/stories/collision_detection/raytrace_example.dart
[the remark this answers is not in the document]
spydon: I think only 1 and 3 are possible here (like I said, I already tried 4). I think I'll go with 3 and then and in a later PR we can possibly introduce 1 on multiple places.

examples/lib/stories/collision_detection/raytrace_example.dart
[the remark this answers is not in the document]
spydon: Ah true, I don't even use `numberOfRays` here (it's 1) so I'll just remove that.

examples/lib/stories/collision_detection/raycast_example.dart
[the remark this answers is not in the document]
spydon: > In this case, in order to create a light-source effect, would you draw these rays on a separate layer, and then apply via a decorator as a transparency mask?

I haven't actually thought about it, but that sounds like a good idea.

packages/flame/lib/src/collisions/sweep.dart
spydon: Ah right, forgot about that, it's the same in many other oop languages.

packages/flame/lib/src/geometry/ray2.dart
spydon: Done!

examples/lib/stories/collision_detection/raytrace_example.dart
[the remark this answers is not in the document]
spydon: Did 3

packages/flame/lib/src/collisions/collision_detection.dart
```diff
@@ -64,64 +66,44 @@ abstract class CollisionDetection<T extends Hitbox<T>> {
...
-  /// Check what the intersection points of two collidables are,
-  /// returns an empty list if there are no intersections.
-  @override
-  Set<Vector2> intersections(
-    ShapeHitbox hitboxA,
-    ShapeHitbox hitboxB,
-  ) {
-    return hitboxA.intersections(hitboxB);
-  }
+  /// [raycastAll] casts rays uniformly around the given [origin] and gives back
```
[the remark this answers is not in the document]
spydon: In that case the helper function barely gives anything at all, then the user can just as well just loop over its list of rays and call `raycast`. But I could add an optional start and end angle, what do you think about that?

spydon:
The `RaycastResult` fields are still not calculated lazily, it showed to be quite complex to get that done and at the same time re-use objects, but I have an open branch where I keep on experimenting on this.
I want to make the docs more visual Pasha-style, but I'll have to do that later if we want to get this into the release tomorrow.

doc/flame/collision_detection.md
```diff
@@ -223,6 +223,135 @@ class MyGame extends FlameGame with HasCollisionDetection {
...
+
+
+### Ray casting
+
+Ray casting is the operation of casting out one or more rays from a point and see if they hit
+anything, in Flame's case, hitboxes.
+
+We provide two methods for doing so, `raycast` and `raycastAll`. The first one just casts out
+a single ray and gets back a result with information about what and where the ray hit, and some
+extra information like the distance, normal and the reflection ray. The second one, `raycastAll`,
```
st-pasha, line 240:
```suggestion
extra information like the distance, the normal and the reflection ray. The second one, `raycastAll`,
```

doc/flame/collision_detection.md
```diff
@@ -223,6 +223,135 @@ class MyGame extends FlameGame with HasCollisionDetection {
...
+works similarly but sends out multiple rays uniformly all around the origin, or within two angles
+from the origin.
```
st-pasha, line 242:
```suggestion
works similarly but sends out multiple rays uniformly around the origin, or within an angle
centered at the origin.
```

doc/flame/collision_detection.md
```diff
@@ -223,6 +223,135 @@ class MyGame extends FlameGame with HasCollisionDetection {
...
+game. After you have added that you can call `collisionDetection.raycast(...)` on your game class.
+
+Example:
+
+```dart
+class MyGame extends FlameGame with HasCollisionDetection {
+  @override
+  void update(double dt) {
+    super.update(dt);
+    final ray = Ray2(Vector2(0, 100), Vector2(1, 0));
```
st-pasha, line 254: Is it too late to change the signature of `Ray2` to have the arguments named? I.e. `Ray2(origin: Vector2(0, 100), direction: Vector2(1, 0))`. Otherwise, it may be confusing as to which vector represents what.

doc/flame/collision_detection.md
```diff
@@ -223,6 +223,135 @@ class MyGame extends FlameGame with HasCollisionDetection {
...
+    final result = collisionDetection.raycast(ray);
+  }
+}
+```
+
+In this example one can see that the `Ray2` class is being used, this class defines a ray from an
+origin position and a direction (which are both defined by `Vector2`s). This particular ray starts
+from `0, 100` and shoots a ray straight to the right.
+
+The result from this operation will either be `null`, if the ray doesn't hit anything, or a
```
st-pasha, line 264:
```suggestion
The result from this operation will either be `null` if the ray didn't hit anything, or a
```

doc/flame/collision_detection.md
```diff
@@ -223,6 +223,135 @@ class MyGame extends FlameGame with HasCollisionDetection {
...
+`RaycastResult` which contains:
+  - Which hitbox that the ray hit
```
st-pasha, line 266:
```suggestion
  - Which hitbox the ray hit
```

wolfenrain (approved):

doc/flame/collision_detection.md
```diff
@@ -223,6 +223,135 @@ class MyGame extends FlameGame with HasCollisionDetection {
...
+  void update(double dt) {
+    super.update(dt);
+    final ray = Ray2(Vector2(0, 100), Vector2(1, 1)..normalize());
+    final results = collisionDetection.raytrace(
+      ray, 
+      maxDepth: 100,
+    );
+    for(final result in results) {
+      if(result.intersectionPoint.distanceTo(ray.origin) > 300) {
+        break;
```
st-pasha, line 331:
```suggestion
    for (final result in results) {
      if (result.intersectionPoint.distanceTo(ray.origin) > 300) {
        break;
```

doc/flame/collision_detection.md
spydon: I had it like that first, but I heard it is a bit more expensive with named arguments, but I have no idea of how much more expensive. So maybe it is worth it?

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
```diff
@@ -129,6 +130,15 @@ mixin ShapeHitbox on ShapeComponent implements Hitbox<ShapeHitbox> {
     return collisionAllowed && aabb.intersectsWithAabb2(other.aabb);
   }
 
+  /// Returns the reflection of the ray against the closest.
```
spydon, line 133:
```suggestion
  /// Returns information about how the ray intersects the shape.
  ```

doc/flame/collision_detection.md
st-pasha: Ok, let's keep as-is then.

spydon: Haha, just updated! It is so much easier to read.

PR #1788, opened 2022-07-08
feat: Add `Ray2` class to be used in raytracing/casting

spydon:
# Description

Introduces the `Ray2` class which later will be used to see whether a ray intersects a hitbox.

## Checklist

## Breaking Change

## Related Issues


packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,101 @@
...
+    var tMin = min(tx1, tx2);
+    var tMax = max(tx1, tx2);
+
+    final ty1 = (box.min.y - origin.y) * directionInvY;
+    final ty2 = (box.max.y - origin.y) * directionInvY;
+
+    tMin = max(tMin, min(ty1, ty2));
+    tMax = min(tMax, max(ty1, ty2));
+
+    return tMax >= tMin;
```
st-pasha, line 54: I think you also need to check that `tMax >= 0`.
Need tests for (a) ray that starts within the aabb, (b) ray that starts outside and such that -ray intersects the aabb.

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,101 @@
...
+
+  Vector2 origin;
+  late Vector2 _direction;
+  Vector2 get direction => _direction;
+  set direction(Vector2 direction) {
+    assert(
+      direction.x.abs() <= 1 && direction.y.abs() <= 1,
+      'direction must be normalized',
+    );
+    _direction = direction;
```
st-pasha, line 25:
```suggestion
    _direction.setFrom(direction);
```
then the `_direction` can be final

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,101 @@
...
+    final tx2 = (box.max.x - origin.x) * directionInvX;
+
+    var tMin = min(tx1, tx2);
+    var tMax = max(tx1, tx2);
+
+    final ty1 = (box.min.y - origin.y) * directionInvY;
+    final ty2 = (box.max.y - origin.y) * directionInvY;
+
+    tMin = max(tMin, min(ty1, ty2));
+    tMax = min(tMax, max(ty1, ty2));
```
st-pasha, line 52: I think it would look prettier if you put the definitions of `tMin` and `tMax` from lines 45-46 directly in here.

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,101 @@
...
+
+/// A ray in the 2d plane.
+///
+/// The [direction] should be normalized.
+class Ray2 {
+  Ray2(this.origin, Vector2 direction) {
+    this.direction = direction;
+  }
+
+  Ray2.empty() : this(Vector2.zero(), Vector2(1, 0));
```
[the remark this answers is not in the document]
spydon: It is used to initialize a `Ray` which gets its values later, it is used in some places.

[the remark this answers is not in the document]
st-pasha: How about calling it `Ray2.default()` or something like that? "empty" is too weird - how can a ray be empty?

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,101 @@
...
+
+  Vector2 origin;
+  late Vector2 _direction;
+  Vector2 get direction => _direction;
+  set direction(Vector2 direction) {
+    assert(
+      direction.x.abs() <= 1 && direction.y.abs() <= 1,
+      'direction must be normalized',
+    );
```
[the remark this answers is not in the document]
spydon: Good idea, done

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,101 @@
...
+    _direction = direction;
+    directionInvX = 1 / direction.x;
+    directionInvY = 1 / direction.y;
```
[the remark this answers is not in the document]
spydon: Yes, it'll be `infintity`, and I think that is fine for the later calculations.

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,101 @@
...
+  Vector2 point(double length, {Vector2? out}) {
+    return ((out?..setFrom(direction)) ?? direction.clone())
+      ..scale(length)
+      ..add(origin);
+  }
+
+  late final Vector2 _v1 = Vector2.zero();
+  late final Vector2 _v2 = Vector2.zero();
+  late final Vector2 _v3 = Vector2.zero();
+  double? lineSegmentIntersection(LineSegment segment) {
```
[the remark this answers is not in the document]
spydon: Agree, is the updated description understandable?

packages/flame/lib/src/geometry/ray2.dart
spydon: The problem is that then we need to create another vector for the initialization of the ray, but maybe that is worth it.

packages/flame/lib/src/geometry/ray2.dart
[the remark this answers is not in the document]
spydon: `default` is a keyword, I renamed it `uninitialized`, but that doesn't feel spot on.
Any more ideas?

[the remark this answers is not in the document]
st-pasha: Maybe .zero?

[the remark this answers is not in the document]
spydon: Sounds good, then it is consistent with the other classes.

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,97 @@
...
+    _v3.setValues(-direction.y, direction.x);
+
+    final dot = _v2.dot(_v3);
+    final t1 = _v2.cross(_v1) / dot;
+    final t2 = _v1.dot(_v3) / dot;
+    if (t1 >= 0 && (t2 >= 0 && t2 <= 1)) {
+      return t1;
+    }
+    return null;
+  }
```
st-pasha, line 83: This method should probably need to be tested too. In particular, I'm not sure whether the sign of `t1` is correct.

packages/flame/test/geometry/ray2_test.dart
```diff
@@ -0,0 +1,128 @@
...
+
+    test('intersectsWithAabb2 southEast', () {
+      final direction = Vector2(0.5, -0.5);
+      final ray = Ray2(Vector2.zero(), direction.normalized());
+      expect(
+        ray.intersectsWithAabb2(Aabb2.minMax(Vector2(1, -2), Vector2(2, -1))),
+        isTrue,
+      );
+    });
+  });
```
st-pasha, line 127: A better test coverage for `intersectsWithAabb2` can be achieved by combining multiple checks into a single test, possibly randomized. For example:
- Make a box, pick a point outside the box and another inside the box. Check that the ray starting from the first point and pointing towards the second intersects the Aabb. Check that the ray going in the opposite direction doesn't.
- Make a box and pick a point inside. Check that a ray going in random direction intersects the box.
- Create a box such as (0,0,10,10), and pick a point on X axis, e.g. (-3,0). Check that all rays with angle less than `arctan(10/3)` hit the box, and all others don't.

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,97 @@
...
+
+  /// Deep clones the object, i.e. both [origin] and [direction] are cloned into
+  /// a new [Ray2] object.
+  Ray2 clone() => Ray2(origin.clone(), direction.clone());
+
+  void setFrom(Ray2 other) {
```
renancaraujo, line 89:
```suggestion
  /// Set the values by copying them from [other].
  void setFrom(Ray2 other) {
```

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,97 @@
...
+  // a division in [intersectsWithAabb2], since a ray will usually be tried
+  // against many bounding boxes it's good to pre-calculate it, which is done
+  // in the direction setter.
+  @visibleForTesting
+  late double directionInvX;
+  @visibleForTesting
+  late double directionInvY;
+
+  // Uses the branchless Ray/Bounding box intersection method by Tavian.
+  // https://tavianator.com/2011/ray_box.html
```
renancaraujo, line 40: Cool

renancaraujo (approved):

wolfenrain (approved):

erickzanardo (approved):

packages/flame/lib/src/geometry/ray2.dart
spydon: You were completely correct, I've updated it now!

packages/flame/lib/src/geometry/ray2.dart
spydon: Added some simple tests for it, please have a look to see whether they are testing what you were wondering about.

spydon:
Hmm, seems like something is still off with `intersectsWithAabb2` from the results I'm getting after rebasing #1785 on this PR.
Possibly the Y sign is flipped somewhere...

spydon:
> Hmm, seems like something is still off with `intersectsWithAabb2` from the results I'm getting after rebasing #1785 on this PR. Possibly the Y sign is flipped somewhere...

Nvm me, it was just the direction inverse that wasn't properly updated when a `Ray2` was modified.

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,106 @@
...
+
+  static final Vector2 _v1 = Vector2.zero();
+  static final Vector2 _v2 = Vector2.zero();
+  static final Vector2 _v3 = Vector2.zero();
+
+  /// Returns where (length wise) on the ray that the ray intersects the
+  /// [LineSegment] or null if there is no intersection.
+  ///
+  /// A ray that is parallel and overlapping with the [segment] is considered to
+  /// not intersect.
```
st-pasha, line 74: I wonder if we can handle this edge case better? For example, adding `if (dot == 0) { ... }`.

Or was this a deliberate choice?

st-pasha (approved):

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,106 @@
...
+  @visibleForTesting
+  late double directionInvY;
+
+  /// Whether the ray intersects the [box] or not.
+  ///
+  /// Caveat: Since this uses the Branchless Ray/Bounding box intersection
+  /// method by Tavian, rays that originate and follow an edge of the AABB will
+  /// be considered as non-intersecting.
+  /// https://tavianator.com/2011/ray_box.html
+  /// https://tavianator.com/2015/ray_box_nan.html
```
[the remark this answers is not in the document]
spydon: 1. True, I actually had most of it as a comment before.
2. Updated
3. Very good idea!

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,106 @@
...
+    final tMax = min(max(tx1, tx2), max(ty1, ty2));
+
+    return tMax > max(tMin, 0);
+  }
+
+  /// Gives the point at a certain length along the ray.
+  Vector2 point(double length, {Vector2? out}) {
+    return ((out?..setFrom(direction)) ?? direction.clone())
+      ..scale(length)
+      ..add(origin);
```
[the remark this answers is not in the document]
spydon: It became nicer to read like that too.

packages/flame/lib/src/geometry/ray2.dart
spydon: That would be nice, what should we return in that case though, would `t1` still be a valid point on the line segment?

spydon: Ah, obviously it can't return `t1` since that would be having a division by zero.
I'll try around a bit, if you have a good idea of how it can be done I'm all ears.

spydon: What do you think about the solution tgat I pushed now?

spydon: Hmm, maybe it makes sense that it doesn't collide with parallel rays, because it will always collide with one of the line segments connecting to it anyways?

packages/flame/lib/src/extensions/double.dart
```diff
@@ -0,0 +1,15 @@
+export 'dart:ui' show Color;
```
st-pasha, line 1: I guess this line can be removed?

packages/flame/lib/src/geometry/ray2.dart
```diff
@@ -0,0 +1,129 @@
...
+      ..sub(segment.from);
+    _v2
+      ..setFrom(segment.to)
+      ..sub(segment.from);
+    _v3.setValues(-direction.y, direction.x);
+
+    final dot = _v2.dot(_v3);
+
+    if (dot == 0) {
+      // ray is parallel to line
```
st-pasha, line 91: Sorry I'm on the phone rn, but I'd do smth like this here:
1. Let dot2=v1.dot(v3). Then if dot2!=0 there's no intersection and we can return null.
2. Otherwise, let t1=direction.dot(segment.from-ray.origin) and t2=direction.dot(segment.to-ray.origin). (These can be expressed in terms of v1 and v2). Then if both t1 and t2 positive, return the min of them; if both negative return null; otherwise, return 0 because the Ray's origin is inside the segment.

This ensures that we return a point closest to Ray's origin, and correctly handle the case when ray and segment are on parallel lines.

spydon: Hmm, maybe it makes sense that it doesn't collide with parallel rays, because it will always collide with one of the line segments connecting to it anyways? I'll remove it again for now and then we can re-add it again later if we choose to.

packages/flame/lib/src/geometry/ray2.dart
st-pasha: I guess it makes sense, esp since the intersection in this case is not a single point but rather a segment, so returning a single point would be wrong.
But we should probably explain this in a comment, because I suspect we'll soon forget what the reason was.

spydon: Good idea, I added that in #1785

PR #1799, opened 2022-07-15
feat: included InheritedComponent

alestiago:
**STATUS: In Development**

# Description

Changes:
- Creates `InheritedComponent` abstract class
- Adds public `dependOnInheritedComponentOfExactType` method.

Aim:
Aims to include the concept of InheritedWidgets into Flame. Looking for O(1) performance for packages such as `flame_bloc`. 

Integrating this concept into the framework will facilitate resolving common issues such as theming. In addition, mixins such as HasGameRef might be made redundant.

Example:
```dart
final component = Component();
  _InheritedComponent(
    children: [
      Component(
         children: [
           component,
         ],
      ),
   ],
;

// Retrieves closest _InheritedComponent in O(1).
component.read<_InheritedComponent>();
```

## Checklist

## Breaking Change

## Related Issues

## Additional Context
- https://api.flutter.dev/flutter/widgets/InheritedWidget-class.html
- https://api.flutter.dev/flutter/widgets/BuildContext/dependOnInheritedWidgetOfExactType.html

### DIFF
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -4,6 +4,7 @@ import 'dart:collection';
 import 'package:collection/collection.dart';
 import 'package:flame/src/cache/value_cache.dart';
 import 'package:flame/src/components/component_set.dart';
+import 'package:flame/src/components/inherited_component.dart';
 import 'package:flame/src/components/mixins/coordinate_transform.dart';
 import 'package:flame/src/components/position_type.dart';
 import 'package:flame/src/game/flame_game.dart';
@@ -240,6 +241,28 @@ class Component {
       addToParent(newParent);
     } else {
       newParent.lifecycle._adoption.add(this);
+    }
+  }
+
+  Map<Type, InheritedComponent>? _inheritedComponents;
+
+  T? dependOnInheritedComponentOfExactType<T extends InheritedComponent>() {
+    return _inheritedComponents?[T] as T?;
+  }
+
+  void _updateInheritance() {
+    final incomingComponents = _parent?._inheritedComponents;
+    if (incomingComponents == null) {
+      _inheritedComponents = null;
+    } else {
+      _inheritedComponents =
+          HashMap<Type, InheritedComponent>.from(incomingComponents);
+    }
+
+    final parent = this.parent;
+    if (parent is InheritedComponent) {
+      _inheritedComponents ??= HashMap<Type, InheritedComponent>();
+      _inheritedComponents![parent.runtimeType] = parent;
     }
   }
 
@@ -736,6 +759,7 @@ class Component {
     assert(_parent!.findGame()!.hasLayout);
     _setLoadingBit();
     onGameResize(_parent!.findGame()!.canvasSize);
+    _updateInheritance();
     final onLoadFuture = onLoad();
     if (onLoadFuture == null) {
       _finishLoading();
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -248,7 +248,9 @@ class Component {
 
   /// Obtains the nearest [InheritedComponent] of the given type [T], which must
   /// be the type of a concrete [InheritedComponent] subclass.
-  T? dependOnInheritedComponentOfExactType<T extends InheritedComponent>() {
+  ///
+  /// Calling this method results in O(1); unlike [ancestors].
+  T? read<T extends InheritedComponent>() {
     return _inheritedComponents?[T] as T?;
   }
 
--- a/packages/flame/lib/src/components/component.dart
+++ b/packages/flame/lib/src/components/component.dart
@@ -246,6 +246,8 @@ class Component {
 
   Map<Type, InheritedComponent>? _inheritedComponents;
 
+  /// Obtains the nearest [InheritedComponent] of the given type [T], which must
+  /// be the type of a concrete [InheritedComponent] subclass.
   T? dependOnInheritedComponentOfExactType<T extends InheritedComponent>() {
     return _inheritedComponents?[T] as T?;
   }
--- a/packages/flame/lib/src/components/inherited_component.dart
+++ b/packages/flame/lib/src/components/inherited_component.dart
@@ -1,0 +1,3 @@
+import 'package:flame/components.dart';
+
+abstract class InheritedComponent extends Component {}
--- a/packages/flame/lib/components.dart
+++ b/packages/flame/lib/components.dart
@@ -7,6 +7,7 @@ export 'src/components/component_set.dart';
 export 'src/components/custom_painter_component.dart';
 export 'src/components/fps_component.dart';
 export 'src/components/fps_text_component.dart';
+export 'src/components/inherited_component.dart';
 export 'src/components/input/joystick_component.dart';
 export 'src/components/input/keyboard_listener_component.dart';
 export 'src/components/isometric_tile_map_component.dart';
--- a/packages/flame/test/components/component_inheritance_test.dart
+++ b/packages/flame/test/components/component_inheritance_test.dart
@@ -8,7 +8,7 @@ class _TestInheritedComponentB extends InheritedComponent {}
 
 void main() {
   group('Component Inheritance', () {
-    group('dependOnInheritedComponentOfExactType', () {
+    group('read', () {
       testWithFlameGame(
         'retrieves an InheritedComponent',
         (game) async {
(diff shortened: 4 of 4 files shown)


packages/flame/lib/src/components/inherited_component.dart
```diff
@@ -0,0 +1,3 @@
+import 'package:flame/components.dart';
+
+abstract class InheritedComponent extends Component {}
```
alestiago, line 3: Shall this be just a typed mixing? Any suggestions on improving this?

```suggestion
mixin InheritedComponent {}
```

packages/flame/lib/src/components/component.dart
```diff
@@ -736,6 +759,7 @@ class Component {
     assert(_parent!.findGame()!.hasLayout);
     _setLoadingBit();
     onGameResize(_parent!.findGame()!.canvasSize);
+    _updateInheritance();
```
alestiago, line 762: Is this the right place to update inheritance? Does this cover change parent use cases? 

I will soon add tests to cover this scenarios.

packages/flame/lib/src/components/component.dart
```diff
@@ -243,6 +244,28 @@ class Component {
...
+    if (incomingComponents == null) {
+      _inheritedComponents = null;
+    } else {
+      _inheritedComponents =
+          HashMap<Type, InheritedComponent>.from(incomingComponents);
+    }
+
+    final parent = this.parent;
+    if (parent is InheritedComponent) {
+      _inheritedComponents ??= HashMap<Type, InheritedComponent>();
```
alestiago, line 264: I think we shall make sure to set _inheritedComponents to null when the component is removed. Wdyt?

alestiago:
(cc: @renancaraujo @felangel @erickzanardo)

packages/flame/lib/src/components/component.dart
```diff
@@ -243,6 +244,28 @@ class Component {
     }
   }
 
+  Map<Type, InheritedComponent>? _inheritedComponents;
```
st-pasha, line 247: This functionality should probably be defined in the `InheritedComponent` class, instead of here.

packages/flame/lib/src/components/component.dart
```diff
@@ -243,6 +244,28 @@ class Component {
...
+
+  T? dependOnInheritedComponentOfExactType<T extends InheritedComponent>() {
```
st-pasha, line 249: Can you please add a description of what this method is supposed to do? The name says "depend on ...", but it doesn't make anything depend on anything, so I was wondering if perhaps it is NYI or just needs a better name.

st-pasha (commented):
It might be useful to convert this into a draft for now, to indicate clearly that this is still a work-in-progress.

packages/flame/lib/src/components/component.dart
alestiago: Thanks @st-pasha for the review.

I'm struggling to see how would the component inheritance would work if the `_inheritedComponents` lived in the `InheritedComponent`. I think, moving the `_inheritedComponent` to the `InheritedComponent` would remove O(1).

Consider a tree structure where:

InheritedComponentA has ComponentA
ComponentA has ComponentB and ComponentC
ComponentC has componendD

ComponentD should have access to InheritedComponentA since it is one of its ancestors.

The `_inheritedComponents` are always inherited from the parent component, regardless if it's an `InheritedComponent` or not. If a parent is an `InheritedComponent` it gets added to the `_inheritedComponents`.

packages/flame/lib/src/components/component.dart
alestiago: I'm mimicking Flutter's approach to InheritedWidget (https://api.flutter.dev/flutter/widgets/BuildContext/dependOnInheritedWidgetOfExactType.html). Perhaps we can drop the `dependOn` since no method marks it as to update its dependant descendants.

What do you think about, `retreiveInheritedComponentOfExactType`? Is not my favourite name, I'm very open to naming suggestions.

packages/flame/lib/src/components/component.dart
spydon: I agree with Pasha, since this change will make the basic component more expensive.

alestiago: @spydon how would it work then? 

Also with the current implementation a component tree with no InheritedComponents are equally expensive.

spydon: Possibly as a mixin, it feels weird having this functionality straight on the raw component at least.

alestiago: I'm struggling to see how a mixin can work here. 

Do you mean a mixin that you have to apply to all descendants? 

So if you had a ThemingComponent that is an InheritedComponent and you have a tree of depth 40, you would apply the mixin to the 40 components?

spydon: I haven't thought through it (currently in Berlin on vacation), but I'm fairly sure we don't want this functionality directly in the core component class at least.

spydon: @flame-admin

alestiago: I'm looking forward to a solution where the search is O(1) without having the functionality living in the core component.

spydon: I'm not going to work out a solution for it, but feel free to propose other solutions.

> @spydon Enjoy the vacation! I was in Berlin some weeks ago and loved it 😍

Thanks, been great so far!

st-pasha:
I'm still not quite sure how this is different from `findParent<T>()`?

Perhaps it would make sense to clarify the overall purpose? Currently the stated goal is to "mirror the concept of InheritedWidget from Flutter". However, the primary purpose of an InheritedWidget in Flutter is to allow downstream widgets to mark themselves as "dependent" so that they will be automatically rebuilt when the InheritedWidget's state changes. I don't see how this functionality can be mirrored in Flame, since there's no notion of "rebuilding", neither it is clear what the "state" of an InheritedComponent will be.

alestiago:
@st-pasha the purpose of the above is to guarantee O(1) search with a Flutter like API.

I believe, `findParent<T>` fails to do so, since it relies on `ancestors`. Which can be a very expensive operation in very deep trees.

Do you know of another way of retrieving an ancestor component in O(1) already in Flame? Is there a way of doing so by composition?

erickzanardo (commented):
I didn't have a deep look on this, I am also kind on vacation on this weekend so looking on my mobile.

But here is my two cents based on what I saw and on what I saw being discussed.

I like the idea of this a lot! This will help a lot flame bloc or any other stuff that needs to read dependencies up on the component tree.

I agree with Pasha that the rebuilt when changed don't make much sense on flame, so I think that mimicking the inherited widget from flutter is not a good idea, at least not on the name as it may be confusing given the context.

I do like otherwise if we mimic Provider.read functionality though. That seems perfect for flame.

About bundling that map on the component class, I also agree that it isn't optimal, we thrive to keep that component smaller as possible to keep performance good. I wonder if we couldn't move that to the flame game instance though, and the ProviderComponent would be a HasGameRef by default

packages/flame/lib/src/components/component.dart
alestiago: I'm happy with `read`, I'll resolve. Feel free to unresolve if the name can be improved.

alestiago:
> https://github.com/flame-engine/flame/pull/1799#pullrequestreview-1041056284

Thanks @erickzanardo for joining.

I agree some renaming needs to occur. I've renamed the method to `.read<T>` I think it is less verbose and descriptive.

> About bundling that map on the component class, I also agree that it isn't optimal, we thrive to keep that component smaller as possible to keep performance good. I wonder if we couldn't move that to the flame game instance though, and the ProviderComponent would be a HasGameRef by default

I agree, I would like to have the `Component` be untouched by these changes. I'm trying to think of a solution that keeps the `Component` class untouched but still guarantees O(1) reads.

What do you think of the following API? (cc: @erickzanardo)
```dart
class ProviderComponent<V> extends Component {
  ProviderComponent({
    required this.value,
    super.children,
    super.priority,
  });

  final V value;

  static T? of<T>(Component component) {
    // Magic.
  }
}

extension ProviderRead on Component {
  T? read<T>() {
    return ProviderComponent.of<T>(this);
  }
}
```

erickzanardo:
Hummm, you need to pass a value to the provider component right? Otherwise it isn't providing anything 😅

Other than that, API looks good to me

alestiago:
Thanks @erickzanardo!

Haha, funny enough I edited and copied the wrong sample. I will update as soon as I get back to the office.

Thanks for the feedback once again.

alestiago:
I'm going on vacation so this PR will have limited activity from me up to around the 10th of August.

alestiago:
Closing for now since I do not plan to continue the work on this until the majority of the team is onboard.

Will probably move the discussion to an issue.

PR #1802, opened 2022-07-16
perf: Use TextElements within the TextComponent

### LINKED ISSUE #1801: Performance issue occurs after #1720
# Current bug behaviour
After #1720(c44272be45eadfabc8f03ef250eb663e59ef2aab), the performance issue came out.
It drops the frame rate from 60 to 20s.

As you can see in the video below, I am using two `TextComponent` for each player.
After I removed these in the same commit, the performance issue was gone.

FYR, I didn't use the new feature of that PR.

# Expected behaviour
Remove some performance issue

# Flutter doctor output

```
[✓] Flutter (Channel stable, 3.0.5, on macOS 12.1 21C52 darwin-arm, locale ko-KR)
    • Flutter version 3.0.5
    • Framework revision f1875d570e (3 days ago), 2022-07-13 11:24:16 -0700
    • Engine revision e85ea0e79c
    • Dart version 2.17.6
    • DevTools version 2.12.2

    • Android SDK 
    • Java binary
    • Java version OpenJDK Runtime Environment (build 11.0.12+0-b1504.28-7817840)

    • CocoaPods version 1.11.3

[✓] Android Studio (version 2021.2)
    • Android Studio
    • Java version OpenJDK Runtime Environment (build 11.0.12+0-b1504.28-7817840)

[✓] IntelliJ IDEA Ultimate Edition (version 2022.1.3)
    • IntelliJ 

[✓] IntelliJ IDEA Ultimate Edition (version 2022.1.3)
    • IntelliJ

[✓] VS Code (version 1.69.1)
    • Flutter extension version 3.44.0

    • SM F711N (mobile) • R3CRC0329PF • android-arm64  • Android 12 (API 31)
    • macOS (desktop)   • macos       • darwin-arm64   • macOS 12.1 21C52 darwin-arm
    • Chrome (web)      • chrome      • web-javascript • Google Chrome 103.0.5060.114

[✓] HTTP Host Availability

• No issues found!
```

* Flame version: c44272be45eadfabc8f03ef250eb663e59ef2aab
* Platform affected: android
I can confirm that it occurs on Android. I fix this after testing the IOS.
* Platform version affected: android 12 (confirmed) 

# More information

### Normal

https://user-images.githubusercontent.com/38072762/179347629-14bce155-fed6-4dd2-9c98-4212fb995726.mp4

### Has frame drop
https://user-images.githubusercontent.com/38072762/179347620-f907cef1-461a-45c9-9988-ef7b4769e52d.mp4

st-pasha:
# Description

`TextElement` is a class that caches the layout of a particular string of text. When drawing any text on the screen, it is important to convert it to text-elements and then cache those, especially if there are many text fragments being displayed simultaneously.

This PR changes the `TextComponent` class so that it computes and stores a `TextElement` instead of using `TextRenderer.render()` directly. This improves the performance of a game, as it avoids costly recalculations of the text layout on every game tick.

## Checklist

## Breaking Change

## Related Issues

Closes #1801


spydon (approved):

packages/flame/lib/src/components/text_component.dart
```diff
@@ -19,35 +33,34 @@ class TextComponent<T extends TextRenderer> extends PositionComponent {
...
-        _textRenderer = textRenderer ?? TextRenderer.createDefault<T>() {
-    updateBounds();
-  }
+  TextElement? _textElement;
 
   @internal
   void updateBounds() {
-    final expectedSize = textRenderer.measureText(_text);
-    size.setValues(expectedSize.x, expectedSize.y);
+    if (_textRenderer is FormatterTextRenderer) {
```
erickzanardo, line 46: Can't we put this logic inside `FormatterTextRenderer` somehow?

It doens't seems right to put logic of a specific renderer on the `TextComponent`

st-pasha: We have 3 `TextRenderer` implementations right now, and I had an idea for a fourth, but all of them are `FormatterTextRenderer`s. Ideally, we would merge `FormatterTextRenderer` API into the `TextRenderer`, but that would be a breaking change for anyone who implements their own `TextRenderer`.

I would argue in favor of merging those two classes anyways, if the team agrees. But that ought to be in a separate PR, at which point we can simply remove the "else" part of this if-statement.

spydon: I vote for merging them too, I don't think there are many (if anyone at all) that has made their own `TextRenderer` implementation.

packages/flame/lib/src/components/text_component.dart
```diff
@@ -1,16 +1,30 @@
...
+    super.children,
+    super.priority,
+  })  : _text = text ?? '',
+        _textRenderer = textRenderer ?? TextRenderer.createDefault<T>() {
+    updateBounds();
+  }
 
   String get text => _text;
-
+  String _text;
```
wolfenrain, line 27:
```dart
  String _text;
  String get text => _text;
```

Normally we have the private at the top and the getter/setter below it right?

wolfenrain (approved):

packages/flame/lib/src/components/text_component.dart
spydon: Doesn't that style become weird when adding dartdocs to the getter though? 🤔

st-pasha: As far as I can see, there is no Dart or Flutter official recommendations about this. The style that I'm currently using is taken from the example here: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#perform-dirty-checks-in-setters. 

In practice, Flutter repo uses several styles: either
```dart
  /// The parent of this node in the tree.
  AbstractNode? get parent => _parent;
  AbstractNode? _parent;

  /// The current value stored in this notifier.
  ///
  /// When the value is replaced with something that is not equal to the old
  /// value as evaluated by the equality operator ==, this class notifies its
  /// listeners.
  @override
  T get value => _value;
  T _value;
  set value(T newValue) {
    if (_value == newValue)
      return;
    _value = newValue;
    notifyListeners();
  }
```
or
```dart
  /// Current code point.
  ///
  /// If the iterator has hit either end, the [_currentCodePoint] is -1
  /// and `_position == _nextPosition`.
  int _currentCodePoint = -1;

  ...

  /// The rune (integer Unicode code point) starting at the current position in
  /// the string.
  ///
  /// The value is -1 if there is no current code point.
  int get current => _currentCodePoint;
```

I guess we just need to decide which style we like more, and put this into our own style guide.

spydon: I like the first one, since we usually don't add dartdocs to both the public facing getter and the underlying value.

PR #1830, opened 2022-08-01
feat: Structured text and text styles

st-pasha:
# Description

This PR introduces the notions of structured text, and text styles, to support rendering of rich text bodies.

Specifically, we recognize that sometimes in games one needs to render pieces of text that are larger than a single word or even a single paragraph. These pieces may include: books, quest descriptions, mission objectives, tutorials, in-game help system, dialogues, etc. Rendering such a piece of text is non-trivial, however. In order to tackle this problem, I break into the following parts:

1. **Text structure**, represented as a tree of `Node`s. The nodes describe the *logical* structure of the text, for example the document may contain a header, and then several paragraphs, and a list, where the list contains some list items, some of which having possibly several paragraphs, etc. This structure is similar to how in HTML the text is marked up with HTML tags.

2. **Text styles** are struct-like classes that contain properties describing how the text is to be styled: font size, font renderer, borders, backgrounds, margins, padding, etc. This representation is also tree-like, so that for example text inside paragraphs can have different style than text within headers, and paragraphs within lists can have different margins. A text style is similar to a stylesheet in HTML.

3. **Text elements** are the result of applying the document style to a document node: they are the "prepared" and laid out pieces, ready to be rendered. Elements are a bit like mini-components, or perhaps text "particles" in a particle system.

Screenshot of the example:
Screen Shot 2022-08-02 at 9 35 02 PM

## Checklist

## Breaking Change

## Related Issues

WIP for #1627


packages/flame/lib/src/text/nodes/paragraph_node.dart
```diff
@@ -0,0 +1,60 @@
...
+import 'package:flame/src/text/elements/group_element.dart';
+import 'package:flame/src/text/formatters/text_painter_text_formatter.dart';
+import 'package:flame/src/text/inline/text_painter_text_element.dart';
+import 'package:flame/src/text/nodes.dart';
+import 'package:flame/src/text/nodes/block_node.dart';
+import 'package:flame/src/text/styles/block_style.dart';
+import 'package:flutter/painting.dart' as painting;
+
+class ParagraphNode extends BlockNode {
+  // ParagraphNode._(this.child);
```
erickzanardo, line 11: Was this forgotten?
```suggestion
```

erickzanardo (approved):

erickzanardo:
@st-pasha can you attach a screen shoot of the example of this PR?

packages/flame/lib/src/text/nodes/paragraph_node.dart
st-pasha: indeed

packages/flame/lib/src/text/styles/document_style.dart
```diff
@@ -0,0 +1,137 @@
...
+
+  DocumentStyle copyWith({
+    double? width,
+    double? height,
+    EdgeInsets? padding,
+    BackgroundStyle? background,
+    BlockStyle? paragraphStyle,
+    BlockStyle? header1Style,
+    BlockStyle? header2Style,
+    BlockStyle? header3Style,
```
spydon, line 106: Could this be a list instead, supporting more than 3 header styles?

spydon (approved):
Lgtm, just one minor comment

packages/flame/lib/src/text/styles/document_style.dart
st-pasha: We could easily add `header4Style` and `header5Style` if you think a more deep documents would be needed. However, I don't want to have _arbitrarily_ deep header levels -- that way lies madness...

Also, bear in mind that these header styles will all have defaults, like `Style(fontScale: 2.0, margins: ...)`. So that by default a header would look like a header, not just a plain text. Of course, these would be overridable if the user don't like the defaults. But they may only have an objection against `header3Style`, say, and be fine with `header1Style` and `header2Style`. Having the styles as separate properties makes overriding them easier.

It also seem to be more in line with how people usually think about the document structure. There is `<H1>`, `<H2>`, and `<H3>` to designate nested sections. But it's not very common to think of them as a "list of headers of consecutive levels".

spydon: Maybe we should at least have 1-6 then, to be in line with HTML and flutter.

st-pasha: Ok, let's do that

PR #1846, opened 2022-08-15
feat: PositionComponent now has a built-in Decorator

### LINKED ISSUE #1829: HasDecorator with PositionComponent
# What could be improved

Currently, the `HasDecorator` mixin works by wrapping the component's `renderTree()` with the decorator's `apply()` method. This works fine for position-independent decorators (such as those that manipulate the paint), but is problematic for decorators that need to know the component's coordinates.

Specifically, the `PositionComponent` in its `renderTree` method applies the coordinate transformation, which affects the component's `render()` and its children. However, this coordinate transform does not affect the decorator, which is applied outside. This is problematic for such decorators as `Rotate3DDecorator` or `Shadow3DDecorator` when the underlying component moves, since it requires some external logic to keep updating the decorators so that they are in sync with the underlying component.

# How to improve

I'm currently considering the following approaches:

- **Option 1**: Add `decorator` property directly to the `PositionComponent`, so that its `renderTree` can apply the decorator after the coordinate transform.
  
  Pros:
    * Cleaner code;
    * No need to explicitly add `HasDecorator` mixin to position components;
 
  Cons:
    * Extra property `decorator` inside the `PositionComponent` class. Though if unused, it will be `null`, and won't take up much resources.

- **Option 2**: Inside the `HasDecorator` mixin check if the current component is a `PositionComponent`, and if it is, then perform `renderTree` which does not invoke `super.renderTree` but instead applies coordinate transform and renders children manually.

  Pros:
    * There is no `decorator` property unless explicitly requested by the user;

  Cons:
    * Duplicates code from `renderTree` methods of `PositionComponent` and `Component`;
    * More bug-prone in case the `renderTree` code is updated in those classes;

I'm leaning towards Option 1 here, but was wondering what do you guys think or if there's perhaps an Option 3 that I'm missing ?

st-pasha:
# Description

This PR adds the `decorator` property to `PositionComponent`, and makes decorators chain-able. That is, decorators function as linked lists, and a new decorator can be appended to the end of such list, and later removed.

The `PositionComponent` starts with the default decorator: the `Transform2DDecorator`, which encapsulates the functionality which previously resided within the `PositionComponent.renderTree`. This allows to:
  - apply/remove additional visual effects via `.decorator.addLast()/.removeLast()`;
  - replace the `Transform2DDecorator` with another one -- for example supporting skew, or isometric coordinates, or 3d rotations.

`HasDecorator` mixin can no longer be applied to a `PositionComponent` because it already has the decorator "built in".

The decorators `Shadow3DDecorator` and `Rotate3DDecorator`, when added to the PositionComponent, will now apply _after_ the coordinate transform of the `PositionComponent`, as desired. This means that, for example, the shadow's base and rotation pivot point can be specified using the local coordinates of the component.

## Checklist

## Breaking Change

  - This is non-breaking compared to the last released version, but it does change the functionality that currently exists in `main`.

## Related Issues

Closes #1829 


spydon:
Hmm, why have it set up as a linked list? Seems like quite a normal usecase that you'd want to remove, add or swap a decorator in the middle of the chain? Couldn't it just be a normal list?

st-pasha:
For the linked list it's pretty straightforward to define methods to add/remove/replace elements in the middle, if needed. So we're not sacrificing any functionality there.

At the same time, the most common use-case is when there are no additional decorators, so the linked list with a single `null` pointer will be more efficient in this case than a regular list of length 1 (which is likely to have some initial capacity > 1).

In addition, a linked list makes it much easier to call the decorators recursively (each decorator wraps the output of subsequent decorators).

spydon:
> For the linked list it's pretty straightforward to define methods to add/remove/replace elements in the middle, if needed. So we're not sacrificing any functionality there.
> 
> At the same time, the most common use-case is when there are no additional decorators, so the linked list with a single `null` pointer will be more efficient in this case than a regular list of length 1 (which is likely to have some initial capacity > 1).
> 
> In addition, a linked list makes it much easier to call the decorators recursively (each decorator wraps the output of subsequent decorators).  

Makes sense, could these methods be added in this PR? Or should that be done separately?

st-pasha:
> could these methods be added in this PR? Or should that be done separately?

I'd prefer not to overburden the PR with functionality that is not needed for it. These methods can be added later as the need arises.

spydon:
Lgtm, but aren't there docs missing?

st-pasha:
True, i'll add some

doc/flame/rendering/decorators.md
```diff
@@ -162,6 +162,28 @@ the rendering of the component. In order to remove this visual effect, simply se
...
+
+It is also possible to replace the root decorator of the `PositionComponent`, if you want to create
+an alternative logic for how the component shall be positioned on the screen.
+
+
+### Multiple decorators
+
+It is possible to apply several decorators simultaneously to the same component: the `Decorator`
+class supports chaining. That is, if you have an existing decorator on a component and you want to
+add another one, then you can call `component.decorator.addLast(newDecorator)` -- this will add
```
spydon, line 182: What happens if `addLast` is called and there already is a decorator added? Is the other one just replaced?

spydon: Probably just add a sentence of what the behaviour is if it is called twice on the same decorator with different decorators?

spydon (approved):

doc/flame/rendering/decorators.md
st-pasha: Added clarification

erickzanardo (approved):

PR #1864, opened 2022-08-23
feat(internal): Added RecycledQueue class

st-pasha:
# Description

This PR introduces `RecycledQueue` class, which is similar to a regular FIFO `Queue` except that it supports recycling of objects instead of having to create/delete them all the time.

This class will be used to build more efficient component lifecycle queues.

## Checklist

## Breaking Change


packages/flame/lib/src/utils/recycled_queue.dart
```diff
@@ -0,0 +1,120 @@
...
+
+  /// The backing container.
+  ///
+  /// Some of the items in this list are considered "active" while others are
+  /// "disposed". The [_startIndex] and [_endIndex] describe the range of items
+  /// in the list which are "active".
+  ///
+  /// Two data layouts are possible: the normal one, when [_endIndex] is larger
+  /// than or equal to the [_startIndex]:
+  ///   [----##############--]
```
spydon, line 43: Do these not have to be in back ticks for the doc parser not to regard them as references to fields?

spydon (approved):
Lgtm! Not too fond of the utils directory becoming a place with a wide mix of things in it though, maybe the two existing files in there could be put in some `math` directory, and this one be placed somewhere else or having the utils directory renamed?

packages/flame/lib/src/utils/recycled_queue.dart
st-pasha: I've added the backticks.

st-pasha:
> and this one be placed somewhere else or having the utils directory renamed?

To me, "utils" sounds a bit like "miscellaneous" (but easier to spell :).
So, where would you suggest the `RecycledQueue` class be located?

spydon:
> To me, "utils" sounds a bit like "miscellaneous" (but easier to spell :). So, where would you suggest the `RecycledQueue` class be located?

Indeed, that's what I'd like to avoid, to not have a place where non-related files live.
Maybe a new directory in `components` called `lifecycle`? And then maybe the `LifecycleManager` could be moved there too?

st-pasha:
I created the directory `components/core` then, seems like it would have a larger scope than mere `lifecycle`.

spydon:
> I created the directory `components/core` then, seems like it would have a larger scope than mere `lifecycle`.

Very nice!

erickzanardo (approved):

PR #1867, opened 2022-08-26
chore!: Remove functions/classes that were scheduled for removal in v1.3.0

st-pasha:
# Description

This functionality should have been deleted before releasing version 1.3.0, but I guess we forgot ;), so better late than never.

## Checklist

## Breaking Change

### Migration instructions

For each deleted function/method/class, the deprecation comment already specifies what functionality should be used instead.


spydon (approved):
Lgtm, seems like I didn't follow my own instructions when making the release. 😅

PR #1876, opened 2022-08-29
perf: SpriteAnimationWidget will re-render only when needed

Hwan-seok:
# Description

As-Is
`SpriteAnimationWidget` is re-rendering even if the frame index is not changed. This wastes the resource.

To-Be
`SpriteAnimationWidget` now only re-renders(call setState) only when the frame index is changed.

Additionally, the `canvas.renderRotated` is called in `SpritePainter` even though the rotation angle is 0. I think it is not needed.
 
## Checklist

## Breaking Change


spydon (approved):

renancaraujo (approved):
Thanks

PR #1877, opened 2022-08-29
fix!: Make all `ComponentSet` modifications internal

spydon:
# Description

These methods should never have been exposed, since you bypass the component lifecycle with them.

## Checklist

## Breaking Change

### Migration instructions

For most methods `Component` has the corresponding methods directly on it already.
For example, instead of using `component.children.addAll` you should do `component.addAll`.

## Related Issues

Depends on #1878 


st-pasha (approved):
I think this is good for now, but in the future it would be safer to define `ComponentSet` interface explicitly, without inheriting from `Iterable` -- this way we can control the exposed API more tightly. Plus, it could make creating alternative ComponentSet implementations easier.

erickzanardo (approved):

PR #1881, opened 2022-09-01
chore: Upgrade the dartdoc version to work with the current flutter version

luanpotter:
# Description

Upgrade the dartdoc version to work with the current flutter version

## Checklist

## Breaking Change

### DIFF
--- a/packages/flame/lib/src/cache/images.dart
+++ b/packages/flame/lib/src/cache/images.dart
@@ -1,6 +1,5 @@
 import 'dart:async';
 import 'dart:convert' show base64, json;
-import 'dart:typed_data';
 import 'dart:ui';
 
 import 'package:flame/src/flame.dart';
--- a/doc/flame/examples/pubspec.yaml
+++ b/doc/flame/examples/pubspec.yaml
@@ -5,7 +5,7 @@ publish_to: none
 
 environment:
   sdk: ">=2.17.0 <3.0.0"
-  flutter: ">=3.0.0"
+  flutter: ">=3.3.0"
 
 dependencies:
   flame: ^1.3.0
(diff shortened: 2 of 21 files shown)


erickzanardo (approved):
I just wonder if this is a `fix` or `chore`, because this is a fix on the dev dependencies, so don't actually affects the user.

luanpotter:
Good call, I didn't know chore; updated.

spydon (approved):

packages/flame/lib/src/cache/images.dart
```diff
@@ -1,6 +1,5 @@
 import 'dart:async';
 import 'dart:convert' show base64, json;
-import 'dart:typed_data';
```
spydon, line 3: Don't we have to bump the min version of Flutter in the pubspec if we remove this?

luanpotter: I'm not sure -- it was just an analyzer warning I fixed.
I can bump the version if you want though, just lmk which.

spydon: It's an analyzer warning in the current version, but would be an error when the line is removed when running with an older version of Flutter right?

spydon: Currently CI builds with the latest stable to make sure that we support the latest stable (which was why we got these analyze issues now). I think it would be fine to just bump the Flutter version in the pubspecs? I don't think there are any blockers for anyone to move to the latest stable flutter version.

luanpotter: So I changed my channel to stable and upgraded. Latest version is 3.3.0. Ours is currently 3.0.0. So I am updating all references.
image

st-pasha:
This PR changes the min Flutter version for Flame from 3.0.0 to 3.3.0 -- this is a significant (and breaking) change, so the title should at least reflect that.

PR #1891, opened 2022-09-06
feat: Add lookAt method for PositionComponent

### LINKED ISSUE #1107: Add orientation property do PositionComponent
## What could be improved
New property `double orientation;` will keep track of where the component is "looking at": is it naturally pointing up, or down, or left, or right? It's easier (and more flexible) to have this as a double and not enum.

Maybe there is a better name for it too.

# Why should this be improved
The `PositionComponent` currently consists of `size`, `anchor`, and `transform` (where transform has position, rotation, and scale). What it's missing is the notion of a natural direction. 

Use cases:
- allow `PositionComponent.angleTo()` method which computes the angle between this component's natural direction and another given component;
- allow `RotateEffect.towards()` which can rotate the component so that it would "look at" another component.

# Any risks?
No.

ufrshubham:
# Description

This PR adds a new method called `lookAt` for `PositionComponent`. It is a convenience method which rotates the component to make it point towards/look at the given target position.

Additionally, this PR also adds a `angleTo` method which can be used to get the calculated angle for `lookAt`. It will be useful if someone want to smoothly rotate towards target using effects or manual lerping.

## Checklist

## Breaking Change

## Related Issues

Fixes #1107


spydon:
@st-pasha didn't you make a function like this once? But for Radiance maybe? I think you had another name for it if I remember correctly.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -332,6 +332,10 @@ class PositionComponent extends Component
   /// The absolute center of the component.
   Vector2 get absoluteCenter => absolutePositionOfAnchor(Anchor.center);
 
+  /// Rotates the component to look at given target.
+  void lookAt(Vector2 target) =>
+      angle = Vector2(0, -1).angleToSigned(target - absolutePosition);
```
spydon, line 337: This assumes that the component is "looking up" from the beginning right? I don't think we can assume that unfortunately.

There was another PR a looong time ago setting the "start angle", I'm not sure what happened to that one. I think that was yours too @st-pasha? Did it get down voted for some reason?

ufrshubham: > This assumes that the component is "looking up" from the beginning right? I don't think we can assume that unfortunately.

Isn't `angle` always measured w.r.t the global up direction?

spydon: It isn't in regard to anything really, imagine if you have a sprite looking down or to the right for example, then if you would use this method then the angle would be off by 180/90 degrees.

spydon:
Found it, #1107 should be done before, or together with this

packages/flame/lib/src/components/position_component.dart
ufrshubham: In such cases I normally introduce a parent `PositionComponent` for the sprite and rotate that parent by 180/90 in opposite direction so that the sprite points upwards. This feels natural to me coming from other game engines. 

That being said, adding something like `orientation` does feel like a good solution as it helps avoid that additional parent component.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -87,6 +88,9 @@ class PositionComponent extends Component
     if (angle != 0) {
       transform.angle = angle ?? 0;
     }
+
+    transform.angle += angularOffset;
```
spydon, line 92: This should not add the angular offset, since that would rotate the canvas.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -98,6 +102,12 @@ class PositionComponent extends Component
   final NotifyingVector2 _size;
   Anchor _anchor;
 
+  /// Should be used to account for any angular offset needed to
+  /// orient the component in desired direction. It can be thought
+  /// of as rotating the local co-ordinate system of the component
+  /// by [angularOffset].
+  double angularOffset;
```
spydon, line 109: Not sure I like this name either, naming is hard! 😅
This should state which direction the component is looking at from the beginning, so the dartdoc is a bit wrong, it shouldn't do any modifications to the local co-ordinate system.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -132,9 +142,9 @@ class PositionComponent extends Component
   /// rotated around its anchor point in the clockwise direction if the
   /// angle is positive, or counterclockwise if the angle is negative.
   @override
-  double get angle => transform.angle;
+  double get angle => transform.angle - angularOffset;
```
spydon, line 145: This should not take `angularOffset` into regards.

ufrshubham:
Okay, now I am confused, or maybe my understanding of this new property is wrong😅. In any application dealing with transformations, angles are always w.r.t something. Just like how position is w.r.t to a origin. In most cases `angle=0` implies a vector pointing up. Looking at how angle works in Flame as well, this seems to be the case. So following that logic, my understanding of `angularOffset` was that it is just to help orient the component in the desired direction. 

For example, if a sprite is pointing downwards (as shown in the image), `angle=0` implies the default orientation of the sprite (which in this case is, pointing down).
image

But with the help of the `angularOffset`, we are trying to allow users to offset the rotation by whatever angle they want. So if they want the `angle=0` to mean that the sprite points up, they can just set the `angularOffset` to `-pi/2` or `pi/2`. That is the reason why I applied the `angularOffset` to `transform.angle` in constructor, getter and setter.

I am not able to understand how this can be achieved without actually changing `angle` of the component.😅

spydon:
The angle will be in regards to the angulaOffset, but you don't have to do canvas transformations to achieve that, because the sprite is already rotated in that direction.

So if you have a sprite looking to the right and set `angularOffset = pi/2` and `angle = pi` the sprite will look to the left, without having to add `angularOffset` to the transform. Does it make sense? :)

ufrshubham:
Okay, I think I get what you mean conceptually. Will have to play around with the code to wrap my head around it. I'll make modifications to this PR accordingly. I'm marking this as draft till I get it working.

packages/flame/lib/src/components/position_component.dart
ufrshubham: Added the suggested `nativeAngle`.

packages/flame/lib/src/components/position_component.dart
ufrshubham: Removed this rotation.

packages/flame/lib/src/components/position_component.dart
ufrshubham: Changed it to `nativeAngle` as suggested in #1107

packages/flame/lib/src/components/position_component.dart
```diff
@@ -132,9 +142,9 @@ class PositionComponent extends Component
...
   @override
-  set angle(double a) => transform.angle = a;
+  set angle(double a) => transform.angle = a + angularOffset;
```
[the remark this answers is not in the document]
ufrshubham: Modified all the angle calculations to not use this new property.

packages/flame/lib/src/components/position_component.dart
ufrshubham: Modified all the angle calculations to not use this new property.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -98,6 +101,19 @@ class PositionComponent extends Component
...
+  double _nativeAngle;
+  double get nativeAngle => _nativeAngle;
+  set nativeAngle(double value) {
+    _nativeAngle = value;
+    _nativeDirection.setFrom(Vector2(0, -1)..rotate(_nativeAngle));
+  }
+
+  /// The forward direction for this component when it is in its
+  /// default state, i.e. when [angle] is equal to zero.
+  final Vector2 _nativeDirection;
```
spydon, line 115: I don't think we want another `Vector2` field for this since it makes the class heavier, I also think there is a rotate method on `Vector2` so one could just do `Vector2(0, -1)..rotate(nativeAngle)` to get this.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -332,6 +348,11 @@ class PositionComponent extends Component
   /// The absolute center of the component.
   Vector2 get absoluteCenter => absolutePositionOfAnchor(Anchor.center);
 
+  /// Rotates the component to look at given target.
+  void lookAt(Vector2 target) {
+    angle = _nativeDirection.angleToSigned(target - absolutePosition);
```
spydon, line 353: This should be possible to calculate without creating a new vector object

spydon (commented):
Almost there! Just two comments

packages/flame/lib/src/components/position_component.dart
ufrshubham: Wouldn't doing this in `lookAt` result in creation of a new Vector2 object? Chances of people calling `lookAt` in update are right.

packages/flame/lib/src/components/position_component.dart
ufrshubham: Just to be clear, you are referring to new vector created because of `target - absolutePosition` and not `_nativeDirection`, right?

packages/flame/lib/src/components/position_component.dart
ufrshubham: Removed _nativeDirection.

packages/flame/lib/src/components/position_component.dart
ufrshubham: Nvm. Got what you were saying. Changed the logic to not depend on additional vectors.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -332,6 +337,15 @@ class PositionComponent extends Component
   /// The absolute center of the component.
   Vector2 get absoluteCenter => absolutePositionOfAnchor(Anchor.center);
 
+  /// Rotates the component to look at given target.
+  void lookAt(Vector2 target) {
+    angle = math.atan2(
+          target.x - absolutePosition.x,
+          absolutePosition.y - target.y,
+        ) -
+        nativeAngle;
```
spydon, line 346: Good shit! 😎

spydon (approved):

packages/flame/lib/src/components/position_component.dart
```diff
@@ -98,6 +99,10 @@ class PositionComponent extends Component
   final NotifyingVector2 _size;
   Anchor _anchor;
 
+  /// The angle where this component is looking at in its
+  /// default state, i.e. when [angle] is equal to zero.
+  double nativeAngle;
```
st-pasha, line 104: This is still a bit unclear: if nativeAngle is 0, then what does it mean? If I want to have my component pointing up, or to the right, then what values should I use?

packages/flame/lib/src/components/position_component.dart
```diff
@@ -332,6 +337,15 @@ class PositionComponent extends Component
   /// The absolute center of the component.
   Vector2 get absoluteCenter => absolutePositionOfAnchor(Anchor.center);
 
+  /// Rotates the component to look at given target.
+  void lookAt(Vector2 target) {
```
st-pasha, line 341: It's probably better to return the angle towards the target, instead of setting it directly. That way we can animate rotation towards the new direction instead of setting it directly.
```dart
double bearingOn(Vector2 target) => ...
```

st-pasha: Also, might be good to document/test what happens when the target is on top of you.

packages/flame/lib/src/collisions/hitboxes/composite_hitbox.dart
```diff
@@ -15,6 +15,7 @@ class CompositeHitbox extends PositionComponent
     super.size,
     super.scale,
     super.angle,
+    super.nativeAngle,
```
st-pasha, line 18: @spydon Do we want to add these new constructor parameters everywhere? They are accessible through a setter anyway.

packages/flame/lib/src/components/position_component.dart
ufrshubham: > double bearingOn(Vector2 target)

Are you also suggesting to rename `lookAt` to `bearingOn`?

As a non-native English speaker, it is not easy to guess that `bearingOn` has something to do with angles/orientation/rotation. Also, 3 of the common engines (Godot, Unreal and Unity) have a `lookAt` method with almost similar features. So keeping this method name similar to them will make it easier to find and understand.

st-pasha: Perhaps "bearingOn" is too cryptic; how about `angleTo()`? The idea is that the `lookAt()` method would essentially be
```dart
void lookAt(Vector2 target) => angle = angleTo(target);
```

But most of the time you wouldn't want to "look at" something directly, but rather apply a rotation effect so that the component would turn to the new orientation smoothly, like
```dart
component.add(RotateTowardsEffect(component.angleTo(target), EffectController(speed: 1)));
```

ufrshubham: Yeah, `angleTo` sounds perfect.

packages/flame/lib/src/components/position_component.dart
ufrshubham: Modified the comments to explain it better with meaning of common ortho angles

packages/flame/lib/src/components/position_component.dart
ufrshubham: Added `angleTo` and modified `lookAt` to reuse it.

packages/flame/lib/src/collisions/hitboxes/composite_hitbox.dart
spydon: Hmm, not sure, let me check with the rest.

ufrshubham: I would like it to be available in at least these components:
- `SpriteComponent`
- `SpriteGroupComponent`
- `SpriteAnimationComponent`
- `SpriteAnimationGroupComponent`

*Just a personal preference 😄

spydon: Maybe that's a good compromise, let's go with that.

ufrshubham: Done. Reduced the changes to sprite related components.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -332,6 +343,23 @@ class PositionComponent extends Component
...
+  /// Returns the angle formed by component's forward vector and a vector
+  /// starting at component's absolute position and ending at target.
+  /// Uses [nativeAngle] to decide the forward direction of the component.
+  /// See [lookAt] to make the component instantly rotate towards target.
+  /// Note: Returns 0 if target is exactly same as [absolutePosition].
+  double angleTo(Vector2 target) {
+    return math.atan2(
+          target.x - absolutePosition.x,
+          absolutePosition.y - target.y,
+        ) -
```
st-pasha, line 355: A bit unclear about the choice of coordinate systems here: why is the current component uses `absolutePosition`, but the target has simple position, shouldn't it use the absolute position as well?

Another concern is the rotation of the parent coordinate frame: we may need to add something like `absoluteAngle - angle` here.

ufrshubham: > why is the current component uses `absolutePosition`, but the target has simple position, shouldn't it use the absolute position as well

Yes, `target` also need to be in absolute coordinate system. But it is just a `Vector2`, so can't really enforce users to pass in absolutePosition. It is similar to how `camera.followVector2()` works. If users supply local position of a component it probably won't work as expected.

Should I change it to pass in a position component so that we can internally get the `absolutePosition`?

packages/flame/lib/src/components/position_component.dart
```diff
@@ -332,6 +343,23 @@ class PositionComponent extends Component
   /// The absolute center of the component.
   Vector2 get absoluteCenter => absolutePositionOfAnchor(Anchor.center);
 
+  /// Returns the angle formed by component's forward vector and a vector
+  /// starting at component's absolute position and ending at target.
+  /// Uses [nativeAngle] to decide the forward direction of the component.
+  /// See [lookAt] to make the component instantly rotate towards target.
+  /// Note: Returns 0 if target is exactly same as [absolutePosition].
```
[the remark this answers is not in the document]
ufrshubham: > * How about "orientation vector"

Habitual of using forward vector because of other engines, but orientation vector sounds easier to understand. Will change this.


> * or is it in CCW?

No, it is CW. Will mention that in the comments as well.



> * The last sentence is inaccurate if the nativeAngle is non-zero. I think a more accurate statement would be that if a target coincides with the current component, then it is treated as being north.

Ohhh! Good catch. Will add a test for this case as well.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -98,6 +99,16 @@ class PositionComponent extends Component
   final NotifyingVector2 _size;
   Anchor _anchor;
 
+  /// The angle where this component is looking at when it is in
+  /// the default state, i.e. when [angle] is equal to zero. Valid range
+  /// is -pi to +pi radians with 0 being vertically up.
```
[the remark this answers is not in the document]
ufrshubham: Removed that line.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -332,6 +343,23 @@ class PositionComponent extends Component
...
+  double angleTo(Vector2 target) {
+    return math.atan2(
+          target.x - absolutePosition.x,
+          absolutePosition.y - target.y,
+        ) -
+        nativeAngle;
+  }
+
+  /// Rotates/snaps the component to look at given target. Use [angleTo] to get the
+  /// calculated used by this method.
```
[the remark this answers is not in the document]
ufrshubham: Modified as suggested.

packages/flame/lib/src/components/position_component.dart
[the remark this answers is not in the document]
ufrshubham: Made the suggested changes.

packages/flame/lib/src/components/position_component.dart
ufrshubham: > Another concern is the rotation of the parent coordinate frame: we may need to add something like `absoluteAngle - angle` here.

Fixed this and added test for nested components as well.

spydon: > Should I change it to pass in a position component so that we can internally get the absolutePosition?

I think that it should be a vector2 as the argument just like it is, but we should specify in the docs that it should be an absolute vector.

ufrshubham: Added comment regarding `target`'s coordinate system.

spydon:
It would be good with some docs for this and a tiny example in the examples directory (or an embedded example in the docs).

doc/flame/components.md
```diff
@@ -27,8 +26,7 @@ void main() {
 The `Component()` here could of course be any subclass of `Component`.
 
 Every `Component` has a few methods that you can optionally implement, which are used by the
-`FlameGame` class. 
-
```
spydon, line 31: Same here

examples/lib/stories/components/look_at_example.dart
```diff
@@ -0,0 +1,114 @@
...
+import 'package:flame/input.dart';
+import 'package:flame/palette.dart';
+import 'package:flame/sprite.dart';
+import 'package:flutter/material.dart';
+
+class LookAtExample extends FlameGame with TapDetector {
+  static const description = 'This example demonstrates how a component can be '
+      'made to look at a specific target using the lookAt method. Tap anywhere '
+      'to change the target point for both the choppers. '
+      'It also shows how nativeAngle can be used to make the component orient '
```
spydon, line 15:
```suggestion
      'It also shows how nativeAngle can be used to make the component oriented '
```

ufrshubham:
> It would be good with some docs for this and a tiny example in the examples directory (or an embedded example in the docs).

Here we go

https://user-images.githubusercontent.com/33748002/189835798-eb023af7-7648-4b62-8a00-b4b1066f83af.mp4

packages/flame/test/components/position_component_test.dart
```diff
@@ -717,6 +717,96 @@ void main() {
...
+
+          expectDouble(comp.angleTo(target), angle, epsilon: 1e-10);
+
+          comp.lookAt(target);
+          expectDouble(comp.angle, angle, epsilon: 1e-10);
+        }
+      });
+
+      test('lookAt corner cases', () {
+        final comp = PositionComponent(position: Vector2(-20, 50));
```
spydon, line 802: No abbreviations, comp -> component

spydon (commented):
Just some tiny things left!

packages/flame/test/components/position_component_test.dart
ufrshubham: Fixed this for mine as well as existing tests too.

packages/flame/lib/src/components/position_component.dart
```diff
@@ -332,6 +342,33 @@ class PositionComponent extends Component
...
+  /// angle is measured in clockwise direction. [target] should be in absolute/world
+  /// coordinate system.
+  ///
+  /// Uses [nativeAngle] to decide the orientation direction of the component.
+  /// See [lookAt] to make the component instantly rotate towards target.
+  ///
+  /// Note: If target coincides with the current component, then it is treated
+  /// as being north.
+  double angleTo(Vector2 target) {
+    return math.atan2(
```
spydon, line 356: Can't you do `absoluteCenter.angleTo(target) - (nativeAngle + absoluteAngle)` here btw?
I think that it should check with the `absoluteCenter` instead of `absolutePosition` since it will be giving a bit strange angles from the edges. And also, what purpose does `- angle` have in the current calculation?

ufrshubham: > Can't you do `absoluteCenter.angleTo(target) - (nativeAngle + absoluteAngle)` here btw?

That would actually produce undesirable rotation if `anchor` is not `Anchor.center`. Think of a sprite component where the sprite is not actually centered. In such case one would set anchor to a custom value and would expect the angle calculation to be about that point.

spydon: True true!

spydon (approved):

packages/flame/lib/src/components/position_component.dart
ufrshubham: > And also, what purpose does `- angle` have in the current calculation?

I subtracted it so that I can find the absoluteAngle of parent. But now that you've mentioned it, it does look wrong. Right now it is not actually returning the *angle to* target. Rather, it returns the value than `angle` should be set to, to make it look at the target.

spydon: I guess just removing it would make the calculation correct?

ufrshubham: I think lookAt will have to be change to `angle += angleTo(target)` and some test will also have to be modified.

ufrshubham: Made these changes. Also updated the example.

erickzanardo (approved):

Issue #1882, opened 2022-09-02
Isometric support in `flame_tiled`


jtmcdole:
# What could be improved
`flame_tiled` (and thus flame) could support Tiled's isometric tile maps.

![image](https://user-images.githubusercontent.com/1924313/188240447-411b8c1f-8db8-4602-a65f-a6190433dd33.png)

# Why should this be improved
Gota have that shadowrun returns isometric feel.

# Any risks?
Doesn't seem like it at first glance; but I'm not an expert.  I didn't look into the painted layer. I didn't try to render objects from the object layer.

# More information

`renderable_tile_map.dart` does have enough information: i.e. the `TileMap`'s `MapOrientation` is known. In the `_RenderableTileLayer._cacheLayerTiles` it is possible to:

1. cache the `halfTile` outside of the loops.
2. Test the orientation in the inner loop to solve for the offsets:

```dart
          if (_map.orientation == MapOrientation.isometric) {
            offsetX = halfTile.x * (tx - ty);
            offsetY = halfTile.y * (tx + ty);
          } else {
            offsetX = ((tx + .5) * size.x);
            offsetY = ((ty + .5) * size.y);
          }

          offsetX += (layer.offsetX * scale);
          offsetY += (layer.offsetY * scale);
```

3. and then switch between how to add the batch transform:

```dart
            if (_map.orientation == MapOrientation.isometric) {
              batch.add(source: src, offset: Vector2(offsetX, offsetY));
            } else {
              final scos = flips.cos * scale;
              final ssin = flips.sin * scale;

              batch.addTransform(...
```

jtmcdole:
Happy to put this up as a change - assumption is that no one is using isometric right now, so it would be "experimental" at best.

jtmcdole:
I'll have a pull request soon once I write some tests - but it works with the RSTransform, so that's nice.

PR #1892, opened 2022-09-06
feat: Hexagonal maps

### LINKED ISSUE #1887: Hexagonal Support in `flame_tiled`
# What could be improved
Support Tiled's Hexagonal maps

image

# Why should this be improved
Gota have that Catan  hexagonal feel.

# Any risks?
Doesn't seem like it at first glance; but **I'm not an expert**. I didn't look into the painted layer. I didn't try to render objects from the object layer.

# More information

Building on #1882 and #1885 - I just wanted to go for completeness.  Hexagons was a bit more though since staggering between X/Y oriented tiles requires some juggling.  The other problem: hexagonal tiles are not square - if your destination tile size in `TiledComponent.load`  is wrong, it looks bad (gaps).

Possibly future improvement: discover the `destTileSize` automatically from the tmx files?

jtmcdole:
# Description

Adds support for Tiled's Hexagonal maps.

## Checklist

## Breaking Change

## Related Issues

Fixes #1887
Similar to #1882


spydon (approved):
Nice, good job! What angle are the hexagonal tiles in when they are angled? Looks like less than for the isometric view?

jtmcdole:
> What angle are the hexagonal tiles in when they are angled

I don't know for these specific tiles, but they are not top-down.  This is OK since Tiled doesn't ask you to define "length of one side" and force you into top-down mode - it asks you for tile width and height. I think this will work for trimetric hexagons (where width = 2 *h of the tile).  I can experiment later today or tomorrow.

erickzanardo (approved):

kurtome:
@jtmcdole It looks like this PR inadvertently removed taking `layer.offsetX` and `layer.offsetY` into account when calculating the tile positions for renderable tile layers. I'm just going to add the offset as a `canvas.translate(offsetX, offsetY)` in the `render` function for that layer as part of #1886, does that sound right to you?

jtmcdole:
@kurtome - it was in the previous PR. Sorry about that oversite!

When you do, can we add some golden tests to make sure this doesn't happen again? I added several goldens already, so it should be easy to add more.

PR #1894, opened 2022-09-09
feat: quad tree broadphase support 

ASGAlex:
# Description

Quad tree broadphase support.

## Checklist

## Breaking Change


packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
```diff
@@ -0,0 +1,103 @@
...
+  ///   [minimumDistanceCheck] if you need some custom behavior
+  ///
+  /// [broadPhaseCheck] checks if objects of different types should to collide.
+  /// The result of calculation is cached so you should not check here any
+  /// dynamical parameters, the function intended to be used as pure type
+  /// checker.
+  /// It usually should not be reimplemented, see
+  /// [CollisionCallbacks.broadPhaseCheck] instead
+  ///
+  void initCollisionDetection({
```
spydon, line 40: `initializeCollisionDetection` or `initializeBroadphase` (no abbreviations)

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
```diff
@@ -0,0 +1,103 @@
...
+  ///   to consider them as possibly collideable. You also can reimplement
+  ///   [minimumDistanceCheck] if you need some custom behavior
+  ///
+  /// [broadPhaseCheck] checks if objects of different types should to collide.
+  /// The result of calculation is cached so you should not check here any
+  /// dynamical parameters, the function intended to be used as pure type
+  /// checker.
+  /// It usually should not be reimplemented, see
+  /// [CollisionCallbacks.broadPhaseCheck] instead
+  ///
```
spydon, line 39:
```suggestion
  /// Initialise the QuadTree.
  ///
  /// - [mapDimensions] describes the collision area coordinates and size.
  ///   Should match to game map position and size.
  /// - [maxObjects] (optional) - maximum objects count in one quadrant.
  /// - [maxLevels] (optional) - maximum nested quadrants.
  /// - [minimumDistance] (optional) - specify minimum distance between objects
  ///   to consider them as possibly colliding. You can also implement the
  ///   [minimumDistanceCheck] if you need some custom behavior.
  ///
  /// The [broadPhaseCheck] checks if objects of different types should
  /// collide.
  /// The result of the calculation is cached so you should not check any
  /// dynamical parameters here, the function is intended to be used as pure type
  /// checker.
  /// It usually should not be overridden, see
  /// [CollisionCallbacks.broadPhaseCheck] instead
```

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
```diff
@@ -0,0 +1,103 @@
...
+  bool minimumDistanceCheck(Vector2 activeItemCenter, Vector2 potentialCenter) {
+    if (minimumDistance != null) {
+      if ((activeItemCenter.x - potentialCenter.x).abs() > minimumDistance! ||
+          (activeItemCenter.y - potentialCenter.y).abs() > minimumDistance!) {
+        return false;
+      } else {
+        return true;
+      }
+    }
+    return true;
```
spydon, line 67:
```suggestion
    return minimumDistance == null || 
       !((activeItemCenter.x - potentialCenter.x).abs() > minimumDistance! ||
         (activeItemCenter.y - potentialCenter.y).abs() > minimumDistance!);
```

This one will need formatting, the `!` could be removed too if the check within `()` is inverted.

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,378 @@
...
+import 'package:flame/extensions.dart';
+
+/// Public interface to QuadTree internal data structures
+/// Allow to read node's data without risc to affect on how collisions
+/// works.
+/// Use [QuadTreeNodeDebugInfo.init] to init class for current collision
+/// detection. You need only this instance no get all another nodes and
+/// hitboxes using [nodes] and [allElements] / [ownElements] methods.
+/// Use [rect] to get node's computed box;
+/// The class might be useful to render debugging info. See examples for details
```
spydon, line 11:
```suggestion
/// Public interface to QuadTree internal data structures.
///
/// Allows to read a node's data without risk of affecting the outcome of
/// collisions.
/// Use [QuadTreeNodeDebugInfo.init] to initialize the class for the
/// current collision detection. You need only this instance to get all
/// another nodes and hitboxes using [nodes] and the
/// [allElements] / [ownElements] methods.
/// Use [rect] to get node's computed box;
/// The class might be useful to render debugging info.
/// See examples for details.
```

packages/flame/lib/src/collisions/quadtree/broadphase.dart
```diff
@@ -0,0 +1,164 @@
...
+  Vector2 potentialCenter,
+);
+
+/// Performs Quad Tree broadphase check
+/// See [HasQuadTreeCollisionDetection.initCollisionDetection] for detail
+/// description of its parameters
+class QuadTreeBroadphase<T extends Hitbox<T>> extends Broadphase<T> {
+  QuadTreeBroadphase({
+    super.items,
+    required Rect mainBoxSize,
```
spydon, line 23: Maybe this should be an `Aabb2` since that is what we are using everywhere else, or is `Rect` more efficient here?

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
```diff
@@ -0,0 +1,103 @@
...
+  ///
+  /// [broadPhaseCheck] checks if objects of different types should to collide.
+  /// The result of calculation is cached so you should not check here any
+  /// dynamical parameters, the function intended to be used as pure type
+  /// checker.
+  /// It usually should not be reimplemented, see
+  /// [CollisionCallbacks.broadPhaseCheck] instead
+  ///
+  void initCollisionDetection({
+    required Rect mapDimensions,
```
spydon, line 41: `Aabb2`?

examples/lib/stories/collision_detection/quadtree_example.dart
```diff
@@ -0,0 +1,434 @@
...
+      }
+      for (final box in nodeElements) {
+        // textPaint.render(canvas, node.id.toString(), box.aabb.min);
+        if (boxPaint != null) {
+          final rect = Rect.fromLTRB(
+            box.aabb.min.x,
+            box.aabb.min.y,
+            box.aabb.max.x,
+            box.aabb.max.y,
+          );
```
spydon, line 427: If we don't have it already we could add an extension method for `Aabb2` that does `toRect`, we have that on many other classes.

examples/lib/stories/collision_detection/quadtree_example.dart
```diff
@@ -0,0 +1,434 @@
...
+  @override
+  void onScroll(PointerScrollInfo info) {
+    camera.zoom += info.scrollDelta.game.y.sign * 0.08;
+    camera.zoom = camera.zoom.clamp(0.05, 5.0);
+  }
+}
+
+//#endregion
+
+//#region Player
```
spydon, line 187: Since this region only entails one class it is probably unnecessary?

examples/lib/stories/collision_detection/quadtree_example.dart
```diff
@@ -0,0 +1,434 @@
...
+to understand that handling hitboxes on edges requires more resources
+Blue lines visualises quad tree quadrants positions 
+
+Use WASD to move player. Use mouse scroll to change zoom.
+Hold direction button and press Space to fire bullet. 
+Notice that bullet fly above water but collides 
+with bricks.  
+Notice than creating a lot of bullets at once leads to generating new quadrants
+on the map because there become more than 25 objects
+Press O button to rescan all tree and optimize it, removing unused quadrants.
```
spydon, line 33:
```suggestion
In this example the standard "Sweep and Prune" algorithm is replaced by  
"Quad Tree". Quad Tree is often a more efficient approach of handling collisions,
it's efficiency is shown especially on huge maps with big amounts of collidable 
components.
Some bricks are highlighted when placed on an edge of a quadrant. It is
important to understand that handling hitboxes on edges requires more
resources.
Blue lines visualise the quad tree's quadrant positions.

Use WASD to move the player and use the mouse scroll to change zoom.
Hold direction button and press space to fire a bullet. 
Notice that bullet will fly above water but collides with bricks.

Also notice that creating a lot of bullets at once leads to generating new
quadrants on the map since it becomes more than 25 objects in one quadrant.

Press O button to rescan the tree and optimize it, removing unused quadrants.
```

examples/lib/stories/collision_detection/quadtree_example.dart
```diff
@@ -0,0 +1,434 @@
...
+  @override
+  KeyEventResult onKeyEvent(
+    RawKeyEvent event,
+    Set<LogicalKeyboardKey> keysPressed,
+  ) {
+    Vector2? displacement;
+    var fireBullet = false;
+    for (final key in keysPressed) {
+      if (key == LogicalKeyboardKey.keyW && player.canMoveTop) {
+        displacement = Vector2(0, -stepSize);
```
spydon, line 144: We shouldn't encourage creating new objects within these kind of methods, use a temporary private vector and use `setFrom`/`setValues` on it instead.

examples/lib/stories/collision_detection/quadtree_example.dart
```diff
@@ -0,0 +1,434 @@
...
+        (collisionDetection as QuadTreeCollisionDetection)
+            .quadBroadphase
+            .tree
+            .optimize();
+      }
+    }
+    if (fireBullet && displacement != null) {
+      final bullet =
+          Bullet(position: player.position, displacement: displacement * 50);
+      add(bullet);
```
spydon, line 172: This will lead to that many bullets per tick are added, which might make faster devices able to add more bullets than slow devices. We might oversee that in this example though if the code becomes a lot longer by adding it in `update` instead.

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,378 @@
...
+  }
+
+  @override
+  String toString() {
+    return 'node $id';
+  }
+}
+
+/// QuadTree calculation class not bound to Flame. Could be used anywhere
+/// outside of Flame, for example in isolate to calculate background logic.
```
spydon, line 80:
```suggestion
/// outside of Flame, for example in isolates to calculate background logic.
```

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,378 @@
...
+///
+/// Usage:
+/// 1. Create new instance
+/// 2. Use [add] to add hitboxes to simulation.
+/// 3. Use [query] to get list of hitboxes potentially collide with specified
```
spydon, line 85:
```suggestion
/// 3. Use [query] to get list of hitboxes that potentially collide.
```

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,378 @@
...
+  String toString() {
+    return 'node $id';
+  }
+}
+
+/// QuadTree calculation class not bound to Flame. Could be used anywhere
+/// outside of Flame, for example in isolate to calculate background logic.
+///
+/// Usage:
+/// 1. Create new instance
```
spydon, line 83:
```suggestion
/// 1. Create new instance.
```

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,378 @@
...
+/// 2. Use [add] to add hitboxes to simulation.
+/// 3. Use [query] to get list of hitboxes potentially collide with specified
+/// 4. Use [remove] to remove hitbox from tree
+/// 5. Use sequence of [remove] and [add] to simulate hitbox movement
+/// 6. Use [isMoved] to determine, does hitbox ever changed it's position
+/// 7. Call [clear] to remove all data.
```
spydon, line 89:
```suggestion
/// 4. Use [remove] to remove a hitbox from the tree.
/// 5. Use sequence of [remove] and [add] to simulate hitbox movement.
/// 6. Use [isMoved] to determine if a hitbox ever changed its position.
/// 7. Call [clear] to remove all data.
```

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,378 @@
...
+///
+/// Use [optimize] to scan all tree and remove unused quadrants.
+///
```
spydon, line 92:
```suggestion
/// Use [optimize] to scan the tree and remove unused quadrants.
```

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,378 @@
...
+    }
+  }
+
+  int _getQuadrant(Rect nodeBox, Rect valueBox) {
+    final center = nodeBox.center;
+    // West
+    if (valueBox.right <= center.dx) {
+      // North West
+      if (valueBox.bottom <= center.dy) {
+        return 0;
```
spydon, line 173: Maybe an enum could be used here to make the code more clear?

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,378 @@
...
+  }
+
+  void add(T hitbox) {
+    final node = _add(_rootNode, 0, mainBoxSize, hitbox, null);
+    _oldPositionByItem[hitbox as ShapeHitbox] = Aabb2.copy(hitbox.aabb);
+    _itemAtNode[hitbox as ShapeHitbox] = node;
+  }
+
+  _Node<T> _add(_Node<T> node, int depth, Rect box, T value, _Node? parent) {
+    // assert(box.containsRect(getBoxOfValue(value)));
```
spydon, line 204: Remove?

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,378 @@
...
+    _Node<T> finalNode;
+    if (_noChildren(node)) {
+      // Insert the value in this node if possible
```
spydon, line 207:
```suggestion
      // Insert the value in this node if possible.
```

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,378 @@
...
+      if (depth >= maxLevels || node.hitboxes.length < maxObjects) {
+        node.hitboxes.add(value);
+        finalNode = node;
+      }
+      // Otherwise, we split and we try again
```
spydon, line 212:
```suggestion
      // Otherwise, we split and we try again.
```

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,378 @@
...
+      else {
+        _split(node, box);
+        finalNode = _add(node, depth, box, value, parent);
+      }
+    } else {
+      final quadrant = _getQuadrant(box, _getBoxOfValue(value));
+      // Add the value in a child if the value is entirely contained in it
```
spydon, line 219:
```suggestion
      // Add the value in a child if the value is entirely contained in it.
```

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,378 @@
...
+        }
+        finalNode = _add(
+          children as _Node<T>,
+          depth + 1,
+          _computeBox(box, quadrant),
+          value,
+          node,
+        );
+      }
+      // Otherwise, we add the value in the current node
```
spydon, line 233:
```suggestion
      // Otherwise, we add the value in the current node.
```

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
```diff
@@ -0,0 +1,81 @@
+import 'package:flame/collisions.dart';
+import 'package:flame/components.dart';
+import 'package:flutter/widgets.dart';
+
+/// Collision detection modification to support Quad Tree.
+/// Do not use standard [items] list fo components. Instead adds all components
+/// into [QuadTreeBroadphase] class.
+///
```
spydon, line 8:
```suggestion
/// Collision detection modification to support a Quad Tree broadphase.
///
/// Do not use standard [items] list for components. Instead adds all components
/// into [QuadTreeBroadphase] class.
```

spydon (commented):
Very good job!
Tests are missing, but I guess that might have been intentional so that you could show the API first.

packages/flame/lib/src/collisions/quadtree/broadphase.dart
ASGAlex: I cant to predict the best solution for end-user, but can confidently say that it would not be Aabb2 class. 
Map dimensions can be get from `TiledComponent` if map loaded from file. For generated map it could be pure width and height.  
It seems to me that Rect is just more obviouse here: it represents an area where collision would be calculated. 
But maybe it's only my opinion that it would be user-friendly :-)

doc/flame/collision_detection.md
```diff
@@ -196,38 +196,118 @@ collisions to the whole hat, instead of for just each hitbox separately.
...
+therefore avoiding O(n²). The broad phase produces a set of potential collisions (a set of
 `CollisionProspect`s), this set is then used to check the exact intersections between hitboxes, this
 is sometimes called narrow phase.
 
 By default Flame's collision detection is using a sweep and prune broadphase step, if your game
 requires another type of broadphase you can write your own broadphase by extending `Broadphase` and
 manually setting the collision detection system that should be used.
 
-For example if you have implemented a broadphase built on a quad tree instead of the standard
+For example if you have implemented a broadphase built on a magic algorithm instead of the standard
```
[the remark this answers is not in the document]
ASGAlex: Maybe some day....

packages/flame/lib/src/collisions/quadtree/broadphase.dart
spydon: Since the `TiledComponent` soon will be a `PositonComponent` and can therefore also be rotated it is a lot easier to handle rotations in `Aabb2` since the bounding box needs to grow when that happens. And the `PositionComponent` already exposes an Aabb2 if I remember correctly, or maybe the rect that it exposes already takes the angle into consideration?

spydon: Nvm, the `toRect` gives back the axis-aligned rect.

ASGAlex: To be honest, I'd prefer not to handle rotations of whole map in this algoritm and just to make bigger "bounding box' instead...

packages/flame/lib/src/collisions/quadtree/broadphase.dart
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        if (_broadphaseCheckCache[activeItem]?[potential] == false) {
+          continue;
+        }
+
+        final asShapePotential = potential as ShapeHitbox;
+
+        if (asShapePotential.isRemoving || asShapePotential.parent == null) {
+          markRemove.add(potential);
```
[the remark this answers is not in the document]
ASGAlex: This is a problem I mentioned before. 
Some components are removed from game but still present in quad tree. I still cant find a roots of this leak. Maybe it is produced by listeners for hitbox's size and position: in theory it still could be called while component already marked to be removed. 
Anyway, I placed this check into broadphase to exclude all possible "dead" objects from later cycles. It is cheaper then to call "optimize()" on whole tree.

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
```diff
@@ -10,8 +10,25 @@ import 'package:meta/meta.dart';
...
 mixin ShapeHitbox on ShapeComponent implements Hitbox<ShapeHitbox> {
+  CollisionType _collisionType = CollisionType.active;
+
+  set collisionType(CollisionType type) {
+    if (_collisionType == type) {
+      return;
+    }
+    _collisionType = type;
+    final broadPhase = _collisionDetection?.broadphase;
+    if (broadPhase is QuadTreeBroadphase && isMounted) {
```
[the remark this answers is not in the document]
ASGAlex: I have no idea, how to do it without creating external function with two arguments: ShapeHitbox and new collision type. It does not look very pretty especially when code like `hitbox.collisionType = CollisionType.active` become useless for quad tree and we forced to change such lines into something like `(collisionDetection as QuadTreeCollisionDetection).quadBroadphase.setHitboxCollisionType(hitbox, CollisionType.active)`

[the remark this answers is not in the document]
ASGAlex: The only way is, I think, to enchance collisionType with ChangeNotifier

packages/flame/lib/src/collisions/quadtree/broadphase.dart
spydon: Makes sense, so the user just have to send in a larger rect if they plan to rotate the component.

packages/flame/lib/src/collisions/quadtree/broadphase.dart
```diff
@@ -0,0 +1,164 @@
...
+            _broadphaseCheckCache[item0 as T] = {};
+          }
+          _broadphaseCheckCache[item0 as T]![item1 as T] = false;
+        }
+      }
+    }
+    return potentials;
+  }
+
+  void updateItemSizeOrPosition(T item) {
```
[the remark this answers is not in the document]
ASGAlex: Does scaling or rotating a component affects scaling and rotating values of component's hitbox? 
I figured out, that changing component position does not affect to hitbox's position. 
So maybe I need to listen to component's values instead of hitbox's?

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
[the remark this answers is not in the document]
ASGAlex: > The only way is, I think, to enchance collisionType with ChangeNotifier

Review this variant at my past commits, please

packages/flame/lib/src/collisions/quadtree/broadphase.dart
```diff
@@ -0,0 +1,164 @@
...
+      markRemove.forEach(tree.remove);
+    }
+
+    if (potentialsTmp.isNotEmpty) {
+      for (var i = 0; i < potentialsTmp.length; i++) {
+        final item0 = potentialsTmp[i].first;
+        final item1 = potentialsTmp[i].last;
+        var keep = broadphaseCheck(item0, item1);
+        if (keep) {
+          keep = broadphaseCheck(item1, item0);
```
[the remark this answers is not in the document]
ASGAlex: I just fugured out that broadphaseCheck already works as you proposed

examples/lib/stories/collision_detection/quadtree_example.dart
ASGAlex: This is exactly what I intended to do: to create as much bullets as possible to force collision system to split screen into additional quadrants... and then you can press "O" and to see how these sections disappears: 
image

packages/flame/lib/src/collisions/quadtree/broadphase.dart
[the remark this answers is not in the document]
spydon: All transforms of the parent affect the hitbox, the collision detection system is looking at the `absolutePosition` not the local position of the hitbox. The local transforms of the hitbox are for how the hitbox should be placed within the parent.

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
```diff
@@ -4,14 +4,31 @@ import 'package:flame/game.dart';
...
 import 'package:flame/src/geometry/shape_intersections.dart'
     as intersection_system;
-import 'package:meta/meta.dart';
+import 'package:flutter/foundation.dart';
 
 /// A [ShapeHitbox] turns a [ShapeComponent] into a [Hitbox].
 /// It is currently used by [CircleHitbox], [RectangleHitbox] and
 /// [PolygonHitbox].
 mixin ShapeHitbox on ShapeComponent implements Hitbox<ShapeHitbox> {
+  final _collisionType = NotifyingCollisionType(CollisionType.active);
```
spydon, line 13: I would make this public, but `@internal`, so that you don't need the add/remove methods below.

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
```diff
@@ -83,6 +94,14 @@ mixin ShapeHitbox on ShapeComponent implements Hitbox<ShapeHitbox> {
       _parentSizeListener?.call();
       hitboxParent.size.addListener(_parentSizeListener!);
     }
+
+    // This should be placed after the hitbox parent listener
+    // since the correct hitbox size is required by the QuadTree.
+    final parentGame = findParent<FlameGame>();
```
spydon, line 100: Use `findGame` instead

packages/flame/lib/src/collisions/collision_callbacks.dart
```diff
@@ -114,6 +127,22 @@ mixin CollisionCallbacks on Component
     onCollisionEndCallback?.call(other);
   }
 
+  /// Works only for the QuadTree collision detection.
+  /// If you need to prevent collision of items of different types -
+  /// reimplement [broadPhaseCheck]. The result of calculation is cached so you
+  /// should not check any dynamical parameters here, the function intended to
+  /// be used as pure type checker.
+  @mustCallSuper
+  bool broadPhaseCheck(PositionComponent other) {
```
spydon, line 136: I'm not too keen on having this in the callbacks mixin, also since it doesn't work for all systems.

packages/flame/lib/src/collisions/quadtree/broadphase.dart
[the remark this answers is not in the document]
spydon: It feels like this should be a fixable bug, afaik the other system doesn't have any lingering hitboxes after their parents have been removed.

packages/flame/lib/src/collisions/quadtree/broadphase.dart
```diff
@@ -0,0 +1,161 @@
...
+          continue;
+        }
+        if (asShapePotential.parent == asShapeItem.parent &&
+            asShapeItem.parent != null) {
+          continue;
+        }
+
+        Vector2 potentialCenter;
+        if (potential.collisionType == CollisionType.passive) {
+          potentialCenter = _getCenterOfHitbox(asShapePotential);
```
spydon, line 86: Why are the centers only cached when the collion type is passive? :thinking:

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,395 @@
...
+/// 1. Create new instance.
+/// 2. Use [add] to add hitboxes to simulation.
+/// 3. Use [query] to get list of hitboxes that potentially collide.
+/// 4. Use [remove] to remove a hitbox from the tree.
+/// 5. Use sequence of [remove] and [add] to simulate hitbox movement.
+/// 6. Use [isMoved] to determine if a hitbox ever changed its position.
+/// 7. Call [clear] to remove all data.
+///
+/// Use [optimize] to scan the tree and remove unused quadrants.
+class QuadTree<T extends Hitbox<T>> {
```
spydon, line 122: This class should be in the top of the file since it is the "main" class of this file.

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
```diff
@@ -0,0 +1,105 @@
...
+    required Rect mapDimensions,
+    required ExternalBroadphaseCheck broadphaseCheck,
+    required ExternalMinDistanceCheck minimumDistanceCheck,
+    int maxObjects = 25,
+    int maxLevels = 10,
+  }) : super(
+          broadphase: QuadTreeBroadphase<ShapeHitbox>(
+            mainBoxSize: mapDimensions,
+            maxObjects: maxObjects,
+            maxLevels: maxLevels,
```
spydon, line 20: `maxDepth`?

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
```diff
@@ -0,0 +1,105 @@
...
+  QuadTreeBroadphase get quadBroadphase => broadphase as QuadTreeBroadphase;
+
+  final _listenerTransform = <ShapeHitbox, VoidCallback>{};
+  final _listenerCollisionType = <ShapeHitbox, VoidCallback>{};
+  final _scheduledUpdate = <ShapeHitbox>{};
+
+  @override
+  void add(ShapeHitbox item) {
+    super.add(item);
+    // ignore: prefer_function_declarations_over_variables
```
spydon, line 35: Why not make a function declaration in here instead?

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
```diff
@@ -0,0 +1,105 @@
...
+    final listenerTransform = () {
+      if (item.isMounted) {
+        _scheduledUpdate.add(item);
+      }
+    };
+    final parent = item.parent;
+    if (parent != null && parent is PositionComponent) {
```
spydon, line 42:
```suggestion
    if (parent is PositionComponent) {
```

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
```diff
@@ -0,0 +1,105 @@
...
+      parent.position.addListener(listenerTransform);
```
spydon, line 43: This won't work for `ComposedHitboxes`.

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
```diff
@@ -0,0 +1,105 @@
...
+    }
+    item.transform.removeListener(listenerTransform);
+    _listenerTransform[item] = listenerTransform;
+
+    // ignore: prefer_function_declarations_over_variables
+    final listenerCollisionType = () {
```
spydon, line 49: Same here, declare a function instead (they can be declared anywhere).

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
```diff
@@ -0,0 +1,105 @@
...
+  final _listenerTransform = <ShapeHitbox, VoidCallback>{};
+  final _listenerCollisionType = <ShapeHitbox, VoidCallback>{};
+  final _scheduledUpdate = <ShapeHitbox>{};
+
+  @override
+  void add(ShapeHitbox item) {
+    super.add(item);
+    // ignore: prefer_function_declarations_over_variables
+    final listenerTransform = () {
+      if (item.isMounted) {
```
spydon, line 37: As long as the item is getting added here, it should always be mounted right, so no need for this check?

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
spydon: And also won't work if the parent has rotated and/or scaled if the anchor is not in the center.
Better just add a listenter to the full `transform`.

packages/flame/lib/src/extensions/aabb.dart
```diff
@@ -0,0 +1,6 @@
+import 'package:flame/extensions.dart';
+
+extension Aabb2Extension on Aabb2 {
+  /// Creates a [Rect] starting in [min] and going the [max]
+  Rect toRect() => Rect.fromLTRB(min.x, min.y, max.x, max.y);
```
spydon, line 5: This doesn't seem to be used anywhere?

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
ASGAlex: But it will become breaking change, because  NotifyingCollisionType is not a subtype of CollisionType enum. This case not similar to NotifyingVector2 for example.

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
ASGAlex: This is not my piece of code, I just changed it's place ;-)
image

packages/flame/lib/src/collisions/collision_callbacks.dart
ASGAlex: But it is actually callback like other collision callbacks are. It is intended to be reimplemented by framework's user. 
And also, the "sweep" algorithm also could be extended by this functionality, but patching original algorithm is out of the scope of this PR, I simply did not analyze it's code. To perform a basic type-checking at broadphase is basically better approach then to perform it after expensive collision check.

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
ASGAlex: This is a listener. It might be called at any time, its probably a wrong function declaration confuses you.

packages/flame/lib/src/extensions/aabb.dart
ASGAlex: I thought, here  https://github.com/flame-engine/flame/pull/1894#discussion_r967194766  you asked me to add this functionality, isn't it?..

doc/flame/collision_detection.md
```diff
@@ -196,38 +196,123 @@ collisions to the whole hat, instead of for just each hitbox separately.
...
-A broad phase is the first step of collision detection where potential collisions are calculated.
-To calculate these potential collisions are a lot cheaper to calculate than to check the exact
-intersections from the directly and it removes the need to check all hitboxes against each other
-and therefore avoiding O(n²). The broad phase produces a set of potential collisions (a set of
+If your game field is small and do not have a lot of collideable components - you don't have to
+worry about the broad phase system that is used, so if the standard implementation is performant
+enough for you, you probably don't have to read this section.
+
+A broad phase is the first step of collision detection where potential collisions are calculated. To
+calculate these potential collisions are a lot cheaper to calculate than to check the exact
```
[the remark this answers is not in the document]
ASGAlex: Strange, that neither my IDE, nor "melos analyse" thinks so...

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
ASGAlex: Anyway this will lead user to use `collisionTypeNotifier` instead of `collisionType` when changing from "sweep" to "quadtree". It will become addigiotal requirement, different from previous collision system. It seems to me like it is bigger evil for framework design than having two additional functions. Or did I missunderstand you?

packages/flame/lib/src/collisions/quadtree/broadphase.dart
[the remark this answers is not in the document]
ASGAlex: I made a simple check: 
image
It also could be noticed at game screen, the two hitboxes without parent still alive.
It to omit this check, the system will sometimes stuck on a heap of dead hitboxes, sometimes a big heaps: 
image

packages/flame/lib/src/collisions/quadtree/broadphase.dart
ASGAlex: Because passive collision type usualy means that the component keeps it position, probably whole game. So it worth to cache it's center. But it is pointless for activelly moving object, which would have "active" collisionType most probably.

ASGAlex: Because passive collision type usualy means that the component keeps it position, probably whole game. So it worth to cache it's center. But it is pointless for activelly moving object, which would have "active" collisionType most probably.

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
ASGAlex: Because a "VoidCallback" limitation. I must know, on what item the transform had been performed. But I can't pass it as variable to callback. So... for every item a new callback instance created, and item instance is passed to callback's field of view. It's tricky, I agree. To change it for more 'pretty-looking' code we should to extend `Transform2D` from something more advanced than pure ChangeNotifier probably.

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
ASGAlex: It seems like it won't work for every component-hitbox hierarchy deeper than 2. For example when adding component inside component inside component inside...  (two days later) -  and then adding a hitbox. 
I think we should keep "default" behavior for `add` function, supporting simple hierarchy, but to create additional function like `void addQuadtreeHitbox(ShapeHitbox hitbox, Component transformedComponent)` where the `transformedComponent` is top-level component on which all transformation should be performed.

doc/flame/collision_detection.md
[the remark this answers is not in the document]
spydon: We don't have any markdown analysis tool in the stack atm, but the requirements are written in the contributors doc.

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
spydon: I think you did misunderstand me, or possibly I misunderstood you. These callbacks should only be sent within the library and not used by the user, right? And if so, the notifier can be made public and marked internal (hidden from the user, but open for us to use internally).

packages/flame/lib/src/collisions/collision_callbacks.dart
spydon: Hmm, I guess it could be seen as a callback. But it should be named something better then, because the user shouldn't even necessarily have to know what a broadphase is.
I can try to think of something tomorrow if you don't come up with something. :)

ASGAlex: I agree, something like "onComponentsTypeCheck" would be much better than current naming. But I'm open for propositions of course 👌

doc/flame/collision_detection.md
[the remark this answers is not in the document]
ASGAlex: Wow, I thought "melos analysis" follows project's codestyle guide 🤔 anyway, I already fixed this.

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
ASGAlex: Ah, got it! Will fix 👌

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
ASGAlex: It looks like I found better solution - to subscribe directly to aabb invalidation.

examples/lib/stories/collision_detection/quadtree_example.dart
```diff
@@ -0,0 +1,443 @@
...
+  }
+
+  late final QuadTreeNodeDebugInfo dbg;
+
+  @override
+  void render(Canvas canvas) {
+    final nodes = dbg.nodes;
+    // final textPaint = TextPaint(
+    //   style: const TextStyle(fontSize: 5.0, color: Colors.lightGreenAccent),
+    // );
```
spydon, line 423: This can be removed right?

examples/lib/stories/collision_detection/quadtree_example.dart
```diff
@@ -0,0 +1,443 @@
...
+      final nodeElements = node.ownElements;
+      Paint? boxPaint;
+      if (!node.noChildren && nodeElements.isNotEmpty) {
+        boxPaint = Paint();
+        boxPaint.style = PaintingStyle.stroke;
+        boxPaint.color = Colors.lightGreenAccent;
+        boxPaint.strokeWidth = 1;
+      }
+      for (final box in nodeElements) {
+        // textPaint.render(canvas, node.id.toString(), box.aabb.min);
```
spydon, line 435: This can be removed right?

packages/flame/lib/src/extensions/aabb.dart
spydon: Ah yes, you are totally right!

examples/lib/stories/collision_detection/quadtree_example.dart
```diff
@@ -0,0 +1,443 @@
...
+const tileSize = 8.0;
+
+class QuadTreeExample extends FlameGame
+    with HasQuadTreeCollisionDetection, KeyboardEvents, ScrollDetector {
+  QuadTreeExample();
+
+  static const description = '''
+In this example the standard "Sweep and Prune" algorithm is replaced by  
+"Quad Tree". Quad Tree is often a more efficient approach of handling collisions,
+it's efficiency is shown especially on huge maps with big amounts of collidable
```
spydon, line 21:
```suggestion
its efficiency is shown especially on huge maps with big amounts of collidable 
```

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
```diff
@@ -0,0 +1,83 @@
...
+          broadphase: QuadTreeBroadphase<ShapeHitbox>(
+            mainBoxSize: mapDimensions,
+            maxObjects: maxObjects,
+            maxDepth: maxDepth,
+            broadphaseCheck: onComponentTypeCheck,
+            minimumDistanceCheck: minimumDistanceCheck,
+          ),
+        );
+
+  QuadTreeBroadphase get quadBroadphase => broadphase as QuadTreeBroadphase;
```
spydon, line 25: Since I think very few (if anyone...) has implemented their own collision detection system I suggest that we instead do a small breaking change in the `CollisionDetection` class so that you can send in a type parameter for the broadphase, so that we don't have to have a separate field for that here.

So the `CollisionDetection` class would look like this instead of how it looks today:
```dart
CollisionDetection<T extends Hitbox<T>, B extends Broadphase<T>> (
  final B broadphase;
  ...
}
```

packages/flame/lib/src/collisions/quadtree/broadphase.dart
```diff
@@ -0,0 +1,161 @@
+import 'dart:collection';
```
spydon, line 1: Since there is another broadphase too I wouldn't name this file `broadphase`, but something more specific.
And maybe we can create a directory to keep the broadphase systems in, and then you can move out the mixins so that they are next to the other mixins in `collisions` (or possibly make a new directory for that too if you think that would be cleaner).

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
```diff
@@ -40,6 +50,9 @@ mixin ShapeHitbox on ShapeComponent implements Hitbox<ShapeHitbox> {
   final List<Transform2D> _transformAncestors = [];
   late Function() _transformListener;
 
+  @internal
+  Function()? onAabbChanged;
```
spydon, line 54: :+1:

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
spydon: See #1908, what do you think? We can discuss that on that PR.

examples/lib/stories/collision_detection/quadtree_example.dart
ASGAlex: I'm not sure, could it be useful for understanding or not: it prints element's current node ID, showing, how it changes while component moving from one node to another. 
But I commented this out, because text rendering burns out all performance optimisations: I have nearly 3 FPS with so huge map and text rendering on every Brick

packages/flame/lib/src/collisions/quadtree/broadphase.dart
ASGAlex: I usually try to deattach different implementations of an basic class from each other. But here we have multiple classes and mixins for each realisation. From my point of view it would be better to: 
 - move all Sweep-related classes/mixins to corresponding dir
 - move all Quadtree-related stuff to another dir
 - put "core" collision libraries somwhere else, maybe just keep at root of collision subdir 

I did not moved sweep files in my PR to avoid problems with branch updating from upstream while working on PR. Just dont like merge conflicts 🙃

examples/lib/stories/collision_detection/quadtree_example.dart
spydon: It sounds like it is too heavy to be useful for now then.

packages/flame/lib/src/collisions/quadtree/broadphase.dart
spydon: I would keep the actual collision detection systems directly in `collisions/` (including `QuadTreeCollisionDetection`) and then the files relating to the broadphase in `collisions/broadphase` and then the specific implementations in `collisions/broadphase/sweep` and `collisions/broadphase/quadtree`.

> I did not moved sweep files in my PR to avoid problems with branch updating from upstream while working on PR. Just dont like merge conflicts 🙃

We can move those in a separate PR. :)

ASGAlex: Yes, I'd prefer to do this in last turn, when all code changes will be performed  👍🏻

packages/flame/lib/src/collisions/quadtree/quadtree_collision_detection.dart
ASGAlex: I ready to merge this, as soon as it will be present in "main"

spydon: Great, now it is merged!

ASGAlex: done 👌

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
```diff
@@ -8,10 +8,20 @@ import 'package:meta/meta.dart';
 
 /// A [ShapeHitbox] turns a [ShapeComponent] into a [Hitbox].
 /// It is currently used by [CircleHitbox], [RectangleHitbox] and
-/// [PolygonHitbox].
+/// [PolygonHitbox]
```
spydon, line 11:
```suggestion
/// [PolygonHitbox].
```

packages/flame/lib/src/collisions/quadtree/broadphase.dart
spydon: That's not what the passive collision type means, say for example bullets, they would usually be passive since there would be alot of them and few things that they should hit, so the player and enemies would be active and the bullets passive to make the collision detection more efficient (this would lead to the bullets not being able to collide with themselves for example).

https://docs.flame-engine.org/1.3.0/flame/collision_detection.html?highlight=collisiontype#collisiontype

ASGAlex: It mostly depends on desired game logic. It might be a bit different, allowing bullets to collide only if they are fired by players of different fractions.

Anyway, it would be useful for collision system to know, what objects are static and what objects are dynamic - to cache its "center". It turns out to be too expensive to calculate it each time with getter. We can introduce another field for it, if you think that "CollisionType" is not a best choice.

spydon: Yeah, we can definitely not use `CollisionType` for this since that is already defined how it works.
Can't you just cache all center positions for now?

I don't think this issue should be handled in this PR, but for future reference, In box2d/forge2d what you are describing is called `BodyType`:
https://github.com/flame-engine/forge2d/blob/main/packages/forge2d/lib/src/dynamics/body_type.dart

packages/flame/lib/src/collisions/quadtree/broadphase.dart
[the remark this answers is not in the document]
spydon: Very strange, so `onRemove` is never called for these hitboxes?

packages/flame/lib/src/collisions/quadtree/broadphase.dart
ASGAlex: I think the best approach would be like with aabb invalidation: let getter get values always from cached variable an recalculate center and store to cache when transform2d changes.

packages/flame/lib/src/collisions/quadtree/broadphase.dart
```diff
@@ -0,0 +1,161 @@
...
+          _broadphaseCheckCache[item0 as T]![item1 as T] = false;
+        }
+      }
+    }
+    return _potentials;
+  }
+
+  void updateTransform(T item) {
+    tree.remove(item, keepOldPosition: true);
+    if (item.collisionType == CollisionType.passive) {
```
spydon, line 121: This check also have to take into consideration that passive hitboxes can move.

packages/flame/lib/src/collisions/quadtree/broadphase.dart
ASGAlex: fixed

packages/flame/lib/src/collisions/quadtree/broadphase.dart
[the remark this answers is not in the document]
ASGAlex: Fixed. It was quadtree "split" function, wich moved hitboxes into new nodes but did not uptates it in indexed HashSet.

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,386 @@
...
+  bool isMoved(T hitbox) {
+    final lastPos = _oldPositionByItem[hitbox];
+    if (lastPos == null) {
+      return true;
+    }
+    if (lastPos.min == hitbox.aabb.min && lastPos.max == hitbox.aabb.max) {
+      return false;
+    } else {
+      return true;
+    }
```
spydon, line 279:
```suggestion
    return lastPos.min != hitbox.aabb.min || lastPos.max == hitbox.aabb.max;
```

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,386 @@
...
+    final list = <T>[];
+    final parent = node.parent;
+    if (parent != null) {
+      list.addAll(parent.hitboxes as List<T>);
+      list.addAll(_getParentItems(parent));
+    }
+    return list;
+  }
+
+  bool isMoved(T hitbox) {
```
spydon, line 270:
```suggestion
  bool hasMoved(T hitbox) {
```

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
```diff
@@ -0,0 +1,100 @@
...
+  ///   to consider them as possibly colliding. You can also implement the
+  ///   [minimumDistanceCheck] if you need some custom behavior.
+  ///
+  /// The [onComponentTypeCheck] checks if objects of different types should
+  /// collide.
+  /// The result of the calculation is cached so you should not check any
+  /// dynamical parameters here, the function is intended to be used as pure
+  /// type checker.
+  /// It usually should not be overridden, see
+  /// [CollisionCallbacks.onComponentTypeCheck] instead
```
spydon, line 42:
```suggestion
  /// It should usually not be overridden, see
  /// [CollisionCallbacks.onComponentTypeCheck] instead.
```

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
```diff
@@ -0,0 +1,100 @@
...
+    if (cd is! QuadTreeCollisionDetection) {
+      throw 'Must be QuadTreeCollisionDetection!';
+    }
+    _collisionDetection = cd;
+  }
+
+  /// Initialise the QuadTree.
+  ///
+  /// - [mapDimensions] describes the collision area coordinates and size.
+  ///   Should match to game map position and size.
```
spydon, line 29:
```suggestion
  ///   Should match to game map's position and size.
```

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
```diff
@@ -0,0 +1,100 @@
...
+  /// - [maxObjects] (optional) - maximum objects count in one quadrant.
```
spydon, line 30:
```suggestion
  /// - [maxObjects] (optional) - maximum amoung of objects in one quadrant.
```

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
```diff
@@ -0,0 +1,100 @@
...
+  /// - [maxLevels] (optional) - maximum nested quadrants.
```
spydon, line 31:
```suggestion
  /// - [maxLevels] (optional) - maximum number of nested quadrants.
```

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
```diff
@@ -0,0 +1,100 @@
...
+
+  double? minimumDistance;
+
+  bool minimumDistanceCheck(Vector2 activeItemCenter, Vector2 potentialCenter) {
+    return minimumDistance == null ||
+        !((activeItemCenter.x - potentialCenter.x).abs() > minimumDistance! ||
+            (activeItemCenter.y - potentialCenter.y).abs() > minimumDistance!);
+  }
+
+  bool onComponentTypeCheck(PositionComponent one, PositionComponent another) {
```
spydon, line 67: Aren't these always `ShapeHitbox`es? Then both this and the collision callback can be changed to that, and then the collision callback should use `hitboxParent` instead of `parent`.

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+      checkParent = true;
+    }
+    final oneParent = one.parent;
+    final anotherParent = another.parent;
+
+    if (checkParent &&
+        oneParent is PositionComponent &&
+        anotherParent is PositionComponent) {
+      return onComponentTypeCheck(oneParent, anotherParent);
```
spydon, line 90: This shouldn't be needed if `hitboxParent` is used.

spydon (commented):
Some small final comments, this is very close to done now! Just tests that are missing.

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
ASGAlex: I can't replace types at function declaration because if I pass "parent" here, it would not to be a ShapeHitbox.

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
ASGAlex: Fixed, but it did not change things a lot, also there is now additional warning because of "protected" keyword on `hitboxParent`

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
spydon: Ah, of course, my bad.

packages/flame/lib/src/collisions/quadtree/quadtree.dart
```diff
@@ -0,0 +1,382 @@
...
+  List<T> get valuesRecursive {
+    final data = <T>[];
+
+    data.addAll(hitboxes);
+    for (final ch in children) {
+      if (ch == null) {
+        continue;
+      }
+      data.addAll(ch.valuesRecursive as List<T>);
+    }
```
spydon, line 347:
```suggestion
    for (final child in children) {
      if (child == null) {
        continue;
      }
      data.addAll(child.valuesRecursive as List<T>);
    }
```

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
ASGAlex: Did you mean "amount"?

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
spydon: Ah, fixed that in #1928 
It does change more than what is visible, because now `CompositeHitbox`es will work in this solution too.

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
spydon: Indeed I did. 😅

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
spydon: #1928 is now merged, so if you update this branch against `main` the warning should go away.

ASGAlex:
About tests. I see that all excisting collision tests using `testCollisionDetectionGame` could be reused - at least to ensure that all mechanics works correctly with both systems. It would be fastest way to cover all existing scenarios. 

What do you think about this approach @spydon ? 

Before I do any commits about this, let me show a small example: 

```dart

class HasQuadTreeCollidablesGame extends FlameGame
    with HasQuadTreeCollisionDetection {}

@isTest
Future<void> testQuadTreeCollisionDetectionGame(
  String testName,
  Future Function(HasCollisionDetection) testBody,
) {
  return testWithGame(
    testName,
    () {
      final game = HasQuadTreeCollidablesGame();
      game.initializeCollisionDetection(
          mapDimensions: const Rect.fromLTWH(0, 0, 1000, 1000),
      );
      return game;
    },
    testBody,
  );
}

group('Raycasting', () {
    Future raycasting0(HasCollisionDetection game) async {
      game.ensureAdd(
        PositionComponent(
          children: [RectangleHitbox()],
          position: Vector2(100, 0),
          size: Vector2.all(100),
          anchor: Anchor.center,
        ),
      );
      await game.ready();

      final ray = Ray2(
        origin: Vector2.zero(),
        direction: Vector2(1, 0),
      );
      final result = game.collisionDetection.raycast(ray);
      expect(result?.hitbox?.parent, game.children.first);
      expect(result?.reflectionRay?.origin, closeToVector(Vector2(50, 0)));
      expect(result?.reflectionRay?.direction, closeToVector(Vector2(-1, 0)));
    }

    testCollisionDetectionGame('one hitbox', raycasting0);
    testQuadTreeCollisionDetectionGame('one hitbox - QuadTree', raycasting0);
```

If this approach is acceptable, we should to find appropriate naming convention for test functions. Something like 'raycasting0' might be not a best choice, but in other hand copying something like "ray from bottom-right to top-left within CircleHitbox" into function name is not a best choice too 😉

spydon:
> About tests. I see that all excisting collision tests using `testCollisionDetectionGame` could be reused

That sounds like a good idea!

I wonder if not the whole files could be reused if we just created a list of tests in them that we could pass in the specific game factories to. Then we wouldn't have to do this on a per test basis.

ASGAlex:
I have done with tests - looks like old tests are reused correctly 🎉

doc/flame/collision_detection.md
```diff
@@ -197,38 +197,124 @@ collisions to the whole hat, instead of for just each hitbox separately.
...
+    if (other is Brick) {
+      removeFromParent();
+    }
+    super.onCollisionStart(intersectionPoints, other);
+  }
+}
+```
+
+After intensive gameplay a map could become over-clusterized with a lot of empty quadrants. Also some
+objects with null `parent` could still be present in the `QuadTree` but being removed from the game's
```
spydon, line 292: This issue with the null parents was resolved right?

packages/flame/lib/src/collisions/quadtree/has_quadtree_collision_detection.dart
```diff
@@ -0,0 +1,96 @@
+import 'package:flame/collisions.dart';
+import 'package:flame/components.dart';
+import 'package:flame/extensions.dart';
+import 'package:flame/game.dart';
+
+/// Mixin should be applied to FlameGame to bring QuadTree collision support.
+/// [initializeCollisionDetection] should be called at [onLoad] function.
+///
```
spydon, line 8:
```suggestion
/// This should be applied to a [FlameGame] to bring QuadTree collision
/// support.
///
/// [initializeCollisionDetection] should be called in the game's [onLoad]
/// method.
```

packages/flame/test/collisions/collision_test_helpers.dart
```diff
@@ -14,6 +18,35 @@ Future<void> testCollisionDetectionGame(
...
+      game.initializeCollisionDetection(
+        mapDimensions: const Rect.fromLTWH(0, 0, 1000, 1000),
+      );
+      return game;
+    },
+    testBody,
+  );
+}
+
+Future<void> runTestRegistry(
```
spydon, line 39:
```suggestion
Future<void> runCollisionTestRegistry(
```
It should have a name that is specific to collision detection since it only handles those cases at the moment.

spydon (approved):
Some very minor last comments.
You've done an amazing job! 🎉

spydon (approved):

wolfenrain (approved):

spydon:
@ASGAlex if you can just fix the merge conflict this PR is ready to be merged! 😍

ASGAlex:
15 minutes 👌

ASGAlex:
@spydon  looks like done =^_^=

spydon:
> @spydon looks like done =^_^=

Great job! I think it ended up quite nice :)

max-zamkovoy:
@ASGAlex congrats! You did great job! :)

packages/flame/lib/src/collisions/quadtree/broadphase.dart
ASGAlex: @spydon  I just opened new PR about it: https://github.com/flame-engine/flame/pull/1943#issue-1388132672

PR #1895, opened 2022-09-10
feat: Add support for isometric staggered maps

### LINKED ISSUE #1893: Support Tiled's Isometric Staggered
# What could be improved
Support the last layout featured in Tiled - Isometric Staggered.  In this mode, the tiles do not follow the isometric coordinate system, but instead resemble the hexagonal staggering.

# Why should this be improved
I'm a completionist.

# Any risks?
None.

# More information
I'll work on this PR

jtmcdole:
# Description

Adds support for Tiled's Isometric Staggered maps
- Also fixes improper scaling on hexagonal maps.

image

## Checklist

## Breaking Change

## Related Issues

fixes #1893
similar to #1887 and #1882


packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -459,11 +565,11 @@ class _RenderableTileLayer extends _RenderableLayer<TileLayer> {
         //  * stagger: Hexagonal tiles move down or right by only a fraction,
         //             specifically 3/4 the width or height, for packing.
         if (_map.staggerAxis == StaggerAxis.y) {
-          offsetX = tx * _map.tileWidth + staggerX + halfDestinationTile.x;
```
jtmcdole, line 462: This was a bug for scaled hexagonal tiles. Fixed here and not copy-pastaed to isometric staggered.

packages/flame_tiled/lib/src/tiled_component.dart
```diff
@@ -73,7 +73,7 @@ class TiledComponent<T extends FlameGame> extends Component with HasGameRef<T> {
     if (tMap.orientation == MapOrientation.hexagonal) {
       if (tMap.staggerAxis == StaggerAxis.y) {
```
spydon, line 74: Do you think this would be easier to read if it was switch case?

spydon (approved):
Good stuff as always! Could you add an image to the PR description with an example of the outcome when using this?

packages/flame_tiled/lib/src/renderable_tile_map.dart
jtmcdole: > Good stuff as always! Could you add an image to the PR description with an example of the outcome when using this?

will do - It seems I actually need to revisit the painting order in X stagger order - if we walk the traditional row/col,
3D like tiles are painted col-first, leading to ugly overlaps.

packages/flame_tiled/lib/src/tiled_component.dart
jtmcdole: Done.

jtmcdole:
Converted to draft.

jtmcdole:
Back on the hotplate!

jtmcdole:
Let me check on the test failure here.  Now that the layer offset is in, I will update some of my tests to have layer diffs.

jtmcdole:
@spydon :  changes in this last pull request:

1. Merged with head and all the goodness there (groups, parallax, opacity)
2. Fixed some minor bugs in group to call all children for refresh and handle resize
3. **cleared the sprite batches of all cached transforms** when resize called.

For 3: there's multiple refreshes at head, so the maps are being painted multiple times for a perf hit. Not any more!

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -463,16 +474,134 @@ class _RenderableTileLayer extends _RenderableLayer<TileLayer> {
...
+      // painted over. See the second pass loop after tx.
+      final xSecondPass = <_Transform>[];
+
+      for (var tx = 0; tx < tileRow.length; tx++) {
+        final tileGid = tileRow[tx];
+        if (tileGid.tile == 0) {
+          continue;
+        }
+
+        final tile = _map.tileByGid(tileGid.tile);
```
kurtome, line 510: in the next version we'll have to add `!` when getting `tileByGid` because `tiled.dart` has some changes in the next version and this returns `Tile?` now

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -454,7 +465,7 @@ class _RenderableTileLayer extends _RenderableLayer<TileLayer> {
     }
   }
 
-  void _cacheHexagonalTiles() {
+  void _cacheIsometricStaggeredTiles() {
```
kurtome, line 468: Is there anything we can do to refactor these giant methods into something easier to read? Maybe even just putting these methods in a different file?

kurtome (commented):
The geometry math is over my head, but overall makes sense to me

packages/flame_tiled/lib/src/renderable_tile_map.dart
jtmcdole: I thought breaking them into individual functions was enough. Renderable layer could have subclasses maybe? I can look in another pr.

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -333,12 +332,24 @@ class _RenderableTileLayer extends _RenderableLayer<TileLayer> {
...
+        _cacheIsometricTiles();
+        break;
+      case MapOrientation.staggered:
+        _cacheIsometricStaggeredTiles();
+        break;
+      case MapOrientation.hexagonal:
+        _cacheHexagonalTiles();
+        break;
+      case MapOrientation.orthogonal:
+      default:
```
spydon, line 350: If you exhaust the enum options here you don't need the default case. It's actually probably better to remove it, because then if you add another value to the enum then the analyzer will complain here that it doesn't check all values, like it should.

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -532,25 +666,40 @@ class _RenderableTileLayer extends _RenderableLayer<TileLayer> {
...
+        final transform = ui.RSTransform(
+          scos,
+          ssin,
+          offsetX + -scos * anchorX + ssin * anchorY,
+          offsetY + -ssin * anchorX - scos * anchorY,
+        );
+
+        // A second pass is only needed in the case of staggery.
+        if (_map.staggerAxis == StaggerAxis.x && staggerY > 0) {
+          xSecondPass.add(_Transform(src, transform, flips.flip, batch));
```
spydon, line 688: It will add some overhead with creating all these new objects, but maybe there is no good way around it.

packages/flame_tiled/lib/src/tiled_component.dart
```diff
@@ -63,35 +63,41 @@ class TiledComponent<T extends FlameGame> extends Component with HasGameRef<T> {
...
+                tileScaled.y + ((tileMap.map.height - 1) * tileScaled.y * 0.75),
+              )
+            : Vector2(
+                tileScaled.x + ((tileMap.map.width - 1) * tileScaled.x * 0.75),
+                (tileMap.map.height * tileScaled.y) + tileScaled.y / 2,
+              );
+
+      case MapOrientation.isometric:
+      case MapOrientation.orthogonal:
+      default:
```
spydon, line 96: Same with default here

spydon (approved):

packages/flame_tiled/lib/src/renderable_tile_map.dart
jtmcdole: You either do this, or you do:

- for (even){}, for(odd){}. // stagger x even 
- for (odd){}, for(even){}. // stagger x odd

 What was told me by some of the Dart VM Guys: create a lot of garbage quickly and give it up - new heap is fast.
 
 Honestly, the transforms on SpriteBatch should be fungible. I should be able to reorder and update individual elements.

packages/flame_tiled/lib/src/renderable_tile_map.dart
jtmcdole: will do

packages/flame_tiled/lib/src/renderable_tile_map.dart
spydon: When did they say this? Because we've had a lot of problems with creating too many objects previously.

jtmcdole: It was certainly a while ago! That said, this is only done on cache refresh - not every frame, so this is totally fine I think.

spydon: Yeah, we can just re-visit this if we notice any performance issues here.

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -463,16 +477,134 @@ class _RenderableTileLayer extends _RenderableLayer<TileLayer> {
...
+        final src = tileset.computeDrawRect(tile).toRect();
+        final flips = SimpleFlips.fromFlips(tileGid.flips);
+        final scale = size.x / src.width;
+        final anchorX = src.width - halfMapTile.x;
+        final anchorY = src.height - halfMapTile.y;
+
+        late double offsetX;
+        late double offsetY;
+
+        // halfTile.x: shfits the map half a tile forward rather than
```
wolfenrain, line 544:
```suggestion
        // halfTile.x: shifts the map half a tile forward rather than
```

wolfenrain (approved):

packages/flame_tiled/test/tiled_test.dart
```diff
@@ -463,6 +463,128 @@ void main() {
...
+    test('y + odd + half', () async {
+      await setupMap(
+        'iso_staggered_overlap_y_odd.tmx',
+        'dirt_atlas.png',
+        Vector2(128 / 2, 64 / 2),
+      );
+
+      expect(component.size, Vector2(576 / 2, 160 / 2));
+
+      final canvasRecorder = PictureRecorder();
```
erickzanardo, line 547: Nit: Maybe you could use Flame layers for a cleaner code?

erickzanardo (approved):

packages/flame_tiled/test/tiled_test.dart
jtmcdole: Hi! I'd love to have cleaner code - is this what you're referring to?

https://pub.dev/documentation/flame/latest/layers/Layer-class.html

And is there any examples in the flame test code I can use to get up to speed on it?

packages/flame_tiled/lib/src/renderable_tile_map.dart
jtmcdole: done

packages/flame_tiled/test/tiled_test.dart
spydon: Maybe this could help: https://examples.flame-engine.org/#/Rendering_Layers
Get to the code by pressing `< >` in the upper right corner.

jtmcdole: The question that I think remains: is this about the test code or is this a suggestion that each RenderableTileLayer is now a FlameLayer?

jtmcdole: (a search for `PreRenderedLayer` and `DynamicLayer` in *test.dart is not turning up anything).

erickzanardo: I was thinking on the tests itself, indeed PreRenderLayer and DunamicLayer isn't used in any tests. But since those classes basically wrap a PictureRecoder, it could make these tests cleaner, since you won't need to instantiate the Canvas and the PictureRecorder yourself.

But again this is just a nit. PR looks good to me already.

spydon: I think it is several years since anyone did any real changes to that file, so from before we required tests unfortunately.

jtmcdole: Yep - I could do a non-lazy thing and move the canvas/picturerecorder/give me an image code into a function since its the same method over and over.. I can do this later tonight (lunch break over):

```dart
Future<Uint8List> generatePngData(Component component, Vector2 size) async {
      final canvasRecorder = PictureRecorder();
      final canvas = Canvas(canvasRecorder);
      component.tileMap.render(canvas);
      final picture = canvasRecorder.endRecording();
      
      final image = await picture.toImageSafe(size.x, size.y);
      return (await image.toByteData(format: ImageByteFormat.png))!
          .buffer
          .asUint8List();      
}
```

spydon: No rush, love that Flame is getting some love even on the lunch breaks. :smile:

jtmcdole: done

PR #1897, opened 2022-09-10
docs: Added Style Guide and Test Writing Guide

### LINKED ISSUE #1772: Integrate STYLEGUIDE.md into the main documentation site
# What could be improved

The STYLEGUIDE.md should be merged into the main documentation site, for better visibility. In addition, the guide needs to be substantially cleaned up as it contains content that is either outdated, or enforced via lint rules. At the same time, it can be expanded to included some new conventions that we have introduced over time.

Currently that guide is not linked from anywhere, so it is unlikely anybody reads that document.

# Any risks?

None.

st-pasha:
# Description

The Style Guide moved into the Documentation site for better visibility. Removed lots of discussion around topics that are already enforced by the formatter/linter. Added new advice regarding a variety of topics.

The Test Writing Guide discusses how to write tests for Flame, and what kinds of tests are available.

## Checklist

## Breaking Change

## Related Issues

Closes #1772 


CONTRIBUTING.md
```diff
@@ -68,18 +68,18 @@ and it will be automatically reflected in the PR.
...
 ### Environment Setup
-Flame uses [Melos] to manage the project and dependencies.
 
-To install Melos, run the following command from your terminal:
+Flame uses [Melos] to manage the project and dependencies. To install Melos, run the following
+command from your terminal:
 
 ```bash
-flutter pub global activate melos
+$ flutter pub global activate melos
```
spydon, line 76: I find that adding these makes it harder for users to copy the line, and I think people understand that they shouldn't run it as root. (And if they don't understand that they should run it in a shell they probably don't know what `$` is anyways 😅)

doc/development/style_guide.md
```diff
@@ -0,0 +1,211 @@
...
+This is a general style guide for writing code within Flame and adjacent projects. We strive to
+maintain the code clean and readable -- both for the benefit of the users who will need to study
+this code in order to understand how a particular feature works or debug something that is breaking,
+and for the benefit of the current and future maintainers.
+
+This guide extends upon the official style guides of Dart and Flutter. Those guides should be
+considered an essential part of this style guide. However, whenever some rules contradict others,
+the following order of preferences apply:
+- **Flame Style Guide** (this document)
+- **[Flutter Style Guide]**
```
spydon, line 12: This one is the internal style guide for the Flutter project as far as I'm aware and shouldn't be included here.

doc/development/style_guide.md
```diff
@@ -0,0 +1,211 @@
...
+$ flutter analyze
+$ flutter format .
+```
+
+
+## Code Structure
+
+### Imports
+
+- Avoid importing groups of APIs internally, for example, importing `lib/effects.dart` just to use
```
spydon, line 30: I don't think we want to enforce this, I think it will just be a lot more imports which doesn't add any readability, and since the compiler has tree shaking it makes no difference if more things would be included in an import.

doc/development/style_guide.md
```diff
@@ -0,0 +1,211 @@
...
+- Never import `package:flutter/cupertino.dart` or `package:flutter/material.dart` -- prefer
+  a much smaller library `package:flutter/widgets.dart` if you're working with widgets.
+
+
+### Exports
+
+- Strongly prefer to have only one public class per file, and name the file after that class.
+  Having several private classes within the file is perfectly reasonable.
+
+- A possible exception to this rule if the "main" class requires some small "helper" classes that
```
spydon, line 46:
```suggestion
- A possible exception to this rule is if the "main" class requires some small "helper" classes that
```

doc/development/style_guide.md
```diff
@@ -0,0 +1,211 @@
...
+  how the class ought to be used.
+
+- Try to make as much of your class' API private as possible, do not expose members "just in case".
+  This makes it much easier to modify/refactor the class later without it being a breaking change.
+
+  Remember to document all your public members! Documenting things is harder than it looks, and one
+  way to avoid the burden of documentation is to make as many variables private as possible.
+
+- If a class exposes a `List<X>` or `Vector2` property -- it is **NOT** an invitation to modify
+  them at will! Consider such properties as read-only, unless the documentation explicitly says that
```
spydon, line 112: A lot of the time we actually want the users to modify the `Vector2`s directly? So that they don't need to create new objects just to update the value.

doc/development/style_guide.md
```diff
@@ -0,0 +1,211 @@
...
+  ```
+
+- If a class has a private member that needs to be exposed via a getter/setter, prefer the following
+  code structure:
+  ```dart
+  class MyClass {
+    MyClass();
+
+    ...
+    int _variable;
```
spydon, line 130: Don't we usually have this in the block below the dartdoc?

doc/development/style_guide.md
```diff
@@ -0,0 +1,211 @@
...
+
+- Use markdown documentation in `doc/` folder to give the high-level overview of the functionality,
+  and especially how it fits into the overall Flame framework.
+
+
+### Dartdocs
+
+- Check the [Flutter Documentation Guide] -- it contains lots of great advice on writing good
+  documentation.
+    - However, disregard the advice about writing in a passive voice.
```
spydon, line 158: Hehe they are not even following that recommendation themselves.

doc/development/style_guide.md
```diff
@@ -0,0 +1,211 @@
...
+  Avoid stating the obvious (or at least only the obvious).
+
+- Constructor documentation may follow either the style of a method (i.e. "Creates ...",
+  "Constructs ..."), or of the class but omitting the name of the class (i.e. "Rectangular-shaped
+  component"). Constructor documentation may be omitted if (1) it's the main constructor of the
+  class, and (2) all the parameters are obvious and coincide with the public members of the class.
+
+  **Do not** use macros to copy the class documentation into the constructor's dartdoc. Generally,
+  the class documentation answers the question "what the class is", whereas the constructor docs
+  answer "how it will be constructed".
```
spydon, line 190: Interesting, aren't they using macros like this in flutter?

STYLEGUIDE.md
```diff
@@ -1,165 +0,0 @@
-# Style Guide
```
spydon, line 1: Can we symlink this, or just keep an empty doc with a link to the site? Because I think GitHub encourages repos to contain this file in the root.

st-pasha: I checked at 
- https://medium.com/code-factory-berlin/github-repository-structure-best-practices-248e6effc405 
- https://github.com/joelparkerhenderson/github-special-files-and-paths

and neither lists `STYLEGUIDE` as a "standard" file of which GitHub is in any way aware.

CONTRIBUTING.md
st-pasha: Right, I'll put them back for the CONTRIBUTING doc. In other docs the `$` sign is not selectable, so it doesn't interfere with anything.

doc/development/style_guide.md
st-pasha: Alright, I'll remove the Flutter Style Guide then.

doc/development/style_guide.md
spydon: So that all private variables would be in one block? It feels like it would be easier to find/see the variable if it was within the same block as the getters and setters, so either between them or under the setter.

STYLEGUIDE.md
spydon: I mixed it up with the `CONTRIBUTING.md` doc.

doc/development/style_guide.md
st-pasha: Come to think of it, I actually like having all private variables in one block. It gives you an ability to look at the class and at a glance see how light or heavy it is. As opposed to the current situation, where in order to see the size of the class, say `Component`, one needs to scroll through a thousand of lines of code.

Compared to that, finding the variable from its getter is both easy (a simple shortcut to go to variable's definition), and rarely needed (since the getter already tells you the name and the type of the variable).

spydon: True, that is a good argument. Maybe add aline about all private variables being in one block then?

spydon (approved):

doc/development/style_guide.md
```diff
@@ -0,0 +1,203 @@
...
+The main documentation site serves as a place where people go to learn about various functionality
+available in Flame. If you're adding a new class, then it must be documented both at the dartdocs
+level, and on the main docs site. The latter serves the purposes of discoverability of your class.
+Without the docs site, your class might never be used (or at least used less than it could have
+been), because the developers would simply not know about it.
+
+When adding the documentation to the main docs site, consider also including an example directly
+into the docs. This will make readers more excited about trying this new functionality.
+
+Check the [Documentation] manual about how to work with the docs site.
```
spydon, line 198: Should we add a few sentences about the markdown structures that we are following too? Like two lines before a new section, max 100 chars on a row etc?

st-pasha: Right, I forgot about those. Will add.

st-pasha: Done.

spydon (approved):

doc/development/testing_guide.md
```diff
@@ -0,0 +1,160 @@
...
+  await tester.pump(const Duration(milliseconds: 20));
+});
+```
+There are some additional methods available on the `tester` controller, for example in order to
+simulate taps, or drags, or key presses.
+
+
+### Golden tests
+
+These tests verify that a component renders correctly. The process of creating a golden test is
```
erickzanardo, line 107: Not only components

erickzanardo (approved):

doc/development/testing_guide.md
```diff
@@ -0,0 +1,160 @@
...
+simple:
```
spydon, line 108:
```suggestion
These tests verify that things render as intended. The process of creating a golden test is
simple:
```

PR #1900, opened 2022-09-11
feat: Tiled component is positionable

### LINKED ISSUE #1890: Make TiledMapComponent positionable
# What could be improved
`TiledMapComponent` extends `Component` - in discord conversations it was agreed that they should extend `PositionComponents`

bonus points: tiles could also be positionable, leading to some interesting animations of tiles entering the world.

# Why should this be improved
No other way to position aside from wrapping the map in a Position component.  No way to position tils either.  If tiles existed as individual components; items could just be easily added to them with "add()".

# Any risks?
Rendering would have to visit objects to do the same as today's:

```dart
_cachedSpriteBatches.values.forEach((batch) {
      batch.render(canvas, paint: _layerPaint);
    });
```

# More information
* I'm interested in working on it.
* Potential breaking change for anyone depending on tiles?

jtmcdole:
# Description

Update `TiledComponent` to extend `PositionComponent` instead of `Component`. With some quirks:

1. Size is computed from the Map and not settable. I've made these operations fail silently for now.
2. Width and Height setters by proxy are ignored.
3. Scale does work

## Checklist

## Breaking Change

I do not believe this is breaking, since tests are still passing. It is only adding functionality to TiledComponent.

## Related Issues

Fixes #1890

### DIFF
--- a/packages/flame_tiled/lib/src/tiled_component.dart
+++ b/packages/flame_tiled/lib/src/tiled_component.dart
@@ -21,21 +21,21 @@ class TiledComponent<T extends FlameGame> extends PositionComponent
   /// [PositionComponent] larger or smaller, change its [scale].
   @override
   set size(Vector2 size) {
-    // Intenitonally left empty
+    // Intentionally left empty.
   }
 
   /// This property **cannot** be reassigned at runtime. To make the
   /// [PositionComponent] larger or smaller, change its [scale].
   @override
   set width(double w) {
-    // Intenitonally left empty
+    // Intentionally left empty.
   }
 
   /// This property **cannot** be reassigned at runtime. To make the
   /// [PositionComponent] larger or smaller, change its [scale].
   @override
   set height(double h) {
-    // Intenitonally left empty
+    // Intentionally left empty.
   }
 
   /// {@macro _tiled_component}
--- a/packages/flame_tiled/lib/src/tiled_component.dart
+++ b/packages/flame_tiled/lib/src/tiled_component.dart
@@ -79,7 +79,7 @@ class TiledComponent<T extends FlameGame> extends PositionComponent
     );
   }
 
-  static NotifyingVector2 _computeSize(RenderableTiledMap tileMap) {
+  static Vector2 _computeSize(RenderableTiledMap tileMap) {
     final tMap = tileMap.map;
 
     final xScale = tileMap.destTileSize.x / tMap.tileWidth;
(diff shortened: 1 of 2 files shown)


packages/flame_tiled/lib/src/tiled_component.dart
```diff
@@ -12,20 +12,42 @@ import 'package:tiled/tiled.dart';
...
+  set width(double w) {
+    // Intenitonally left empty
+  }
+
+  /// This property **cannot** be reassigned at runtime. To make the
+  /// [PositionComponent] larger or smaller, change its [scale].
+  @override
+  set height(double h) {
+    // Intenitonally left empty
+  }
```
ufrshubham, line 39: You can mark these setters with `@internal` so that users are warned when they try to use them.

jtmcdole: Tiled Component is the public API for this package.  `@internal` would throw an error for exposing it in the public API, no?  If you add `TiledComponent` to a list of `PositionComponent`, you wouldn't get that warning.

from pub.dev api:
> the declaration is declared in a package's public API, or is exposed from a package's public API, or

ufrshubham: Oh, my bad. I forgot that this is flame_tiled 😅

spydon:
#1895 is now merged  :)

jtmcdole:
> #1895 is now merged :)

updated - took a bit to unbork the stack (dear future self: `git rebase --onto main tiledComponentIsPositionable~ tiledComponentIsPositionable` does the trick)

jtmcdole:
recorded from chrome (obs was choppy - animation is great): 
https://user-images.githubusercontent.com/1924313/189997161-ac7f9ee7-55c2-4963-87b2-00174ab5b1ff.mp4

packages/flame_tiled/lib/src/tiled_component.dart
```diff
@@ -12,20 +12,42 @@ import 'package:tiled/tiled.dart';
...
   RenderableTiledMap tileMap;
 
-  /// The logical size of the component. The game assumes that this is the
-  /// approximate size of the object that will be drawn on the screen.
-  late final Vector2 size = _computeSize();
+  /// This property **cannot** be reassigned at runtime. To make the
+  /// [PositionComponent] larger or smaller, change its [scale].
+  @override
+  set size(Vector2 size) {
+    // Intenitonally left empty
```
spydon, line 24:
```suggestion
    // Intentionally left empty.
```

packages/flame_tiled/lib/src/tiled_component.dart
```diff
@@ -12,20 +12,42 @@ import 'package:tiled/tiled.dart';
...
+  }
+
+  /// This property **cannot** be reassigned at runtime. To make the
+  /// [PositionComponent] larger or smaller, change its [scale].
+  @override
+  set width(double w) {
+    // Intenitonally left empty
```
spydon, line 31:
```suggestion
    // Intentionally left empty.
```

packages/flame_tiled/lib/src/tiled_component.dart
```diff
@@ -12,20 +12,42 @@ import 'package:tiled/tiled.dart';
...
+  }
+
+  /// This property **cannot** be reassigned at runtime. To make the
+  /// [PositionComponent] larger or smaller, change its [scale].
+  @override
+  set height(double h) {
+    // Intenitonally left empty
```
spydon, line 38:
```suggestion
    // Intentionally left empty.
```

packages/flame_tiled/lib/src/tiled_component.dart
```diff
@@ -57,7 +79,7 @@ class TiledComponent<T extends FlameGame> extends Component with HasGameRef<T> {
     );
   }
 
-  Vector2 _computeSize() {
+  static NotifyingVector2 _computeSize(RenderableTiledMap tileMap) {
```
spydon, line 82: Does it need to be a `NotifyingVector2` here?

spydon (approved):

packages/flame_tiled/lib/src/tiled_component.dart
jtmcdole: You cannot change the type from position component .

spydon: `PositionComponent` takes in a `Vector2` for super though, where this is the only place that you use this?

jtmcdole: Not implementing `PositionComponent` means we'd have to implement: 

```
        AnchorProvider,
        AngleProvider,
        PositionProvider,
        ScaleProvider,
        CoordinateTransform
```
And a lot of other logic from there. If PositonComponent only gave you a Vector2 - and had another way to register an observer on the vector, then maybe that's a work around?

spydon: No no, I'm not suggesting you should not extend `PositionComponent`, I'm simply asking to change the return of this static method to `Vector2` since everything will still work like that. 😅
(Size will still be a `NofifyingVector2` in the end, but this method doesn't need to expose that)

jtmcdole: /me squints

I totally missed that context. Sorry! Ill see if I can do that at lunch time.

PR #1906, opened 2022-09-14
fix: Take scale into account

### LINKED ISSUE #1904: Minor Bug: Layer offsets in tiled map don't respect scaling
# Current bug behaviour
TiledMaps with destination tile sizes smaller/larger than the Tiled map have offsets that do not scale.

Two examples:
image

The offsets calculated in #1886 need to include scale.

# Expected behaviour
Small or larger maps look like regular maps... just smaller or bigger.

# Steps to reproduce
Use a TiledMap - with a destination tile size different from the actual map. The map should included layer offsets.

# More information

I'll put up a fix. Offsets should also not be calculated per-frame.

jtmcdole:
# Description

- Take scale into account when calculating offsets
- Memoize the work of offsets - don't calculate it each time.
- Tests - everyone loves tests.

In scaled destination tile size, this would render the pyramids in different locations: 
shifted_scaled_regular

Example pre-fix wrongness:
shifted_scaled_smaller_maskedDiff

## Checklist

## Breaking Change

## Related Issues

Fixes #1904


spydon (approved):

spydon:
@kurtome @ufrshubham could you review this PR too? :)

kurtome:
LGTM, although can you double check that `_applyParallaxOffset` looks right, it takes `camera.zoom` into account so I'm not sure if additional change are needed there. Maybe not since it's all relative anyways...

jtmcdole:
> LGTM, although can you double check that `_applyParallaxOffset` looks right, it takes `camera.zoom` into account so I'm not sure if additional change are needed there. Maybe not since it's all relative anyways...

Sure thing. Probably needs a test as well.

jtmcdole:
> > LGTM, although can you double check that `_applyParallaxOffset` looks right, it takes `camera.zoom` into account so I'm not sure if additional change are needed there. Maybe not since it's all relative anyways...
> 
> Sure thing. Probably needs a test as well.

So it doesn't look like it uses the destination tile size into account.. and I'm not sure it should. Yes, destination tile scale might seem like a cheap "scale" or "zoom", but it could just be sizing different maps to be the same for mixing. I don't know what the original intent was behind having it passed in versus reading it from the map.... and now that Tiled Component is positionable - it's also scalable. 

That mixed with my unfamiliarity with parallax and I'm not sure.

kurtome:
> > > LGTM, although can you double check that `_applyParallaxOffset` looks right, it takes `camera.zoom` into account so I'm not sure if additional change are needed there. Maybe not since it's all relative anyways...
> > 
> > 
> > Sure thing. Probably needs a test as well.
> 
> So it doesn't look like it uses the destination tile size into account.. and I'm not sure it should. Yes, destination tile scale might seem like a cheap "scale" or "zoom", but it could just be sizing different maps to be the same for mixing. I don't know what the original intent was behind having it passed in versus reading it from the map.... and now that Tiled Component is positionable - it's also scalable.
> 
> That mixed with my unfamiliarity with parallax and I'm not sure.

My hunch is that it will work fine as is because parallax is really a function of solely the current Viewport/Camera relative to the 0,0 position of the map. And `0.5` parallax is going to do the same thing regardless of the tile size (I think)

ufrshubham:
> parallax is really a function of solely the current Viewport/Camera relative to the 0,0 position of the map

This is correct. Tile size won't matter for parallax. I'll try to play around with the changes just to be sure. But if I don't post anything within the next few days, consider a LGTM from my side.

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -237,51 +245,39 @@ class RenderableTiledMap {
...
-  double get parallaxY {
-    if (parent != null) {
-      return parent!.parallaxY * layer.parallaxY;
-    } else {
-      return layer.parallaxY;
-    }
-  }
+  late double parallaxX = layer.parallaxX * (parent?.parallaxX ?? 1);
+
+  late double parallaxY = layer.parallaxY * (parent?.parallaxY ?? 1);
```
ufrshubham, line 280: Any specific reason to not have these as getters?

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -237,51 +245,39 @@ class RenderableTiledMap {
 
 abstract class _RenderableLayer<T extends Layer> {
   final T layer;
+  final Vector2 destTileSize;
+  final TiledMap map;
```
ufrshubham, line 249: Is it possible to not have these in here? It look like it is unnecessarily increasing all the derived classes as well. For example, `_UnrenderableLayer` is not even using them.

packages/flame_tiled/lib/src/renderable_tile_map.dart
```diff
@@ -237,51 +245,39 @@ class RenderableTiledMap {
...
 
   bool get visible => layer.visible;
 
   void render(Canvas canvas, Camera? camera);
 
-  void handleResize(Vector2 canvasSize) {}
+  void handleResize(Vector2 canvasSize);
 
-  void refreshCache() {}
+  void refreshCache();
```
ufrshubham, line 267: Maybe better to leave these with a default implementation. Will avoid some empty overrides in derived classes.

ufrshubham (commented):
Just some structural comments. Overall the changes look solid!

packages/flame_tiled/lib/src/renderable_tile_map.dart
jtmcdole: Which ones? Offset, Opacity, or Parallax? 

Main reason: performance.  As getters, you're calculating this every single render loop. As late-calculations, they are calculated once and cached.

packages/flame_tiled/lib/src/renderable_tile_map.dart
jtmcdole: Yes, I think it's possible. Let work on it.

jtmcdole: Actually; they are needed for the scale and offset calculations.  Tile, Image, and Group uses these. `_UnrenderableLayer` is just unlucky.  `RenderableTiledMap` is allocated after the layers are created... maybe we just update the layers to point to `RenderableTiledMap` as the parent map?

packages/flame_tiled/lib/src/renderable_tile_map.dart
jtmcdole: `_RenderableLayer` is abstract. Having these as default implementations actual lead to the bug that this is fixing.

packages/flame_tiled/lib/src/renderable_tile_map.dart
ufrshubham: As per our discussion on discord, this is few as these values never change.

packages/flame_tiled/lib/src/renderable_tile_map.dart
ufrshubham: Got idea. Also, unrenderable layer is basically the object layer and technically it can contain renderable tile objects. So this change is actually useful for adding rendering support for the object layer in future.

ufrshubham:
These changes LGTM 🔥

PR #1908, opened 2022-09-15
feat: Add broadphase generics to CollisionDetection

spydon:
# Description

This adds generics for the broadphase used in the collision detection.
This should simplify things a bit in #1894.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/flame/lib/src/collisions/collision_detection.dart
+++ b/packages/flame/lib/src/collisions/collision_detection.dart
@@ -7,8 +7,9 @@ import 'package:flame/geometry.dart';
 ///
 /// If the [HasCollisionDetection] mixin is added to the game, [run] is
 /// called every tick to check for collisions
-abstract class CollisionDetection<T extends Hitbox<T>> {
-  final Broadphase<T> broadphase;
+abstract class CollisionDetection<T extends Hitbox<T>,
+    B extends Broadphase<T>> {
+  final B broadphase;
   List<T> get items => broadphase.items;
   final Set<CollisionProspect<T>> _lastPotentials = {};
 
--- a/packages/flame/lib/src/collisions/has_collision_detection.dart
+++ b/packages/flame/lib/src/collisions/has_collision_detection.dart
@@ -3,12 +3,13 @@ import 'package:flame/game.dart';
 
 /// Keeps track of all the [ShapeHitbox]s in the component tree and initiates
 /// collision detection every tick.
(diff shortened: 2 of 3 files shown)


spydon:
@ASGAlex

ASGAlex:
Since we're talking about breaking changes... 

To be honest, all these generics in collision detection system looks overcomplicated. The `<T extends Hitbox<T>>` is specified everywhere, but in real practice only the `ShapeHitbox` makes sense. Why should't we to use this base class instead generic type? If to specify `ShapeHitbox` explicitly in the `Broadphase` and `CollisionDetection` classes, then we will not need to pass these generics everywhere. And it would allow to avoid unnecesary type conversions at many places. 

As I can see, there are no hitboxes at the Flame which are not extended from ShapeHitbox. I also tried to replace all generics with basic types - ShapeHitbox and Broadphase - and everithing works well, at least in examples.

spydon:
> Since we're talking about breaking changes...
> 
> To be honest, all these generics in collision detection system looks overcomplicated. The `<T extends Hitbox<T>>` is specified everywhere, but in real practice only the `ShapeHitbox` makes sense. Why should't we to use this base class instead generic type? If to specify `ShapeHitbox` explicitly in the `Broadphase` and `CollisionDetection` classes, then we will not need to pass these generics everywhere. And it would allow to avoid unnecesary type conversions at many places.
> 
> As I can see, there are no hitboxes at the Flame which are not extended from ShapeHitbox. I also tried to replace all generics with basic types - ShapeHitbox and Broadphase - and everithing works well, at least in examples.

This is because the collision detection system is fairly new, if we would only be able to use `ShapeHitbox` then the system is narrowed down to only be able to be used within FCS, if someone would want to have collision detection in for example Oxygen they can't have their hitboxes extend from `ShapeHitbox`. But there are changes coming to that in the future, where we'll use pure shapes as hitboxes instead of components, but right now since we need to know about the absolute transforms we have to go this route.

ASGAlex:
>But there are changes coming to that in the future

Nice! Hope someday it will be reworked because with adding more dependencies we aggravates this "hell of generics" :-) 

As about this PR - i merged it into my branch, changes really helps to make life easier for end-user. Some previous type conversions become unnecesary. So I strongly confirm this PR usefulness 👍🏻

erickzanardo (approved):

PR #1909, opened 2022-09-16
feat: FlameIsolate - a neat way of handling threads

lohnn:
# Description

Adding a bridge library for [integral_isolates](https://pub.dev/packages/integral_isolates) adding support for components to run CPU intensive code with a function similar to Flutter's compute function, but with a long lived isolate. Lifecycle is handled by the game loop, where the isolate would live between onMount and onRemove.

## Checklist

## Breaking Change

### Gallery example

In the Gallery you can find a minimal example of a program that calculates prime numbers. Tap the brown rectangle to switch between running the calculation in an isolate or synchronously. Running sync will cause noticeable stutter:

https://user-images.githubusercontent.com/960157/196875186-2a2dadca-f8b6-48bf-bffb-becd40e08643.mov

### Flame Isolate example application

The example application for the FlamIsolate package runs an A* path finding algorithm that tries to find an optimal path for the workers to move from their position to a randomly selected bread. This example can be toggled between running in the isolate, as well as using the Flutter `compute` function and synchronously.
If you run the computation in the compute function, you will see that there is no handling of back pressure, and that multiple isolates will be spawned at once (and of course killed off once done). Running synchronously will cause stutter.

https://user-images.githubusercontent.com/960157/196877058-b6ad4a8d-2ca9-4919-a579-338e30c46d64.mp4


lohnn:
Tests are not yet added, but I'm intending on adding some tests for this, hence the Draft status :)

packages/flame_tiled/LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021 Blue Fire
+Copyright (c) 2022 Johannes Löhnn
```
spydon, line 3: Accidental change I guess. 😄

packages/flame_isolate/README.md
```diff
@@ -0,0 +1,40 @@
+The power of [integral_isolates](https://pub.dev/packages/integral_isolates) in your
+[Flame](https://pub.dev/packages/flame) game.
+
+## Usage
```
spydon, line 4: Two empty lines before a new section
(We don't have a markdown linter running atm...)

packages/flame_tiled/LICENSE
lohnn: ... yes...

st-pasha:
If my understanding is correct, this PR adds the `FlameIsolated` component mixin, which essentially adds 2 lines of code:
```
isolated = Isolated(strategy)..init();
...
isolated.dispose();
```

Then, there are several questions:
1. Is this actually worth it creating an entire new package? (which would require some level support in writing documentation, tests, fixing bugs, answering questions on Discord, etc)
2. Does the `FlameIsolated` mixin promote a good programming practice? I imagine some users might decide to add it to their components, and accidentally create a quagmire of threads all running in the background and slowing down the main app (or violating some kind of policy on the number of threads that can be created).
3. I would imagine the best practice for dealing with long-running computationally intensive tasks would be to have a dedicated thread (isolate) at the app level (i.e. attached to the game instance). This would mean creating an `Isolated` in game's `onLoad`, and then make it globally available through the `game` reference. 

It seems to me like this could be a simple documentation recipe, instead of a full-blown package.

jtmcdole:
> integral_isolates 0.1.0+1 
> Published [6 hours ago](https://pub.dev/packages/integral_isolates)

That's a fresh package!

ASGAlex:
I'm sorry to interrupt you conversation 🙃

Isolates might be useful for heavy calculations, yes, but it is extrimely useful when computing not just a function, but preserves state between computations. For example, moving collision detection into separate process. Or calculate player's route by "a star algorithm". But passing every time whole set of data to long running isolate is very expensive, you will spend all possible performance optimisation for type conversion and copying data.

So I might to be mistaken, but it seems this package has this problem. There is no much use of solution like simple `compose` function as we already have one 🤷🏼‍♂️ The thing might be much more useful is something more specific to game process, a "background service" with internal state, just receives small updates from main process and sends result in response.

jtmcdole:
> . But passing every time whole set of data to long running isolate is very expensive

This is not entirely true - or at least on the roadmap last I read it. They were planning on letting you send objects over a port and have it "Transfer" ownership without copying.  I'd love to see some numbers on it though.

E.g.:
https://api.dart.dev/stable/2.18.1/dart-isolate/TransferableTypedData-class.html

ASGAlex:
Anyway you will spend your time for type conversions. So it is better to exchange as small amount of data as possible.
For example, to send component's position to isolate, I need to convert it to Vector2 from NotifyingVector2. And this means copying the value into new object. If I would have thousands of objects....

lohnn:
@st-pasha that is, so far, correct. How it will evolve in the future I cannot say right now.

1. In regards to having thin wrapper packages. While I agree it adds some more maintenance work, most of it will be covered by the underlying package.
2. It is always possible to do stupid things. Just because this wrapper (or the underlying package) allows for easily creating isolates without thinking too much about it, does not mean that you can document it properly and explain how you should go about using it. To be honest, I think you could go as much bananas with just adding too many rays in raycasting/tracing, or physics objects.
The documentation for this library, as well as the underlying one, is being created and I'll try to make sure to explain how to use it, and the potential pitfalls. Possibly with a few examples.
3. I believe there's a few ways of thinking about best practice for this package. Attaching the isolate to the game instance would mean that everyone that uses the isolate through the game reference would share the same job queue. I would personally use an isolate per calculation type. Like one for controlling NPCs (figuring out where to go, what to do and path finding), one for world generation if it is an endless game and so on.

Regarding if it should just be a documentation recipe, and not a full blown package, I leave that up to the Flame team. Maybe @spydon has any wise words?

lohnn:
Regarding all this talk about sending data being expensive. Yes there is a cost with copying data between isolates. This package inherently has the same downsides as with all isolates in Dart. However if you downsize (copy) your data into smaller objects, the cost of sending it is not that significant. And thousands of objects should most likely be fine to copy, CPUs are quite fast nowadays.
And as @jtmcdole said, if the data type you copy to is a [typed data](https://api.dart.dev/stable/2.18.1/dart-typed_data/TypedData-class.html) of some sorts, you can transfer ownership and skip the second copy between threads.

spydon:
> Regarding if it should just be a documentation recipe, and not a full blown package, I leave that up to the Flame team. Maybe @spydon has any wise words?

I don't think that the wrapper being thin at the moment is a big problem, it's still a good convenience mixin for the user. And it will also be easier for us to help debugging it when people have problems since they will all follow the same structure.

>  The documentation for this library, as well as the underlying one, is being created and I'll try to make sure to explain how to use it, and the potential pitfalls. Possibly with a few examples.

Sounds good, this will also make it easier for the user not to fall in any of the common pitfalls that they could have fallen into if this was not a package. 

> Does the FlameIsolated mixin promote a good programming practice? I imagine some users might decide to add it to their components and accidentally create a quagmire of threads

We shouldn't force our more advanced users to use worse patterns just because new users can use them to make mistakes. I think it makes more sense to have this on a component level than a game level if one would want different types of tasks, just like @lohnn describes. 

What one could do is turn this into its completely separate `Component` instead of being a mixin, then it should be harder for users to accidentally add a lot of them since they wouldn't extend from that component in the same extent (hopefully...). But that wouldn't be as user friendly as the mixin is, so I would vote for keeping it as a mixin.

examples/lib/stories/bridge_libraries/flame_isolate/simple_isolate_example.dart
```diff
@@ -0,0 +1,183 @@
...
+import 'package:flame/effects.dart';
+import 'package:flame/extensions.dart';
+import 'package:flame/game.dart';
+import 'package:flame_isolate/flame_isolate.dart';
+import 'package:flutter/material.dart';
+
+class SimpleIsolateExample extends FlameGame with HasTappables {
+  static const String description = '''
+    This example showcases a simple FlameIsolate example, making easy to 
+    continually run heavy load possible without stutter.
```
spydon, line 13:
```suggestion
    This example showcases a simple FlameIsolate example, making it easy
    to continually run heavy load without stutter.
```

examples/lib/stories/bridge_libraries/flame_isolate/simple_isolate_example.dart
```diff
@@ -0,0 +1,183 @@
...
+    
+    Tap the brown square to swap between running heavy load in in an isolate or
+    synchronous.
+    
+    The selected backpressure strategy used for this example is
+    DiscardNewBackPressureStrategy. This strategy discards all new jobs added to
```
spydon, line 19:
```suggestion
    `DiscardNewBackPressureStrategy`. This strategy discards all new jobs added to
```

examples/lib/stories/bridge_libraries/flame_isolate/simple_isolate_example.dart
```diff
@@ -0,0 +1,183 @@
...
+    the queue if there is already a job in the queue.
+  ''';
+
+  @override
+  void onMount() {
```
spydon, line 24: This should be `onLoad`, otherwise it could add many of the things that are added in here

doc/flame_isolate/isolate.md
```diff
@@ -0,0 +1,47 @@
+# FlameIsolate
```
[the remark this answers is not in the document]
lohnn: I've now added some well needed notes about performance.

doc/flame_isolate/isolate.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+
+### Performance note
+
+Keep in mind that every component with `FlameIsolate` mixin that you create and add to your game
+will create a new isolate. This means you will probably want to create a manager component to
+manage a lot of "dumber" components. Think of it like ants, where the queen controls the worker
+ants. If every individual worker ant got it's own isolate, it would be a total waste of power,
+hence you would put it on the queen, which in turn tells all the worker ants what to do.
```
spydon, line 42: Love the analogy :smile:

doc/flame_isolate/isolate.md
```diff
@@ -0,0 +1,58 @@
...
+
+A simple example of this can be found in the example application for the FlameIsolate package.
+
+
+### Backpressure Strategies
```
spydon, line 47: Maybe add a sentence explaining what backpressure is. And also a list of which strategies that are available.

examples/lib/main.dart
```diff
@@ -36,6 +37,7 @@ void main() {
   addComponentsStories(dashbook);
   addEffectsStories(dashbook);
   addExperimentalStories(dashbook);
+  addFlameIsolateExample(dashbook);
```
spydon, line 40: Since this is a bridge package it should be together with the bridge packages in the bottom (line 51)

examples/lib/stories/bridge_libraries/flame_isolate/simple_isolate_example.dart
```diff
@@ -0,0 +1,183 @@
...
+
+  final _textPaint = TextPaint(
+    style: const TextStyle(
+      fontSize: 10,
+    ),
+  );
+
+  @override
+  void render(Canvas canvas) {
+    final rect = Rect.fromLTWH(0, 0, width, height);
```
spydon, line 154: Define the rect outside of the `render` method, otherwise there are 60 new rect objects created per second.

examples/lib/stories/bridge_libraries/flame_isolate/simple_isolate_example.dart
```diff
@@ -0,0 +1,183 @@
...
+    canvas.drawRect(rect, _paint);
+
+    _textPaint.render(
+      canvas,
+      computeType.description,
+      rect.topLeft.toVector2(),
```
spydon, line 160: Same here

examples/lib/stories/bridge_libraries/flame_isolate/simple_isolate_example.dart
```diff
@@ -0,0 +1,183 @@
...
+    );
+
+    _textPaint.render(
+      canvas,
+      '${_primeData.key} is${_primeData.value ? '' : ' not'} a prime number',
+      rect.center.toVector2(),
```
spydon, line 166: and here

packages/flame_isolate/README.md
```diff
@@ -0,0 +1,46 @@
+The power of [integral_isolates](https://pub.dev/packages/integral_isolates) in your
```
spydon, line 1: Maybe start the readme with a simple `# Flame Isolate` or `# flame_isolate` (and then add a # to the other sections.

packages/flame_isolate/README.md
```diff
@@ -0,0 +1,46 @@
...
+      isolate(recalculateWorld, worldData).then(updateWorld);
+    }
+    ...
+  }
+  ...
+}
+```
+
+
+## Backpressure Strategies
```
spydon, line 28: When you add more info about the backpressure strategies for the docs that info could be copied over here too.

packages/flame_isolate/example/LICENSE
```diff
@@ -0,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
```
spydon, line 1: How come the example licence is GPLv2 instead of MIT?

packages/flame_isolate/example/lib/brains/worker_overmind.dart
```diff
@@ -0,0 +1,169 @@
...
+
+  @override
+  Future onMount() {
+    calculateTasks();
+    _assignTaskInterval = Timer(0.2, repeat: true, onTick: _assignTasks)
+      ..start();
+    return super.onMount();
+  }
+
+  // TODO(lohnn): Would, in reality, also be moved to isolate
```
spydon, line 35: What should be done with this TODO?

packages/flame_isolate/example/lib/brains/worker_overmind_hud.dart
```diff
@@ -0,0 +1,55 @@
...
+
+  final textPaint = TextPaint(
+    style: const TextStyle(
+      fontSize: 15,
+    ),
+  );
+
+  @override
+  void render(Canvas canvas) {
+    final rect = toRect();
```
spydon, line 45: Same thing with the rect and vector2 here

packages/flame_isolate/example/lib/colonists_game.dart
```diff
@@ -0,0 +1,98 @@
...
+    } else if (event.data.logicalKey == LogicalKeyboardKey.numpadSubtract &&
+        event is RawKeyDownEvent) {
+      camera.zoom = max(camera.zoom - 0.1, 0.1);
+    }
+    return super.onKeyEvent(event, keysPressed);
+  }
+
+  @override
+  void update(double dt) {
+    final direction = Vector2(_rightForce - _leftForce, _downForce - _upForce);
```
spydon, line 85: The definition of this vector can be moved outside of `update` and then you can reuse it by `setValues`.

packages/flame_isolate/example/lib/standard/int_vector2.dart
```diff
@@ -0,0 +1,31 @@
...
+
+  const IntVector2(this.x, this.y);
+
+  /// Manhattan distance on a square grid
+  int distanceTo(IntVector2 b) {
+    return (x - b.x).abs() + (y - b.y).abs();
+  }
+
+  @override
+  bool operator ==(Object o) => o is IntVector2 && x == o.x && y == o.y;
```
spydon, line 18: Since this class is const, I think this is already automatically done.

packages/flame_isolate/example/lib/terrain/grass.dart
```diff
@@ -0,0 +1,20 @@
...
+class Grass extends PositionComponent with Terrain {
+  static final _color = Paint()..color = const Color(0xff567d46);
+  static final _debugColor = Paint()..color = Colors.black.withOpacity(.5);
+
+  @override
+  void render(Canvas canvas) {
+    final rect = size.toRect();
+    canvas.drawRect(rect, _color);
+
+    final rect2 = Rect.fromCenter(center: rect.center, height: 10, width: 10);
```
spydon, line 14: These two rects can be defined outside of `render` too

packages/flame_isolate/example/lib/units/worker.dart
```diff
@@ -0,0 +1,153 @@
...
+      ],
+      stepTime: 0.1,
+    ),
+  };
+
+  bool _walkingLeft = false;
+
+  @override
+  // ignore: avoid_setters_without_getters
+  set currentDirection(MoveDirection direction) {
```
spydon, line 102: This would be better suited as a proper method instead of a setter I think

packages/flame_isolate/example/lib/units/worker.dart
```diff
@@ -0,0 +1,153 @@
...
+  void update(double dt) {
+    _currentAnimation.update(dt);
+    //DONE: Move to object
+    //DONE: V2 path finding
+
+    // TODO(lohnn): Check if object is close enough
+    // TODO(lohnn): V2 path finding
+    // TODO(lohnn): Pick object up
+    // TODO(lohnn): Move to destination
+    // TODO(lohnn): Drop object
```
spydon, line 126: Should these TODO's just be removed?

packages/flame_isolate/example/lib/units/worker.dart
```diff
@@ -0,0 +1,153 @@
...
+import 'package:flame/flame.dart';
+import 'package:flutter/material.dart' hide Animation;
+import 'package:flutter_isolates_example/colonists_game.dart';
+import 'package:flutter_isolates_example/constants.dart';
+import 'package:flutter_isolates_example/objects/colonists_object.dart';
+import 'package:flutter_isolates_example/standard/int_vector2.dart';
+import 'package:flutter_isolates_example/standard/pair.dart';
+import 'package:flutter_isolates_example/units/actions/movable.dart';
+
+class Worker extends PositionComponent
```
spydon, line 11: Use a `SpriteAnimationGroupComponent` instead (you are basically re-implementing the logic of that component in here)

packages/flame_isolate/lib/flame_isolate.dart
```diff
@@ -0,0 +1,55 @@
...
+/// class MyGame extends FlameGame with FlameIsolate {
+///   @override
+///   void update(double dt) {
+///     if (shouldRecalculate) {
+///       isolate(recalculateWorld, worldData).then(updateWorld);
+///     }
+///     return super.update(dt);
+///   }
+/// }
+/// ```
```
spydon, line 25: Maybe a short note about the performance impact of adding too many of these here too

spydon (commented):
Good job! Can you add a video to the PR description of the two examples running?

packages/flame_isolate/example/LICENSE
lohnn: Due to the beautiful worker asset. Added a small note about that in the README.

packages/flame_isolate/example/lib/standard/int_vector2.dart
lohnn: Seems like it is not, actually. Get a `Null check operator used on a null value` when trying to run the game if it is not present.

spydon (approved):

packages/flame_isolate/example/lib/units/worker.dart
```diff
@@ -0,0 +1,149 @@
...
+      [getSprite(0, 4)],
+      stepTime: 1,
+    ),
+    MoveDirection.up: SpriteAnimation.spriteList(
+      [
+        getSprite(0, 0),
+        getSprite(1, 0),
+        getSprite(2, 0),
+        getSprite(3, 0),
+        getSprite(4, 0),
```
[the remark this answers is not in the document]
lohnn: That one works on a row basis, right? Also, currently my `getSprite` function crops out the padding from the sprite.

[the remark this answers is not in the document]
spydon: Ah, or just use `SpriteSheet` ofc:
```dart
    final spritesheet = SpriteSheet.fromColumnsAndRows(
      image: image,
      columns: 5,
      rows: 5,
    );
    final animation = spritesheet.createAnimation(
        stepTime: 0.1,
        row: 0,
        loop: true,
        from: 0,
        to: 5
    );
```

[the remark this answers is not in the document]
lohnn: But then I can't crop out the padding?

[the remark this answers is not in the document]
spydon: Aah, didn't notice you were doing that, we should support that on sprite sheet! Maybe that would be your next PR? :)

Consider this discussion resolved then.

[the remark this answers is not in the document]
lohnn: Great idea! Need based development, gotta love it 😁

wolfenrain (approved):

PR #1915, opened 2022-09-18
feat: Add support for styles propagating through the text node tree

st-pasha:
# Description

This PR continues the work for enabling rich text support within Flame.
Here I add support for different text fragments having different `TextStyle`s, and allow those styles to be inheritable within the text node tree.

Screen Shot 2022-09-17 at 10 46 15 PM

## Checklist

## Breaking Change

## Related Issues

WIP for #1627 


packages/flame/lib/src/text/formatters/text_painter_text_formatter.dart
```diff
@@ -1,22 +1,22 @@
+import 'package:flame/src/text/elements/debug_text_painter_text_element.dart';
+import 'package:flame/src/text/elements/text_painter_text_element.dart';
 import 'package:flame/src/text/formatters/text_formatter.dart';
-import 'package:flame/src/text/inline/debug_text_painter_text_element.dart';
-import 'package:flame/src/text/inline/text_painter_text_element.dart';
-import 'package:flutter/rendering.dart';
+import 'package:flutter/rendering.dart' as flutter;
```
spydon, line 4: How come this is needed everywhere now? If there is a clash between two classes I would `hide` the one that isn't used instead.

packages/flame/lib/src/text/styles/text_style.dart
```diff
@@ -0,0 +1,57 @@
+import 'dart:ui';
+
+import 'package:flame/src/text/formatters/text_formatter.dart';
+import 'package:flame/src/text/formatters/text_painter_text_formatter.dart';
+import 'package:flame/src/text/styles/style.dart';
+import 'package:flutter/rendering.dart' as flutter;
+import 'package:meta/meta.dart';
+
+@immutable
+class TextStyle extends Style {
```
spydon, line 10: We shouldn't name things something that clashes with flutter classes, that will be confusing both for us and for our users.

packages/flame/lib/src/text/formatters/text_painter_text_formatter.dart
st-pasha: Ok, i removed the prefix.

packages/flame/lib/src/text/styles/text_style.dart
st-pasha: There are 2 different `TextStyle` classes in Flutter, so it's already confusing... Also, I'm not sure what could be a different name here.

spydon: We definitely shouldn't add a third then. 😅
People will so easily import the wrong one.
The easiest would be to just use a synonym for either text or style, but there aren't really many options out there...
Maybe `ParagraphStyle`, but it doesn't really have the same meaning. I'll try to think of something if you don't come up with anything.

spydon: I tried to gather up suggestions for names for this and I got `TextAppearance` from two people, but I think that would be problematic too since it extends `Style`.

spydon: What about injecting some word between `Text*Style`, like `TextLookStyle` but with another word that look?

spydon: `TextRenderingStyle` was suggested by Erick.

st-pasha: That seems... too specific. It could have made sense if we had, say, `TextRenderingStyle` and `TextUpdatingStyle` (or whatever) -- then it would have been appropriate. But all we have is just a `TextStyle`, the style of the text. It's like, imagine if we had `CoordinatePositionComponent` instead of just `PositionComponent` -- that would be too weird.

We could try to rough up the word "text" a little. Say, `TxtStyle`, or `TexteStyle`, or `TekstStyle`, or something like that.

spydon: We don't allow abbreviations of words, and the misspelling seems even weirder to me. 
Is there no other word that we could inject into the middle then?

spydon: `TextFormatStyle` or something like that maybe?

> imagine if we had `CoordinatePositionComponent` instead of just `PositionComponent`

If we had two other `PositionComponent`s in the standard libs we would have to choose another name for that one too of course, that one would be incredibly much easier to rename than the `TextStyle` though, it would probably just be named `TransformComponent` or any synonym to that, but the problem that we have here is that it is much harder to come up with synonyms for `Text`.

lohnn: Why not just go simple route and prefix it with Flame? `FlameTextStyle`

st-pasha: (I have a suspicion that the reason there are two `TextStyle`s is because Flutter developers were facing a similar dilemma and couldn't find a better name :)

The `FlameTextStyle` is a possibility, but then all other styles don't have `Flame` prefix in them, so it looks a bit inconsistent. It seems though that _some_ kind of rule-breaking is unavoidable here, and then the question is whether we'd want `FlameTextStyle`, or `FTextStyle`, or `TxtStyle`.

And as far as synonym search goes, we could ponder about `InlineStyle` or `ContentStyle`, though I'm not sure myself about those.

spydon: > (I have a suspicion that the reason there are two TextStyles is because Flutter developers were facing a similar dilemma and couldn't find a better name :)

Most likely. The cliché about naming being hard really isn't exaggerated. :sweat_smile: 

`ContentStyle` is not too bad I think.

st-pasha: I'm leaning towards `FlameTextStyle` here. At least it'll be obvious that this is like `TextStyle` but for Flame. With `ContentStyle` it wouldn't be immediately obvious what it represents.

st-pasha: However, I'd like to file a complaint with whoever invented the English language for not putting in more synonyms to the word "text". This was an inexcusable lack of foresight on their part.

spydon: Yeah, let's go with `FlameTextStyle` and explain in the dartdocs why we had to give it an inconsistent name.

st-pasha: It's done!

spydon (approved):

packages/flame/lib/src/text/styles/text_style.dart
```diff
@@ -0,0 +1,55 @@
+import 'package:flame/src/text/formatters/text_formatter.dart';
+import 'package:flame/src/text/formatters/text_painter_text_formatter.dart';
+import 'package:flame/src/text/styles/style.dart';
+import 'package:flutter/rendering.dart';
+import 'package:meta/meta.dart';
+
+@immutable
+class FlameTextStyle extends Style {
```
spydon, line 8: Just change the name of the file too

packages/flame/lib/src/text/nodes/paragraph_node.dart
```diff
@@ -1,59 +1,29 @@
...
-    }
-    if (!startNewLine) {
-      verticalOffset += lines.last.metrics.height;
-    }
-    verticalOffset += style.padding.bottom;
-    final bg = makeBackground(style.background, parentWidth, verticalOffset);
-    final elements = bg == null ? lines : [bg, ...lines];
-    return GroupElement(parentWidth, verticalOffset, elements);
+  static const defaultStyle = BlockStyle(
+    margin: EdgeInsets.all(6),
```
erickzanardo, line 20:
```suggestion
    margin: const EdgeInsets.all(6),
```

erickzanardo (approved):

packages/flame/lib/src/text/nodes/paragraph_node.dart
spydon: It is already `const` since it is assigned to a `static const` right?

st-pasha: Yeah, if you add `const` there it gives the `unnecessary_const` lint warning.

PR #1931, opened 2022-09-21
feat: Adding callbacks for EffectController (#1926)

### LINKED ISSUE #1926: [flame] Allow InfiniteEffectController be stopped
# What could be improved
It should be possible that a running `InfiniteEffectController` can be stopped.

# Why should this be improved
Since `InfiniteEffectController` simplifies running a repeating effect (also without knowing the amount of steps yet) it would be great if that controller could be stopped from outside by considering the 2 possible end cycles (after completing `duration` and/or `reverseDuration`), i.e. by not just removing the effect from its parent in the middle of running the effect.

# Any risks?
My current suggestion would allow defining 2 callback functions in the EffectController factory:
```
factory EffectController({
    double? duration,
    double? speed,
    Curve curve = Curves.linear,
    double? reverseDuration,
    double? reverseSpeed,
    Curve? reverseCurve,
    bool infinite = false,
    bool alternate = false,
    int? repeatCount,
    double startDelay = 0.0,
    double atMaxDuration = 0.0,
    double atMinDuration = 0.0,
    VoidCallback? onPeak,
    VoidCallback? onReversePeak,
  })
```

# More information
A new `CallbackController` would be created and inserted in EffectController just before checking `atMaxDuration` and `atMinDuration`:
```
// ON PEAK CALLBACK
if (onPeak != null) {
  items.add(CallbackController(onPeak, 1.0));
}

// AT-MAX
if (atMaxDuration != 0) {
  items.add(PauseEffectController(atMaxDuration, progress: 1.0));
}
```
```
// ON REVERSE PEAK CALLBACK
if (onReversePeak != null) {
  items.add(CallbackController(onReversePeak, 0.0));
}

// AT-MIN
if (atMinDuration != 0) {
  items.add(PauseEffectController(atMinDuration, progress: 0.0));
}
```
The `CallbackController` looks like this:
```
import 'dart:ui';

class CallbackController extends DurationEffectController {
  CallbackController(this.callback, this._progress) : super(0.0);

  final VoidCallback callback;

  final double _progress;

  @override
  double get progress => _progress;

  @override
  bool get completed => false;


  @override
  double advance(double dt) {
    callback();
    return 0.1;
  }

  @override
  double recede(double dt) {
    callback();
    return 0.1;
  }
}
```

With that implementation a cycle will be finished before `onPeak` or `onReversePeak` callbacks will be invoked.
I'm curious what you think about it - i've talked about this suggestion quite a lot to spydon and DevKage on Discord channel.
These callbacks can of course also be used for other purposes and therefore will not only improve `InfiniteEffectController`.

rivella50:
# Description
Adds two callback functions to EffectController factory which are invoked before atMaxDuration or atMinDuration runs.

## Checklist

## Breaking Change?
No

### Migration instructions

## Related Issues
Closes #1926 


packages/flame/lib/src/effects/controllers/effect_controller.dart
```diff
@@ -71,6 +74,10 @@ abstract class EffectController {
   /// If the animation is finite and there are no "backward" or "atMin" stages
   /// then the animation will complete at `progress == 1`, otherwise it will
   /// complete at `progress == 0`.
+  ///
+  /// Before [atMaxDuration] and [atMinDuration] a callback function can be
```
spydon, line 78:
```suggestion
  /// Before [onPeak] and [onReversePeak] a callback function can be
```

packages/flame/lib/src/effects/controllers/callback_controller.dart
```diff
@@ -0,0 +1,31 @@
+import 'dart:ui';
+import 'package:flame/effects.dart';
+
+/// callback will be invoked after the effect's duration or reverseDuration
```
spydon, line 4:
```suggestion
/// The [callback] will be invoked once the effect's [progress] is reached.
/// 1.0 to match `duration` or 0.0 to match when the `reverseDuration`
/// has been reached.
```

spydon (commented):
Lgtm, just some tests missing to verify that the callbacks are called at the right time.

packages/flame/lib/src/effects/controllers/callback_controller.dart
```diff
@@ -0,0 +1,32 @@
...
+  @override
+  double get progress => _progress;
+
+  @override
+  bool get completed => true;
+
+  @override
+  double advance(double dt) {
+    callback();
+    return 0.00001;
```
spydon, line 24: Shouldn't this (and recede) just return `dt`?

spydon: 0 I mean

rivella50: 0 would mean that the controller still runs which blocks the effect. But returning dt seems to work fine too.

spydon: I guess this is just to pass on the dt to the next controller in the chain.

st-pasha: This should be `dt`, see the docs:

> /// If the controller is still running, the return value will be 0. If it
  /// already finished, then the return value will be the "leftover" part of
  /// the [dt]. That is, the amount of time [dt] that remains after the
  /// controller has finished. In all cases, the return value can be positive
  /// only when `completed == true`.

rivella50: It has been changed to return dt meanwhile.

doc/flame/effects.md
```diff
@@ -350,6 +350,8 @@ EffectController({
     int? repeatCount,
     bool infinite = false,
     double startDelay = 0.0,
+    VoidCallback? onPeak,
+    VoidCallback? onReversePeak,
```
st-pasha, line 354: The "reversePeak" term seem kinda confusing. Maybe we should call these callbacks `onMax/onMin`, or `onReachMax/onReacMin`, or `onArriveAtMax/onArriveAtMin`, etc -- but to remain in concord with the existing terminology of `atMaxDuration/atMinDuration`.

packages/flame/lib/src/effects/controllers/callback_controller.dart
```diff
@@ -0,0 +1,32 @@
+import 'dart:ui';
+import 'package:flame/effects.dart';
+
+/// The [callback] will be invoked once the effect's [progress] is reached.
+/// 1.0 to match `duration` or 0.0 to match when the `reverseDuration`
+/// has been reached.
+class CallbackController extends DurationEffectController {
+  CallbackController({required this.callback, required double progress})
```
st-pasha, line 8:
```suggestion
  CallbackController(this.callback, {required double progress})
```
This would be a slightly simpler API, since for a CallbackController the main argument must obviously be the callback.

doc/flame/effects.md
spydon: `onMin` and `onMax` I think is good, with some good docs of course.

packages/flame/lib/src/effects/controllers/callback_controller.dart
```diff
@@ -0,0 +1,32 @@
+import 'dart:ui';
+import 'package:flame/effects.dart';
+
+/// The [callback] will be invoked once the effect's [progress] is reached.
+/// 1.0 to match `duration` or 0.0 to match when the `reverseDuration`
+/// has been reached.
+class CallbackController extends DurationEffectController {
+  CallbackController(this.callback, {required double progress})
```
st-pasha, line 8:
```suggestion
/// This effect controller invokes the [callback] function and then immediately finishes. Use it as part of a more complex effect to insert callbacks at certain parts of that effect.
class CallbackController extends DurationEffectController {
  /// Creates a controller that invokes the given [callback] function.
  ///
  /// The [progress] parameter specifies the progress level of the effect at the time when the callback is invoked. It is the responsibility of the user to ensure that this progress level is contiguous with respect to the progress of the overall effect.
  CallbackController(this.callback, {required double progress})
```
(will need formatting for line length).

doc/flame/effects.md
st-pasha: Here the method names are still old.

st-pasha (approved):

spydon (approved):

PR #1933, opened 2022-09-21
fix: OpacityEffect rounding error calculation

### LINKED ISSUE #1574: Inconsistency with OpacityEffect.fadeOut
# Current bug behaviour

When using `OpacityEffect.fadeOut()` the final alpha/opacity of target does not reach 0 sometimes. Most of the times it is not even visible, but I've seen cases where the alpha stays at 30 or even 73, making the target still visible.

# Expected behaviour

At the end of fadeOut target should be completely hidden.

# Steps to reproduce

Add fadeOut effect to simple rectangle component and print out the alpha/opacity values from `onFinishCallback`. Here is a simple code to do this:

```dart
main() {
  runApp(
    MaterialApp(
      home: Scaffold(
        body: GameWidget(
          game: SimpleGame(),
        ),
      ),
    ),
  );
}

class SimpleGame extends FlameGame {
  @override
  Future<void>? onLoad() async {
    camera.viewport = FixedResolutionViewport(Vector2(640, 360));

    final rect = RectangleComponent(
      position: size / 2,
      size: Vector2.all(32),
    );

    add(rect);

    rect.add(
      OpacityEffect.fadeOut(
        LinearEffectController(3),
      )..onFinishCallback = () {
          print(rect.getAlpha());
          print(rect.getOpacity());
        },
    );
    return super.onLoad();
  }
}
```

# Flutter doctor output

```
Output of: flutter doctor -v

[√] Flutter (Channel stable, 2.10.5, on Microsoft Windows [Version 10.0.22000.556], locale en-US)
    • Flutter version 2.10.5 at C:\apps\android\flutter
    • Framework revision 5464c5bac7 (8 days ago), 2022-04-18 09:55:37 -0700
    • Engine revision 57d3bac3dd
    • Dart version 2.16.2
    • DevTools version 2.9.2

[√] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
    • Android SDK at C:\apps\android
    • Platform android-31, build-tools 30.0.2
    • ANDROID_HOME = C:\apps\android
    • ANDROID_SDK_ROOT = C:\apps\android\cmdline-tools\tools
    • Java binary at: C:\apps\openjdk-11.0.2_windows-x64_bin\jdk-11.0.2\bin\java
    • Java version OpenJDK Runtime Environment 18.9 (build 11.0.2+9)

    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[√] Visual Studio - develop for Windows (Visual Studio Community 2022 17.1.5)
    • Visual Studio at C:\Program Files\Microsoft Visual Studio\2022\Community
    • Visual Studio Community 2022 version 17.1.32414.318
    • Windows 10 SDK version 10.0.22000.0

[!] Android Studio (not installed)
    • Android Studio not found; download from https://developer.android.com/studio/index.html
      (or visit https://flutter.dev/docs/get-started/install/windows#android-setup for detailed instructions).

[√] VS Code (version 1.66.2)
    • VS Code at C:\Users\Ryuzaki\AppData\Local\Programs\Microsoft VS Code
    • Flutter extension version 3.38.1

    • Windows (desktop) • windows • windows-x64    • Microsoft Windows [Version 10.0.22000.556]
    • Chrome (web)      • chrome  • web-javascript • Google Chrome 100.0.4896.127
    • Edge (web)        • edge    • web-javascript • Microsoft Edge 91.0.864.54

[√] HTTP Host Availability

```

# Log information

```
```

# More information

ufrshubham:
# Description
When using OpacityEffect.fadeOut, the value of opacity at the end of the effect was non zero. I suspect that the rounding error calculation is wrong and this PR tried to simplify that logic.

Proof that this PR actually fixes the problem can be seen at the test results of [793415f](https://github.com/flame-engine/flame/actions/runs/3099879070/jobs/5019538100) and [5f3c591](https://github.com/flame-engine/flame/actions/runs/3099914265/jobs/5019612017). First one just added a test for `OpacityEffect.fadeOut` and second added the actual fix.

_**Disclaimer**: I don't fully understand this part of the code, so be extra vigilant while reviewing this PR._

## Checklist

## Breaking Change?

## Related Issues
Closes #1574 


packages/flame/lib/src/effects/opacity_effect.dart
```diff
@@ -68,14 +68,8 @@ class OpacityEffect extends ComponentEffect<HasPaint> {
...
     final deltaAlpha =
         (_alphaOffset * deltaProgress) + _roundingError * deltaProgress.sign;
-    final remainder = deltaAlpha.remainder(1.0).abs();
-    _roundingError = remainder >= 0.5 ? -1 * (1.0 - remainder) : remainder;
     var nextAlpha = (currentAlpha + deltaAlpha).round();
-    if (nextAlpha < 0) {
-      _roundingError += nextAlpha.abs();
-    } else if (nextAlpha > 255) {
-      _roundingError += nextAlpha - 255;
-    }
```
spydon, line 78: I have a feeling I wrote this code, I wonder why I made it so damn complicated. :sweat_smile:

spydon (commented):
Can you add a test that runs the same effect several cycles too?

ufrshubham:
> Can you add a test that runs the same effect several cycles too?

Oh, you mean by resetting the same opacity effect without removing it from the parent?

packages/flame/lib/src/effects/opacity_effect.dart
ufrshubham: It does show your name in GitLens 😄

image

spydon:
> Oh, you mean by resetting the same opacity effect without removing it from the parent?

Without resetting it, so letting it run for N-times and then check the values (`RepeatedEffectController` should simplify this).

ufrshubham:
> `RepeatedEffectController` should simplify this

Used `SequencedEffect` instead of `RepatedEffectController` as `RepatedEffectController` didn't allow me to capture the opacity values at the end of each cycle.

spydon:
> Used `SequencedEffect` instead of `RepatedEffectController` as `RepatedEffectController` didn't allow me to capture the opacity values at the end of each cycle.

That won't have the effect that I'm looking for in the test though, because that adds several effects, I want to make sure that all values are correct within the same effect when it runs through several cycles.
You could also just set it to `infinite` and step it forward with `game.update(dt)`.

ufrshubham:
> You could also just set it to `infinite` and step it forward with `game.update(dt)`, I guess that is actually a better option because now when I look at the `RepatedEffectController` I see that it uses `setToStart` on the effect.

Sorry, a bit confused here. So you are saying set `infinite` to true in `EffectController` and not the `SequenceEffect`, right?
Also, if I do that, I won't be able capture the value of opacity exactly at the end of a cycle. We don't have any callback for that, and since dt is varying, I can't add a check for `elapsedTime == durationOfOneCycle`.

For example, if I add an infinite fade out effect with duration as 3 seconds and make it run with `game.udpate(dt)`, how will I get to know the value of opacity exactly when 3 seconds mark is reached. If for nth frame elapsedTime is 2.9 and in (n+1)th frame it becomes 3.1, wouldn't the opacity change back to 1 in that frame.

spydon:
> For example, if I add an infinite fade out effect with duration as 3 seconds and make it run with `game.udpate(dt)`, how will I get to know the value of opacity exactly when 3 seconds mark is reached. If for nth frame elapsedTime is 2.9 and in (n+1)th frame it becomes 3.1, wouldn't the opacity change back to 1 in that frame.

You just call `game.update(3.0)` or split it up in smaller chunks and then check the opacity.

ufrshubham:
> You just call game.update(3.0) or split it up in smaller chunks and then check the opacity.

I've modified the test to use an infinite effect with fixed time-step. Hope this is what you were looking for.

spydon (approved):

PR #1939, opened 2022-09-24
chore: Add sphinx melos target and fix txt -> text

spydon:
# Description

Sphinx was warning that `txt` doesn't` exist, because the correct tag is `text`.
Also added a melos target to build the docs.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -51,6 +51,10 @@ scripts:
     run: melos exec flutter pub run dartdoc
     description: Run dartdoc checks for all packages.
 
+  sphinxdoc:
+    run: cd doc/_sphinx && make html
+    description: Create the sphinx html docs.
+
   test:select:
     run: melos exec -- flutter test
     select-package:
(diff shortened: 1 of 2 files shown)


melos.yaml
```diff
@@ -51,6 +51,10 @@ scripts:
     run: melos exec flutter pub run dartdoc
     description: Run dartdoc checks for all packages.
 
+  sphinxdoc:
+    run: cd doc/_sphinx && make html
```
st-pasha, line 55: Is there a way to make this work even if you're currently in a subdirectory? Say, if `cwd = packages/flame`?

spydon: Yes, but first of all the makefile should be rewritten to depend on paths in relation to where the file is, not where the shell is.

spydon: Fixed it so that it can run from any directory

erickzanardo (approved):

PR #1948, opened 2022-09-28
feat: Allow followBodyComponent to follow centre of mass (#1947)

### LINKED ISSUE #1947: Allow flame_forge2d followBodyComponent to follow body's center of mass instead of position
# What could be improved

The flame_forge2d camera allows following a BodyComponent with the followBodyComponet method. This method only allows following the body's position vector. It can be improved by optionally allowing the body's center of mass to be followed instead.

# Why should this be improved

Rotations in forge2d occur about the body's center of mass, not always its visual center (the position vector). If the two differ then the whole scene appears to wobble as the body rotates, even if it is otherwise stationary.

# Any risks?

The proposed addition is an optional named parameter useCenterOfMass for followBodyComponet that will default to false, so no breaking changes will be introduced. To follow center of mass instead, pass "useCenterOfMass: true". Internally this just directs Flame's followVector2 function to reference body.worldCenter rather than body.position, with no impact on performance.

# More information

I have prepared a PR.

mattmyne:
# Description

This PR adds the optional named parameter useCenterOfMass to followBodyComponent. It defaults to false so no breaking changes. Setting true allows camera following that does not "wobble" when an otherwise stationary BodyComponent (with an offset centre of mass) rotates.

Also removed reference to component anchor in documentation/comments since BodyComponent does not use an anchor.

## Checklist

## Breaking Change?

## Related Issues

Closes #1947 

### DIFF
--- a/packages/flame_forge2d/lib/forge2d_camera.dart
+++ b/packages/flame_forge2d/lib/forge2d_camera.dart
@@ -12,16 +12,18 @@ extension Forge2DCameraExtension on Camera {
   /// [relativeOffset] (default to the center).
   /// [worldBounds] can be optionally set to add boundaries to how far the
   /// camera is allowed to move.
-  /// The component is "grabbed" by its anchor (default top left).
-  /// So for example if you want the center of the object to be at the fixed
-  /// position, set the components anchor to center.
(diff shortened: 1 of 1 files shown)


spydon (approved):
Lgtm!

PR #1971, opened 2022-09-30
feat: Draft for new Layout components Row and Column

### LINKED ISSUE #1944: Row and Column for Flame Components
# Problem to solve
With `Row` and `Column` classes children `Component`s could be placed and layouted correctly without overlapping and without having to care about repeating size and position calculations.
In my games i often have `Component`s (Sprites, buttons, etc.) which have to be placed either in rows or columns, where it's always quite an effort to calculate their positions and sizes.
The goal of the suggested two classes would be to make these calculations transparent for the developer.

# Proposal
Flutter-like `Row` and `Column` classes with alignment options for both axes (i.e. `mainAxisAlignment` and `crossAxisAlignment`).
Some ideas:
- the new classes could extend `PositionComponent` themselves
- children can possibly be all `PositionComponent` extending classes (since getting their size is important)
- adding a child dynamically to a `Row` or `Column` would need to re-layout the component
- another new class for adding gaps would make sense too (like `SizedBox`) - ok this can be accomplished by just adding a `PositionComponent` with a `size` defined

rivella50:
# Description
This is a draft for the planned row and column components for Flame.
At the moment it only covers the horizontal direction class `RowComponent`.
As soon as we are clear what features these 2 components will have, the class `ColumnComponent` will be added accordingly.
How to use `RowComponent` can be seen in issue #1944 .

To be discussed:
- how shall the layouting routine behave if the children don't have the same anchor type?
- additional parameters involved?

## Checklist

## Related Issues
Closes #1944


st-pasha:
I'm worried about overriding the `add()` and `remove()` methods, since we don't really provide a guarantee that these are the only ways to add or remove children components. For example, there are also `addAll()`, `removeAll()`, `addToParent()`, `removeFromParent()`, `changeParent()`, adding via the constructor; more methods like these can be added in the future. 

I believe a reasonable alternative would be to:
- not subscribe to size changes at all: it is a very rare operation after all, especially for the kinds of components that get put in rows/columns;
- make the `layout()` method public, so that in a rare circumstance that a child does need to change its size, it would simply instruct the parent to re-do the layout manually;
- create a new lifecycle callback `Component.onChildrenChanged()` which would be invoked whenever children components are added/removed (<-- this should probably go in a separate PR);
- do not require that children of `LayoutComponent` must be `PositionComponent`s -- what if someone wants to add an `Effect` for example? The approach taken in `layoutChildren()` is better: simply perform a filter. And even that can be improved: instead of looking for `PositionComponent`s, use `PositionProvider` / `SizeProvider` (the `PositionComponent` satisfies both of these interfaces).

rivella50:
@st-pasha Thanks for the feedback. Makes all sense imo.

rivella50:
@st-pasha Regarding the filter in layoutChildren: You mean filtering like
```
final list = children.where((element) => element is PositionProvider && element is SizeProvider);
```
and then casting to one of the interfaces where needed, e.g.
```
for (final child in list) {
  (child as PositionProvider).position = Vector2(currentPosition.x, currentPosition.y);
  currentPosition.x += (child as SizeProvider).size.x + gap;  // width cannot be used anymore
}
```

rivella50:
@spydon Can we discuss how to proceed with this draft?
I.e. what's missing in functionality in order to be released?
Writing tests won't be a problem.

erickzanardo (commented):
Two small nits but I think this good for an initial approach.

I have two suggestions though

 - The code between Column and Row components are very similar, I wonder if we could share some methods/logic between them to avoid repetition.
 - I would suggest to move these components to the experimental namespace, that we way we have the freedom to explore this new functionality without fear of breaking changes.

rivella50:
@erickzanardo Thanks for the feedback. I've refactored the 3 classes as follows:

- method `layoutChildren` has been moved up to class `LayoutComponent`
- a switch statement is used now
- all 3 classes have been moved to the `experimental` namespace

erickzanardo:
Awesome. From my part. I would say this is ready to receive tests and examples.

Once we have `RowComponent` and `ColumnComponent` on main, we can start expanding this, like an `ExpandedComponent`, a `Flex` one and so on.

rivella50:
Great, thank you, Erick. Then i will write some tests for it.
I also hope that these components will receive many more extensions.

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,123 @@
...
+    layoutChildren();
+  }
+
+  void layoutChildren() {
+    final list = children.whereType<PositionComponent>().toList();
+    if (list.isEmpty) {
+      return;
+    }
+    final currentPosition = Vector2.zero();
+    final componentsDimension = list.fold<double>(
```
spydon, line 37: This should be set to the `PositionComponent`'s width or height right? And then the one that isn't set here should be the `max` width/height of the child components.

spydon (commented):
Just the `size`s for the components that are missing I think (and tests for the `size`s), otherwise it looks good!

packages/flame/lib/src/experimental/layout_component.dart
rivella50: The size of the parent component is checked and handled here:
https://github.com/flame-engine/flame/blob/7486750248cba830bfb900632f9471fa9635d375/packages/flame/lib/src/experimental/layout_component.dart#L45-L48

spydon: `dimensionAvailable` should possibly be the parent's `size`, or it's own `size`, it shouldn't check anything regarding the canvas size or such.
So either we let the `size` of the `LayoutComponent` grow dynamically with the children added to it, or it has a set size from the beginning, but to make it more flutter like I think that it should set the size itself and then just be constrained by the parent's `size`.
Because right now the layout component's `size` will be (0, 0) now since it isn't updated when children are added. The setter of `size` should probably also be overridden so that it can't be set too if we choose to go with a dynamic size.

rivella50: But can the parent not also have a zero size?
I had it in mind like this: if the developer defines a size for LayoutComponent then this size will be taken for the available dimension calculation, else it just takes the canvas' width or height for the end edge.

spydon: Why would it take the canvas size? A component can live anywhere in the tree.
To lower the confusion as much as possible for these ones I think we should follow the Flutter style with having the constraints go down and the sizes go up (or in this case just set it itself).

rivella50: Ok, but how would you retrieve the width (or height) for the case if the developer hasn't defined an explicit size for the `LayoutComponent` in order for being able to have a value for `dimensionAvailable`?
`parent.size` cannot be used because `parent` is a `Component` and doesn't necessarily have a `size`.

rivella50: ... and growing with the size of its children doesn't make much sense for alignment types like spaceBetween imo.

spydon: You take the first available parent that is a `PositionComponent` (or a `SizeProvider` when we have that, not sure why we don't have that currently) with `findParent<PositionComponent>` and if that returns null then the size is unbounded, so the row or column can grow infinitely.

rivella50: Meanwhile i've changed it to this:
```
/// Either we have a defined size or we just use the summed width/height of
/// the current children.
final dimensionAvailable = size[vectorIndex] != 0.0
    ? size[vectorIndex]
    : componentsDimension;
```
With your suggestion you wouldn't allow to set an explicit `size` for the `LayoutComponent`, but only for its parents up the hierarchy.
What if i would check for an explicit `size` of the `LayoutComponent` _and_ if that is zero look for a `size` in the first found `SizeProvider` up the hierarchy? I wouldn't forbid setting a `size` for the `LayoutComponent`.

spydon: > What if i would check for an explicit size of the LayoutComponent and if that is zero look for a size in the first found SizeProvider up the hierarchy? I wouldn't forbid setting a size for the LayoutComponent.

I think this will be prone to bugs, what if the user shrinks the size down to 0? Then it will all of a sudden start using `componentsDimension` instead. Maybe take in a boolean instead that specifies whether it should be manually or automatically sized? That boolean could be set to true if a `size` is sent in through the constructor.

We'd also have to connect a listener to `size` in the cases that it is manually sized, and in the cases that it is automatically sized we have to connect a listener to the ancestor size, so that we can do `layoutChildren` when the bounds change.

rivella50: It gets more and more complicated...
Pasha suggested that the developer should call layoutChildren if the size is manually changed, i.e. not to use a listener for that case.

spydon: I don't see the point of laying that responsibility on the user when we have the possibility to easily add a listener to the bounds with:
```dart
final bounds = isManuallySized ? size : findParent<PositionComponent>;
bounds.addListener(layoutChildren);
```

st-pasha: Perhaps let's not overcomplicate this?

In the first version of the LayoutComponent the `size` will have to be provided explicitly. In the next version (in a separate PR), we will relax that requirement and allow `size` to be omitted, in which case it will be derived from a parent.

rivella50: As you like. It's already implemented though.

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,143 @@
...
+/// A relayout is performed when
+///  - a change in this component's children takes place
+///  - the size of this component or a parent changes
+///  - the [gap] parameter is changed
+abstract class LayoutComponent extends PositionComponent with HasGameRef {
+  LayoutComponent(
+    this.direction,
+    this.mainAxisAlignment,
+    this._gap,
+    this.isManuallySized,
```
spydon, line 18: I think this should take a `size` too and automatically set `isManuallySized` to true if it is provided and then omit the boolean for the constructor, but having it as mutable so that the user can change it afterwards.

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,147 @@
...
+abstract class LayoutComponent extends PositionComponent with HasGameRef {
+  LayoutComponent(
+    this.direction,
+    this.mainAxisAlignment,
+    this._gap,
+    Vector2? size,
+  ) : super(
+          size: size,
+        ) {
+    isManuallySized = size != null;
```
spydon, line 22: This doesn't have to be done in the body, it can be done directly after the arguments and then the variable doesn't have to be late.

packages/flame/lib/src/experimental/column_component.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:flame/src/experimental/layout_component.dart';
+import 'package:flutter/widgets.dart';
+import 'package:vector_math/vector_math_64.dart';
+
+/// Allows laying out children in a column by defining a [MainAxisAlignment]
+/// type.
+class ColumnComponent extends LayoutComponent {
+  ColumnComponent({
+    MainAxisAlignment mainAxisAlignment = MainAxisAlignment.start,
```
st-pasha, line 9: Perhaps call this simply `alignment`?

packages/flame/lib/src/experimental/column_component.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:flame/src/experimental/layout_component.dart';
+import 'package:flutter/widgets.dart';
+import 'package:vector_math/vector_math_64.dart';
+
+/// Allows laying out children in a column by defining a [MainAxisAlignment]
+/// type.
+class ColumnComponent extends LayoutComponent {
+  ColumnComponent({
```
st-pasha, line 8: It would be good to add docs here, as it is not clear what the parameters mean. Especially `size` is the tricky one here...

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,146 @@
...
+import 'package:flutter/widgets.dart';
+
+enum Direction { horizontal, vertical }
+
+/// Super class for [RowComponent] and [ColumnComponent]
+/// A relayout is performed when
+///  - a change in this component's children takes place
+///  - the size of this component or a parent changes
+///  - the [gap] parameter is changed
+abstract class LayoutComponent extends PositionComponent {
```
st-pasha, line 13: Is this class intended to be user-facing? That is, do we anticipate a need for a derived class that is neither RowComponent nor ColumnComponent? If not, then this could be declared `@internal` and un-exported.

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,146 @@
...
+  LayoutComponent(
+    this.direction,
+    this.mainAxisAlignment,
+    this._gap,
+    Vector2? size,
+  )   : isManuallySized = size != null,
+        super(
+          size: size,
+        );
+  final Direction direction;
```
st-pasha, line 23: looks like it would be easier to just have a boolean `isHorizontal` here.

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,146 @@
...
+    final list = children.whereType<PositionComponent>().toList();
+    if (list.isEmpty) {
+      return;
+    }
+    final currentPosition = Vector2.zero();
+    final componentsDimension = list.fold<double>(
+      0,
+      (previousValue, element) => direction == Direction.horizontal
+          ? previousValue + element.width
+          : previousValue + element.height,
```
st-pasha, line 61: `previousValue + element.size[vectorIndex]`.
A simple for-loop might be more understandable here...

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,146 @@
...
+    /// the current children.
+    final componentBounds = _bounds;
+    final dimensionAvailable =
+        componentBounds != null && componentBounds[vectorIndex] != 0.0
+            ? componentBounds[vectorIndex]
+            : componentsDimension;
+    switch (mainAxisAlignment) {
+      case MainAxisAlignment.start:
+        for (final child in list) {
+          child.position = Vector2(currentPosition.x, currentPosition.y);
```
st-pasha, line 76: `child.position.setFrom(currentPosition)`

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,146 @@
...
+          currentPosition[vectorIndex] += direction == Direction.horizontal
+              ? child.width + gap
+              : child.height + gap;
```
st-pasha, line 79: `child.size[vectorIndex] + gap`

packages/flame/test/experimental/layout_component_test.dart
```diff
@@ -0,0 +1,287 @@
...
+  void _initChildren() {
+    firstText = TextComponent(
+      text: 'First',
+    )
+      ..textRenderer = TextPaint(
+        style: const TextStyle(
+          fontSize: 22,
+        ),
+      )
+      ..size = Vector2(50, 50);
```
st-pasha, line 27: You don't need a `TextComponent` here: just a simple `PositionComponent(size: Vector2(50,50))` will do.
(Similarly for secondText and thirdText).

packages/flame/test/experimental/layout_component_test.dart
```diff
@@ -0,0 +1,287 @@
...
+      await _initScene(game, MainAxisAlignment.center, Vector2(50, 50), 10);
+      final startPosition = (rowComponent.size.x - componentsWidth - 20) / 2;
+      expect(firstText.absolutePosition.x, startPosition + 50);
+      expect(firstText.position.x, startPosition);
+      expect(secondText.position.x, startPosition + 60);
+      expect(thirdText.position.x, startPosition + 170);
+    });
+  });
+
+  group('ColumnComponent', () {
```
st-pasha, line 173: Seems like there is a lot of repetition here with the `RowComponent` - how about we merge them? Something like
```dart
for (var k = 0; k <= 1; k++) {
  group(k == 0 ? 'RowComponent' : 'ColumnComponent', () {
    Future<void> initScene(...) async {
      game.add(k == 0? RowComponent(...) : ColumnComponent(...));
    }

    testWithFlameGame('...', (game) async {
      ...
      expect(firstComponent.position[k], 0);
      expect(secondComponent.position[k], 60);
      expect(thirdComponent.position[k], 170);
    });
  });
}
```

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,146 @@
...
+///  - the [gap] parameter is changed
+abstract class LayoutComponent extends PositionComponent {
+  LayoutComponent(
+    this.direction,
+    this.mainAxisAlignment,
+    this._gap,
+    Vector2? size,
+  )   : isManuallySized = size != null,
+        super(
+          size: size,
```
st-pasha, line 21: The `size` of the component must be set even if the user didn't specify it explicitly. Without that, the children components wouldn't be able to receive tap events.

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,146 @@
...
+          ? previousValue + element.width
+          : previousValue + element.height,
+    );
+    final vectorIndex = direction == Direction.horizontal ? 0 : 1;
+    final gapDimension = gap * (list.length - 1);
+
+    /// Either we have a defined size or we just use the summed width/height of
+    /// the current children.
+    final componentBounds = _bounds;
+    final dimensionAvailable =
```
st-pasha, line 69: Most of the variables in this method are confusingly named. When naming variables, I find it easier to first try to explain what the variable represents, and then convert the first few words of that explanation into a name.

For example, consider `componentsDimension`. An explanation for this variable would sound like "the total width/height of all children components". So the variable name could be something like `totalSizeOfComponents`. Similarly, `gapDimension` -> `totalGapsSize`, `dimensionAvailable` -> `availableSpace`, `freeSpacePerComponent` -> `extraGapBetweenComponents`, etc. Another good synonym for "dimension" is "extent".

packages/flame/lib/src/experimental/column_component.dart
rivella50: What if we would like to add a `crossAxisAlignment` later?

packages/flame/lib/src/experimental/layout_component.dart
rivella50: It's all experimental at the moment, so we just don't know yet.

packages/flame/lib/src/experimental/column_component.dart
st-pasha: then we could call it `crossAlignment`

spydon: It's pretty nice to keep the same naming as for the rows and columns for Flutter though, even though their naming isn't optimal.

st-pasha: It is superfluous though: when you say `alignment: MainAxisAlignment.center`, you're already mentioning that this is a main axis alignment. Saying it twice sounds tautological.

Also, Flutter has `mainAxisSize`, whereas we use simple `size`.

spydon: Yeah, I agree it's a bad name. I still think I lean to having it consistent with Flutter though.

`mainAxisSize` is not the same thing as our `size`.

packages/flame/lib/src/experimental/layout_component.dart
spydon: Passing that will be the same as not passing anything, it already defaults to the zero vector.
I think that the `size` should just be set in `layoutChildren`.

rivella50: This results in a stackoverflow error since there is a listener on size which calls `layoutChildren`.

st-pasha: The listener ought to be on parent's size, whereas here we'd be setting the component's own size.
Note that the size doesn't have to be set in the constructor - we can defer setting it to the layout stage

rivella50: But if we don't have a listener on the layout component's `size` as well there won't be a call to `layoutChildren` on resizing it.

packages/flame/lib/src/experimental/column_component.dart
spydon: Let's go with the better name (`alignment`) and skip being consistent with Flutter, I think our users will easily be able to figure it out anyways.

doc/flame/components.md
```diff
@@ -1104,6 +1104,21 @@ Check the example app
 for details on how to use it.
 
 
+## LayoutComponent
```
spydon, line 1107: If we won't expose this we should change the docs to be about `RowComponent` and `ColumnComponent` instead.

packages/flame/lib/src/experimental/layout_component.dart
spydon: > Perhaps let's not overcomplicate this?
> 
> In the first version of the LayoutComponent the `size` will have to be provided explicitly. In the next version (in a separate PR), we will relax that requirement and allow `size` to be omitted, in which case it will be derived from a parent.

When re-reviewing this I agree, maybe it's a bit overkill to listen to the parent's size.
Then we can scrap `isManuallySized` too, so that the component only grows to its size, can we restrict the user from changing the `size` variable from the outside though? Then we only need to do the layout when children are added.

Later we could add a `WrapComponent` so that components can be organized within a set size.

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -32,17 +26,15 @@ abstract class LayoutComponent extends PositionComponent {
...
-      isManuallySized ? size : findParent<PositionComponent>()?.size;
-
   @override
-  Future<void> onLoad() async {
-    final size = _bounds;
-    size?.addListener(layoutChildren);
+  set size(Vector2 size) {
+    if (_allowSetSize) {
+      this.size = size;
+    }
```
st-pasha, line 33: otherwise, some kind of error should be reported to the user, right?

rivella50: By throwing an exception or how is this normally done within Flame?

spydon: I think an assertion is fine

packages/flame/lib/src/experimental/layout_component.dart
spydon: What do you think about this @rivella50? :)

rivella50: @spydon Sure, why not. I've already removed setting the `size` and `isManuallySized` last weekend.
Just to be clear: Without being able to set the `size` (parent or `LayoutComponent`) the alignment types spaceXY don't make sense anymore.

spydon: > Without being able to set the size (parent or LayoutComponent) the alignment types spaceXY don't make sense anymore.

Hmm, that's true, a shame to drop that one...
But maybe it is worth dropping it to just have a simple first version?

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,125 @@
...
+  LayoutComponent(
+    this.direction,
+    this.alignment,
+    this._gap,
+  );
+  final Direction direction;
+  final MainAxisAlignment alignment;
+  bool _allowSetSize = false;
+
+  /// gap between components
```
spydon, line 21:
```suggestion
```

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,125 @@
...
+enum Direction { horizontal, vertical }
+
+/// Super class for [RowComponent] and [ColumnComponent]
+/// A relayout is performed when
+///  - a change in this component's children takes place
+///  - the [gap] parameter is changed
+abstract class LayoutComponent extends PositionComponent {
+  LayoutComponent(
+    this.direction,
+    this.alignment,
```
spydon, line 14: So this should be removed?

rivella50: No no, the gap still works since it leads to an additional growing of the size.
But the spaceXY modes calculate the available space of the existing components and since there is no free space to be distributed they don't have an effect.

spydon: Not the gap, the alignment (comments refer to the line above)

rivella50: Some alignment modes still work, e.g. start and end and possibly center as well.

spydon: Then we should use an enum of our own for this that only includes the modes that we support.

wolfenrain (approved):

packages/flame/test/experimental/layout_component_test.dart
```diff
@@ -0,0 +1,147 @@
...
+      size: Vector2(100, 100),
+    );
+
+    thirdComponent = PositionComponent(
+      size: Vector2(150, 150),
+    );
+  }
+
+  for (var k = 0; k <= 1; k++) {
+    group(k == 0 ? 'RowComponent' : 'ColumnComponent', () {
```
luanpotter, line 31: would this be better as:
```
for (final name in ['RowComponent', 'ColumnComponent']) {
```
or, maybe even better, extract the block to a function and call it twice?

```
group('RowComponent', testLayoutComponent(RowComponent))
group('ColumnComponent', testLayoutComponent(ColumnComponent))

```

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -0,0 +1,94 @@
...
+    }
+    final currentPosition = Vector2.zero();
+    final vectorIndex = direction == Direction.horizontal ? 0 : 1;
+    final totalSizeOfComponents = list.fold<double>(
+      0,
+      (previousValue, element) => previousValue + element.size[vectorIndex],
+    );
+    _allowSetSize = true;
+    size[vectorIndex] = totalSizeOfComponents;
+    _allowSetSize = false;
```
luanpotter, line 56: I don't think you need this variable -- can you just user `super.size`?

example:
```
class A {
  double size = 1.0;
}

class B extends A {
  @override
  set size(double size) {
    assert(false, 'Setting the size is currently unsupported.');
  }
  
  void foo() {
    // this.size = 2.0; -> this breaks
    super.size = 2.0;
  }
}

void main() {
  final b = B()..foo();
  print(b.size);
}
```

packages/flame/lib/src/experimental/row_component.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:flame/src/experimental/layout_component.dart';
+
+/// Allows laying out children in a row by defining a [LayoutComponentAlignment]
+/// type.
+class RowComponent extends LayoutComponent {
+  RowComponent({
+    LayoutComponentAlignment alignment = LayoutComponentAlignment.start,
```
luanpotter, line 7: can't you use `super.alignment` here?

packages/flame/test/experimental/layout_component_test.dart
```diff
@@ -0,0 +1,147 @@
...
+      size: Vector2(50, 50),
+    );
+
+    secondComponent = PositionComponent(
+      size: Vector2(100, 100),
+    );
+
+    thirdComponent = PositionComponent(
+      size: Vector2(150, 150),
+    );
```
luanpotter, line 27: nit: you can use Vector2.all

doc/flame/components.md
```diff
@@ -1104,6 +1104,18 @@ Check the example app
 for details on how to use it.
 
 
+## RowComponent & ColumnComponent
```
luanpotter, line 1107: if it is experimental, should we flag it in the doc, similar to how it is done [elsewhere](https://github.com/flame-engine/flame/blob/47970224f8c9c90718c54301ee69d9cddcced87b/doc/flame/camera_component.md)?

spydon:
Do you want to continue on this PR @rivella50 or should we take it over? :)
(Great work so far btw!)

rivella50:
@spydon Feel free to take it over. My current project doesn't let me much free time at the moment.

spydon:
I'll close this for now since it's been open for so long, but we should definitely look at this when creating these components based on the `AlignComponent`.

PR #1976, opened 2022-10-01
feat: Adding onChildrenChanged

### LINKED ISSUE #1973: Add Component.onChildrenChanged() lifecycle method
# Problem to solve

Currently a component has no way of knowing when it acquires new children or loses existing ones. Which is a shame really, because how it can take good care of its children if it doesn't know what happens to them?

# Proposal

Add a new lifecycle method `.onChildrenChanged()` which would fire whenever new children are added to the component, or existing children removed.

The method could have an argument `child`, which would indicate the component that caused the lifecycle event. If `child.parent == this` then the child was just added to the current component; otherwise it was removed.

# More information
This would be useful for #1944

rivella50:
# Description
In order to get notified about a change in a parent's children this new function will be called now.
This feature will help simplifying e.g. the implementation of issue #1944 (Row and Column components).

## Checklist

## Breaking Change?

## Related Issues
Closes #1973


alestiago:
@rivella50 can you add some tests?

rivella50:
@alestiago Sure, if someone can tell me that's the way to go i will add the needed tests.

packages/flame/lib/src/components/core/component.dart
```diff
@@ -462,6 +462,10 @@ class Component {
     _children?.forEach((c) => c.updateTree(dt));
   }
 
+  /// This method will be invoked from lifecycle if [component] has been added
+  /// to or removed from its parent children list.
+  void onChildrenChanged(Component component) {}
```
spydon, line 467: I think it would be nice to add an enum here that says whether the component was added or removed, so that the user doesn't have to look at the parent to know that. I would also rename `component` to `child` here.

rivella50: Thanks for the feedback. How about that enum:
```
ComponentOnChildrenChangedType { add, remove }
```
and added in component.dart ?

spydon: I think `ChildrenChangedType` is enough. Yeah we can have it directly in `component.dart` for now, just remember to export it in the barrel-file.

packages/flame/lib/src/components/core/component.dart
```diff
@@ -897,6 +901,8 @@ class Component {
 
 typedef ComponentSetFactory = ComponentSet Function();
 
+enum ChildrenChangedType { add, remove }
```
spydon, line 904: Since this has already happened I think this enum should be in past tense:
```suggestion
enum ChildrenChangedType { added, removed }
```

spydon (commented):
Looks good! Some docs in `doc/components.md` would be good describing the new lifecycle method.

spydon:
Lifecycle master @st-pasha, could you review this too? :smile:

packages/flame/lib/src/components/core/component.dart
rivella50: Sure, no problem. I just thought why having past tense twice, since the enum name is already in past tense.

spydon: I missed that! But I think it is better to have the enum as `ChildrenChangeType` and then the values as past tense, I think it can even be `ChildChangeType` since the change only regards the change of the specific child that we are passing in to the method.

rivella50: Ok, i can switch it.
But i slightly disagree with renaming the enum to ChildChangeType since the change refers to the children list we are interested here. Passing the added or removed child component is just an additional info @st-pasha suggested.

spydon: Alright, that makes sense, keep it as `Children` then. :)

spydon:
Is this still to be considered a draft? If not you can press "Ready for review". :)

rivella50:
Meanwhile i would say we are not in draft mode anymore ;-)

packages/flame/lib/src/components/core/component.dart
```diff
@@ -897,6 +901,8 @@ class Component {
 
 typedef ComponentSetFactory = ComponentSet Function();
 
+enum ChildrenChangeType { added, removed }
```
alestiago, line 904: Do we need this? 

Can't be just check if the given child if the child has the parent or not?

How do we envision this to work for reparenting?

packages/flame/test/components/component_test.dart
```diff
@@ -344,6 +344,51 @@ void main() {
           expect(wrapper.contains(child), true);
         },
       );
+
```
alestiago, line 347: Can we make a group for "onChildrenChanged"?

packages/flame/lib/src/components/core/component.dart
rivella50: @spydon  suggested this additional type parameter for exactly not having to do this check.
Regarding reparenting: I'm not sure if i understand your question correctly. It also works for reparenting, whereas the old parent gets notified with a `removed` type, and the new parent with the `added` type.

packages/flame/test/components/component_test.dart
rivella50: Sure

packages/flame/lib/src/components/core/component.dart
rivella50: @spydon  suggested this additional type parameter for exactly not having to do this check.
Regarding reparenting: I'm not sure if i understand your question correctly. It also works for reparenting, whereas the old parent gets notified with a `removed` type, and the new parent with the `added` type (there is also a test for that case).

spydon: This is a convenience design for the user to be able to easier determine whether a child has been added or removed. I don't think it would be clear to the user that they can check what the parent is to know whether it has been added or removed and this has no performance loss. 

> How do we envision this to work for reparenting?

It will follow the rest of the lifecycle flow that happens when reparenting.

spydon (approved):

luanpotter (approved):
LGTM

wolfenrain (approved):

PR #1981, opened 2022-10-01
feat: onComponentTypeCheck support for ShapeHitbox

ASGAlex:
# Description

`onComponentTypeCheck` was only supported for components but not for hitboxes, because ShapeHitbox implements `GenericCollisionCallbacks` instead `CollisionCallbacks`.

Specifying `onComponentTypeCheck` for individual hitboxes might be useful in cases where `Component` have hitboxes for collisions and also some utility hitboxes (for example, to check available directions for pathfinding). Colliding between this two types of hitboxes should not lead to movement restrictions and so on. But without `onComponentTypeCheck` in `ShapeHitbox` we can not to perform such check.

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame/lib/src/collisions/broadphase/quadtree/has_quadtree_collision_detection.dart
+++ b/packages/flame/lib/src/collisions/broadphase/quadtree/has_quadtree_collision_detection.dart
@@ -68,16 +68,16 @@ mixin HasQuadTreeCollisionDetection on FlameGame
 
   bool onComponentTypeCheck(PositionComponent one, PositionComponent another) {
     var checkParent = false;
-    if (one is CollisionCallbacks) {
-      if (!(one as CollisionCallbacks).onComponentTypeCheck(another)) {
+    if (one is GenericCollisionCallbacks) {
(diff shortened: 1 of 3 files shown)


spydon (approved):

ASGAlex:
@spydon  I also implemented `onComponentTypeCheck` for Sweep algorithm. But after performing a benchmark I decided that it absolutely not worth in performance, it only makes code more complicated. Maybe we should to document this somewhere? To avoid new "issues" after years, when everithing will be forgotten 😉

spydon:
> @spydon I also implemented `onComponentTypeCheck` for Sweep algorithm. But after performing a benchmark I decided that it absolutely not worth in performance, it only makes code more complicated. Maybe we should to document this somewhere? To avoid new "issues" after years, when everithing will be forgotten wink

It is probably worth having it just for consistency, because it will also avoid callbacks being called for collisions that the user doesn't care about, even though the performance gain isn't that high?

ASGAlex:
Ok, I pushed what I have for now: https://github.com/flame-engine/flame/pull/1996 
Let's to discuss realization there. This code does not looks good for me but fixing it in right way will lead to breaking changes, I think

PR #1982, opened 2022-10-02
feat: Flame TiledAtlas

### LINKED ISSUE #1898: Flame Tiled: Batch caching by image name leads to incorrect painting.
# Current bug behaviour

Tiled maps multiple images (either multiple atlases, or multiple single-sourced images) will not render correctly in staggered modes. It may not operate correctly in other modes as well.

# Expected behaviour

Maps that match what Tiled renders.

# Steps to reproduce

1. Create a tiled map (isometric, staggered, x-asis, even or odd).
2. Use multiple, oversized images for the tileset (for overlapping).
  a. A good CC-0 set: [kenny.nl](https://www.kenney.nl/assets/isometric-miniature-dungeon)
3. Place images such that the overlap.

# More information

What Flame does today is optimized for batch rendering (performance), but that means map authors need to know about the limitations. A quick rundown of what happens:

1. `flame_tiled` creates a `SpriteBatch` for every image in the tiled map.  Images can be N*(Atlas||Single Image).
2. These batches are stored in a map keyed on the _source file name_ (`_cachedSpriteBatches`)
3. While sweeping each layer, and then sweeping each row, and then sweeping each tile in the row, the batch is looked up.
4. "cache tiles": transforms generated from the layer->row->tile are stored for each image, not for each tile  (`batch.addTransform`)
5. render loop: walk over all visible layers.  walk over all cached batches (remember: this is image keys, not tile keyed). call sprite batch render.

What could be done about it?

1. You tell everyone in documentation and perhaps runtime warning: don't use multiple images/atlases when generating tiled maps.
2. We could collapse every image into an image atlas on load and have only one atlas to draw from for every layer.
3. Convert RenderableTileMap to not actual render - just do all the parallax / shifting / tile **layout**.  Leave the render step to another class that flips between sprite batching or sprite painting depending on map conditions.

Option 1 seems harsh, but not uncommon - its an optimization build step.
Option 2 could lead to increased load times, higher memory usage - flips could double that if incorrectly handled. 2D-rect-bin packing is NP-hard.
Option 3 could be a bit of work, but also might lead to easier to animate tiles, or positioning players on the maps.

The goal is to go from randomly having this: 
image

To having this:
image

Which is achievable with option 1 for now.

jtmcdole:
# Description

Today: `flame_tiled` creates a series of sprite batches for each image associated with a tiled_map and layer.  This is only performant for maps that have a single atlas image - something easily overlooked while using Tiled. Worse, this causes painting order issues like that in #1898.  This can be easily seen in staggered tiled maps (isometric, staggered, or hexagonal), or even orthogonal if the tiles are wider than the map tiles.  In the following example; the tiles are taller than the isometric tiles:

painting out of order

Flame Tile uses sprite batching for performance - to paint correctly in staggered, we paint tiles in red-line order, even though we visit them in green-line ordering.  As seen above, this have some cached batches (the barrels) not obeying the order because they are in a different image.

stagger_draw

### Performance

* Windows 11
* <8K textures
* 200x200 tile isometric staggered map
* Same "seed" for reproducible randomness
* 4 Layers - ground, floor, furniture, walls

**Debug**: `flutter run -d windows` 

* HEAD: 14fps
* ATLAS: 60fps

**Release**: `flutter run -d windows --release` 

* HEAD: 14fps
* ATLAS: 80~90fps

(flutter is vsync locked; maximum FPS would be 100 for my monitor)

So:

Allocate an atlas per map to keep sprite paint orders correct. That atlas is shared between all layers, but sprite batching is still generated for each layer. If you already use one atlas permap, this will have no change. Multiple tileset images will see both a performance gain from single batch renders, and correct painting on staggered maps.

- Fix animations with offsets
- Remove TiledLayer._loadImages - no longer used.
- Animations that are in other layers need also translate by offset
- Parallel load atlas images

Cheap 2D rect bin packing

- Sort: images from large->small
- Keep searching for smallest rect, but input rects should be big->small

Other changes:

- Moved some testing classes around for better code sharing
- Fixed tests for external tsx loading. It is dangerous to override global static values...

## Checklist

## Breaking Change?

## Related Issues

Fixes #1898

## Important Notes

1. Rectangle bin packing is NP-Complete. Don't spend a whole lot of time trying to solve it.
2. Chrome on **Android** has a hardcoded texture size limit of 4k. We do not scale atlas images down to meet that requirement - which means you could have missing images.  There are no warnings for this before this PR or with this PR - but I do try to pack in those constraints (kIsWeb).
3. 8K textures on Windows 11 native can lead to a bad time. I'm filing a bug with Flutter to investigate.  No problems seen on Windows 11 Chrome, or Windows 10 fullscreen.


jtmcdole:
This is a big change; I'm in no rush, so take your time.  Happy to make changes.

spydon:
There should only be one conventional commit tag so I'll remove the `fix` one :)

spydon:
@kurtome @ufrshubham could we have your reviews here too?

packages/flame_tiled/lib/src/tile_atlas.dart
```diff
@@ -0,0 +1,125 @@
...
+    final key = atlasKey(imageList);
+    if (atlasMap.containsKey(key)) {
+      return atlasMap[key]!.clone();
+    }
+
+    if (imageList.length == 1) {
+      // we already have an atlas
+      final tiledImage = imageList.first;
+      final image = (await Flame.images.load(tiledImage.source!)).clone();
+      Flame.images.add(key, image);
```
ufrshubham, line 82: This will cache the same image at 2 different keys. Was that the intent here? If not, then this can be written as:
```dart
final image = (await Flame.images.load(tiledImage.source!, key: key)).clone();
```

jtmcdole: Good catch.  My original intent was that every map gets its own handle; I moved things around and `atlasMap` above does that job. 

Only concern: If someone was referencing this exact image elsewhere in their game (e.g. shared sprite sheets), would you expect there to be an "atlas{...}" image and the source image cached?   We look only for the atlas-key above and return a clone - we _could_ also just return the normal image here as cloned.

jtmcdole: ah, we still need the key since spritebatch generates #flips.

kurtome:
Only had time to skim it but the approach makes sense to me.

But, wouldn't it be better to fail hard rather than silently if the images don't fit into the 4096 square texture? Or automatically create a second atlas?

packages/flame_tiled/lib/src/rectangle_bin_packer.dart
```diff
@@ -0,0 +1,76 @@
...
+        search.right,
+        found.bottom,
+      );
+      bins.add(newRect);
+    }
+    bins.sort(_compare);
+    return found;
+  }
+
+  int _compare(Rect a, Rect b) {
```
kurtome, line 72: nit: rename this or add a comment explaining why we need to sort by height then width

packages/flame_tiled/lib/src/rectangle_bin_packer.dart
```diff
@@ -0,0 +1,76 @@
...
+  late final List<Rect> bins = [Rect.fromLTWH(0, 0, maxX, maxY)];
+
+  RectangleBinPacker({
+    this.maxX = kIsWeb ? 4096 : 8192,
+    this.maxY = kIsWeb ? 4096 : 8192,
+  });
+
+  /// Finds a free space for a rectangle of lengths [width] and [height] in
+  /// the atlas.
+  Rect pack(double width, double height) {
```
kurtome, line 28: Can you link to any page that explains the bin packing algorithm? (new to me)

packages/flame_tiled/lib/src/tile_atlas.dart
```diff
@@ -0,0 +1,127 @@
...
+        if (image?.source != null) {
+          imageSet.add(image!);
+        }
+      }
+    }
+    return imageSet;
+  }
+
+  /// Loads all the tileset images for the [map] into one [TiledAtlas].
+  static Future<TiledAtlas> loadImages(tiled.TiledMap map) async {
```
kurtome, line 65: nit: name this "load" since it's the main public constructor for this class any other name is a bit misleading?

packages/flame_tiled/lib/src/tile_atlas.dart
```diff
@@ -0,0 +1,127 @@
...
+  final Map<String, Offset> offsets;
+
+  /// The single batch operation for this atlas.
+  final SpriteBatch? batch;
+
+  /// Image key for this atlas.
+  final String key;
+
+  /// Track one atlas for all images in the Tiled map.
+  TiledAtlas._(this.atlas, this.offsets, this.key)
```
kurtome, line 26: I'm assuming this is how you do private constructors in Dart? That's cool!

Maybe include a note to use the static "load" method in the comment?

jtmcdole:
> But, wouldn't it be better to fail hard rather than silently if the images don't fit into the 4096 square texture? Or automatically create a second atlas?

A second atlas would break the painting order just as multiple images would. We could assert(?) but it does work well on other web browser.  It would be better if Flutter told us what the maximum texture size is.

packages/flame_tiled/lib/src/rectangle_bin_packer.dart
jtmcdole: https://en.wikipedia.org/wiki/Rectangle_packing#Packing_different_rectangles_in_a_minimum-area_rectangle

https://www.david-colson.com/2020/03/10/exploring-rect-packing.html

packages/flame_tiled/lib/src/tile_atlas.dart
jtmcdole: I'm going to name it `fromTiledMap` to match `RenderableTiledMap`

spydon (approved):
Great job! And example for the examples site would be good to have too, but doesn't necessarily need to be in this PR.

ufrshubham:
LGTM 🤘🏻

PR #1992, opened 2022-10-02
feat: Added SpriteFont class

### LINKED ISSUE #1935: Create SpriteFont class
Currently we have `SpriteFontTextFormatter`, but not the `SpriteFont` class. Encapsulating the font  information into a dedicated class would make it easier to create multiple formatters from the same font definition (i.e. for different font sizes, colors, etc); and also to support sprite-fonts from the `TextStyle` class.

The change would affect the construction of the `SpriteFontTextFormatter`. We could make it slow via the regular deprecation cycle, or fast by making a breaking change.

st-pasha:
# Description

This PR adds the `SpriteFont` class, which encapsulates reusable information about a sprite font. Text renderers / text formatters can be created based on this `SpriteFont`.

In addition:
- `GlyphInfo` removed (was internal), and `GlyphData` deprecated -- instead, there is now a consolidated `Glyph` class;
- Sprite font now supports extended Unicode characters and ligatures (when a combination of letters is represented as a single glyph)
  - This feature can also be used to define simple name substitutions. For example, you can declare that "[gp]" or "&gp;" would represent a gold coin icon;
- Sprite font supports variable-width fonts;
- Sprite font supports glyphs where the source rect is different from the logical rect;
- Sprite font renderer can now control the color of the text rendered.

## Checklist

## Breaking Change?

## Related Issues

Closes #1935


packages/flame/lib/src/text/common/glyph.dart
```diff
@@ -0,0 +1,104 @@
+import 'package:flame/src/text/common/glyph_data.dart';
+import 'package:flame/src/text/common/sprite_font.dart';
+import 'package:meta/meta.dart';
+
+/// [Glyph] describes a single character/ligature/icon within a [SpriteFont].
+///
+/// A glyph has an associated "glyph box", which is the box where the glyph is
+/// logically located. Here "logically" means that it includes not only the
+/// character itself, but also some padding around it as necessary for the
+/// character's to look nice within the text. For all glyphs in a font, their
```
spydon, line 10:
```suggestion
/// characters to look nice within the text. For all glyphs in a font, their
```

packages/flame/lib/src/text/common/glyph.dart
```diff
@@ -0,0 +1,104 @@
...
+/// and descent, but possibly different widths.
+///
+/// The properties [left], [top], [width] and [height] describe the location of
+/// the glyph box within the source image.
+///
+/// In addition to the logical "glyph box", a glyph may also have a "source
+/// box", which describes a rectangle within the source image where the glyph's
+/// pixels are actually located. The source box may be larger or smaller than
+/// the glyph box. It will be larger if the glyph has a particularly large
+/// flourish that trespasses upon other character's space; or smaller if the
```
spydon, line 21:
```suggestion
/// flourish that trespasses upon other characters' space; or smaller if the
```

packages/flame/lib/src/text/common/glyph.dart
```diff
@@ -0,0 +1,104 @@
...
+  Glyph(
+    this.char, {
+    required this.left,
+    required this.top,
+    double? width,
+    double? height,
+    double? srcLeft,
+    double? srcTop,
+    double? srcRight,
+    double? srcBottom,
```
spydon, line 34: Can't these be non-nullable and default to -1 and then the assert just look on the -1?

packages/flame/lib/src/text/common/glyph_data.dart
```diff
@@ -1,15 +1,19 @@
+@Deprecated('This class will be removed in 1.5.0; use Glyph instead')
```
spydon, line 1: 1.6.0 right? Since 1.4.0 is released now

packages/flame/lib/src/text/common/sprite_font.dart
```diff
@@ -0,0 +1,66 @@
+import 'dart:ui';
+
+import 'package:flame/src/text/common/glyph.dart';
+import 'package:meta/meta.dart';
+
+class SpriteFont {
```
spydon, line 6: Missing dartdocs

packages/flame/lib/src/text/sprite_font_renderer.dart
```diff
@@ -24,10 +25,13 @@ import 'package:flame/src/text/text_renderer.dart';
 /// the opacity of the paint's color will make the text semi-transparent.
 class SpriteFontRenderer
     extends FormatterTextRenderer<SpriteFontTextFormatter> {
+  @Deprecated('Use SpriteFontRenderer.fromFont() instead; this constructor '
+      'will be removed in 1.5.0')
```
spydon, line 29:
```suggestion
      'will be removed in 1.6.0')
```

spydon (approved):
Lgtm! Some minor comments

packages/flame/lib/src/text/common/glyph_data.dart
st-pasha: Right, forgot to update these after the release.

packages/flame/lib/src/text/common/glyph.dart
st-pasha: It wouldn't simplify the assert, since we'd still need to check that either all four are -1 or all four are >= 0.

Also, from the semantic perspective, if a property is nullable then it sends a clear message that the property can be omitted. However, if the property has the default value of `-1` then I feel the user may want to know what does that mean. However, the value of `-1` is temporary, its only purpose is to store "null" value in a non-nullable field, it will be replaced by a real value once the `Glyph` is sent into the `SpriteFont`.

spydon: Fair enough!

packages/flame/lib/src/text/common/sprite_font.dart
st-pasha: added

erickzanardo (approved):

PR #1996, opened 2022-10-02
feat: onComponentTypeCheck for Sweep

ASGAlex:
# Description

Add onComponentTypeCheck functionality to Sweep.

## Checklist

## Breaking Change?

This is a quession for discussion. I pushed not best impllementation but it avoids breaking changes. We should to decide here, what to do.

### DIFF
--- a/packages/flame/lib/src/collisions/broadphase/sweep/sweep.dart
+++ b/packages/flame/lib/src/collisions/broadphase/sweep/sweep.dart
@@ -31,6 +31,9 @@ class Sweep<T extends Hitbox<T>> extends Broadphase<T> {
       final currentMin = currentBox.min.x;
       for (var i = _active.length - 1; i >= 0; i--) {
         final activeItem = _active[i];
+        if (_broadphaseCheckCache[activeItem]?[item] == false) {
+          continue;
+        }
         final activeBox = activeItem.aabb;
         if (activeBox.max.x >= currentMin) {
           if (item.collisionType == CollisionType.active ||
--- a/packages/flame/lib/src/collisions/broadphase/sweep/sweep.dart
+++ b/packages/flame/lib/src/collisions/broadphase/sweep/sweep.dart
@@ -1,10 +1,14 @@
 import 'package:flame/collisions.dart';
 
 class Sweep<T extends Hitbox<T>> extends Broadphase<T> {
-  Sweep({super.items});
+  Sweep({super.items, this.broadphaseCheck});
 
   late final List<T> _active = [];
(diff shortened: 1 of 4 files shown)


packages/flame/lib/src/collisions/broadphase/sweep/sweep.dart
```diff
@@ -27,18 +31,50 @@ class Sweep<T extends Hitbox<T>> extends Broadphase<T> {
...
     }
+    if (_potentialsTmp.isNotEmpty) {
+      for (var i = 0; i < _potentialsTmp.length; i++) {
+        final item0 = _potentialsTmp[i].first;
+        final item1 = _potentialsTmp[i].last;
+        final checkFunction = broadphaseCheck;
+        var checkSuccess = false;
+        if (checkFunction != null &&
+            item0 is ShapeHitbox &&
+            item1 is ShapeHitbox) {
```
spydon, line 59: We don't want to have type checks in here, everything should be based of T.
Possibly the sweep could be chained with another step before it that has a tighter restriction on its generics.

spydon (commented):
Not too fond of this draft unfortunately, it breaks a lot of the generalization of the class. But maybe that logic can be broken out and wrap the sweep instead and then that class could be used by the `StandardCollisionDetection`.

ASGAlex:
> Not too fond of this draft unfortunately, it breaks a lot of the generalization of the class. But maybe that logic can be broken out and wrap the sweep instead and then that class could be used by the `StandardCollisionDetection`.

I'm not happy too. Code looks ugly, it is strictly dependent on PositionComponents and ShapeHitbox, it breaks some tests... Right now i measuring it's performance because I still have questions with these. 

That's why I said that maybe it would be better to leave this unimplemented :-)

spydon:
> > Not too fond of this draft unfortunately, it breaks a lot of the generalization of the class. But maybe that logic can be broken out and wrap the sweep instead and then that class could be used by the `StandardCollisionDetection`.
> 
> I'm not happy too. Code looks ugly, it is strictly dependent on PositionComponents and ShapeHitbox, it breaks some tests... Right now i measuring it's performance because I still have questions with these.
> 
> That's why I said that maybe it would be better to leave this unimplemented :-)

It feels like it should be possible to do this and still follow the generics, since the `GenericCollisionCallbacks` also have this method. But the `ExternalBroadphaseCheck` typedef can't be used since it is bound to `PositionComponent`.

spydon:
I'll close this for now and maybe we can re-do something similar in the future

PR #2035, opened 2022-10-06
docs: make the language tag non moving in code blocks

### LINKED ISSUE #2025: Documentation code blocks' "dart" position changes
The flame documentation is good which contains code blocks that shows the code language also. But, there is a slight problem in it while scrolling the block horizontally. The language tag also moves along with it.

Maybe below screenshot will better explain.

# Current bug behaviour

image

# Expected behaviour

The language tag (dart here) should be fixed to the right side and shouldn't move.

# Steps to reproduce

Go to flame documentation page => https://docs.flame-engine.org/

And click on any section that contains code block, try to change width of window to be able to scroll. Then scroll the code block.

# Flutter doctor output

Not needed

Not needed

# Log information

Not needed

# More information

I will update here once I search for more information on this issue.

I am interested in working on a PR for this.

immadisairaj:
the language tag for code blocks used to move along with scroll of code. Now, with this commit, the tag will be fixed at the top and doesn't move with scroll

# Description

This commit makes a few css style changes in order that the language tags for code blocks in documentation site remains same even when horizontally scrolled.

Along with the above, this commit adds right padding to the code as to not come under the language tag when reaching the end.

Screenrecording:

https://user-images.githubusercontent.com/40348358/194294111-2d966f9e-2600-45b6-ab25-b15bf9d57770.mov

## Checklist

## Breaking Change?

## Related Issues
Closes #2025

### DIFF
--- a/doc/_sphinx/theme/flames.css
+++ b/doc/_sphinx/theme/flames.css
@@ -599,6 +599,8 @@ div.document :is(h1, h2, h3, h4) > code {
   position: relative;
   overflow-x: auto;
   padding: 2px;
+  padding-top: 18px;
+  margin-top: -18px;
 }
 
 pre {
@@ -612,18 +614,18 @@ pre {
   line-height: 125%;
   margin: 0;
   padding: 10px;
-  padding-right: 41px;
+  border-top-right-radius: 0;
 }
 
 div[class^="highlight-"] pre:before {
   background: black;
-  border-bottom-left-radius: 5px;
+  border-top-left-radius: 5px;
   border-top-right-radius: 5px;
   color: #58d0ff;
   font-size: 80%;
   padding: 1px 8px;
   position: absolute;
-  right: 0;
+  right: 1px;
   top: 0;
 }
 
--- a/doc/_sphinx/theme/flames.css
+++ b/doc/_sphinx/theme/flames.css
@@ -607,7 +607,6 @@ pre {
   margin: 0;
   padding: 10px;
   padding-right: 41px;
-  position: unset;
 }
 
 div[class^="highlight-"] pre:before {
--- a/doc/_sphinx/theme/flames.css
+++ b/doc/_sphinx/theme/flames.css
@@ -593,6 +593,7 @@ div.document :is(h1, h2, h3, h4) > code {
 
  div[class^="highlight-"] {
   position: relative;
+  overflow-x: auto;
 }
 
 pre {
--- a/doc/_sphinx/theme/flames.css
+++ b/doc/_sphinx/theme/flames.css
@@ -591,6 +591,10 @@ div.document :is(h1, h2, h3, h4) > code {
  * Code
  *----------------------------------------------------------------------------*/
 
+ div[class^="highlight-"] {
+  position: relative;
+}
+
 pre {
   background: #202020;
   border: none;
@@ -602,7 +606,8 @@ pre {
   line-height: 125%;
   margin: 0;
   padding: 10px;
-  position: relative;
+  padding-right: 41px;
+  position: unset;
 }
 
 div[class^="highlight-"] pre:before {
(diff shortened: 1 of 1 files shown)


spydon (approved):

doc/_sphinx/theme/flames.css
```diff
@@ -602,7 +606,7 @@ pre {
   line-height: 125%;
   margin: 0;
   padding: 10px;
-  position: relative;
+  padding-right: 41px;
```
lex111, line 609: Just want to point out that this change will causes empty space and increases scrollable area at all. Are you sure that's a good idea?

| Before   | After    |
| -------- | -------- |
| image | image |

immadisairaj: Yes, I have already mentioned before why I have added the extra padding.

Let's take an example of one line:
Screenshot 2022-10-06 at 16 49 57

We won't be able to look at what is there below the tag (in above).

So, adding the padding will make it look like the below:
Screenshot 2022-10-06 at 16 51 09

This makes a better readability (like 2nd image)

lex111: This works well for a one-line code block, but increases scrollable area for a multi-line code block. Wouldn't it have been better to add some opacity for a code block lang?

spydon: Maybe we could just put it above the box, instead of in the box? I don't think opacity would make it very readable it it is above the code either.

immadisairaj: Then we should probably replace the padding right to padding top? That makes the tag be inside the box.
But, the issue is that it will add extra space on top for all boxes (which will happen even when moving the tag outside).

I feel padding right would be a good option as most won't be scrolling to the complete right and even if they scroll, it doesn't look so odd from what I feel

spydon: I was thinking without any padding, but putting it outside of the box with a negative margin or such.

immadisairaj: You mean it like this?

Screenshot 2022-10-06 at 17 12 17

immadisairaj:
Sorry that I didn't catch this issue before. 😅 Fixed it with the latest commit.

Before the latest commit:
Screenshot 2022-10-06 at 17 47 39

After the latest commit:
Screenshot 2022-10-06 at 17 48 19

doc/_sphinx/theme/flames.css
spydon: Exactly

immadisairaj: NVM.. I'll add a negative margin top outside 😅

st-pasha:
Now the code tag can overlap with the text above :)
How about we add `top: 7px` so that it is partially within the code block and partially sticks outside? Also in that case it can be made all-rounded: `border-radius: 5px`.

immadisairaj:
> Now the code tag can overlap with the text above :)

Yep, was wondering the same. So, put the screenshot.

We could add the 7px as you've suggested. And, I think we could keep the old style rounded border (top right and bottom left) instead of all round. Like moving it to the top right side keeping same rounded border. How will that be?

Anyway, can you please let me know what looks good and what you feel is better to put. I'm going to bed. Will make a change and push the commit first thing in the morning based on what you've discussed and decided. 🙂 Thank you.

spydon:
> Now the code tag can overlap with the text above :)
> How about we add `top: 7px` so that it is partially within the code block and partially sticks outside? Also in that case it can be made all-rounded: `border-radius: 5px`.
> 
> Screen Shot 2022-10-06 at 9 31 15 AM
> 

I think there is enough space for it not to overlap the text that much as can be seen in the screenshot, but I think your suggested solution looks better anyways! :)

immadisairaj:
Updated Screenshot:
Screenshot 2022-10-07 at 07 38 27

st-pasha (approved):

spydon (approved):

Issue #1545, opened 2022-04-13
SizeEffect on PolygonComponent


mel-mouk:
Hi, I came across an issue while playing around with built-in effect for a small game.

# Current bug behaviour
When I use SizeEffect.to the actual effect is a move, not a resize. I had a look at the source code and tried to override it to make some test. The code is actually updating the size and the component's properties does not show anything weird so I don't really get what's going on

# Expected behaviour
Expected behaviour would be a resize of the component

# Steps to reproduce
Create a simple Game object, add a Rectangle of some size. On tap, add a SizeEffect. Minimalist source code to reproduce below :
```
class SimpleGame extends FlameGame with TapDetector {
  late Vector2 canvasSize;
  late RectangleComponent testSquare;
  @override
  void onGameResize(Vector2 size) {
    super.onGameResize(size);
    canvasSize = size;
  }

  @override
  Future<void> onLoad() async {
    testSquare = RectangleComponent(
        position: canvasSize / 2,
        size: canvasSize / 2,
        paint: BasicPalette.blue.paint(),
        anchor: Anchor.center
    );
    add(testSquare);
    await super.onLoad();
  }

  @override
  bool onTapDown(TapDownInfo event) {
    testSquare.add(SizeEffect.to(Vector2(10, 10), EffectController(duration: 1)));
    return true;
  }
}

main() {
  final game = SimpleGame();

  runApp(
    GameWidget(
      game: game,
    ),
  );
}
```

# Flutter doctor output
```
Flutter (Channel stable, 2.8.1, on macOS 12.2 21D49 darwin-x64, locale en-FR)
    • Flutter version 2.8.1 at /Applications/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 77d935af4d (4 months ago), 2021-12-16 08:37:33 -0800
    • Engine revision 890a5fca2e
    • Dart version 2.15.1
(...)
[✓] Xcode - develop for iOS and macOS (Xcode 13.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • CocoaPods version 1.10.1

[✓] Android Studio (version 2020.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
[...]

PR #2061, opened 2022-10-10
fix!: PolygonComponent no longer modifies _vertices

mattmyne:
# Description

PolygonComponent used to modify the contents of the _vertices list passed to it by offsetting them all by _topLeft. This prevented the user from using the same vertices list again, such as in another call to PolygonComponent. This PR stops PolygonComponent from directly modifying the list vectors and instead add the _topLeft offset where needed within internal code.

This is a breaking change since PolygonComponent's .vertices getter will now return the original, unmodified vertices rather than with the offset added.

## Checklist

## Breaking Change?

Yes, to obtain the previous values of PolygonComponent .vertices subtract the polygon's top-left point vector from all vectors in the list returned the new output.

## Related Issues

### DIFF
--- a/packages/flame/lib/src/geometry/shape_intersections.dart
+++ b/packages/flame/lib/src/geometry/shape_intersections.dart
@@ -80,9 +80,10 @@ class CirclePolygonIntersections
       intersectionPoints.addAll(circle.lineSegmentIntersections(line));
     }
     if (intersectionPoints.isEmpty && (circle.isSolid || polygon.isSolid)) {
-      final outerShape = circle.containsPoint(polygon.vertices.first)
-          ? circle
-          : (polygon.containsPoint(circle.position) ? polygon : null);
+      final outerShape =
+          circle.containsPoint(polygon.vertices.first - polygon.topLeft)
+              ? circle
+              : (polygon.containsPoint(circle.position) ? polygon : null);
       if (outerShape != null && outerShape.isSolid) {
         final innerShape = outerShape == circle ? polygon : circle;
         return {innerShape.absoluteCenter};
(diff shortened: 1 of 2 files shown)


spydon:
`containsLocalPoint` also has be modified

packages/flame/lib/src/geometry/shape_intersections.dart
```diff
@@ -80,9 +80,10 @@ class CirclePolygonIntersections
       intersectionPoints.addAll(circle.lineSegmentIntersections(line));
     }
     if (intersectionPoints.isEmpty && (circle.isSolid || polygon.isSolid)) {
-      final outerShape = circle.containsPoint(polygon.vertices.first)
-          ? circle
-          : (polygon.containsPoint(circle.position) ? polygon : null);
+      final outerShape =
+          circle.containsPoint(polygon.vertices.first - polygon.topLeft)
```
spydon, line 84: Yeah this is definitely not correct, how sloppy of me to not write a test for this case.

spydon (approved):
Let's merge this even though the solid circle - polygon intersection logic is wrong (since it was already wrong before). Did you want to write up a PR for that fix too? :)

packages/flame/lib/src/geometry/shape_intersections.dart
mattmyne: Yup I can put together a PR for the circle-polygon intersection (will need to get up to speed on how the others work first though). Since this PR only maintains previous behaviour I agree merge should be good!

mattmyne:
@ufrshubham I've just tried running the minimal code to reproduce provided in #1545 and it throws an exception when clicking: "Unsupported operation: Can only apply this effect to SizeProvider". So I don't know.

ufrshubham:
> @ufrshubham I've just tried running the minimal code to reproduce provided in #1545 and it throws an exception when clicking: "Unsupported operation: Can only apply this effect to SizeProvider". So I don't know.

Ohh, my bad! I misread the description and thought this PR was fixing that issue 😅

PR #2062, opened 2022-10-10
feat!: Add OpacityProvider

### LINKED ISSUE #2053: OpacityEffect should allow using generic `paintId`
# Problem to solve

- `OpacityEffect`s cannot be applied on components using HasPaint with custom `paintId`.
   If a component is using multiple paints, with a custom type for `paintId`, it is not possible to use opacity effect. This is because right now opacity effect supports only string type `paintId`.

- Single `OpacityEffect` cannot be applied to multiple paints easily.
  This is because `OpacityEffect` only accepts a single paintId.

# Proposal

- Add a generic type argument for OpacityEffect and use that type as `paintId`
- Convert single `paintId` into a list of id

# More information

I'll work on this.

ufrshubham:
# Description

This PR adds a new `OpacityProvider` interface which should be implemented by components that want to use `OpacityEffect`. This makes `OpacityEffect` decoupled from `HasPaint`.

## Checklist

## Breaking Change?

### Migration instructions

`OpacityEffect`'s `paintId` string input is replaced by `target` property.
E.g.: `paintId: 'bluePaint'` should now be replaced with `target: component.opacityProviderForPaint('bluePaint')`

## Related Issues

Closes #2053 


ufrshubham:
@st-pasha, this draft fixes the issue of using generic a single `paintId`. But for allowing `OpacityEffect` to operate on multiple `paints`, I think we will have to allow multiple targets for `OpacityEffect`. 

With single `target`, I am not sure what the `opacity` and `alpha` getters should return. It is quite possible that all paints do not have same opacity.

packages/flame/lib/src/effects/provider_interfaces.dart
```diff
@@ -50,3 +50,12 @@ abstract class SizeProvider {
...
   set size(Vector2 value);
 }
+
+/// Interface for a component that can be affected by opacity effects.
+abstract class OpacityProvider {
+  double get opacity;
+  set opacity(double value);
+
+  int get alpha;
+  set alpha(int value);
```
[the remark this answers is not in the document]
ufrshubham: They are slightly different representation of the same idea. `alpha` is an integer in the range [0-255] and opacity is a double in the range [0-1]. I added both of them just to avoid making changes to `OpacityEffect`'s calculations as they use alpha right now.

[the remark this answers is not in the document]
st-pasha: I've updated the logic of `OpacityEffect` to use opacities instead, in PR #2064.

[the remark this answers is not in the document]
ufrshubham: Cool! I'll update mine once that PR is merged.

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -97,4 +98,27 @@ mixin HasPaint<T extends Object> on Component {
   void tint(Color color, {T? paintId}) {
     getPaint(paintId).colorFilter = ColorFilter.mode(color, BlendMode.srcATop);
   }
+
+  @override
+  double get opacity => paint.color.opacity;
+
+  @override
+  set opacity(double value) => paint.color = paint.color.withOpacity(value);
```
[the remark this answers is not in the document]
ufrshubham: Done

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -97,4 +98,32 @@ mixin HasPaint<T extends Object> on Component {
...
+  _ProxyOpacityProvider(this.paintId, this.target);
+
+  final T paintId;
+  final HasPaint<T> target;
+
+  @override
+  double get opacity => target.getOpacity(paintId: paintId);
+
+  @override
+  set opacity(double value) => target.setOpacity(opacity, paintId: paintId);
```
st-pasha, line 128:
```suggestion
  set opacity(double value) => target.setOpacity(value, paintId: paintId);
```
I guess it means we don't cover this provider tests?

packages/flame/lib/src/effects/opacity_effect.dart
```diff
@@ -1,75 +1,75 @@
...
       controller,
-      paintId: paintId,
+      target: target,
       onComplete: onComplete,
     );
   }
 
   double _opacityOffset;
   double _roundingError = 0.0;
-  final String? paintId;
```
st-pasha, line 63: Looks like this `paintId` was buggy, so good riddance. Generally, `HasPaint` is supposed to have a `<T>` generic parameter, so that `paintId` would be of type `T?`. But the class was declared as `ComponentEffect<HasPaint>`, thus `T=Object`. But when `T=Object` the paint ids are not allowed (there is an assert for that).

Not sure how (and whether) it all worked before, but good thing we don't have to worry about that anymore.

spydon (approved):

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -97,4 +98,32 @@ mixin HasPaint<T extends Object> on Component {
...
+
+  @override
+  set opacity(double value) {
+    paint.color = paint.color.withOpacity(value);
+    for (final paint in _paints.values) {
+      paint.color = paint.color.withOpacity(value);
+    }
+  }
+
+  OpacityProvider opacityProviderForPaint(T paintId) {
```
st-pasha, line 113: Maybe call it `opacityOfPaint(paintId)` for short?

`OpacityEffect(target: component.opacityOfPaint('bg'))` sounds nicer.

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -97,4 +98,32 @@ mixin HasPaint<T extends Object> on Component {
...
+    return _ProxyOpacityProvider(paintId, this);
+  }
+}
```
st-pasha, line 116: How about adding a provider for all opacities at once? This seems to be a common case, where you have a multi-paint component, and want to fade it out as a whole.

packages/flame/lib/src/components/mixins/has_paint.dart
spydon: It sounds like you'd get back the `double` from `opacityOfPaint` though?

st-pasha: Well, you'd get something that has `.opacity` property; it's a bit like getting an `Opacity` object instead of a simple value. And there is definitely no confusion when used as a target of an `OpacityEffect`.

spydon: There is no confusion when used as a target, but the user will get this method suggested when doing other things on the component too. Since we have no opacity object I don't think any user would expect that.

ufrshubham: A short name would have been better, but just like @spydon said, `opacityOfPaint` will be a misnomer. How about `opacityProviderOf`? The input parameter already suggests that it operates on paint

packages/flame/lib/src/components/mixins/has_paint.dart
ufrshubham: > This seems to be a common case, where you have a multi-paint component, and want to fade it out as a whole.

This use case is already handled by `opacity` setter of `HasPaint`. Even if we add a provider for all paints, how do we handle `opacity` getter if all paints do not have the same `opacity`?

packages/flame/lib/src/effects/opacity_effect.dart
ufrshubham: Yeah, I don't think most users know that `HasPaint` can actually hold multiple paints. Even our `Flower` class from examples stores multiple paints within itself, even though it has `HasPaint`  mixin 😅

packages/flame/lib/src/components/mixins/has_paint.dart
ufrshubham: Oh!! Good catch! I've updated this and a test to check this. I think we can create an issue for adding unit tests for `HasPaint` mixin.

packages/flame/lib/src/components/mixins/has_paint.dart
st-pasha: I would say that if you have a class with 2 paints, one with opacity 90% and another at 25%, then `get`ting the overall opacity should return 90% (the most opaque component), and `set`ting it to, say 50%`, should make the first paint go to 50% and the second paint to (25/90*50)% -- that is, the ratio of the opacities should remain the same.

This means the opacity provider has to store these ratios at the time of construction, and then use them later on.

packages/flame/lib/src/effects/provider_interfaces.dart
```diff
@@ -50,3 +50,9 @@ abstract class SizeProvider {
   Vector2 get size;
   set size(Vector2 value);
 }
+
+/// Interface for a component that can be affected by opacity effects.
```
st-pasha, line 54: It would be good to document that setting the opacity to some `x` would not necessarily produce the same `x` when reading it back -- the implementation is allowed to store the opacity as an integer `alpha` which would incur some rounding errors.

packages/flame/lib/src/components/mixins/has_paint.dart
spydon: Isn't it overkill to have logic for that, the user can just handle the colors separately if they need a set ratio between the opacities?

It also gets a bit weird since the colors with less opacity never can get above what they started at if another color started on 1.0.

st-pasha: > Isn't it overkill to have logic for that in here?

Not an overkill, since the logic will be in a separate `OpacityProvider` class.

> It also gets a bit weird since the colors with less opacity never can get above what they started at if another color started on 1.0.

Why would that be a problem? If you draw something with a faint color on top of a solid color -- for example, a watermark on a page -- then you'd want it to keep roughly the same opacity ratio relative to the whole component. On the other hand, if you wanted to make the water mark more or less visible -- then the OpacityEffect can be applied to that individual paint.

spydon: > Why would that be a problem? If you draw something with a faint color on top of a solid color -- for example, a watermark on a page -- then you'd want it to keep roughly the same opacity ratio relative to the whole component. On the other hand, if you wanted to make the water mark more or less visible -- then the OpacityEffect can be applied to that individual paint.

Sure, but that is only one usecase, I'm sure there are others.

> then the OpacityEffect can be applied to that individual paint.

True, that would get a very strange behavior if it wasn't done in this manner, I'm convinced.

packages/flame/lib/src/effects/provider_interfaces.dart
ufrshubham: This sounds more like implementation details. No need for the interface to make any comments on that.

packages/flame/lib/src/components/mixins/has_paint.dart
ufrshubham: > for example, a watermark on a page -- then you'd want it to keep roughly the same opacity ratio relative to the whole component

So if such a component is completely faded out, both paint opacities will become zero. After this, if the component is faded in using a new `OpacityProvider`, it will not maintain the original ratios.

> `get`ting the overall opacity should return 90% (the most opaque component)

This approach doesn't really define an absolute true opacity. We are returning the opacity of most **opaque** paint and that can change anytime depending on how other opacity effects modify the paints. Even paints can be added and removed by the users anytime. I think this will involve a lot of book keeping to maintain the ratios.

packages/flame/lib/src/effects/provider_interfaces.dart
st-pasha: Indeed, it's an implementation detail. But it is a non-obvious detail. So the interface should tell the implementers that this is something which is allowed; and to the consumers that this is something that they could expect when using this API.

packages/flame/lib/src/components/mixins/has_paint.dart
st-pasha: Surely there could be some corner cases when the effect doesn't work quite right. And if someone encounters such case in real life, they will have to find a way to deal with that problem (and perhaps even submit a PR).

However, from a practical standpoint, the most common scenario is that when there is a component with several paints (such as shape with background and border, and perhaps some text), then it is useful to be able to fade out (or fade in) that component as one whole, which involves maintaining ratios between opacities of individual paints.

packages/flame/lib/src/effects/provider_interfaces.dart
ufrshubham: Done. Added the requested comments.

packages/flame/lib/src/components/mixins/has_paint.dart
ufrshubham: Added a `_MultiPaintOpacityProvider` which allows creation of a provider based on a list of paint ids

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -127,3 +148,51 @@ class _ProxyOpacityProvider<T extends Object> implements OpacityProvider {
...
+        : null;
+  }
+
+  final List<T>? paintIds;
+  final HasPaint<T> target;
+  late final List<double>? _opacityRatios;
+
+  @override
+  double get opacity {
+    var maxOpacity = 0.0;
```
st-pasha, line 172:
```suggestion
    var maxOpacity = target.getOpacity();
```

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -127,3 +148,51 @@ class _ProxyOpacityProvider<T extends Object> implements OpacityProvider {
...
+        ? List<double>.generate(
+            paintIds!.length,
+            (index) =>
+                target.getOpacity(paintId: paintIds!.elementAt(index)) /
+                maxOpacity,
+          )
+        : null;
+  }
+
+  final List<T>? paintIds;
```
st-pasha, line 166:
```suggestion
  final List<T?>? paintIds;
```
since `null` is also a valid paint id

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -110,9 +112,28 @@ mixin HasPaint<T extends Object> on Component implements OpacityProvider {
...
+  /// modifies other paints such that their respective opacity ratio with most
+  /// opaque paint is maintained.
+  ///
+  /// If [paintIds] is null or empty, all the paints are used for creating the
+  /// [OpacityProvider].
+  ///
+  /// Note: Each call results in a new [OpacityProvider] and hence the cached
+  /// opacity ratios are calculated using opacities when this method was called.
+  OpacityProvider opacityProviderOfList({List<T>? paintIds}) {
+    return _MultiPaintOpacityProvider(paintIds, this);
```
st-pasha, line 135: well `paintIds` is null, this should default to using all paints, as the description implies.

packages/flame/lib/src/components/mixins/has_paint.dart
ufrshubham: That results in incorrect output. `getOpacity()` returns the opacity of `paint` which is not part of the `_paints` map. So if users provide a list of `paintIds`, `maxOpacity` will first get initialised with the default value of 1 from `paint`s opacity. As a result the result of `max` inside the loop will always be 1.

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -97,4 +100,91 @@ mixin HasPaint<T extends Object> on Component {
...
+  final HasPaint<T> target;
+
+  @override
+  double get opacity => target.getOpacity(paintId: paintId);
+
+  @override
+  set opacity(double value) => target.setOpacity(value, paintId: paintId);
+}
+
+class _MultiPaintOpacityProvider<T extends Object> implements OpacityProvider {
```
spydon, line 155: Where/how is this used?

ufrshubham: `opacityProviderOfList()` uses this. Using that method, users can specify a list of paintIds on which the provider should operate.

ufrshubham:
> It would be good with a few lines in the docs describing how to use this with the effect. :)

Updated the docs and also modified the `Flower` class so that it can be used to demonstrate `OpacityProvider` in one of the live examples.

spydon (approved):

doc/flame/examples/lib/flower.dart
```diff
@@ -50,7 +53,10 @@ class Flower extends PositionComponent with TapCallbacks {
   @override
   void render(Canvas canvas) {
     for (var i = 0; i < _paths.length; i++) {
-      canvas.drawPath(_paths[i], _paints[i]);
+      canvas.drawPath(
+        _paths[i],
+        getPaint(FlowerPaint.values.elementAt(i)),
```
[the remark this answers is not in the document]
ufrshubham: > This will be a good usage of the layers in #2072 later!

Assuming this gets merged soon 😅

erickzanardo (approved):

doc/flame/effects.md
```diff
@@ -376,6 +378,9 @@ uses multiple paints, the effect can target any individual color using the `pain
 final effect = OpacityEffect.to(
   0.2,
   EffectController(duration: 0.75),
+  target: component.opacityProviderOfList(
```
erickzanardo, line 381: Should we mention and give an example of passing a target and not passing a target since this is optional?

spydon:
Great job as always @ufrshubham! :)

doc/flame/effects.md
ufrshubham: Added this:

OpacityEffectTargetDemo

doc/flame/examples/lib/flower.dart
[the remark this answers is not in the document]
mattmyne: I'm a bit concerned how well all this is going to play with paintLayers. It seems only the main paint and paintIds are taken into account (understandably given paintLayers isn't merged yet!) but hopefully this will be soon (maybe?). I haven't looked at this PR till now myself but - what do you think?

PR #2064, opened 2022-10-11
refactor: OpacityEffect now uses opacity instead of alpha internally

st-pasha:
# Description

Using `opacity` instead of integer `alpha` simplifies the code (and also reduces round-off errors which may accumulate for longer-running effects).

## Checklist

## Breaking Change?


spydon:
We had some problems with this solution, but I don't remember what it was, seems like all the rounding errors are handled fine here right? Because internally it will always be 0-255.
Some more tests might be good, if I remember correctly there is even a test for `OpacityEffect` that isn't running currently.

ufrshubham:
> We had some problems with this solution, but I don't remember what it was, seems like all the rounding errors are handled fine here right? Because internally it will always be 0-255.

Yeah, for me the "fade out" test was failing when I tried to make these changes. I think it was due to the rounding error which seems to be handled correctly in this PR.

packages/flame/test/effects/opacity_effect_test.dart
```diff
@@ -242,11 +242,7 @@ void main() {
         game.update(dt);
       }
       game.update(1000 - totalTime);
-      // TODO(spydon): The loop above has an average of 100fps.
-      // It should change from 0-255 in 1s so it will change alpha with an
-      // average of 255/100=2.5 per tick, which should not result in a need of
-      // an epsilon value this high.
-      expectDouble(component.getOpacity(), 1.0, epsilon: 100 * _epsilon);
+      expectDouble(component.getOpacity(), 1.0, epsilon: _epsilon);
```
spydon, line 245: Ah exactly, this was it.

st-pasha: It appears the test now works much more reliably after switching to opacities. At least the epsilon is much smaller.

spydon (approved):

erickzanardo (approved):

PR #2067, opened 2022-10-11
fix: solid circles and polygons intersection

mattmyne:
# Description

This PR fixes polygons inside a solid circle and circles inside a solid polygon not being correctly detected.
(This relates to Flame's collision detection, not Forge2D)

## Checklist

## Breaking Change?

## Related Issues


spydon:
Could you add a test that verifies that it works too?

mattmyne:
Yes working on that now. I figured the tests would be a separate PR rather than rolling it into this bug fix?

packages/flame/test/geometry/shape_intersections_test.dart
```diff
@@ -0,0 +1,302 @@
+import 'package:flame/collisions.dart';
+import 'package:flame/components.dart';
+import 'package:flame/geometry.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('CircleCircleIntersections', () {
```
spydon, line 7: There should be a file where these tests can be incorporated, where the rest of the intersection tests are. Since we have another concept of shapes currently too it gets a bit confusing if this file is named shape tests.

spydon (approved):
Looks good! Only nit is that the tests should be moved

PR #2070, opened 2022-10-12
fix: Add missing hitbox parameters

mattmyne:
# Description

This PR adds the `position` parameter to `PolygonHitbox` and `isSolid` parameters to all hitbox classes' `.relative` named constructors.

Also returns `PolygonComponent's` internal `topLeft` parameter to private `_topLeft` as no longer required outside of the class file.

## Checklist

## Breaking Change?

## Related Issues


mattmyne:
Adding the `position` parameter to `PolygonHitbox` also means the same vertices list used for (an optional) parent `PolygonComponent` can be reused in the `PolygonHitbox` constructor with `position: Vector2.zero()`. Previously `position` would be fixed at `null` and the hitbox would be offset by topLeft, with no way of correcting except to clone and modify the original vertices list.

packages/flame/lib/src/collisions/hitboxes/circle_hitbox.dart
```diff
@@ -28,8 +28,11 @@ class CircleHitbox extends CircleComponent with ShapeHitbox {
     required super.parentSize,
     super.angle,
     super.anchor,
+    bool isSolid = false,
```
spydon, line 31:
```suggestion
    super.isSolid,
```

packages/flame/lib/src/collisions/hitboxes/polygon_hitbox.dart
```diff
@@ -8,6 +8,7 @@ class PolygonHitbox extends PolygonComponent
     with ShapeHitbox, PolygonRayIntersection {
   PolygonHitbox(
     super.vertices, {
+    super.position,
```
spydon, line 11: Have you tried this? I don't think it will play along nicely with the vertices since the position is already set from the vertices.

packages/flame/lib/src/collisions/hitboxes/polygon_hitbox.dart
```diff
@@ -28,9 +29,12 @@ class PolygonHitbox extends PolygonComponent
     required super.parentSize,
     double super.angle = 0,
     super.anchor,
+    bool isSolid = false,
```
spydon, line 32:
```suggestion
    super.isSolid,
```

packages/flame/lib/src/collisions/hitboxes/rectangle_hitbox.dart
```diff
@@ -31,10 +31,13 @@ class RectangleHitbox extends RectangleComponent
     required super.parentSize,
     super.angle,
     super.anchor,
+    bool isSolid = false,
```
spydon, line 34:
```suggestion
    super.isSolid,
```

packages/flame/lib/src/collisions/hitboxes/polygon_hitbox.dart
mattmyne: Yes I have, and it works as expected. Leaving position unmodified results in the hitbox's vertices being offset by topLeft compared to any parent. Providing a position of zero aligns it to the parent (when using the same vertices as the parent), any offset shows a visual offset by that amount - as position already does with CircleHitbox.

packages/flame/lib/src/collisions/hitboxes/circle_hitbox.dart
mattmyne: That's better. I copied how it has been done already in polygon_hitbox but I'll modify that too to use super instead.

packages/flame/lib/src/collisions/hitboxes/polygon_hitbox.dart
spydon: Could you add a small test for this? :)

packages/flame/lib/src/collisions/hitboxes/circle_hitbox.dart
mattmyne: Actually dart doesn't let me use super for this. Perhaps because it's not a direct parameter of the parent?

spydon: Ah, it's from the mixin. But you don't have to set it in the body anyways, you can set it after `:` and before super.

mattmyne: I can't get that to work.

mattmyne: I think it's because isSolid isn't a field in the class itself, even though it can be accessed in the body through this. Can we keep it as it is then.

packages/flame/lib/src/collisions/hitboxes/polygon_hitbox.dart
mattmyne: Added (7f20899)

packages/flame/lib/src/collisions/hitboxes/circle_hitbox.dart
spydon: Alright, I thought it got the fields before evaluation those, let's keep it like it is then.

spydon (approved):

PR #2073, opened 2022-10-12
feat: Add paint layers to HasPaint and associated component renders

### LINKED ISSUE #2072: Allow second paint for ShapeComponents
# Problem to solve
To have both a fill and outline on a `ShapeComponent` (`PolygonComponent`, `RectangleComponent` or `CircleComponent`) a `ShapeComponent` has to be created with the fill, then an entirely new `ShapeComponent` has to be added to draw the outline on top. This is an unnecessary duplication of code and resources.

# Proposal
Add a new optional named parameter `paintSecond` to `ShapeComponent` and derivatives. In the shape's render function, it will check if `paintSecond` is not null and draw again with this second paint if so.

# More information

I have prepared a PR that adds this feature.

mattmyne:
# Description

Add optional named property `paintLayers` to `HasPaint` mixin. Adding `Paint` classes to this list provides a reference for a component's `render` function to draw with multiple paints on top of each other. The main `paint` property points to the first element in the `paintLayer` and remains backwards compatible. `ShapeComponent` and extensions now expose `paintLayers` in their constructors. `PolygonComponent`, `RectangleComponent` and `CircleComponent` `render` functions have been updated to draw with `paintLayers`.

This removes the need to add an entirely new `ShapeComponent` (and extensions) just to render with multiple paints, e.g. a fill and an outline.

Support functions for `paintLayers` have also been added, such as usage with `HasPaint` paint collections, now also available on shape and sprite library components.

## Checklist

## Breaking Change?

## Related Issues

Closes #2072 


spydon (commented):
I don't think this is worth adding directly in the `ShapeComponent`. Possibly as a mixin, but also `HasPaint` should be used so I'm not sure what the right approach is here, possibly a completely new component wrapping a shape component.

mattmyne:
@spydon My thinking here is if ShapeComponent has a paint parameter, then it's here that paintSecond should be added as it directly enhances paint. It's very little additional code and has no performance impact if kept at its default of null. ShapeComponent seems to be the clear place where having both a fill and outline is not only useful but many might consider expected?

kurtome:
I guess the alternative is to override `render` for your shape, which isn't terrible but is a bit unfortunate for something standard like a shape's border...

mattmyne:
Great, I do think a feature like this will be valuable and save users boilerplate code. I suggested `paintSecond` as the second paint might not always be used for the border (potentially with a shader, or fill might be wanted drawn above outline so it goes all the way to the polygon edge). I started with `paint` so autocomplete would show it as an option alongside just typing `paint` but I agree that's not a super-important reason. Thoughts?

I'll look to how to add this to the HasPaint mixin instead, though I thought HasPaint acted on the component's own paint variable rather than provide one itself?

ufrshubham:
> I'll look to how to add this to the HasPaint mixin instead, though I thought HasPaint acted on the component's own paint variable rather than provide one itself?

`HasPaint` is the one that actually stores a main `paint` and a map for additional `paints`. You can store multiple paints in that map using `setPaint()` and retrieve that paint using `getPaint()` while rendering.

spydon:
> Great, I do think a feature like this will be valuable and save users boilerplate code. I suggested `paintSecond` as the second paint might not always be used for the border (potentially with a shader, or fill might be wanted drawn above outline so it goes all the way to the polygon edge). I started with `paint` so autocomplete would show it as an option alongside just typing `paint` but I agree that's not a super-important reason. Thoughts?

Maybe `foregroundPaint`?

spydon:
> `foregroundPaint` works. Also maybe `topPaint` (might confuse with top/bottom of screen, though "top paint" is recognised in real painting!)

I think there is a larger amount of our users that are more familiar with programming terms than actual real painting. 😉 I for one have never heard about top paint, but I'm also extremely bad at painting. 😅

mattmyne:
`foregroundPaint` it is then 😅 (I should have said "top coat" rather than top paint actually).
Work has got busy again but I'll look to using HasPaint to implement this as soon as I can. I'm assuming as another paint with a dedicated reference in HasPaint's map, following @ufrshubham 's comment.

st-pasha:
Shapes simply need a list of paints instead of a single paint or two paints. Because there are clear applications for more than two.

mattmyne:
> Shapes simply need a list of paints instead of a single paint or two paints. Because there are clear applications for more than two.

That's a much more flexible idea. A paintLayers list could be provided by HasPaint that contains a list of the paint id's to render with, after the main paint.

The render function of HasPaint components (eventually PolygonComponent, CircleComponent etc.) will first render with paint as before. It should then iterate through the paintLayers list and draw again using the paint from HasPaint's getPaint(paintLayers[count]).

So to just add a second/foreground paint, you'd add it to HasPaint as before:
setPaint('myOutline', outlinePaint)
..then to render it, add its id to the new paintLayers list:
paintLayers.add('myOutline')

...this will cause it to be drawn in the second paintLayers render loop by the final render functions. For render functions developed by users there'll be no change - it'll just draw using main paint as before. But updated ones can iterate through paintLayers to implement this new functionality if needed.

mattmyne:
Maybe `paintLayers` list should be called `renderPaints` to better match HasPaint naming?

spydon:
Sure, sounds like a good idea!

mattmyne:
An issue with HasPaint. It seems you can't use its paint storage (setPaint, getPaint etc.) without assigning a type to the index parameter when your class uses the mixin. So rather than:

`abstract class ShapeComponent extends PositionComponent with HasPaint {`

you must use something like:

`abstract class ShapeComponent extends PositionComponent with HasPaint<String> {`

..which would allow you to use `String` types for the paint id/index.

That's fine, I can see why it was put in, allowing people to use custom enums to reference paints etc.

The trouble is, _all_ the components that use the `HasPaint` mixin in the library only use it in its generic `HasPaint` form, effectively disabling the use of paint indexes for anything built off `ShapeComponent` or `SpriteComponent` etc.

Am I missing something? Is there a way of setting the id type after the `HasPaint` mixin has already been declared in base class?

If not, either using `HasPaint` needs a rethink or `HasPaint` itself needs a redesign 😕

mattmyne:
OK, I think the best way to maintain backwards compatibility and keep `HasPaint's` typed index ability is to either:

1) Set a fixed type for paint indexes in all library classes that currently mixin `HasPaint`. Anyone using these classes won't be using `HasPaint` indexes yet anyway (since `HasPaint` won't let them) so **no breaking change**. People using their own drawing classes extending directly from `PositionComponent` can continue mixin `HasPaint<own_type>` as before.

2) Remove the `_assertGenerics()` check in `HasPaint` that throws an exception if you haven't cast an index type from the default `Object` type. No breaking changes either (since users won't have been able to do this before) and would allow library classes to use paint indexes without modification. However, those indexes wouldn't have a specific type. Is that really a problem?

For either case, these classes (and their extensions) will get the new ability to use indexed paints:

`SpriteAnimationComponent`
`SpriteAnimationGroupComponent`
`SpriteComponent`
`SpriteGroupComponent`
`ShapeComponent`

If we go with 1, suggestions as to the new standard paint index type? `String`? `int`?

mattmyne:
I went with option 2. This allows `ShapeComponent` to use any type for `paintId`. No breaking changes, `HasPaint` can still have typed `paintId` for users' own use of the mixin as before. Now `ShapeComponent` and extensions can benefit from `paintId` collections too.

But the real new feature here is `paintLayers` support in `HasPaint`. This has been fully implemented in `ShapeComponent` and all extensions. Both `render` functions for shapes (in `CircleComponent` and `PolygonComponent`) have had the minor change necessary to draw using paintLayers. Simply from:

```dart
canvas.drawCircle((size / 2).toOffset(), radius, paint);
```
to
```dart
for (final currentPaint in paintLayers) {
  canvas.drawCircle((size / 2).toOffset(), radius, currentPaint);
}
```

I've chosen to keep Sprite-related classes as they are for now since layers add little value over the current single `paint` and it's a bit less obvious how to incorporate the layer drawing alongside the image. They still have `layerPaints` available internally through `HasPaint`, but their constructors don't include `layerPaints` and their `render` functions still use `paint` rather than iterate through `layerPaints`.

Everything works as before with the standard `paint`. However, you can now do:

```dart
final fillPaint = BasicPalette.orange.paint();
final outlineThickPaint = Paint()
  ..color = BasicPalette.green.color
  ..style = PaintingStyle.stroke
  ..strokeWidth = 6;
final outlineThinPaint = Paint()
  ..color = BasicPalette.blue.color
  ..style = PaintingStyle.stroke
  ..strokeWidth = 2;

add(CircleComponent(
  radius: 30,
  paintLayers: [fillPaint, outlineThickPaint, outlineThinPaint], // new!
));
```
circle-layers

You can now use the paint collection features of `HasPaint` with `ShapeComponent` and paint layers too:

```dart
final circle = CircleComponent(radius: 30);

// add paints to collection
circle.setPaint('fill', fillPaint);
circle.setPaint('outline', outlineThinPaint);
circle.setPaint('another', anotherPaint);

// set paint layers using paintIds
circle.setPaintLayers(['fill', 'outline']);

add(circle);

// later, you can change the circle's outline paint with:
circle.setPaint('outline', outlineThickPaint);

// by default setPaint also updates any corresponding paint in paintLayers
// (to stop this pass updatePaintLayers = false)

circle.setOpacity(0.5, paintId: 'outline');
```

The previous `paint` acts the same as before but always references the first element in `paintLayers` (with default value if `paintLayers` is empty).

I've added tests for `paintLayers` and associated functions.
Everything is backwards compatible.

Shape drawing is so much easier and prettier 🤩

st-pasha:
I feel like we're encumbering the `HasPaint` mixin too much here, with the additional list of paints. Now any component that was using this mixin previously will also start carrying a list (even if it is empty, it still consumes memory).

IMHO, a better approach here would be to declare a separate mixin `HasPaints`, which would contain a list of paints. This would be much cleaner, and will give users a choice whether they want to go with a single-paint or multi-paint alternative.

spydon:
> I feel like we're encumbering the `HasPaint` mixin too much here, with the additional list of paints. Now any component that was using this mixin previously will also start carrying a list (even if it is empty, it still consumes memory).

Agree, it also carries two more `Paint` instances now.

> IMHO, a better approach here would be to declare a separate mixin `HasPaints`, which would contain a list of paints. This would be much cleaner, and will give users a choice whether they want to go with a single-paint or multi-paint alternative.

That sounds like a good alternative, maybe naming it `HasPaintLayers` to differentiate it more from `HasPaint` which already can carry multiple paints.

mattmyne:
The main aim of this PR is to add paint layers to shapes.

Are you both happy this new `HasPaintLayers` mixin be added to `ShapeComponent`?

My original intent was to add a paint layer just to `ShapeComponent`, but it was requested I do this through `HasPaint` instead.

I agreed to do this 5 days ago with several unanswered comments discussing it and only now after I've put three full nights of work in has it been decided not to use `HasPaint` afterall. I'd prefer to do no further work on this until we're quite clear on the basic implementation aim.

mattmyne:
If we're (understandably) concerned about even small memory usage in `HasPaint` can I also suggest `HasPaint` isn't appropriate as it stands.

The only part of `HasPaint` used by the whole built-in shape and sprite components is the `paint` property. None of them can use its paint collection features (as they use `HasPaint` untyped) yet are all carrying around an unused map to support this. All its functions only support these collection indexes making them redundant too.

If we really want to save memory then I suggest all library components using `HasPaint` drop it and just store a Paint parameter. That's all they're currently using.

`HasPaint` should be renamed `HasPaintCollections` to reflect what it's actually providing and left for those (few?) currently using its collection features on custom classes to rename.

No `HasPaintLayers` is needed. `ShapeComponent` can just store a `paintLayers` list and provide getter/setter for `paint` to keep backwards compatibility. Circle and Polygon renders will iterate layerPaint to implement.

Much simpler. Removing `HasPaint` from sprites and shapes will save a considerable amount of memory for scenes with many of these instances.

spydon:
> Are you both happy this new `HasPaintLayers` mixin be added to `ShapeComponent`?

I have requested that some more people from blue fire have a look at this too, but I think that it would be a good way forward.

> I agreed to do this 5 days ago with several unanswered comments discussing it and only now after I've put three full nights of work in has it been decided not to use `HasPaint` afterall. I'd prefer to do no further work on this until we're quite clear on the basic implementation aim.

Sorry for that, that's why it's usually good to discuss the intended solution in an issue first, but sometimes you also just realize things when you see it in code when reviewing.

> The only part of HasPaint used by any of the built-in shape and sprite components is the paint property. None of them can use its paint collection features (as they use HasPaint untyped) yet are all carrying around an unused map to support this.

The map should be null by default, and only initialized when the first paint is added. Honestly I thought it was already.
We should also remove the generics assertion so that one can still use the other paints without having to put a type on it.

mattmyne:
> > Are you both happy this new `HasPaintLayers` mixin be added to `ShapeComponent`?
> 
> I have requested that some more people from blue fire have a look at this too, but I think that it would be a good way forward.

I'll wait for more input before doing anything then.

> > I agreed to do this 5 days ago with several unanswered comments discussing it and only now after I've put three full nights of work in has it been decided not to use `HasPaint` afterall. I'd prefer to do no further work on this until we're quite clear on the basic implementation aim.
> 
> Sorry for that, that's why it's usually good to discuss the intended solution in an issue first, but sometimes you also just realize things when you see it in code when reviewing.

This PR has an issue (#2072) but your comments were to this PR not the issue. I clearly stated my intention to use `HasPaint` in the comments after the discussion before I started the work.

> > The only part of HasPaint used by any of the built-in shape and sprite components is the paint property. None of them can use its paint collection features (as they use HasPaint untyped) yet are all carrying around an unused map to support this.
> 
> The map should be null by default, and only initialized when the first paint is added. Honestly I thought it was already. We should also remove the generics assertion so that one can still use the other paints without having to put a type on it.

I don't know enough of the underlying memory management of dart to know if this allocates any memory:
```dart
final Map<T, Paint> _paints = {};
```
but if it doesn't then I can modify `HasPaint` to initialise `paintLayers` as `= []` instead of a default empty paint and handle this in function code instead. Would that not alleviate the memory concerns?

If it the paint collection functionality and paint layers functionality don't actually use more memory when unused compared to a single `Paint` property then perhaps it's simpler to just keep everything in `HasPaint` even though I agree with @st-pasha that it's a bit overloaded.

spydon:
> I don't know enough of the underlying memory management of dart to know if this allocates any memory:
> ```dart
> final Map<T, Paint> _paints = {};
> ```
> but I'm fairly sure it does (a map class is still created).

It does, that's why it should be null and handled in the getter and setter instead.

spydon:
I discussed this a bit with @erickzanardo and @wolfenrain and it is probably better to just have a new list of layers on HasPaint, but which is null and private by default, and handled by getters and setters. Then the `ColorEffect` and the `OpacityEffect` can also be used on those colors by including them like we do with the `layers` in #2062, but that's outside of the scope of this PR.

mattmyne:
Using null initialisation and getters/setters for `paintLayers` is fine. When you say keeping `paintLayers` private though do you just mean full access though a public getter/setter or a more restrictive read-only style (like the `_paint` map)?

If read-only that's what I originally tried but ends up needing a whole load of wrapper functions for add/remove/clear layers and a performance hit of having to create a new `UnmodifiableListView` every call to `render` to pass out the current layers. Mixin cannot create a `final UnmodifiableListView` as it can't use its own `_paintLayers` as a constructor for it, so has to be created new every get call (not great).

spydon:
> Using null initialisation and getters/setters for `paintLayers` is fine. When you say keeping `paintLayers` private though do you just mean full access though a public getter/setter or a more restrictive read-only style (like the `_paint` map)?

I think that we can just initialize it with an empty list on the getter if it is previously null since the user will most likely use that functionality if it tries to get the list, so no waste in carrying the list then. I don't think we need to be as restrictive as with the map, the getter and setter will just be there to ensure that the initial value can be null.

mattmyne:
OK, this should implement the "only create when used" for paint collection and paintLayers. Added more tests.

spydon:
> OK, this should implement the "only create when used" for paint collection and paintLayers. Added more tests.

Good job, will review later today!
Can you update the description of the PR since that is what will go into the commit message?

mattmyne:
Done, feel free to edit further if required.

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -3,19 +3,47 @@ import 'dart:ui';
...
 /// by the [paint] attribute and other paints can be manipulated/accessed
 /// using [getPaint], [setPaint] and [deletePaint] by a paintId of generic type
-/// [T], that can be omitted if the component only have one paint.
+/// [T], that can be omitted if the component only has one paint.
+/// [paintLayers] paints should be drawn in list order during the render. The
+/// main Paint is the first element.
 mixin HasPaint<T extends Object> on Component {
-  final Map<T, Paint> _paints = {};
+  Map<T, Paint>? _paintsCollection;
+  Map<T, Paint> get _paints => _paintsCollection ?? (_paintsCollection = {});
```
spydon, line 16:
```suggestion
  late Map<T, Paint>? _paints = {};
```

packages/flame/lib/src/geometry/shape_component.dart
```diff
@@ -16,8 +16,14 @@ abstract class ShapeComponent extends PositionComponent with HasPaint {
...
     Paint? paint,
+    List<Paint>? paintLayers,
   }) {
-    this.paint = paint ?? this.paint;
+    if (paint != null) {
+      this.paint = paint;
+    }
+    if (paintLayers != null) {
+      this.paintLayers = paintLayers;
+    }
```
spydon, line 26:
```suggestion
    this.paint = paint ?? this.paint;
    this.paintLayers = paintLayers ?? this.paintLayers;
```

spydon (approved):
Looks good! Just some minor nits

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -24,8 +23,142 @@ void main() {
...
+    test('paintLayers defaults to [paint]', () {
+      final comp = _MyComponent();
+
+      const color = Color(0xFFE5E5E5);
+      comp.paint = Paint()..color = color;
+
+      expect(
+        (comp.paintLayers.first == comp.paint) &&
+            (comp.paintLayers.length == 1),
+        true,
```
erickzanardo, line 35:
```suggestion
        isTrue,
```

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -24,8 +23,142 @@ void main() {
...
+    test('setting paintLayers.first sets paint', () {
+      final comp = _MyComponent();
+
+      const color = Color(0xFFE5E5E5);
+      comp.paintLayers.first = Paint()..color = color;
+
+      expect(
+        (comp.paintLayers.first == comp.paint) &&
+            (comp.paintLayers.length == 1),
+        true,
```
erickzanardo, line 48:
```suggestion
        isTrue,
```

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -24,8 +23,142 @@ void main() {
...
+    test('setting paintLayers.first when length > 1 sets paint', () {
+      final comp = _MyComponent();
+
+      const color = Color(0xFFE5E5E5);
+      const secondColor = Color(0xFF123456);
+      const thirdColor = Color(0xFFABABAB);
+      comp.paintLayers = [Paint()..color = color, Paint()..color = secondColor];
+      comp.paintLayers.first = Paint()..color = thirdColor;
+
+      expect(comp.paint.color, thirdColor);
```
erickzanardo, line 61:
```suggestion
      expect(comp.paint.color, equals(thirdColor));
```

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -24,8 +23,142 @@ void main() {
...
+    test('setting paint sets paintLayers.first when length > 1', () {
+      final comp = _MyComponent();
+
+      const color = Color(0xFFE5E5E5);
+      const secondColor = Color(0xFF123456);
+      const thirdColor = Color(0xFFABABAB);
+      comp.paintLayers = [Paint()..color = color, Paint()..color = secondColor];
+      comp.paint = Paint()..color = thirdColor;
+
+      expect(comp.paintLayers.first.color, thirdColor);
```
erickzanardo, line 73:
```suggestion
      expect(comp.paintLayers.first.color, equals(thirdColor));
```

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -24,8 +23,142 @@ void main() {
...
+      final secondPaint = Paint()..color = secondColor;
+      final thirdPaint = Paint()..color = thirdColor;
+
+      final circle = CircleComponent(
+        radius: 10,
+        paint: firstPaint,
+        paintLayers: [secondPaint, thirdPaint],
+      );
+
+      expect(circle.paint.color, secondColor);
```
erickzanardo, line 91:
```suggestion
      expect(circle.paint.color, equals(secondColor));
```

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -37,68 +170,189 @@ void main() {
...
+        const anotherColor = Color(0xFFABABAB);
+        comp.setPaint('test', Paint()..color = color, updatePaintLayers: false);
+        comp.setPaint(
+          'anotherTest',
+          Paint()..color = anotherColor,
+          updatePaintLayers: false,
         );
+        comp.addPaintLayer('test');
+
+        expect(comp.paintLayers[1].color, color);
```
erickzanardo, line 236:
```suggestion
        expect(comp.paintLayers[1].color, equals(color));
```

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -37,68 +170,189 @@ void main() {
...
-        comp.deletePaint(_MyComponentKeys.foreground);
+        final testPaint = Paint()..color = const Color(0xFFE5E5E5);
+        comp.setPaint('test', testPaint);
+        comp.setPaintLayers(['test']);
+
+        const newColor = Color(0xFF123456);
+        final newPaint = Paint()..color = newColor;
+        comp.setPaint('test', newPaint);
+
+        expect(comp.paintLayers[0].color, newColor);
```
erickzanardo, line 253:
```suggestion
        expect(comp.paintLayers[0].color, equals(newColor));
```

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -37,68 +170,189 @@ void main() {
...
+
+        const testColor = Color(0xFFE5E5E5);
+        final testPaint = Paint()..color = testColor;
+        comp.setPaint('test', testPaint);
+        comp.setPaintLayers(['test']);
+
+        final newPaint = Paint()..color = const Color(0xFF123456);
+        comp.setPaint('test', newPaint, updatePaintLayers: false);
+
+        expect(comp.paintLayers[0].color, testColor);
```
erickzanardo, line 270:
```suggestion
        expect(comp.paintLayers[0].color, equals(testColor));
```

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -37,68 +170,189 @@ void main() {
...
+      'updatePaintLayers is false',
+      () {
+        final comp = _MyComponent();
+
+        const color = Color(0xFFE5E5E5);
+        comp.setPaint('test', Paint()..color = color);
+        comp.addPaintLayer('test');
+        comp.deletePaint('test', updatePaintLayers: false);
+
+        expect(comp.paintLayers.length, 2);
```
erickzanardo, line 299:
```suggestion
        expect(comp.paintLayers.length, equals(2));
```

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -37,68 +170,189 @@ void main() {
...
+        const color = Color(0xFFE5E5E5);
+        const anotherColor = Color(0xFFABABAB);
+        const thirdColor = Color(0xFF123456);
+        comp.setPaint('test', Paint()..color = color);
+        comp.setPaint('anotherTest', Paint()..color = anotherColor);
+        comp.setPaint('thirdTest', Paint()..color = thirdColor);
+        comp.setPaintLayers(['test', 'anotherTest', 'thirdTest']);
+        comp.removePaintIdFromLayers('anotherTest');
+
+        expect(comp.paintLayers[1].color, thirdColor);
```
erickzanardo, line 317:
```suggestion
        expect(comp.paintLayers[1].color, equals(thirdColor));
```

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -37,68 +170,189 @@ void main() {
...
+        comp.setPaint('thirdTest', Paint()..color = thirdColor);
+
+        comp.setPaintLayers(['thirdTest', 'test']);
 
         expect(
-          () => comp.getPaint(_MyComponentKeys.foreground),
-          throwsArgumentError,
+          (comp.paintLayers[0].color == thirdColor) &&
+              (comp.paintLayers[1].color == color),
+          true,
```
erickzanardo, line 338:
```suggestion
          isTrue,
```

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -3,19 +3,47 @@ import 'dart:ui';
...
+
+  /// List of paints to use (in order) during render.
+  List<Paint> get paintLayers {
+    if ((_paintLayers == null) || _paintLayers!.isEmpty) {
+      return _paintLayers = [_paint];
+    }
+    return _paintLayers!;
+  }
+
+  set paintLayers(List<Paint> pl) {
```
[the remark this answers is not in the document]
mattmyne: Sorry I missed that rule! Happy to accept the change but isn't it missing the line below too where it's assigned?

packages/flame/lib/src/components/mixins/has_paint.dart
mattmyne: Thanks, I didn't know `late` could do that too.

packages/flame/lib/src/components/mixins/has_paint.dart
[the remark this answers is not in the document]
spydon: Yeah, so probably easier to just change it locally :sweat_smile:

packages/flame/lib/src/geometry/shape_component.dart
mattmyne: I changed this on the last commit as to me it reads that if paintLayers in the constructor is null then get this.paintLayers. If the getter is called in `HasPaint` then it allocates memory and creates the `List`. We're trying to avoid this if paintLayers are not used?

packages/flame/test/components/mixins/has_paint_test.dart
mattmyne: Thanks for these, I'm not yet very familiar with dart testing

packages/flame/lib/src/geometry/shape_component.dart
mattmyne: [left unchanged after agreed in discord this.paintLayers should not be read if constructor paintLayers null to prevent get in HasPaint allocating a List unnecessarily]

packages/flame/test/components/mixins/has_paint_test.dart
```diff
@@ -37,68 +170,189 @@ void main() {
     );
 
     test(
-      'getPaint throws exception when used on genericless component',
+      'setPaint sets a paint',
       () {
-        final comp = _MyComponent();
+        final comp = _MyComponentWithType();
```
spydon, line 175: Seems like it was like this before you started this PR, but all `comp` variables should be named `component`.

spydon (approved):
Lgtm! Good job!

packages/flame/test/components/mixins/has_paint_test.dart
mattmyne: It was before my PR but good point - I've added a commit that does this.

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -3,19 +3,46 @@ import 'dart:ui';
...
+      return _paint;
+    }
+    return paintLayers.first;
+  }
+
+  set paint(Paint newPaint) {
+    if ((_paintLayers == null) || _paintLayers!.isEmpty) {
+      _paint = newPaint;
+    }
+    paintLayers.first = newPaint;
```
erickzanardo, line 45: Ha, this is a new for me

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -3,19 +3,46 @@ import 'dart:ui';
...
+    return _paintLayers!;
+  }
+
+  set paintLayers(List<Paint> paintLayers) {
+    _paintLayers = paintLayers;
+  }
+
+  /// Main paint. The first paint in the [paintLayers] list.
+  Paint get paint {
+    if ((_paintLayers == null) || _paintLayers!.isEmpty) {
```
erickzanardo, line 35: This is where it gets confusing to me. Why do the layers have precedence over the paint map?

On the other hand, I also can't argue why would the map have over the layers.

I wonder if we shouldn't consider layers just a different thing from the paint layers, and let the components decide how they will use?

packages/flame/lib/src/components/mixins/has_paint.dart
mattmyne: What part?

erickzanardo: I am just expressing my surprise moment of realizing you can set the first element of the list by assing a value on the .frist hahah

packages/flame/lib/src/components/mixins/has_paint.dart
mattmyne: The paint map is independent of the main paint, and is only used for paint collections? As for the null and isEmpty checks, it's to prevent a get on paintLayers if it's not been set yet as there's no need to allocate the List for it if not used by the component itself, saving memory. Otherwise if it's not empty then paintLayers[0] holds the main paint. This is because when iterating through paintLayers during a component's `render` it avoids needing to duplicate drawing code specially for the special case of the first layer, paint.

I'm confused about the meaning of considering layers a different thing from paint layers? In HasPaint there is only one layer concept, paintLayers?

packages/flame/lib/src/components/mixins/has_paint.dart
mattmyne: Ah yes, I only found this out recently too. I think it looks neater than [0] (and maybe runs a little faster without index lookup?) so went with it!

packages/flame/lib/src/components/mixins/has_paint.dart
spydon: > This is where it gets confusing to me. Why do the layers have precedence over the paint map?

The paint map doesn't really have any natural first item? But I was thinking about that too, that maybe the basic paint shouldn't be included in the paint layers at all. So that there basically are three completely separate features in here; the normal paint, the paint map and the paint layers.

> I wonder if we shouldn't consider layers just a different thing from the paint layers, and let the components decide how they will use?

The components now decide how they will use the paint layers.
What do you mean by "consider layers just a different thing from the paint layers"?

mattmyne (commented):
Sorry, I didn't mean to start another code review! Just see my two individual replies.

packages/flame/lib/src/components/mixins/has_paint.dart
mattmyne: > > This is where it gets confusing to me. Why do the layers have precedence over the paint map?
> 
> The paint map doesn't really have any natural first item? But I was thinking about that too, that maybe the basic paint shouldn't be included in the paint layers at all. So that there basically are three completely separate features in here; the normal paint, the paint map and the paint layers.

The reason is to save duplication of drawing code in components that use paint layers in their render. Otherwise they'll have to do one set of drawing using `paint` then duplicate that drawing iterating through `paintLayers`. Making `paint` tied to the first element of `paintLayers` makes things easier downstream I think.

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: Definitely neater! I will use it from now on.

packages/flame/lib/src/components/mixins/has_paint.dart
erickzanardo: Indeed the paint map doesn't have a first because this is a map, which has no order. But HasPaint does have a concept of a main paint (the one that we store on the variable, which is set without an id.

I would argue that the main paint can be considered part of the paint layers. But should the first layer be considered the main paint? That is where I think I am confused.

I think that this would be less confusing if the `get paint` remained untouched, and if we agree that the main paint can be considered a layer, then it could be included on the paintLayers getter and the components that use layers would remain working the same.

mattmyne: Until `paintLayers` is used, `paint` is a completely separate property as before anyway. Once `paintLayers` are used, then I think it'll be more confusing if `paintLayers` didn't actually contain all the paint layers...

mattmyne: It seems what you're suggesting @erickzanardo is already what's happening? Otherwise I'm a bit confused what the issue with the implementation is? Is the concern whether the main paint should be drawn first or last? Draw first with subsequent paint layers being drawn on top seemed the natural order to me, but do others disagree?

packages/flame/lib/src/components/mixins/has_paint.dart
mattmyne: There's also a .last which I think is also pretty neat.

spydon:
Mattmyne didn't have to finish this up so I took it over.
I've separated the different functionalities in `HasPaint` so the `paint` is no longer related to the `paintLayers` and the `paintLayers` is also not connected to the map of paints.

@ufrshubham could you review if the opacity things make sense?

ufrshubham:
> @ufrshubham could you review if the opacity things make sense?

Sure. I'll take a look at this

spydon:
> It has become a bit complex, but LGTM. 
> 
> It feels like `HasPaint` is trying to do a lot of things that can probably be separated out into smaller mixins. Maybe sometime in future it will need to be broken into:
> - `HasPaint` : For holding a single paint
> - `HasPaintMap`: For holding a map of paint Ids and paints
> - `HasPaintLayers`: For holding an ordered list of paints
> 
> with each mixin implementing `OpacityProvider`

I agree, I'll write down in the V2 suggestions that we should clean up this mixin. Since the map can't be taken out now.

packages/flame/lib/src/geometry/circle_component.dart
```diff
@@ -30,8 +32,19 @@ class CircleComponent extends ShapeComponent implements SizeProvider {
...
     super.anchor,
     super.paint,
+    super.paintLayers,
   }) : super(size: Vector2.all(relation * min(parentSize.x, parentSize.y)));
 
+  @override
+  @mustCallSuper
+  Future<void> onLoad() async {
+    void updateCenterOffset() => _centerOffset = Offset(size.x / 2, size.y / 2);
+    size.addListener(updateCenterOffset);
```
erickzanardo, line 42: This should be a different PR so we could have change log entry specifically for it. Wdyt?

spydon: It is related to this PR though, it's just making it more efficient to render the paint layers (and it's all internal, so a separate changelog entry isn't really needed).

erickzanardo: won't this make the circle component be in fact resizable? Because iirc, none of the shape components are right now.

Or maybe I am confusing something...

spydon: It was already resizable before (all components having the `SizeProvider` mixin are proven to work with resizing), the only thing this does is storing the center point so that it can render from the center. But I see now that it doesn't take the anchor into consideration, so good that we started discussing here. :sweat_smile:

erickzanardo: Alright, so maybe I am confusing with some old bug. Gonna approve it!

erickzanardo (approved):

packages/flame/lib/src/geometry/circle_component.dart
spydon: It seems like the render function was already wrong before. I'll merge this and fix the other bug tomorrow in a separate PR, for a better changelog. ;)

PR #2080, opened 2022-10-14
feat: Adding posibility to execute future once component is removed.

### LINKED ISSUE #1956: Add Component.removed
# Problem to solve
We need a way to subscribe/listen for a component removal.

# Proposal
All Flame components already provide a `mounted` property that returns `Future` that is completed when the component is mounted on the tree.

We could follow this approach (both in concept and implementation) and implement an equivalent for a `removed` property.

idlehart:
# Description
This corresponds to mounted/isMounted lifecycle event but for the removed state.

## Checklist

- [] I have updated/added relevant examples in `examples` or `docs`.

## Breaking Change?

## Related Issues

Closes #1956


erickzanardo (approved):

packages/flame/lib/src/components/core/component.dart
```diff
@@ -206,6 +207,11 @@ class Component {
   void _setRemovingBit() => _state |= _removing;
   void _clearRemovingBit() => _state &= ~_removing;
 
+  /// Whether the component is removed.
```
spydon, line 210:
```suggestion
  /// Whether the component has been removed.
```
Since this will be false before it has been added to a parent.

packages/flame/lib/src/components/core/component.dart
```diff
@@ -816,12 +832,19 @@ class Component {
         component._clearMountedBit();
         component._clearRemovingBit();
         component._parent = null;
+
```
spydon, line 835:
```suggestion
```

spydon (approved):

packages/flame/test/components/component_test.dart
```diff
@@ -28,6 +28,91 @@ void main() {
         await expectLater(component.mounted, completes);
       });
 
+      testWithFlameGame(
+          'component removed times out when set before game is ready',
```
st-pasha, line 32: This shouldn't be the expected behavior though, right? If such future is awaited, the game will enter into an infinite loop...

packages/flame/test/components/component_test.dart
```diff
@@ -28,6 +28,91 @@ void main() {
...
+        final removed = child.removed;
+
+        parent.removeFromParent();
+        game.update(0);
+        await expectLater(removed, completes);
+        expect(child.isRemoved, true);
+      });
+
+      testWithFlameGame(
+        'component removed times out when changing parent from a null parent',
```
st-pasha, line 95: Not sure this is an accurate description.
Changing parent from null to A is the same as adding the component to A. So, after that the component will end up mounted -- which means the `removed` future is still pending, and will complete once the component will get removed.

packages/flame/test/components/component_test.dart
spydon: What do you suggest should be returned? Returning a completed feature is also wrong, since the component haven't been removed yet. Why will it enter an infinite loop?

This test seems wrong though, because here the component is actually removed and then it still times out?
And the description of the test should be changed, once we figure out what the intended behavior should be.

st-pasha: I would expect the future to complete upon the component's removal even if it was created before the game was mounted.

If for some reason that's impossible, then it would be better to throw an `assert` error at the time of the future creation, if it can be detected that this future would not operate properly.

spydon: Agree, I think the test is just wrong. I wonder why it times out though.

packages/flame/test/components/component_test.dart
idlehart: Ah right, this test is useless. I will remove it.

packages/flame/test/components/component_test.dart
idlehart: @spydon i found why the `expectLater(removed, completes)` just times out. When trying to run `_finishRemoving` that I added to the `propagateToChildren` call that I added in `_remove()`. The `_lifecycleManager` is null so it never arrives at the completer for `removed` on the component. 

However, not sure why the `_lifecycleManager` is null in this stage.

idlehart: I pushed a change that should keep the `_lifecyclemanager` alive until its `_removedCompleter` has run `.complete()` and been set to null. Let me know what you think.

packages/flame/lib/src/components/core/component.dart
```diff
@@ -727,7 +740,8 @@ class Component {
   void processPendingLifecycleEvents() {
     if (_lifecycleManager != null) {
       _lifecycleManager!.processQueues();
-      if (!_lifecycleManager!.hasPendingEvents) {
+      if (!_lifecycleManager!.hasPendingEvents &&
+          _lifecycleManager!._removedCompleter == null) {
```
spydon, line 744: This check can't be in `hasPendingEvents` since the component will never mount if that is the case right?

idlehart: I suppose you can do something like this: 

``` 
 bool get hasPendingEvents {
    return (_children.isNotEmpty ||
            _removals.isNotEmpty ||
            _adoption.isNotEmpty ||
            _mountCompleter != null ||
            _loadCompleter != null) &&
        _removedCompleter == null;
  }
```

idlehart: Or no that wouldn't work for other things utilizing `hasPendingEvents` I think.

idlehart:
Not sure why the `flex_game` is failing the pipline at the moment. Is there a known issue?

spydon:
> Not sure why the `trex_game` is failing the pipline at the moment. Is there a known issue?

It failed to get some dependencies or something for it, it became green when I re-ran it.

packages/flame/test/components/component_test.dart
```diff
@@ -28,6 +28,65 @@ void main() {
         await expectLater(component.mounted, completes);
       });
 
+      testWithFlameGame(
+          'component removed times out when set before game is ready',
```
st-pasha, line 32:
```suggestion
          'component.removed completes even if obtained before the game was ready',
```

packages/flame/lib/src/components/core/component.dart
```diff
@@ -963,6 +997,9 @@ class _LifecycleManager {
   /// Queue for moving components from another parent to this one.
   final Queue<Component> _adoption = Queue();
 
+  /// Wheter or not there are any pending lifecycle events for this component.
```
st-pasha, line 1000:
```suggestion
  /// Whether or not there are any pending lifecycle events for this component.
```

packages/flame/lib/src/components/core/component.dart
```diff
@@ -206,6 +207,11 @@ class Component {
   void _setRemovingBit() => _state |= _removing;
   void _clearRemovingBit() => _state &= ~_removing;
 
+  /// Whether the component has been removed.
```
st-pasha, line 210:
```suggestion
  /// Whether the component has been removed. Originally this flag is `false`,
  /// but it becomes `true` after the component was mounted and then removed
  /// from its parent. The flag becomes `false` again when the component is
  /// mounted to a new parent.
```

packages/flame/lib/src/components/core/component.dart
```diff
@@ -219,6 +225,13 @@ class Component {
   Future<void> get mounted =>
       isMounted ? Future.value() : lifecycle.mountFuture;
 
+  /// A future that completes when this component has been removed.
```
st-pasha, line 228:
```suggestion
  /// A future that completes when this component is removed from its parent.
```

st-pasha (approved):

PR #2103, opened 2022-10-21
feat: Parser for jenny

st-pasha:
# Description

This PR adds ability to parse Yarn files and create a `YarnProject`.
Not all syntaxes are covered yet, but a significant part is, and this should be enough to get started on creating a `DialogueRunner`.

This PR also renames the package into `jenny` / `flame_jenny`, as discussed in #2126.

## Checklist

## Breaking Change?

## Related Issues


packages/flame_yarn/lib/src/parse/parse.dart
```diff
@@ -0,0 +1,788 @@
...
+    final lhs = parsePrimary();
+    return _parseExpressionImpl(lhs, 0);
+  }
+
+  /// This is an implementation function for the Operator-Precedence parsing
+  /// algorithm. It consumes expressions of the form `LHS op RHS op...`, where
+  /// the precedence of operators must be greater or equal to [minPrecedence].
+  /// The initial [lhs] sub-expression is provided, and the parsing position
+  /// should be at the start of the next operator.
+  Expression _parseExpressionImpl(Expression lhs, int minPrecedence) {
```
spydon, line 481: `lhs` and `rhs` should preferably have better names.

spydon: Maybe left `left` and `right` would be enough?

spydon (approved):
Very cool, it's like its own little programming language

packages/flame_yarn/lib/src/parse/parse.dart
st-pasha: It's a kinda of a standard terminology though when it comes to expressions. In fact, these terms even have their own wikipedia page: https://en.wikipedia.org/wiki/Sides_of_an_equation :)

spydon: For sure, but is it really used that much in parser terminology when it is actually written code?
I can't remember us using that terminology when coding parsers in any uni CS courses, but that was quite some time ago so I might be wrong.

spydon: And `lhs` in the context here, wouldn't it make more sense to be named just `expression`?
This discussion is quite some bike-shedding though and as you've seen I've already approved the PR.

st-pasha: They did use this terminology [here](https://en.wikipedia.org/wiki/Operator-precedence_parser#Pseudocode) where I borrowed the algorithm from, so it seemed natural to me to use the same.

spydon:
When having a look at this again, it seems like most of this code would be awesome as a separate package, since almost nothing so far seem specific to Flame, and then have the bridge package using that package (same structure as radiance will have once we get flame_radiance). Any reason that you put everything in here?

st-pasha:
The functionality works in stages: first we load the data from `.yarn` files and create an internal representation; then we use that representation in order to run the dialogue in a Flame game. The second part I'm still working on atm.

packages/flame_jenny/README.md
```diff
@@ -0,0 +1,6 @@
+# flame_jenny
+
+Bridge package for `jenny`.
```
spydon, line 3: Maybe add a sentence here explaining what yarn spinner is too, since this is what will show up on pub.dev

packages/flame_jenny/jenny/README.md
```diff
@@ -1,3 +1,3 @@
-# Throstle
+# Jenny
 
 [Yarn Spinner](https://docs.yarnspinner.dev/) support within Dart ecosystem.
```
spydon, line 3: Same here

spydon (approved):

packages/flame_jenny/README.md
st-pasha: I have an expanded version of README in PR #2113 (which is in the future from this one).

spydon (approved):

PR #2129, opened 2022-10-28
feat: Added glow effect using maskFilter

### LINKED ISSUE #1027: Glow effect
# Problem to solve
A new paint effect for glowing effects

# Proposal
As saw in the last FlameCon on the Orbit game talk, We could use Paint maskFilter to create a glow effect

https://api.flutter.dev/flutter/dart-ui/Paint/maskFilter.html

dipakp2726:
# Description
A new paint effect for glowing effects using maskFilter.

## Checklist

## Breaking Change?

## Related Issues

Closes #1027


munsterlander:
How do I change the size and color of the glow effect? Glow is supposed to be like a neon light, right?

dipakp2726:
I think I have misunderstood something,
I have used maskFilter but it's not glowing,🤔

dipakp2726:
https://user-images.githubusercontent.com/36990991/198687564-71897c60-a4f3-452d-9824-64d10ebfdb41.mov

munsterlander:
Awesome!

packages/flame/lib/src/effects/glow_effect.dart
```diff
@@ -0,0 +1,25 @@
...
+
+import 'package:flame/components.dart';
+import 'package:flame/effects.dart';
+
+class GlowEffect extends ComponentEffect<HasPaint> {
+  GlowEffect(this.style, this.blurValue, super.controller, {this.paintId});
+
+  final BlurStyle style;
+  final double blurValue;
+  final String? paintId;
```
st-pasha, line 11: This variable can probably be removed.
1. It's not used anyways;
2. Strictly speaking it has to be arbitrary type `T`;
3. A better approach would be to pass this parameter via the `PaintProvider` interface (which we don't have yet), similarly to how `OpacityEffect` works.

packages/flame/lib/src/effects/glow_effect.dart
```diff
@@ -0,0 +1,25 @@
+import 'dart:ui';
+
+import 'package:flame/components.dart';
+import 'package:flame/effects.dart';
+
+class GlowEffect extends ComponentEffect<HasPaint> {
+  GlowEffect(this.style, this.blurValue, super.controller, {this.paintId});
+
+  final BlurStyle style;
```
st-pasha, line 9: Most of the time this would be `outer`, so why not make this parameter optional with the default value of `BlurStyle.outer`?

packages/flame/lib/src/effects/glow_effect.dart
```diff
@@ -0,0 +1,25 @@
...
+  final double blurValue;
```
st-pasha, line 10: For a user, it would be kinda confusing to have a `GlowEffect` with a `blurValue`. Why not call it `glowStrength` or something like that?

st-pasha (commented):
Very nice addition!

But it seems like this needs to have a well-considered documentation. For example, the effect applies to shapes but not sprites. It makes a certain **paint** glow -- so the component has to be drawing something with a paint. We need to investigate how it would apply to "stroke" paints, or to transparent paints (say, can I draw a transparent outline of a sprite and make that outline glow?)

packages/flame/lib/src/effects/glow_effect.dart
dipakp2726: renamed to `strength`

packages/flame/lib/src/effects/glow_effect.dart
dipakp2726: added `PaintProvider` interface and passes using it.

dipakp2726:
> We need to investigate how it would apply to "stroke" paints


example on stroke paints
https://user-images.githubusercontent.com/36990991/198811437-c5c55671-285f-43d8-aa92-a02c7ab081d4.mov

dipakp2726:
> say, can I draw a transparent outline of a sprite and make that outline glow?)

example on outline.
ps: it not exact outline.

https://user-images.githubusercontent.com/36990991/198812733-bc7e520a-6f49-4df9-a66d-d27a3a49c039.mov

examples/lib/stories/effects/glow_effect_example.dart
```diff
@@ -0,0 +1,58 @@
...
+  ''';
+
+  @override
+  Future<void> onLoad() async {
+    final paint2 = Paint()
+      ..color = const Color(0xff39FF14)
+      ..style = PaintingStyle.stroke;
+
+    add(
+      Ember(
```
spydon, line 24: I think you can skip the ember here since the GlowEffect doesn't seem to work very well on it.

examples/lib/stories/effects/glow_effect_example.dart
```diff
@@ -0,0 +1,58 @@
...
+}
+
+class GlowEffectExample extends FlameGame with TapDetector {
+  static const String description = '''
+    In this example we show how the `GlowEffect` can be used.
+  ''';
+
+  @override
+  Future<void> onLoad() async {
+    final paint2 = Paint()
```
spydon, line 19:
```suggestion
    final paint = Paint()
```

examples/lib/stories/effects/glow_effect_example.dart
```diff
@@ -0,0 +1,39 @@
...
+      CircleComponent(
+        radius: 50,
+        position: Vector2(300, 400),
+        paint: paint,
+      )..add(
+          GlowEffect(
+            10.0,
+            EffectController(
+              duration: 3,
+              // reverseDuration: 1.5,
```
spydon, line 32: This should be removed, or uncommented

packages/flame/lib/src/components/mixins/has_paint.dart
```diff
@@ -14,9 +15,10 @@ import 'package:meta/meta.dart';
 /// [T], that can be omitted if the component only has one paint.
 /// [paintLayers] paints should be drawn in list order during the render. The
 /// main Paint is the first element.
-mixin HasPaint<T extends Object> on Component implements OpacityProvider {
+mixin HasPaint<T extends Object> on Component
+    implements OpacityProvider, PaintProvider {
   late final Map<T, Paint> _paints = {};
-  Paint paint = BasicPalette.white.paint();
+  Paint _paint = BasicPalette.white.paint();
```
spydon, line 21: Just override the paint here instead of making it private.

spydon:
There should also be docs for it, and an interactive example, but other than that it looks good!

dipakp2726:
https://user-images.githubusercontent.com/36990991/198830431-a855a8c4-23de-4314-b6d5-30f85d60101f.mov

doc/flame/effects.md
```diff
@@ -111,6 +111,7 @@ offset is relative to the current position of the target:
 ```
 
 ```dart
+
```
spydon, line 114: Why are you adding these lines in this PR?

dipakp2726: i think its added by ide format.

spydon: Your IDE seems to have some trouble with formatting, these lines should not be added. The other formatting fixes look fine though.

spydon: Ah you fixed it, good!

doc/flame/effects.md
```diff
@@ -431,6 +431,31 @@ Currently this effect can only be applied to components that have a `HasPaint` m
...
+```{flutter-app}
+:sources: ../flame/examples
+:page: glow_effect
+:show: widget code infobox
+:width: 180
+:height: 160
+```
+
+```dart
+
```
spydon, line 449:
```suggestion
```

doc/flame/examples/lib/glow_effect.dart
```diff
@@ -0,0 +1,29 @@
...
+
+class GlowEffectExample extends FlameGame {
+  @override
+  Future<void> onLoad() async {
+    final paint = Paint()..color = const Color(0xff39FF14);
+
+    add(
+      CircleComponent(
+        radius: canvasSize.y / 4,
+        position: canvasSize / 2,
```
spydon, line 15:
```suggestion
        radius: size.y / 4,
        position: size / 2,
```

packages/flame/lib/src/effects/glow_effect.dart
```diff
@@ -0,0 +1,30 @@
...
+
+import 'package:flame/effects.dart';
+import 'package:flame/src/effects/provider_interfaces.dart';
+
+/// Change the MaskFilter on Paint of a component over time.
+///
+/// This effect applies incremental changes to the MaskFilter on Paint of a
+/// component and
+/// requires that any other effect or update logic applied to the same component
+/// also used incremental updates.
```
spydon, line 11: These lines should be formatted so that they fill up to 80 chars and then line break, not break in the middle of a line.

spydon (approved):

st-pasha (approved):

st-pasha:
Oh lol -- I just realized that this filter **replaces** the color with a glow. So if you had a shape with a solid color before, that color will now disappear and there will only be glow.

Perhaps we could mark this effect as `@experimental` for now, and then figure out whether we need a different approach, or just document this behavior carefully?

dipakp2726:
if we make `BlurStyle.solid` then both color and glow will stay.

st-pasha:
Yeah, but it would hardly look like a glow then -- more like a blurred-out shape.
For a "real" glow I think we need a second paint.

spydon:
> Oh lol -- I just realized that this filter **replaces** the color with a glow. So if you had a shape with a solid color before, that color will now disappear and there will only be glow.
> 
> Perhaps we could mark this effect as `@experimental` for now, and then figure out whether we need a different approach, or just document this behavior carefully?

We should make the effect target a color then right? So that the component can have several colors defined.
It doesn't really fit that well in with how the `PaintProvider` is created in this PR.
Any ideas of how to improve on this?

st-pasha:
I guess the first thing we have to acknowledge is that "glow effect" is not an `Effect`. We use the term "Effect" to describe a change in some property over time. At the same time, a glow "effect" is a static change in appearance of the target.
A more appropriate term here would probably be `Decoration`

spydon:
Maybe it could be modified to be pulsating or such? But yeah, maybe this shouldn't have been an effect at all.

PR #2161, opened 2022-11-10
feat: Add `maintainState` property to Route

### LINKED ISSUE #2160: Add property to Route to allow the page to rebuild when the route is revisited
When a Route is popped from the stack, its state is maintained by the Router.  Therefore, if the Route is pushed back onto the stack, the state of the Route's child Component is exactly where it was when the Route was popped.  There are some use cases where it would be beneficial for the Route's build method to be called each time the Route is pushed on the stack, thereby, rebuilding the Route's child Component.

As discussed on Discord (10/31/2022), the proposal is to add a boolean property to the Route component named pageTransient.  When true (by default), the current behavior remains where the state of the child Component is retained.  If pageTransient is set to false, the child Component of the Route will be rebuilt when the Route is pushed onto the stack again.

Please assign issue to me and I can submit the PR for this.

Thanks!!!!!

pastordougdev:
# Description

Add a boolean property to the Route component named pageTransient. When true (by default), the current behavior remains where the state of the child Component is retained. If pageTransient is set to false, the child Component of the Route will be rebuilt when the Route is pushed onto the stack again.

## Checklist

- [ x] The title of my PR starts with a [Conventional Commit] prefix (`fix:`, `feat:`, `docs:` etc).
- [ x] I have followed the [Contributor Guide] when preparing my PR.
- [x ] I have updated/added tests for ALL new/updated/fixed functionality.
- [ x] I have updated/added relevant documentation in `docs` and added dartdoc comments with `///`.
- [ x] I have updated/added relevant examples in `examples` or `docs`.

## Breaking Change?

- [ x] No, this PR is not a breaking change.

## Related Issues

Closes #2160 


pastordougdev:
I updated the doc page based on the output of the markdown-lint step.  I'm not sure what error is causing the failure in the analyze step??

pastordougdev:
Ok, found the analyzer issue.

packages/flame/lib/src/components/route.dart
```diff
@@ -36,6 +37,13 @@ class Route extends PositionComponent with ParentIsA<RouterComponent> {
   /// route underneath doesn't need to be rendered.
   final bool transparent;
 
+  /// If false, the route will not maintain the state of this route's page
+  /// component.  By default, once a route becomes active, the component
+  /// built by the build routine is maintained by the route after the route
+  /// is popped off the stack.  Setting [pageTransient] to false will drop the
+  /// page component when the route is popped off the stack.
+  final bool pageTransient;
```
st-pasha, line 45: I'm reading this description, and just can't wrap my head around it.

"Transient" means temporary, short-lived, impermanent. So, if the page is transient (`pageTransient` is true), that has to mean that the page will be short-lived, that it will be deleted once the route is popped.

However, this is not the default behavior. The default behavior of a route is that a page is persistent: once built, it will remain attached to the route and will not need to be recreated in the future. Thus, the default behavior is that a page is persistent, that is `pageTransient = false`.

It appears then that the property is currently doing the opposite of what it's supposed to be doing.

pastordougdev: You're right - I don't know how I flipped that in my head.  Let me redo this to make more sense and fix the wording.

spydon: Maybe it could be flipped around, and be called `pagePermanent` or something like that instead? I think that would also be clearer to the users what it does

spydon: @pastordougdev what do you think about flipping it around?

spydon:
@pastordougdev we're doing a release today, any chance you'd have the time to flip around the variables like mentioned in this comment? https://github.com/flame-engine/flame/pull/2161#discussion_r1023136254

spydon (approved):

st-pasha:
I think `maintainState` is a bit problematic, because in the future the Route may contain more state than just a page, and then the property will become a misnomer.

How about `keepPage`?

erickzanardo:
But won't maintainState fit for whatever state it has?

I like this name as it is in sync with Flutter's parameter name

erickzanardo (approved):

spydon:
> I think `maintainState` is a bit problematic, because in the future the Route may contain more state than just a page, and then the property will become a misnomer.
> 
> How about `keepPage`?

Wouldn't one want to keep any potential future state in Route together with the page if one has this variable set?

pastordougdev:
Hi all!  Sorry I got behind on this - got busy with some family and work things.  I like the logic flipped around like this.  I think feature will be very useful.

PR #2174, opened 2022-11-20
fix!: Correct coordinate system for a circular viewport

st-pasha:
# Description

- The coordinate system within a `CircularViewport` component was inconsistent: it assumed the center of the viewport has coordinates (0, 0), whereas that point should be at the top-left corner of the bounding box.
- Allow having an elliptic viewport.
- Added tests for `CircularViewport` (previously there weren't any).
- Added debug-mode rendering for the circular viewport.
- All camera-related files moved into the `camera/` directory.

## Checklist

### Migration instructions

- If you're using a `CircularViewport` and attach HUD elements to this viewport, they will now have to be shifted by `Vector2.all(radius)`.

## Related Issues


packages/flame/lib/camera.dart
```diff
@@ -0,0 +1,12 @@
+export 'src/camera/behaviors/bounded_position_behavior.dart'
```
spydon, line 1: Hmm, by creating this barrel file it is no longer clear to the user that the camera component is in the experimental stage, seems like a change outside of the scope of this PR?

spydon (approved):

packages/flame/lib/camera.dart
st-pasha: Ok, I've removed the export file.

PR #2176, opened 2022-11-21
feat: Added CameraComponent.withFixedResolution() constructor

st-pasha:
# Description

This PR adds new constructor for `CameraComponent` which simulates the old-style `FixedResolutionViewport`.

## Checklist

## Breaking Change?


spydon (commented):
Why have a constructor for this instead of a real viewport class? Feels inconsistent that some viewports would be set as a `CameraComponent` constructor and that some set the viewport field.

packages/flame/lib/src/camera/camera_component.dart
```diff
@@ -45,6 +46,29 @@ class CameraComponent extends Component {
...
+    required World world,
+    required double width,
+    required double height,
+    List<Component>? hudComponents,
+  }) {
+    return CameraComponent(
+      world: world,
+      viewport: FixedAspectRatioViewport(aspectRatio: width / height)
+        ..addAll(hudComponents ?? []),
+      viewfinder: Viewfinder()..visibleGameSize = Vector2(width, height),
```
spydon, line 68: I understand that `visibleGameSize` is the reason for this being a constructor instead of a class, but doesn't this field belong in the viewport class?

st-pasha: Not really, because it controls the zoom level, which is part of the Viewfinder.

Think about it this way: Viewport is the window through which we look at the game World. The properties of a viewport are: its shape, size, and position on the screen. Once Viewport has a fixed position on the screen, everything that happens inside it is controlled by the Viewfinder: which game point to "look at", the angle of rotation, and the zoom level.

spydon: But you agree that this feels inconsistent? Is there any other way we can design this so that the `FixedResolutionViewport` can be defined as a viewport?

spydon: For a normal physical camera maybe, but for pretty much every other game engine that I know of you set the fixed resolution at the viewport.

spydon: I think at least? I'm not well versed in other game engines.

spydon: Maybe not, seems to be very different how this is done in different engines.

packages/flame/lib/src/camera/camera_component.dart
```diff
@@ -45,6 +46,29 @@ class CameraComponent extends Component {
...
+  /// possible on all sides while maintaining the [width]:[height] aspect ratio.
+  /// The zoom level will be set such in such a way that exactly [width] x
+  /// [height] pixels are visible within the viewport. The viewfinder will be
+  /// initially set up to show world coordinates (0, 0) at the center of the
+  /// viewport.
+  factory CameraComponent.withFixedResolution({
+    required World world,
+    required double width,
+    required double height,
+    List<Component>? hudComponents,
```
spydon, line 62: This should probably be added to the normal constructor too if we have it on this constructor.

spydon (approved):

erickzanardo (approved):

packages/flame/lib/src/camera/camera_component.dart
```diff
@@ -42,9 +43,33 @@ class CameraComponent extends Component {
...
+  /// Create a camera that shows a portion of the game world of fixed size
+  /// [width] x [height].
+  ///
+  /// The viewport will be centered within the canvas, expanding as much as
+  /// possible on all sides while maintaining the [width]:[height] aspect ratio.
+  /// The zoom level will be set such in such a way that exactly [width] x
+  /// [height] pixels are visible within the viewport. The viewfinder will be
+  /// initially set up to show world coordinates (0, 0) at the center of the
+  /// viewport.
+  factory CameraComponent.withFixedResolution({
```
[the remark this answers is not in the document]
spydon: For the docs site that is, the dartdocs are good as they are!

spydon:
@st-pasha do you think you'd be able to add a line in the docs about this method before we do the release today, so that we can include this one? :)

packages/flame/lib/src/camera/camera_component.dart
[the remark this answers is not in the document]
st-pasha: done

PR #2185, opened 2022-11-24
feat: Add command <<local>>

st-pasha:
# Description

The `<<local>>` command can be used to introduce local variables -- those whose scope is limited to the node in which the command was used. The syntax of the command is similar to that  of `<<declare>>`.

## Checklist

## Breaking Change?

## Related Issues


spydon:
You don't have to put `jenny` in the scope, Melos already knows that since it is changes within that package.

spydon (approved):

st-pasha:
But what about other developers who are looking at the PR queue, maybe they would like to see this hint too?

spydon:
> But what about other developers who are looking at the PR queue, maybe they would like to see this hint too?

Better to use labels for that

PR #2190, opened 2022-11-25
feat: `clampLength` for `Vector2` extension

spydon:
# Description

With this method you can clamp the length of a vector, instead fitting it in a box like `clamp` and `clampScalar` does.

## Checklist

## Breaking Change?

## Related Issues


packages/flame/lib/src/extensions/vector2.dart
```diff
@@ -80,6 +80,20 @@ extension Vector2Extension on Vector2 {
...
+  /// Clamps the [length] of this vector.
+  ///
+  /// This means that if the length is less than [min] the length will be set to
+  /// [min] and if the length is larger than [max], the length will be set to
+  /// [max]. If the length is in between [min] and [max], no changes will be
+  /// made.
+  void clampLength(double min, double max) {
+    if (length2 > max * max) {
+      scaleTo(max);
+    } else if (length2 < min) {
```
st-pasha, line 92: Should be min squared.

packages/flame/lib/src/extensions/vector2.dart
```diff
@@ -80,6 +80,20 @@ extension Vector2Extension on Vector2 {
...
   }
 
+  /// Clamps the [length] of this vector.
+  ///
+  /// This means that if the length is less than [min] the length will be set to
+  /// [min] and if the length is larger than [max], the length will be set to
+  /// [max]. If the length is in between [min] and [max], no changes will be
+  /// made.
+  void clampLength(double min, double max) {
+    if (length2 > max * max) {
```
st-pasha, line 90: Perhaps length2 can be stored in a variable so that it won't be calculated twice?
Or maybe: precompute length and then use it with scale() function (since scaleTo will have to compute the length again)

packages/flame/lib/src/extensions/vector2.dart
spydon: Good catch, updated!

erickzanardo (approved):

PR #2246, opened 2022-12-31
fix: Only use initialized game for tests and remove setMount from onGameResize

### LINKED ISSUE #1351: Tests should use only properly initialized game instances

spydon:
# Description

Since we used `setMounted` in `onGameResize` the `mounted` feature didn't work properly when invoked from for example `onLoad`. This PR makes sure that all tests passes without using `setMounted` in `onLoad`.

Some `GameWidget` tests I want a second opinion on, or ideas of how to make those tests better, because when setting mounted explicitly it kind of cheats on part of those tests.

## Checklist

## Breaking Change?

## Related Issues

Closes #1351 


packages/flame_test/lib/src/flame_test.dart
```diff
@@ -89,9 +95,12 @@ class GameTester<T extends Game> {
     game.onGameResize(size);
 
     await game.onLoad();
+    game.onMount();
     game.update(0);
     if (game is FlameGame && makeReady) {
       await game.ready();
+      // ignore: invalid_use_of_visible_for_testing_member
```
luanpotter, line 102: do we need these since this is a test file?

luanpotter (approved):
LGTM with one comment

packages/flame_test/lib/src/flame_test.dart
spydon: For some reason the analyzer doesn't recognise it as a test.

renancaraujo (approved):

packages/flame_test/lib/src/flame_test.dart
```diff
@@ -158,6 +167,12 @@ class GameTester<T extends Game> {
           await _pump(gameWidget, tester);
           await tester.pump();
 
+          if (makeReady && game is FlameGame) {
+            // ignore: invalid_use_of_visible_for_testing_member
+            game.setMounted();
```
st-pasha, line 172: This looks suspicious to me: if the `game` is being put into a `GameWidget`, which is pumped through a `tester`, then this should be no different than the "normal" use of `FlameGame`. In other words, `setMounted` should not be necessary here...

Which makes me wonder - have we checked that the examples are still working ok?

spydon: Agree, which examples are you referring to?

st-pasha: I just tried running `examples/lib/main.dart` from this branch, and immediately got an error:
```
Assertion failed: file:///flame/packages/flame/lib/src/components/core/component.dart:1024:14
child._parent!.isMounted
```

spydon: Interesting, so somehow it isn't called in the order I expected then I guess, or not called at all... Any clue what could be wrong? I guess that would solve the sketchy widget tests too

spydon: Now I'm doing `setMounted` explicitly from the `GameWidget` instead after `onMount`, do you think that is the correct way to do it?

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -195,12 +195,16 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
...
+        assert(game.hasLayout);
+        final onLoad = game.onLoadFuture;
         if (onLoad != null) {
           await onLoad;
         }
-        currentGame.onMount();
+        game.onMount();
+        if (game is FlameGame) {
+          game.setMounted();
+        }
```
spydon, line 207: @st-pasha @renancaraujo is this the way to go, or do you have any other ideas?

packages/flame_test/lib/src/flame_test.dart
spydon: Moving the discussion here:
https://github.com/flame-engine/flame/pull/2246/files#r1062345184

packages/flame/lib/src/game/game_widget/game_widget.dart
st-pasha: I wanted to suggest to set this flag directly in `onMount()`, but then I'm not even sure what `onMount` even means for the `Game` class (if it's distinct from attaching to the widget tree). So, perhaps we could set this flag at the end of the `onLoad()`?

Frankly, I'm completely confused here.

spydon: `onLoad` should only run once of the game is removed from the widget and added again, so that would not render the correct behaviour. And running it directly in `onMount` would mean that the user would need to make sure they always run `super.onMount` last to get the correct behaviour, so I think the current solution is better than that.

st-pasha: We can go with the current approach for now, and then if we think of a better way in the future, we can always refactor.

packages/flame/test/events/flame_game_mixins/has_draggable_components_test.dart
```diff
@@ -2,26 +2,28 @@ import 'package:flame/components.dart';
...
 import 'package:flame/game.dart';
+import 'package:flame_test/flame_test.dart';
 import 'package:flutter_test/flutter_test.dart';
 
 void main() {
+  final draggableTester = GameTester(_GameWithHasDraggableComponents.new);
+  final dualDraggableTester = GameTester(_GameWithDualDraggableComponents.new);
+
   group('HasDraggableComponents', () {
-    testWidgets(
```
st-pasha, line 9: I don't think these changes to the tests are actually needed. They don't reduce the boilerplate in any way, plus they make the tests weaker.

When a test runs through `testWidgets()`, I can see how the `game` instance is created, and then put inside a `GameWidget`, and that widget is pumped through the test framework. When there is a "helper" function involved, I'll start to worry that it performs some "sneaky" initialization of the `game` instance (as it did before), and the code wouldn't actually work if it was used by an actual user who has no access to `@internal` methods.

packages/flame/lib/src/game/game_widget/game_widget.dart
st-pasha: I see that `FlameGame` now calls `setMounted()` inside its `mount()` method -- so perhaps these lines are no longer needed?

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -195,12 +195,16 @@ class _GameWidgetState<T extends Game> extends State<GameWidget<T>> {
...
-        assert(currentGame.hasLayout);
-        final onLoad = currentGame.onLoadFuture;
+        final game = currentGame;
+        assert(game.hasLayout);
+        final onLoad = game.onLoadFuture;
         if (onLoad != null) {
           await onLoad;
         }
-        currentGame.onMount();
+        game.onMount();
```
st-pasha, line 204: should this be `game.mount()` now?

spydon: It should be indeed, why are the tests passing...

packages/flame/lib/src/game/game_widget/game_widget.dart
spydon: Missed removing that one, removed now.

packages/flame/lib/src/game/game_widget/game_widget.dart
spydon: Since the other one was still there of course, fixed now.

packages/flame/test/events/flame_game_mixins/has_draggable_components_test.dart
spydon: Yeah, since there is no need for the initialization hack anymore I can revert these.
But it would be nice to be consistent with how we do the tests, and now `testGameWidget` is used on some places and `testWidgets` is used on some.

packages/flame/lib/src/game/game_widget/game_widget.dart
spydon: @st-pasha everything should be updated now, can you have another look? :)

st-pasha (approved):
Looks all good!

Issue #2177, opened 2022-11-22
Need a pushReplacementNamed and pushReplacement for `RouterComponent`


ufrshubham:
# Problem to solve

As of now, there is no way to push and replace current route with a new one in a single call. It can be achieved using two separate pop() and push() calls, but only if there are more than 2 routes on the stack.

My use case involves an `OverlayRoute` as the initial route and a normal component based route. To navigate from the initial route to the component based one, I have to first pop() the initial route (because `OverlayRoute`s are always rendered on top). But in doing so, the `length > 1` assertion gets triggered.

# Proposal

It would be good to have `pushReplacementNamed` and `pushReplacement` style APIs for RouterComponent, similar to flutter's navigator.

# More information

spydon:
Should I assign you to this @ufrshubham or do you want to take it @st-pasha?

ufrshubham:
> Should I assign you to this @ufrshubham or do you want to take it @st-pasha?

I am completely new to the routing system, so it would be better if Pasha takes a look at this.

munsterlander:
@spydon You can assign this to me.  I think I have it all done.

PR #2249, opened 2023-01-02
feat!: Adds new route methods `pushReplacement`, `pushReplacementNamed`, and `pushReplacementOverlay`

munsterlander:
# Description

This PR adds the aforementioned methods to the router component so a user can replace the current route / overlay with something else without having to call `pop` then `push`.  Additionally, it corrects an error in the router documents as `popPage()` does not exist as the call is just `pop()`.

## Checklist

## Breaking Change?

- [X] Yes, this PR is **barely** a breaking change.

## Related Issues

Closes: https://github.com/flame-engine/flame/issues/2177


packages/flame/lib/src/components/router_component.dart
```diff
@@ -104,6 +104,34 @@ class RouterComponent extends Component {
     _adjustRoutesVisibility();
   }
 
+  /// Pops the current route and places the route [name] on top of the
+  /// navigation stack.
+  ///
+  /// If the route is already in the stack, it will be simply moved to the top.
+  /// Otherwise the route will be mounted and added at the top. We will also
+  /// initiate building the route's page if it hasn't been built before. If the
+  /// route is already on top of the stack, this method will do nothing.
```
spydon, line 113:
```suggestion
  /// Otherwise the route will be mounted and added at the top. The route's
  /// page will also start building if it hasn't been built before. If the
  /// route is already on top of the stack, this method will do nothing.
```
Avoid "we" in dartdocs

packages/flame/lib/src/components/router_component.dart
```diff
@@ -104,6 +104,34 @@ class RouterComponent extends Component {
...
+  ///
+  /// The method calls the [Route.didPush] callback for the newly activated
```
spydon, line 115:
```suggestion
  /// This method calls the [Route.didPush] callback for the newly activated
```

packages/flame/lib/src/components/router_component.dart
```diff
@@ -123,6 +151,28 @@ class RouterComponent extends Component {
...
+  /// Pops the current route and puts a new [route] on top of the navigation
+  /// stack.
+  ///
+  /// The route may also be given a [name], in which case it will be cached in
+  /// the [routes] map under this name (if there was already a route with the
+  /// same name, it will be overwritten).
+  ///
+  /// The method calls [Route.didPush] for this new route after it is added and
+  /// also calls the [Route.didPop] callback for the popped route.
+  void pushReplacement(Route route, {String name = ''}) {
```
spydon, line 163: `name` should probably be nullable instead of an empty string by default.

spydon (commented):
Good job! Some minor comments.

packages/flame/lib/src/components/router_component.dart
munsterlander: Thats what you guys wrote, lol!  Love that git blame.

packages/flame/lib/src/components/router_component.dart
munsterlander: I just copied the existing `pushRoute` to implement.  @st-pasha wrote that code originally, so I am not certain as to why it was implemented as an empty string, but the code calls:

```
 void pushRoute(Route route, {String name = ''}) {
    route.name = name;
    if (name.isNotEmpty) {
      _routes[name] = route;
    }
```
So it only pushes the name to the stack if its not empty.  I can go either way, as its the same outcome, but maybe he had a reason for it??

spydon: Maybe he did, I can't come up with a good reason for using a string when it is not going to be used though, and theoretically an empty string is also a valid name for a route.

munsterlander: I can change the code and run some tests to see if there is any impact.

munsterlander: Ok, the answer is pretty easy, because in dart you can't have an optional named parameter of type string with a default value of null.  So you have to make it a required parameter which then causes other issues for unnamed routes.  Regardless, with it being an empty string, the if statement prevents the unnamed route from being added to the stack list, so you do not have multiple empty string route names or even one empty string route name.  Thus you can push an overlay or route that isn't named and once it is popped, it is destroyed.

spydon: You can have optional parameters of default value null, it is the default without you having to add anything.
Check here:
https://zapp.run/edit/flame-z03w062x03x0?

munsterlander: Ok, I must have seriously misread that then.  I have modified the code as necessary and all tests pass.  I will be curious what happens long term, because `route.dart` defines name like:

```
 /// The name of the route (set by the [RouterComponent]).
  String get name => _name;
  late String _name;
  @internal
  set name(String value) => _name = value;
  ```
  If its never set, which my code update prevents unless its not null, then is there anything that would ever call `route.name` and expect it to not be null when it could be because it was never set.  Regardless, I think its all working right.

Comment (spydon) [reply packages/flame/lib/src/components/router_component.dart]: Ooh right, this was already in there, I missed that.
Let's see what @st-pasha says.

File: packages/flame/lib/src/components/router_component.dart [hunk 4]
```diff
@@ -104,18 +104,68 @@ class RouterComponent extends Component {
...
+  /// navigation stack.
+  ///
+  /// If the route is already in the stack, it will simply be moved to the top.
+  /// Otherwise the route will be mounted and added at the top. The route's
+  /// page will also start building if it hasn't been built before. If the
+  /// route is already on top of the stack, this method will do nothing.
+  ///
+  /// This method calls the [Route.didPush] callback for the newly activated
+  /// route and also calls the [Route.didPop] callback for the popped route.
+  void pushReplacementNamed(String name) {
```
st-pasha, line 117: Maybe it would be simpler to add a boolean flag to the `pushNamed()` method above?
```dart
void pushNamed(String name, {bool replaceCurrent = false}) {
  ...
}
```
As far as I can see, the difference between the two methods is just a single line, so it would make further maintenance easier to have just a single method instead of two.

packages/flame/lib/src/components/router_component.dart
```diff
@@ -104,18 +104,68 @@ class RouterComponent extends Component {
...
+  /// Pops the current route and puts a new [route] on top of the navigation
+  /// stack.
+  ///
+  /// The route may also be given a [name], in which case it will be cached in
+  /// the [routes] map under this name (if there was already a route with the
+  /// same name, it will be overwritten).
+  ///
+  /// The method calls [Route.didPush] for this new route after it is added and
+  /// also calls the [Route.didPop] callback for the popped route.
+  void pushReplacement(Route route, {String? name}) {
```
st-pasha, line 163: (same here)

packages/flame/lib/src/components/router_component.dart
```diff
@@ -188,6 +254,18 @@ class RouterComponent extends Component {
     pop();
   }
 
+  /// Local method to bypass [pop]'s assert
+  void _popWithoutMinimumAssert() {
+    assert(
+      _routeStack.isNotEmpty,
+      'There must be a route to pop.',
+    );
```
st-pasha, line 262: I think this assert can be removed. Because
- it can never trigger as far as we know (there's always at least one route on the stack);
- even if the stack becomes empty through some unknown means, the error will trigger on the line below anyways;
- the method's name says "WithoutAssert", yet here's the assert, so it's a bit awkward.

packages/flame/lib/src/components/router_component.dart
st-pasha: I don't quite remember why I made the default name as the empty string. Maybe at first I thought the name must be required, and then when no-name routes were added later, it was the easiest way? Anyways, I don't mind if we make the name nullable. In which case the `name` getter should return `String?`, because we don't want it to throw an exception if the route has no name.

munsterlander: This would be a breaking change then or no?

spydon: It would, but quite minor, I don't think it would affect anyone.

munsterlander: Ok.  I will update accordingly.  I'm at CES so it will be a day or two before I get to it.

spydon: Cool, no rush!

packages/flame/lib/src/components/router_component.dart
spydon: It's pretty good to use the same names as exist in Flutter though:
https://api.flutter.dev/flutter/widgets/Navigator/pushReplacementNamed.html

packages/flame/lib/src/components/router_component.dart
spydon: And this one has this counter-part:
https://api.flutter.dev/flutter/widgets/Navigator/pushReplacement.html

packages/flame/lib/src/components/router_component.dart
munsterlander: I was going to do the flag originally,  but was like that definitely would be breaking, so avoided it, but then saw it in flutter so went that route. Unfortunately, I reversed word order but will fix that.

spydon: It wouldn't be breaking since the flag is optional, but I think it is better to match Flutter's method names anyways.

st-pasha: Ok, makes sense

packages/flame/lib/src/components/router_component.dart
munsterlander: Yeah, I was kind of torn on this one. I didn't have it at first, hence the name but I was concerned it would somehow get called on an empty stack and I didn't want to throw an error if it could be asserted. I agree it should never be empty, but didn't know if it should be overly cautious or not. I can remove it.

packages/flame/lib/src/components/route.dart
```diff
@@ -45,10 +45,10 @@ class Route extends PositionComponent with ParentIsA<RouterComponent> {
   final bool maintainState;
 
   /// The name of the route (set by the [RouterComponent]).
-  String get name => _name;
-  late String _name;
+  String? get name => _name;
+  late String? _name;
```
st-pasha, line 49: `late` can be removed now.

st-pasha (commented):
For an `OverlayRoute` we can try to define
`@override String get name => super.name!;`, because overlays should always have names

spydon (approved):

doc/flame/router.md
```diff
@@ -105,9 +108,12 @@ final router = RouterComponent(
 Overlays that were defined within the `GameWidget` don't even need to be declared within the routes
 map beforehand: the `RouterComponent.pushOverlay()` method can do it for you. Once an overlay route
 was registered, it can be activated either via the regular `.pushNamed()` method, or via the
-`.pushOverlay()` -- the two method will do exactly the same, though you can use the second one to
+`.pushOverlay()` -- the two methods will do exactly the same, though you can use the second one to
 make it more clear in your code that an overlay is being added instead of a regular route.
 
+The current overlay can be replaced using `pushReplacementOverlay`.  This method executes
+`pushReplacementName` or `pushReplacement` based on the status of the overlay being pushed.
```
st-pasha, line 115: `pushReplacementNamed`

st-pasha (approved):

doc/flame/router.md
munsterlander: Fixed

PR #2261, opened 2023-01-09
perf: Avoid Vector2 creation in `Sprite.render`

spydon:
# Description

This creates two new private vector2 objects that can be re-used inside of `Sprite.render` so that we don't have to create new objects in there.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/sprite.dart
+++ b/packages/flame/lib/src/sprite.dart
@@ -107,11 +107,7 @@ class Sprite {
       _tmpRenderPosition.setZero();
     }
 
-    if (size != null) {
-      _tmpRenderSize.setFrom(size);
-    } else {
-      _tmpRenderSize.setFrom(srcSize);
-    }
+    _tmpRenderSize.setFrom(size ?? srcSize);
 
     _tmpRenderPosition.setValues(
       _tmpRenderPosition.x - (anchor.x * _tmpRenderSize.x),
--- a/packages/flame/lib/src/sprite.dart
+++ b/packages/flame/lib/src/sprite.dart
@@ -82,8 +82,8 @@ class Sprite {
   }
 
   // Used to avoid the creation of new Vector2 objects in render.
-  late final _tmpRenderPosition = Vector2.zero();
-  late final _tmpRenderSize = Vector2.zero();
+  static final _tmpRenderPosition = Vector2.zero();
+  static final _tmpRenderSize = Vector2.zero();
 
   /// Renders this sprite onto the [canvas].
   ///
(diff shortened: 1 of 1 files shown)


packages/flame/lib/src/sprite.dart
```diff
@@ -97,18 +101,30 @@ class Sprite {
...
+      _tmpRenderPosition.setFrom(position);
+    } else {
+      _tmpRenderPosition.setZero();
+    }
+
+    if (size != null) {
+      _tmpRenderSize.setFrom(size);
+    } else {
+      _tmpRenderSize.setFrom(srcSize);
+    }
```
st-pasha, line 114:
```suggestion
    _tmpRenderSize.setFrom(size ?? srcSize);
```

packages/flame/lib/src/sprite.dart
```diff
@@ -81,6 +81,10 @@ class Sprite {
     );
   }
 
+  // Used to avoid the creation of new Vector2 objects in render.
+  late final _tmpRenderPosition = Vector2.zero();
+  late final _tmpRenderSize = Vector2.zero();
```
st-pasha, line 86: Probably no need to make them `late`.

wolfenrain (approved):

packages/flame/lib/src/sprite.dart
spydon: Lol, really is time for me to go to bed :joy: 
Thanks!

packages/flame/lib/src/sprite.dart
spydon: No harm in doing so right? If some user has some big collection of sprites that aren't being rendered yet they have a little bit less overhead.

st-pasha: Well, the overhead is that there needs to be a runtime check whether the `late` variable has actually been initialized or not. And I'm not sure whether such a check is debug-mode only, or if it is also done in production...

spydon: I don't think such a check is even done in debug-mode? Then we'd have better stack traces for `LateInitializationError` right?

spydon: That must be very minimal? But I don't know how it is done.
I'd say it is worth the trade-off of not carrying around two `Vector2` before they are used.
I can do some benchmarking tomorrow.

st-pasha: Another possibility is to declare them `static`. Then you have only one object instance no matter how many sprites

spydon: True, single thread ftw 😄

spydon, line 86:
```suggestion
  static final _tmpRenderPosition = Vector2.zero();
  static final _tmpRenderSize = Vector2.zero();
```

PR #2270, opened 2023-01-13
feat: Add a `canSee` method to the `CameraComponent`

wolfenrain:
# Description

Adding `inView` method so developers can more easily check if a given component is visible from a camera point of view.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/doc/flame/camera_component.md
+++ b/doc/flame/camera_component.md
@@ -173,6 +173,19 @@ The `visibleWorldRect` is a cached property, and it updates automatically whenev
 moves or the viewport changes its size.
 
 
+### Check if a component is visible from the camera point of view
+
+The `CameraComponent` has a method `inView` which can be used to check if a component is visible
+from the camera point of view. This is useful for example to cull components that are not in
+view.
+
+```dart
+if (!camera.inView(component)) {
+   component.removeFromParent(); // Cull the component
+}
+```
+
+
 ## Comparison to the traditional camera
 
 Compared to the normal [Camera](camera_and_viewport.md), the `CameraComponent`
--- a/doc/flame/camera_component.md
+++ b/doc/flame/camera_component.md
@@ -179,12 +179,13 @@ moves or the viewport changes its size.
 
 ### Check if a component is visible from the camera point of view
 
-The `CameraComponent` has a method `isInView` which can be used to check if a component is visible
-from the camera point of view. This is useful for example to cull components that are not in
+The `CameraComponent` has a method called `sees` which can be used
+to check if a component is visible from the camera point of view.
+This is useful for example to cull components that are not in
 view.
 
 ```dart
-if (!camera.isInView(component)) {
+if (!camera.sees(component)) {
    component.removeFromParent(); // Cull the component
 }
 ```
--- a/doc/flame/camera_component.md
+++ b/doc/flame/camera_component.md
@@ -175,12 +175,12 @@ moves or the viewport changes its size.
 
 ### Check if a component is visible from the camera point of view
 
-The `CameraComponent` has a method `inView` which can be used to check if a component is visible
+The `CameraComponent` has a method `isInView` which can be used to check if a component is visible
 from the camera point of view. This is useful for example to cull components that are not in
 view.
 
 ```dart
-if (!camera.inView(component)) {
+if (!camera.isInView(component)) {
    component.removeFromParent(); // Cull the component
 }
 ```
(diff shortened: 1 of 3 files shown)


luanpotter (approved):
LGTM

packages/flame/lib/src/camera/camera_component.dart
```diff
@@ -271,4 +271,10 @@ class CameraComponent extends Component {
       boundedBehavior.bounds = bounds;
     }
   }
+
+  /// Returns true if the given [component] is currently visible through this
+  /// camera.
+  bool inView(PositionComponent component) {
+    return visibleWorldRect.overlaps(component.toAbsoluteRect());
```
spydon, line 278: We can probably make this more efficient in the long run, but it might not be possible now.

spydon (approved):

packages/flame/lib/src/camera/camera_component.dart
```diff
@@ -271,4 +271,10 @@ class CameraComponent extends Component {
       boundedBehavior.bounds = bounds;
     }
   }
+
+  /// Returns true if the given [component] is currently visible through this
+  /// camera.
+  bool inView(PositionComponent component) {
```
st-pasha, line 277: I think Dart's naming convention would be to call this method `isInView`: https://dart.dev/guides/language/effective-dart/design#prefer-a-non-imperative-verb-phrase-for-a-boolean-property-or-variable

st-pasha (approved):

packages/flame/lib/src/camera/camera_component.dart
wolfenrain: Changed it!

st-pasha: Oops -- sorry, my bad... Now it sounds like the camera is in view of a component: `if (camera.isInView(component)) ...`. Perhaps `hasInView` would sound better?

wolfenrain: Hmm, what about `sees`? `if (camera.sees(component)) ...`

st-pasha: Yeah, that's even better

wolfenrain: Changed it!

packages/flame/lib/src/camera/camera_component.dart
```diff
@@ -278,4 +278,9 @@ class CameraComponent extends Component {
       boundedBehavior.bounds = bounds;
     }
   }
+
+  /// Returns true if this camera is able to see the [component].
+  bool sees(PositionComponent component) {
```
spydon, line 283: I think `isInView` was clearer

st-pasha: It would have been clearer if it was a Component's method:
```
if (component.isInView(camera)) { ... }
```
However, when it is a method on camera, then the statement reads as *"If camera is in view of component ..."*, which is confusing.

packages/flame/lib/src/camera/camera_component.dart
aytunch: Would it be useful to return the overlapping `Rect` instead of a `bool`? Maybe we need to make a decision according to how much it is being "seen"?

st-pasha: That could probably be another helper method.

packages/flame/lib/src/camera/camera_component.dart
spydon: Makes sense. Hmm... I don't think `sees` is super clear either, maybe `canSee`? Let me think until tomorrow (maybe you'll come up with some other alternatives too).

spydon: Well today, but when I wake up 😅

st-pasha: `canSee` is not bad, I like it

wolfenrain: Fixed it to `canSee`, jeez naming things is hard :D

doc/flame/camera_component.md
```diff
@@ -177,6 +177,20 @@ The `visibleWorldRect` is a cached property, and it updates automatically whenev
 moves or the viewport changes its size.
 
 
+### Check if a component is visible from the camera point of view
+
+The `CameraComponent` has a method called `canSee` which can be used
+to check if a component is visible from the camera point of view.
+This is useful for example to cull components that are not in
+view.
```
spydon, line 185:
```suggestion
The `CameraComponent` has a method called `canSee` which can be used to check
if a component is visible from the camera point of view. 
This is useful for example to cull components that are not in view.
```
It wasn't filling out up to 80

spydon (approved):

PR #2278, opened 2023-01-21
docs: Create "dart" domain extension

st-pasha:
# Description

In Sphinx, there is a concept of a "domain", which enables documenting library object such as classes and functions, and then referencing them from any other place in the documentation. This PR adds such a domain for the Dart language.

With the new approach, a class/function can be documented using the following directive:
``````
```{dartdoc}
:package: flame
:symbol: GameWidget
:file: src/game/game_widget/game_widget.dart
```
``````

The documentation will then be automatically extracted from the referenced file using the `dartdoc_json` tool (published as a separate package), and then rendered inside the page. The screenshot below shows an example of how DialogueRunner class from Jenny would be rendered:

Screenshot 2023-01-20 at 7 07 57 PM

Once a symbol has been documented using the `dartdoc` directive, it can be referenced from other places in the documentation as
```
{ref}`DialogueRunner`
```

## Checklist

## Breaking Change?

### CODE REVIEW

**Comment by munsterlander** [comment]
This is really cool!

File: .markdownlint.yaml [hunk 1]
```diff
@@ -233,6 +233,6 @@ MD051: true
 MD052: true
 
 # MD053/link-image-reference-definitions - Link and image reference definitions should be needed
-MD053:
+MD053: false
   # Ignored definitions
-  ignored_definitions: ["//"]
+  # ignored_definitions: ["//"]
```
spydon, line 238: I guess this will be removed when the PR isn't draft anymore?

packages/flame_jenny/jenny/lib/src/dialogue_view.dart
```diff
@@ -7,13 +7,35 @@ import 'package:jenny/src/structure/dialogue_choice.dart';
...
+///
+/// The class defines a number of "event handler" methods, which can be
+/// overridden in subclasses in order to respond to the corresponding event.
+/// Each method has a default no-op implementation, which means you only need
+/// to override those methods that you care about.
+///
+/// Most of the event handler methods return [FutureOr], which means they can
+/// be implemented either synchronously or asynchronously. In the latter case
+/// the dialogue runner will wait for the future to resolve before proceeding
+/// (futures from several dialogue views will be awaited simultaneously).
```
spydon, line 30: Were these Jenny fixes supposed to be in this PR? 🤔

st-pasha: Yes; the changes in Jenny are only related to doc-comments. I needed to test the new Dart domain feature on something, and Jenny classes seemed like a good fit. The end result is that some content from the docs site ends up being transferred into the code. Which is good, because it means the docs can now only be written in one place, and don't have to be duplicated.

So what you see as changes in the Jenny code, is really just me combining docs with the code doc-comments.

spydon: Sounds good, didn't read it that carefully on the first pass.

spydon (approved):
This will be really nice!

PR #2281, opened 2023-01-23
ci: Added Melos command "doc-setup"

st-pasha:
# Description

* The purpose of this new command is to make it easier to install things needed for doc development.
* Also added `jinja2` into the requirements file -- currently it is installed manually by the doc-publish script (https://github.com/flame-engine/flame-docs-site/blob/main/publish.sh#L58)

## Checklist

## Breaking Change?

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -51,6 +51,19 @@ scripts:
     run: melos exec flutter pub run dartdoc
     description: Run dartdoc checks for all packages.
 
+  doc-setup:
+    run: |
+      echo '\033[1;32mChecking python version:\033[m' &&
+      python --version &&
+      echo '\n\033[1;32mInstalling required python modules:\033[m' &&
+      (python -c 'import sys; sys.exit(0 if sys.version_info>=(3,8) else 2)' ||
+       (echo '\033[1;31mError: Python 3.8+ is required\033[m' && false)) &&
+      python -m pip install -r "$MELOS_ROOT_PATH/doc/_sphinx/requirements.txt" &&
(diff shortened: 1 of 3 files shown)


melos.yaml
```diff
@@ -51,6 +51,19 @@ scripts:
...
+  doc-setup:
+    run: |
+      echo '\033[1;32mChecking python version:\033[m' &&
+      python --version &&
+      echo '\n\033[1;32mInstalling required python modules:\033[m' &&
+      (python -c 'import sys; sys.exit(0 if sys.version_info>=(3,8) else 2)' ||
+       (echo '\033[1;31mError: Python 3.8+ is required\033[m' && false)) &&
+      python -m pip install -r "$MELOS_ROOT_PATH/doc/_sphinx/requirements.txt" &&
+      echo '\n\033[1;32mInstalling dartdoc_json:\033[m' &&
+      dart pub global activate dartdoc_json &&
```
spydon, line 63: We should be able to run this without `global` I think, and we should probably also check whether `dart` exists on the path and if not use `flutter` instead.

st-pasha: `dart pub activate dartdoc_json` wouldn't work -- or what did you have in mind?

spydon: Not sure what I had in mind, I've never done that.
But feels like there must be a way to not install it globally on the system.
Maybe it would actually work to download it as a dev_dependency as you suggested before and then point to where it is downloaded. But also maybe it is not worth it...
Maybe just keep it like it is. :smile:

spydon (approved):

munsterlander:
Just FYI - this command does not work on Windows.  I am looking into a fix.

PR #2284, opened 2023-01-24
refactor: Remove "items" variable from core Broadphase class.

### LINKED ISSUE #2283: Remove "items" variable from core Broadphase class
# What could be improved
More complex algorithms require more complex data structures. The basic `Broadphase` class contains the `items` variable that most probably will be unuseful in algorithms that are more complex than Sweep.
Also, the Flame forces other broadphase implementations to fill the `item` with unneeded data. And also, which is more important, it forces to remove each hitbox from `items` whereas such operation becomes extremely slow for a big count of objects because `items` is just a simple `List`

# Why should this be improved
Keeping items in the core class affects performance for hitbox removal operation and consumes additional memory.

# Any risks?
This change will be breaking, if somebody counts on the `items` variable and uses QuadTree or any own collision detection system implementation. 

# More information
I figured out this while running a performance test for another broadphase implementation. All performance improvements was riuned just thanks to this variable :-)

ASGAlex:
# Description
`items` variable from core `Broadphase` class is replaced to abstract getter. Also methods for adding and removing items are added. All this together allows to preserve backward compatibility and allow broadphase / collision detection  developer to specify own algorithm of adding/removing hitboxes and algorithm of accessing a full list of hitboxes.

## Checklist

## Breaking Change?

## Related Issues

Closes #2283

### DIFF
--- a/packages/flame/lib/src/collisions/broadphase/broadphase.dart
+++ b/packages/flame/lib/src/collisions/broadphase/broadphase.dart
@@ -6,9 +6,9 @@ import 'package:meta/meta.dart';
 /// actual intersections are calculated.
 ///
 /// Currently there are two implementations of [Broadphase]:
-/// - [Sweep] - simplest but slowest system, yet nice for small count of objects
-/// - [QuadTree] - work faster but requires additional setup and works only
-///   with fixed-size maps. See [HasQuadTreeCollisionDetection] for details
+/// - [Sweep] is the simplest but slowest system, yet nice for small amounts of hitboxes.
+/// - [QuadTree] usually works faster, but requires additional setup and works only
+///   with fixed-size maps. See [HasQuadTreeCollisionDetection] for details.
 abstract class Broadphase<T extends Hitbox<T>> {
   Broadphase();
 
--- a/packages/flame/lib/src/collisions/broadphase/broadphase.dart
+++ b/packages/flame/lib/src/collisions/broadphase/broadphase.dart
@@ -6,9 +6,11 @@ import 'package:meta/meta.dart';
 /// actual intersections are calculated.
 ///
 /// Currently there are two implementations of [Broadphase]:
-/// - [Sweep] is the simplest but slowest system, yet nice for small amounts of hitboxes.
-/// - [QuadTree] usually works faster, but requires additional setup and works only
-///   with fixed-size maps. See [HasQuadTreeCollisionDetection] for details.
+/// - [Sweep] is the simplest but slowest system, yet nice for small amounts of
+///   hitboxes.
+/// - [QuadTree] usually works faster, but requires additional setup and works
+///   only with fixed-size maps. See [HasQuadTreeCollisionDetection] for
+///   details.
 abstract class Broadphase<T extends Hitbox<T>> {
   Broadphase();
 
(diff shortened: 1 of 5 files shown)


renancaraujo (approved):

packages/flame/lib/src/collisions/broadphase/broadphase.dart
```diff
@@ -5,17 +5,33 @@ import 'package:meta/meta.dart';
 /// by doing a rough estimation of which hitboxes that can collide before their
 /// actual intersections are calculated.
 ///
-/// Currently there is only one implementation of [Broadphase] and that is
-/// [Sweep].
+/// Currently there are two implementations of [Broadphase]:
+/// - [Sweep] - simplest but slowest system, yet nice for small count of objects
+/// - [QuadTree] - work faster but requires additional setup and works only
+///   with fixed-size maps. See [HasQuadTreeCollisionDetection] for details
```
spydon, line 11:
```suggestion
/// Currently there are two implementations of [Broadphase]:
/// - [Sweep] is the simplest but slowest system, yet nice for small amounts of hitboxes.
/// - [QuadTree] usually works faster, but requires additional setup and works only
///   with fixed-size maps. See [HasQuadTreeCollisionDetection] for details.
```
Added `usually`, because when I tried it on a map where a lot of tree recalculations had to be done it actually turned out to be slower. But maybe you've done some more recent benchmarks?

packages/flame/lib/src/collisions/broadphase/broadphase.dart
```diff
@@ -5,17 +5,33 @@ import 'package:meta/meta.dart';
...
-
-  Broadphase({List<T>? items}) : items = items ?? [];
+  Broadphase();
 
   /// This method can be used if there are things that needs to be prepared in
   /// each tick.
   void update() {}
 
+  /// Returns a flat List of items regardless of what data structure is used to
+  /// store collision information
```
spydon, line 20:
```suggestion
  /// store collision information.
```

packages/flame/lib/src/collisions/broadphase/broadphase.dart
```diff
@@ -5,17 +5,33 @@ import 'package:meta/meta.dart';
...
+  List<T> get items;
+
+  /// Add item to broadphase. Should be called in [CollisionDetection] class
+  /// while adding a hitbox into collision detection system
```
spydon, line 24:
```suggestion
  /// Adds an item to the broadphase. Should be called in a
  /// [CollisionDetection] class while adding a hitbox into its collision
  /// detection system.
```

packages/flame/lib/src/collisions/broadphase/broadphase.dart
```diff
@@ -5,17 +5,33 @@ import 'package:meta/meta.dart';
...
+  void add(T item);
+
+  void addAll(Iterable<T> items) => items.forEach(add);
+
+  /// Remove item from broadphase. Should be called in [CollisionDetection]
+  /// class while removing a hitbox into collision detection system
```
spydon, line 30:
```suggestion
  /// Removes an item from the broadphase. Should be called in a
  /// [CollisionDetection] class while removing a hitbox from its collision
  /// detection system.
```

spydon (approved):
Lgtm! Just some grammatical comments

packages/flame/lib/src/collisions/broadphase/broadphase.dart
ASGAlex: What kind of tree recalculations? 
Adding and removing components is not cheap and QuadTree creates additional listeners to track updates, whereas Sweep just adds item into list. So it is known... "sacrifice" of performance on component's adding stage with hope to have faster calculations after.

spydon: A lot of fast moving hitboxes is what I'm talking about, not adding and removing hitboxes.

PR #2286, opened 2023-01-24
ci: Remove lint rule invariant_booleans

st-pasha:
The rule is deprecated, and is scheduled for removal in a future version of Dart linter: https://dart-lang.github.io/linter/lints/invariant_booleans.html


wolfenrain:
Oh I opened #2287 as well, will close that in a bit in favor of this, tho we should update the title as this is a change on a package and not just CI related

wolfenrain (approved):

Gustl22:
@spydon can we do a minor release for this package? AudioPlayers depends on this and this makes the CI fail :)

PR #2295, opened 2023-01-26
feat: Game splash screen with loading progress. 

### LINKED ISSUE #1776: Implement "Loader" component
# Problem to solve

Often, a game needs some time to load its resources/assets. In this case some kind of a loading screen is necessary.

Currently, the only support from Flame for this functionality comes from the `loadingBuilder` parameter of the `GameWidget`. However, this solution may be insufficient for the following reasons:
- it is non-trivial to pass the loading progress from the game to the loading widget;
- since the loading page is inherently dynamic, one may want to use Flame's rendering for it instead of relying on Flutter animations or custom painters;
- it is unclear how to use the loadingBuilder (or whether it is even possible) to show loading progress when loading internal components, such as a particular level within the game.

# Proposal

How exactly to solve this is TBD. Should this be a specialized `Route` within the Navigator? Or a mixin to a page? Or a page wrapper? Or a child?

I'd be interested to hear how people solve this problem right now.

ASGAlex:
# Description

As mentioned in #1776, "it is non-trivial to pass the loading progress from the game to the loading widget".  With this API you can do it in few steps: 

1. Declare immutable message class
2. Implement getter for `progressNotifier` in your game class and pass the message's type into template parameter
3. Create your own stateful or stateless widget with `LoadingWidgetMixin`  and call `widget.initMessageListener` to initialize stream listeners (I don't know exactly a use-case for stateless widget, but whatewer, you just can to use it).
4. Implement custom `build` for progress screen. It should to return `widget.gameWidget` at the moment when we want to switch from loading screen to the game
5. Just call `progressNotifier.reportLoadingProgress` with a message in game's `onLoad` function. 

This approach doen not restricts developer on which data format to use for messaging, what component to use for loading screen. Every third-party "loading progress" component should be integrated easily. 

As a bonus, there is a special mixin `ProgressListener` for components.  So you can continue to use `progressNotifier.reportLoadingProgress` after the game is loaded and components with this mixin will receive messages and... do something :-)

It does not breaks default `loadingBuilder`, it works just as alternative, if everything is setup.

## Checklist

## Breaking Change?

## Related Issues
Closes #1776 (at least at core level)


examples/lib/stories/experimental/loading_screen.dart
```diff
@@ -0,0 +1,239 @@
...
+        progress = 101;
+
+  final String message;
+  final int progress;
+}
+
+/// The loading screen widget with animated progress bar anf fade effect
+/// when displaying the game.
+class LoadingScreenExampleWidget extends StatefulWidget
+    with LoadingWidgetMixin<ProgressMessage> {
```
aytunch, line 123: Is there a special reason why we add the `LoadingWidgetMixin` here instead of the `State` class just below?
Usually I see mixins are added to State part of the `StatefulWidget`s next to `SingleTickerProviderStateMixin`

examples/lib/stories/experimental/loading_screen.dart
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  /// This is just an example of sending messages to game components.
+  @override
+  void update(double dt) {
+    anyProgressEmulation += dt;
+    final probability = Random().nextInt(100);
+    if (probability < 1) {
+      messagesSent++;
+      progressNotifier.reportLoadingProgress(
```
aytunch, line 92: What is the best way to separate component and game `progressNotifier`s?
let's say i want to have a 2 progress systems.

1. Game Loading progress (will be used by Flutter)
2. Levels loading progress (will be used by Components)

What is the preferred way to implement this? I don't want to receive component messages from Flutter and game load progress from within Components.

examples/lib/stories/experimental/loading_screen.dart
```diff
@@ -0,0 +1,239 @@
...
+    _controller.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    /// This is the game's widget. You should manually decide, when it is the
+    /// time to show it. But in return you have ability to add some effects!
+    if (showGame) {
+      return Opacity(
```
aytunch, line 187: Would wrapping the `FlameGame` in `Opacity` decrease performance even when the value is 1.0?
I remember `Opacity` being very cpu intensive: https://docs.flutter.dev/perf/best-practices#minimize-use-of-opacity-and-clipping
Maybe using a `AnimatedCrossFade` is more performant?

ASGAlex: Maybe, it was just fast insight when I added it - to demonstrate possibilities. 
You even could rebuild the widget without Opacity or any unnecessary widgets at all, after the animation finish

examples/lib/stories/experimental/loading_screen.dart
ASGAlex: I did not understand clearly your point. If you have two progress systems - well, ok, it will work. 
If the game is still loading, there are no messages from the component's lifecycle, because the lifecycle still has not started. 
If the game is already loaded and you have a message from the component - you could just ignore it in your loading widget, setting a variable to track the difference between states, for example... 
I see I missed it in my example, and every future message triggers `setState` - will fix it.

examples/lib/stories/experimental/loading_screen.dart
ASGAlex: The reason is an attempt to hide all stream setups from the end-user. 
I want to pass the already initialized `_LoadingProgressReceiver` to a user-defined widget, so I need to access this widget at build time. I haven't such access to the widget's state. Also, maybe your loading screen is just a stateless wrapper-widget? In this cause, the mixin will not work at all.

ASGAlex: Ah, got it, my trick does not work here. Thanks! 
Dont waste your time on reading my code now, please:  it has very critical problem!

examples/lib/stories/experimental/loading_screen.dart
ASGAlex: I changed it to `AnimatedSwitcher` now.

examples/lib/stories/experimental/loading_screen.dart
ASGAlex: Done. 
Components with mixin are replaced by custom class with builder functions. 
Streams and futures are kept under hood. After loading is finished, unnecessary StreamBuilder is removed from widgets tree, so performance should not be affected. 

And yes, bugs seems to be fixed now =)

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -39,6 +40,8 @@ class GameWidget<T extends Game> extends StatefulWidget {
   /// The text direction to be used in text elements in a game.
   final TextDirection? textDirection;
 
+  final LoadingWidgetBuilder? loadingProgressBuilder;
```
st-pasha, line 43: What's the difference between this and `loadingBuilder`?

ASGAlex: In loadingBuilder you should setup subscription on stream manually. With this class everything is already done, you build function just have additional "message" parameter. Also the base class takes care on removing StreamBuilder from widget tree when onLoad is finished - to prevent tree rebuilding on new messages. And allows you to setup transition between loading screen and game widget.

Of course everything could be done manually, even with loadBuilder (except animated transition). A new separate class allows user to forget about concrete messaging implementation and just to build own widget. It also allows to define strict message type - this is more flexible than receiving strings and less dirty than receiving "dynamic"

st-pasha: I get the idea about the `LoadingWidget` class, but now `GameWidget` will have two parameters: `loadingBuilder`, and `loadingProgressBuilder`, which I feel would be confusing to users. Can't they be combined into one?

packages/flame/lib/src/widgets/loading_widget_builder.dart
```diff
@@ -0,0 +1,66 @@
+import 'dart:async';
```
st-pasha, line 1: This file should probably reside in the game/game_widget directory.

packages/flame/lib/src/game/game_widget/game_widget.dart
st-pasha: I feel the API would be much simpler if we could say: `GameWidget` has the `loadingBuilder` parameter, where you can pass any widget you want. But if you create that widget based on `LoadingWidget`, then it will also receive loading progress events from the game.

ASGAlex: > But if you create that widget based on LoadingWidget, then it will also receive loading progress events from the game.

The problem is that loadingProgressBuilder is not a builder of a widget, it is pure mutable class. This is necessary for make stream setup under hood. But if to refactor in to Widget-based class, it become immutable and all widget's parameters should be passed only to constructor. So this will force user to do unnecessary copy-paste of boilerplate code. That's why `loadingProgressBuilder` and `loadingBuilder` are not compatible in this way.

I thought that `loadingProgressBuilder` could replace `loadingBuilder`  at future because it provides same functionality. To make things easier for users, who won't have any progress tracking, I thing it would be possible to subclass a special LoadingWidgetBuilder class with only `loadingBuilder`-like function as parameter. 

So, we have two options here: to make this change breaking but with easy migration steps. Or to make it non-breaking, by keeping both builders.

st-pasha: > This is necessary for make stream setup under hood. But if to refactor in to Widget-based class, it become immutable and all widget's parameters should be passed only to constructor.

I'm not sure I understand what the problem is. So, suppose there is a `LoadingProgressWidget`, which has a function `getProgressMessage(context)`, which finds `GameWidget` from the build context, retrieves the `game` object from there, then looks up the last progress message from the `Game`. The API could look something like this:
```dart
class MyLoadingScreenWidget extends LoadingProgressWidget<double> {
  @override Widget build(BuildContext context) {
    final message = getProgressMessage(context);
    return Center(child: Text(message.toString()));
  }
}

void main() {
  runApp(
    GameWidget(
      game: game,
      loadingBuilder: MyLoadingScreenWidget(),
    )
  );
}
```

The `Game` would then be responsible for driving rebuilds of the game widget using its `refreshWidget` method.

ASGAlex: Yes, this should work too. But only advantage of this approach is backward compatibility, I think. Here is the list of disadvantages I see: 

**First.**  You can't implement transition between loading screen and game widget after game is loaded. To do that, you should to pass something into `MyLoadingScreenWidget` constructor, but you can't, besause `Stack(children: stackedWidgets);` does not exists here. You can fix it by using factory finction instead passing widget itself, like this:  `loadingBuilder: (Widget gameWidget) => MyLoadingScreenWidget(gameWidget),`. But with this you will force all users to repeat this unnecessary code and implement necessary parameter in constructor. And, as far as I know, IDE won't generate constructor for you automatically. Whereas with `loadingProgressBuilder` things become much easier: all variables are set in GameWidget builder and you dont care aboun constructor. And IDE will notice you about functions needed to be reimplemented. 

**Second.**  You will need different messaging implementation for widgets and for components. `reportLoadingProgress` can send messages both to loading widget and to components with `ProgressListener` mixin. Also streams is universal solution to subscribe any listener on it, for example to write logs or to create in-game debugging console (with cheatcodes, of course (-: ). With `HasProgressNotifier` you have `messagingStream` getter which returns broadcast stream. You can do with it whatever you want. But with approach 
>The Game would then be responsible for driving rebuilds of the game widget using its refreshWidget method 

we going to do separate implementations for loading screen, for components, etc. 
So with streams I hope to solve many posiible problems at core level.

st-pasha: > But only advantage of this approach is backward compatibility, I think.

Another big advantage here is code simplicity. A complicated code creates maintenance problems in the long term (or even short term), as it increases chances of interference when new code needs to be added into the `GameWidget` class. Ideally, it would be nice to have the new functionality tightly encapsulated within one or two classes or mixins, without affecting the rest of the code too much. But the current approach modifies `GameWidget` quite significantly, which means it's likely that future changes to `GameWidget` may run into conflict with this functionality.

> First. You can't implement transition between loading screen and game widget after game is loaded.

True, but this functionality is independent from displaying the loading progress itself. This is something that could be implemented separately, and even have a separate parameter in the `GameWidget`: say, `loadingBuilderTransition`.

> Second. You will need different messaging implementation for widgets and for components.

This is true. But, there are really no components that would want to listen to game loading progress events. Simply because at the time the game is loading, no components are alive yet. And for logging purposes, you could simply add the needed code into the function that emits those progress messages in the first place. It would be quite simpler than creating the stream, then subscribing to the stream and unsubscribing from the stream.

As for loading components themselves (as described in #1776), I don't think streams are necessary either. Because, again, there would be only one component (`LoaderComponent`) interested in those "events", and it could just go and query the progress itself.

I understand that coming from Flutter, it seems natural to have streams everywhere to send messages to widgets to inform them of something that's happening, and forcing them to rebuild that way. But this is a Flutter approach, where everything is static and needs to be prodded by an external force in order to change.

But Flame is not like that. Flame is more dynamic, every component lives its own life, and updates and re-renders on every tick. A `LoaderComponent` doesn't need to subscribe to any events or streams, it can simply read `child.progress` value and draw itself based on that value, and that's it.

ASGAlex: >But the current approach modifies GameWidget quite significantly
Can't agree with this: there is just additional parameter and several lines of code to work with it. The rest of changes are related to formatting code and there is really nothing significant.

>This is something that could be implemented separately, and even have a separate parameter in the GameWidget: say, loadingBuilderTransition.

And this will change GameWidget much more than current implementation, where all possible things are bundled into `LoadingWidgetBuilder` and game's mixin `HasProgressNotifier`. 
Unfortunately, there is no way to avoid modification of `GameWidget`, if we want to implement transition after loading screen. 

>As for loading components themselves

Oh, no, my proposition was not about to use it while loading component themselves. In general, I thinking about such causes: b

>For example, there could be a game where the player travels on a world map, encounters a monster, and then goes into special fighting UI -- the UI, and the monster assets need to be loaded at this moment. Or the player enters a dungeon, and now you need to load the dungeon visuals. Etc.

In such situation components loading will be initialized from Game level, so we would not call progress notifier from component which loads itself. It is planned to be called from game, when game triggers loading process for a set of additional components.


But well, I see your point and I think I can't offer something useful for fixing the issue in the way you describe 👌🏼

PR #2296, opened 2023-01-27
fix: RouterComponent replace methods to correctly handle previous/nextRoute

spydon:
# Description

Before when you only had two routes you'd get:
```
Bad state: No element

dart:core                                                  List.last
package:flame/src/components/router_component.dart 260:30  RouterComponent._popWithoutMinimumAssert
package:flame/src/components/router_component.dart 122:7   RouterComponent.pushReplacementNamed
test/components/router_component_test.dart 20:14           main.<fn>.<fn>
```

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/components/router_component.dart
+++ b/packages/flame/lib/src/components/router_component.dart
@@ -257,7 +257,7 @@ class RouterComponent extends Component {
   /// Local method to bypass [pop]'s assert
   void _popWithoutMinimumAssert() {
     final route = _routeStack.removeLast();
-    route.didPop(_routeStack.last);
+    route.didPop(route);
     route.removeFromParent();
   }
 
--- a/packages/flame/lib/src/components/router_component.dart
+++ b/packages/flame/lib/src/components/router_component.dart
@@ -119,7 +119,7 @@ class RouterComponent extends Component {
     if (route == currentRoute) {
       return;
     } else {
-      _popWithoutMinimumAssert();
+      _replaceRoute(route);
     }
     if (_routeStack.contains(route)) {
       _routeStack.remove(route);
@@ -165,7 +165,7 @@ class RouterComponent extends Component {
       route.name = name;
       _routes[name] = route;
     }
-    _popWithoutMinimumAssert();
+    _replaceRoute(route);
     add(route);
     _routeStack.add(route);
     _adjustRoutesOrder();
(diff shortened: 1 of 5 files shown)


erickzanardo (approved):

munsterlander (commented):
LGTM - odd that made it through, but it comes from pop with the same tests just modified.

Edit to clarify:

I just want to confirm that `pop()` does not have the same issue. 

```dart
void pop() {
    assert(
      _routeStack.length > 1,
      'Cannot pop the last route from the Router',
    );
    final route = _routeStack.removeLast();
    _adjustRoutesOrder();
    _adjustRoutesVisibility();
    route.didPop(_routeStack.last);
}
```

spydon:
> I just want to confirm that `pop()` does not have the same issue.

It does indeed, good catch!
I'll write some tests that tests this.

spydon:
Ah wait, this is wrong, `onPop` should return the next route, not the route that was popped.

munsterlander:
> Ah wait, this is wrong, `onPop` should return the next route, not the route that was popped.

Ah! When I was adding this, I was confused why that was there, but not knowing the full router creation origin story, I was like well there had to be a reason.   The tests passed for what wa expected, but good to know.

spydon:
> Ah! When I was adding this, I was confused why that was there, but not knowing the full router creation origin story, I was like well there had to be a reason. The tests passed for what was expected, but good to know.

Fixed it and added some extra tests that should cover this. :)

packages/flame/lib/src/components/router_component.dart
```diff
@@ -255,9 +239,9 @@ class RouterComponent extends Component {
   }
 
   /// Local method to bypass [pop]'s assert
-  void _popWithoutMinimumAssert() {
+  void _replaceRoute(Route nextRoute) {
```
st-pasha, line 242: `_removeTopRoute`? This method doesn't actually replace anything.

packages/flame/test/components/router_component_test.dart
```diff
@@ -4,6 +4,47 @@ import 'package:flame_test/flame_test.dart';
 import 'package:flutter/widgets.dart' hide Route, OverlayRoute;
 import 'package:flutter_test/flutter_test.dart';
 
+class _TestRoute extends Route {
```
st-pasha, line 7: Could you please move this helper class to the bottom of the file? That way, when you open file you can immediately see `void main()` and can press the green arrow to start the tests.

st-pasha (approved):

packages/flame/lib/src/components/router_component.dart
spydon: Done

packages/flame/test/components/router_component_test.dart
spydon: Done

Issue #2263, opened 2023-01-10
RouterComponent.pushOverlay() causes concurrent modification exception.


pastordougdev:
# Current bug behavior
In the update method of a component in the component tree, I am detecting the game over condition and calling the router's pushOverlay() method to push a game over dialog.  When pushOverlay is called, the remainder of the component tree finishes the update methods for the current game loop tick as is normal, but the router is going ahead with adding the overlay route to the component tree, resulting in a concurrent modification exception.

# Expected behavior
Expected behavior is to be able to push the route, allow the game tick to complete and then the overlay route be rendered on the screen.

# Steps to reproduce
1 - call the router component's pushOverlay method from a component in the tree, such as below:
```dart
@override
  void update(double dt) {
    debugger(when: gameStatus == GameStatus.gameOver);
    if (gameStatus == GameStatus.gameOver) {
      print('game is over');
      if (!_showingOverlay) {
        _showingOverlay = true;
        gameRef.router.pushOverlay('gameOver');
      }
      return;
    }
```

# More environment information

# Log information

```
======== Exception caught by scheduler library =====================================================
The following ConcurrentModificationError was thrown during a scheduler callback:
Concurrent modification during iteration: Instance of 'SplayTreeSet<LinkedHashSet<Component>>'.

When the exception was thrown, this was the stack: 
#0      _SplayTreeIterator.moveNext (dart:collection/splay_tree.dart:763:7)
#1      ExpandIterator.moveNext (dart:_internal/iterable.dart:477:21)
#2      IterableMixin.forEach (dart:collection/iterable.dart:45:23)
#3      Component.updateTree (package:flame/src/components/core/component.dart:479:16)
#4      Component.updateTree.<anonymous closure> (package:flame/src/components/core/component.dart:479:33)
```

# More information

I have not determined a cause or fix yet, but I will be happy to assist with a PR for this issue.

munsterlander:
Have you tried returning: `return super.update(dt);`?

st-pasha:
It's possible that this is caused by `Component.reorderChildren()` which may modify the component tree during the `update()` which traverses the component tree.

munsterlander:
Yes, we have confirmed that is it, but we are looking at how you can invoke this without tripping it, unless you think there is a true bug here in that call?

munsterlander:
[...]

Issue #1851, opened 2022-08-19
ConcurrentModificationError using changeParent


tboyce021:
# Current bug behaviour

Calling `changeParent` throws a `ConcurrentModificationError`.

# Expected behaviour

Calling `changeParent` shouldn't throw an exception.

# Steps to reproduce

Run the following game and wait 5 seconds:
```
import 'package:flame/components.dart';
import 'package:flame/game.dart';
import 'package:flutter/widgets.dart';

void main() {
  runApp(
    GameWidget(
      game: ChangeParentGame(),
    ),
  );
}

class ChangeParentGame extends FlameGame {
  @override
  Future<void>? onLoad() async {
    final childComponent = Component();
    final parentComponent = Component();

    add(childComponent);
    add(parentComponent);

    Future.delayed(
      const Duration(seconds: 5),
      () => childComponent.changeParent(parentComponent),
    );
  }
}
```

# Flutter doctor output

```
[√] Flutter (Channel stable, 3.0.5, on Microsoft Windows [Version 10.0.22000.856], locale en-US)
    • Flutter version 3.0.5 at D:\Users\username\AppData\Local\Programs\flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision f1875d570e (5 weeks ago), 2022-07-13 11:24:16 -0700
    • Engine revision e85ea0e79c
    • Dart version 2.17.6
    • DevTools version 2.12.2

[√] Android toolchain - develop for Android devices (Android SDK version 32.0.0-rc1)
    • Android SDK at D:\Users\username\AppData\Local\Android\sdk
    • Platform android-32, build-tools 32.0.0-rc1
    • Java binary at: D:\Program Files\Android\Android Studio\jre\bin\java
    • Java version OpenJDK Runtime Environment (build 11.0.12+7-b1504.28-7817840)
    • All Android licenses accepted.

[√] Chrome - develop for the web
    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[√] Visual Studio - develop for Windows (Visual Studio Community 2022 17.3.0)
    • Visual Studio at C:\Program Files\Microsoft Visual Studio\2022\Community
    • Visual Studio Community 2022 version 17.3.32804.467
    • Windows 10 SDK version 10.0.19041.0

[√] Android Studio (version 2021.2)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin can be installed from:
[...]

PR #2300, opened 2023-01-28
refactor: Added ComponentTreeRoot

st-pasha:
# Description

* The `ComponentTreeRoot` is a special Component designed to sit at the root of a component tree. For example, `FlameGame` is a `ComponentTreeRoot`.
* The class contains a game-wide queue (`RecycledQueue`) for component lifecycle events. Currently this supports only `Component.add` functionality, but in the future we'll extend it to other functions as well: remove, re-parent, and re-balance.
* The new approach is supposed to be more efficient: instead of each component having their own queues of events, now there is just one central queue, which also supports recycling of elements, which means no unnecessary objects are created on every tick.
* The new approach should be more robust as well: previously, each component's queue was processed during that component's `updateTree` cycle -- now, all lifecycle events are resolved from one location, before the component tree is updated or rendered. If during the update phase the user schedules new components to be added/removed, then those changes will be always deferred to the next game tick. Thus, we have a guarantee that during a single game tick the structure of a component tree will remain constant.
  * This is, of course, pending future refactoring to move the remaining lifecycle handling functionality into the ComponentTreeRoot. 

## Checklist

## Breaking Change?

## Related Issues

Prerequisite for #1851
Prerequisite for #2263


spydon:
Interesting, will review tomorrow!

packages/flame/lib/src/components/core/component.dart
```diff
@@ -806,21 +824,28 @@ class Component {
...
+    _lifecycleManager?.processQueues();
+    _parent!.onChildrenChanged(this, ChildrenChangeType.added);
+  }
+
+  void _reAddChildren() {
     if (_children != null) {
-      _children!.forEach(
-        (child) => child._mount(parent: this, existingChild: true),
-      );
+      final children = List<Component>.from(_children!, growable: false);
```
spydon, line 835: I wonder if this could be possible to do without creating a new list, or possibly have a temporary list variable outside of the method, or maybe there are no use-cases when this method will be called a lot?

packages/flame/lib/src/components/core/component.dart
```diff
@@ -1053,7 +1046,7 @@ class _LifecycleManager {
       oldParent?.onChildrenChanged(child, ChildrenChangeType.removed);
       child._parent = owner;
       child._mount();
-      owner.onChildrenChanged(child, ChildrenChangeType.added);
+      // owner.onChildrenChanged(child, ChildrenChangeType.added);
```
spydon, line 1049: ?

packages/flame/lib/src/components/core/component_tree_root.dart
```diff
@@ -0,0 +1,119 @@
...
+    _queue.addLast()
+      ..kind = _LifecycleEventKind.add
+      ..child = child
+      ..parent = parent;
+  }
+
+  bool get hasLifecycleEvents => _queue.isNotEmpty;
+
+  void processLifecycleEvents() {
+    assert(_blocked.isEmpty);
```
spydon, line 31: Add message to assertion

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -44,6 +45,9 @@ class FlameGame extends Component with Game {
   @override
   Vector2 get size => camera.gameSize;
 
+  @override
+  ComponentTreeRoot get fcsRoot => this;
```
spydon, line 49: Maybe it could just be called `root`, since we don't prefix anything else in the system with `fcs`?

packages/flame/lib/src/game/game.dart
```diff
@@ -209,6 +210,10 @@ abstract class Game {
   /// not be on the flutter widget tree anymore.
   void onDetach() {}
 
+  /// The root component of the component tree, if the game uses a Flame
+  /// Component System, or `null` otherwise.
+  ComponentTreeRoot? get fcsRoot => null;
```
spydon, line 215: A little sad that FCS spills over into `Game`, is it not possible to avoid this?

spydon (commented):
Good stuff! It feels like this will both be easier to comprehend and to do optimizations and other things with. :)

packages/flame/lib/src/components/core/component.dart
```diff
@@ -543,17 +547,23 @@ class Component {
...
     );
-    _parent = parent;
-    parent.lifecycle._children.add(this);
-    if (!isLoaded && (parent.findGame()?.hasLayout ?? false)) {
-      return _startLoading();
+    child._parent = this;
+    final game = findGame();
+    if (isMounted) {
+      assert(game != null);
+      game!.fcsRoot!.enqueueAdd(child, this);
```
luanpotter, line 560: do we need the assert if we are doing `!` anyway?
and if so do we need an assert on fcsRoot too?

packages/flame/lib/src/components/core/component.dart
```diff
@@ -748,10 +759,25 @@ class Component {
     }
   }
 
+  @internal
+  LifecycleEventStatus handleLifecycleEventAdd(Component parent) {
+    assert(!isMounted);
```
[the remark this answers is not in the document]
st-pasha: I believe our current practice is to add assertion messages only in cases when it is a user-caused error. For example, adding a component to two parents at once. In this case the message would tell the user what they did wrong, and how they can fix it.

At the same time, we have a small number of "internal" assertions, which should never trigger as far as we know. If such assertion fails, it would indicate a bug in Flame, some execution flow that we didn't think of. In these cases having no error message helps, because it makes the user go on Discord and start asking questions, and this is how we'd know there is a bug.

This assert falls into the second category. It also serves as a kind of "documentation" stating that I expect this method to be called when the component is not mounted yet.

packages/flame/lib/src/components/core/component.dart
st-pasha: The `_mount()` method already does 
```
_parent!.onChildrenChanged(this, ChildrenChangeType.added);
```
at the end, so this call would create a second notification.

packages/flame/lib/src/game/game.dart
st-pasha: Hmm, I'll try to convert this into an interface.

packages/flame/lib/src/components/core/component.dart
st-pasha: I converted this to use a static temporary list.

packages/flame/lib/src/game/flame_game.dart
st-pasha: My thinking here was like this:
- this is a deeply internal property that will probably not be of interest to the users.
- at the same time, it is common for users to add lots of their own properties into their derived Game classes. As such, if this was simply `root`, there would be a risk of a name clash for some users.

packages/flame/lib/src/components/core/component.dart
st-pasha: good point

packages/flame/lib/src/components/core/component.dart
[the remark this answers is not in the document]
spydon: Alright, I think in some cases we should also add messages to our internal asserts so that it becomes easier for the us to know what 2e have done wrong, but I think these two are so self explanatory that it is not needed.

packages/flame/lib/src/components/core/component.dart
spydon: So it should be removed then? :)

packages/flame/lib/src/game/flame_game.dart
spydon: I think it is better for us to take the cleaner word `root` anyways, even though it might clash with any user definitions, they users can easily add a more descriptive name to their variable too.

It is not that bad here, but I just don't want it to spread so that we start prepending fcs to all kinds of things, so prefer killing the convention in its cradle. 😅

packages/flame/lib/src/components/core/component.dart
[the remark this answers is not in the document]
st-pasha: sure, removed

packages/flame/lib/src/components/core/component.dart
st-pasha: How about changing the name to `componentTreeRoot`? That way there is no "fcs" prefix, and no chance for name clash.

packages/flame/lib/src/game/flame_component_system_game.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:flame/src/components/core/component.dart';
+import 'package:flame/src/components/core/component_tree_root.dart';
+import 'package:meta/meta.dart';
+
+/// An interface which should be implemented by any `Game` subclass that uses
+/// the Flame Component System (i.e. it contains a single Component tree).
+@internal
+abstract class FlameComponentSystemGame {
```
spydon, line 8: Pretty much everything in `FlameGame` is FCS + `Game`, do we really need this then?
If someone wants to implement their own `FCS` they can just do `implements FlameGame`?

packages/flame/lib/src/components/core/component.dart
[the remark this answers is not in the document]
spydon: I think you misunderstood me, I meant that the asserts are good and self explanatory, so no message is needed (but the asserts themselves can stay. :sweat_smile:

packages/flame/lib/src/game/flame_component_system_game.dart
st-pasha: We'd need to have this for the `BareFlameGame` class, so that it can share the interface with `FlameGame`, but without inheriting any unnecessary stuff.

packages/flame/lib/src/components/core/component.dart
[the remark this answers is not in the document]
st-pasha: Lol, I meant to comment this for line 1056; I guess I accidentally posted in a wrong thread :)

packages/flame/lib/src/game/flame_component_system_game.dart
spydon: I don't think we should consider `BareFlameGame` before it exists, it may very well never come into existence and we deprecate the camera from `FlameGame` before. And if it does come into existence it is better to add this interface on that PR.

st-pasha: Ok, I removed this class for now.

spydon (approved):

packages/flame/lib/src/components/core/component.dart
```diff
@@ -806,21 +823,40 @@ class Component {
...
+  /// At the end of mounting, we remove all children components and then re-add
+  /// them one-by-one. The reason for this is that before the current component
+  /// was mounted, its [children] may have contained components in arbitrary
+  /// state -- initial, loading, unmounted, etc. However, we don't want to
+  /// have such components in a component tree. By removing and then re-adding
+  /// them, we ensure that they are placed in a queue, and will only be placed
+  /// into [children] once they are fully mounted.
+  void _reAddChildren() {
+    if (_children != null && _children!.isNotEmpty) {
+      assert(_tmpChildren.isEmpty);
```
luanpotter, line 844: is this to avoid some race condition?

packages/flame/lib/src/components/core/component_tree_root.dart
```diff
@@ -0,0 +1,119 @@
...
+
+  /// The event was fully processed and can now be removed from the queue.
+  done,
+}
+
+enum _LifecycleEventKind {
+  unknown,
+  add,
+  remove,
+  // move,
```
erickzanardo, line 99: Should those be removed? Or have a TODO in them?

st-pasha: These are just to show that the enum will contain more values in the near future.

packages/flame/lib/src/components/core/component.dart
st-pasha: This is a precondition: I expect `_tmpChildren` to be empty at this point (and if not, the logic would fail). It shouldn't be possible to have this list non-empty here, because I'm clearing it after every use, on line 851. Still, if someone were to add a new method that also uses `_tmpChildren`, leaving it non-empty, then this assert would alert of danger.

packages/flame/lib/src/components/core/component_tree_root.dart
erickzanardo: I would say to remove them now and put it back when they get implemented for real.

This comments without context don't add much to the code, but rather creates confusion if they are really needed or not.

erickzanardo (commented):
Just wandering about the name, not sure I liked `ComponentTreeRoot`, I think it breaks the naming convention with the other components like `PositionComponent`, `SpriteComponent`, etc.

Probably making the name in reverse? `TreeRootComponent` or even just `RootComponent`. Wdyt?

Other than that, LGTM

packages/flame/lib/src/components/core/component_tree_root.dart
st-pasha: Ok, I've removed them.

st-pasha:
> Just wandering about the name, not sure I liked ComponentTreeRoot, I think it breaks the naming convention with the other components like PositionComponent, SpriteComponent, etc. Probably making the name in reverse? TreeRootComponent or even just RootComponent. Wdyt?

The idea behind the name `ComponentTreeRoot` is that it means "the root of the component tree". If we were to follow the convention, then this would become `ComponentTreeRootComponent`, which kinda sounds silly. At the same time, the `TreeRootComponent` is not specific enough: this is a component which is a root of a tree, but what kind of tree?

So, I think it might be ok to leave the name as-is, ignoring the convention, especially since this is largely an internal class, so the users wouldn't get confused by the name because they wouldn't be using this class directly.

erickzanardo (approved):

Issue #1457, opened 2022-03-16
Component.priority discussion


st-pasha:
This is a discussion for the evolution of the component's `.priority` system.

### Current situation

Every `Component` has the `int priority` property, which determines that component's sorting order within its parent's `children`. This approach works for many games, but for some becomes a hindrance, or even necessitates creation of a parallel "priority-like" system.

### Ideal future state

The "priority" mechanism of the `Component` is replaceable and highly customizable, just like everything else in the Flame engine. Maybe it becomes a mixin, or in some other way.

The following priority systems should be provided:
  - no priority -- in some simplest games the priority system may not be needed at all;
  - `int priority` -- same as the current system;
  - `double priority` -- similar to the current approach, but uses floating-point priorities. This is useful in cases when the "priority" is just the distance to the camera, in which case it makes sense to have this distance to be `double`;
  - computed priority -- the order of elements changes dynamically on every game tick. This is useful for games with isometric projection, where priority is just a function of (x, y) world coordinates of the component;
  - range priority -- used in games where the components are "fat" in z-direction: instead of having a fixed distance to the camera, the components may contain sub-parts at different distances from the camera;
  - user-defined priority -- users should be able to implement their own priority mechanism if they wish so.

### How to get there

The specifics of the design of the component priority system **need to be discussed**. In particular, we would want to minimize the amount of breaking changes for our users (though understandably this will be breaking), and if possible go through the deprecation cycles.

One aspect of the transition, which we can implement right away, is to **deprecate** the `priority` argument in the `Component` and derived classes, and instead recommend using the `priority` setter, which has been implemented recently. Thus, we would recommend users to replace
```dart
MyComponent(priority: 7)
```
with
```dart
MyComponent()..priority = 7;
```

spydon:
Another thing that we have noticed that is troubling many that is related to this discussion is the need for a priority system where you can globally prioritize components (possibly as another priority system). For example if you have two children (`a` and `b`) with different parents, then `a` should be able to be rendered in front of `b` without having to render `a`'s parent in front of `b`'s parent.

I can't think of a way to easily accommodate for such a change with the current tree structure in FCS currently though.

raulmabe:
Hello! I am new to Flame and I was trying to set the priority of my components in order to achieve what you said @spydon.
I thought the priority system was broken until I saw this discussion and read the following line:

> Every Component has the int priority property, which determines that component's sorting order within its parent's children

[...]

PR #2303, opened 2023-01-28
feat: Add replaceChild and replaceAll methods [WIP]

luanpotter:
# Description

This adds two utility methods, `replaceChild` and `replaceAll`, as per suggested:
image

There are some points to consider about the implementation:

* should we wait for the removal? remove methods are not awaitable currently
* what if we are trying to replace something with itself?
* what if the removal portion fails (component doesn't exist as child, etc)?
* what if the addition portion fails (component already child of someone else, etc)

@erickzanardo what do you think of the questions above, since you had the idea originally?

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame/lib/src/components/core/component.dart
+++ b/packages/flame/lib/src/components/core/component.dart
@@ -619,6 +619,24 @@ class Component {
     parent = newParent;
   }
 
+  /// Replaces a [replaceFrom] child of this component with [replaceWith].
+  /// I.e., removes [replaceFrom], and adds [replaceWith] instead.
+  FutureOr<void> replaceChild(Component replaceFrom, Component replaceWith) {
+    if (replaceFrom == replaceWith) {
+      // nothing to do
+      return Future.value();
+    }
+    remove(replaceFrom);
+    return add(replaceWith);
+  }
+
+  /// Same as [replaceChild], but takes a map of replacements.
+  /// Removes the key and adds the value instead.
+  FutureOr<void> replaceAll(Map<Component, Component> replacements) {
+    removeAll(replacements.keys);
+    return addAll(replacements.values);
+  }
+
   //#endregion
 
   //#region Hit Testing
--- a/packages/flame/test/components/component_test.dart
+++ b/packages/flame/test/components/component_test.dart
@@ -652,6 +652,55 @@ void main() {
       );
     });
 
+    group('Replacing components', () {
+      testWithFlameGame('replaces single child', (game) async {
+        final child1 = Component();
+        final child2 = Component();
+
+        final parent = Component();
+        await game.ensureAdd(parent);
+        expect(parent.isMounted, true);
+
+        await parent.add(child1);
+        game.update(0); // children are only updated on the next tick
+        expect(parent.contains(child1), true);
+        expect(parent.contains(child2), false);
+
+        parent.replaceChild(child1, child2);
+        game.update(0); // children are only updated on the next tick
+        expect(parent.contains(child1), false);
+        expect(parent.contains(child2), true);
+      });
+
+      testWithFlameGame('replaces a map', (game) async {
+        final child1 = Component();
+        final child2 = Component();
+        final child3 = Component();
+        final child4 = Component();
+        final child5 = Component();
+
+        final parent = Component();
+        await game.ensureAdd(parent);
+        expect(parent.isMounted, true);
+
+        await parent.addAll([child1, child2, child3]);
+        game.update(0); // children are only updated on the next tick
+        expect(parent.contains(child1), true);
+        expect(parent.contains(child2), true);
+        expect(parent.contains(child3), true);
+        expect(parent.contains(child4), false);
+        expect(parent.contains(child5), false);
+
+        parent.replaceAll({child1: child5, child3: child4});
+        game.update(0); // children are only updated on the next tick
+        expect(parent.contains(child1), false);
+        expect(parent.contains(child2), true);
+        expect(parent.contains(child3), false);
+        expect(parent.contains(child4), true);
+        expect(parent.contains(child5), true);
+      });
+    });
+
     group('descendants()', () {
       testWithFlameGame(
         'descendants in a deep component tree',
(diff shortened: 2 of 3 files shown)


packages/flame/test/components/component_test.dart
```diff
@@ -652,6 +652,55 @@ void main() {
       );
     });
 
+    group('Replacing components', () {
```
luanpotter, line 655: since we are reusing the remove and add methods, other side-effects and consequences and lifecycle statuses are already tested individually

packages/flame/lib/src/components/core/component.dart
```diff
@@ -619,6 +619,24 @@ class Component {
     parent = newParent;
   }
 
+  /// Replaces a [replaceFrom] child of this component with [replaceWith].
+  /// I.e., removes [replaceFrom], and adds [replaceWith] instead.
+  FutureOr<void> replaceChild(Component replaceFrom, Component replaceWith) {
+    if (replaceFrom == replaceWith) {
+      // nothing to do
+      return Future.value();
+    }
```
luanpotter, line 628: could just allow it to trigger any lifecycle hooks
or could throw an error

packages/flame/lib/src/components/core/component.dart
```diff
@@ -619,6 +619,24 @@ class Component {
...
+    remove(replaceFrom);
+    return add(replaceWith);
+  }
+
+  /// Same as [replaceChild], but takes a map of replacements.
+  /// Removes the key and adds the value instead.
+  FutureOr<void> replaceAll(Map<Component, Component> replacements) {
```
luanpotter, line 635: the map could contain things like:
* replacements with itself
* replacement cycles (a -> b -> c)
* replacements to the same value
is it worth validating such cases, and if so, what should we do?

packages/flame/lib/src/components/core/component.dart
st-pasha: If there is nothing to do, you can simply `return;`

luanpotter: compiler wasn't happy; let me try again

st-pasha:
I believe the `replace()` method is a much needed functionality, I believe I've seen several cases in the past where it might have proved really useful.

However, a simple `..remove(child)..add(newChild)` combo is too trivial. A much more useful functionality would be if the `replace()` method could replace a component without losing its order within the parent's list of children. That is, an actual replacement.

It would be more challenging, of course, since we can't really rely on `priority` field here. But it should be doable, probably needing a new method within the `OrderedSet`.

As for `replaceAll()` method -- its name suggests that it replaces all components within the parent. So why not make it like that? `replaceAll(Iterable<Component> newChildren)` would simply remove all current children and add new ones. This would be simpler to implement than the `replace()`.


> should we wait for the removal? remove methods are not awaitable currently

I'd say the `replace()` method should not be async at all, for several reasons:
- this would create an impression that the returned future resolves once the new component has fully replaced the old one, which isn't true;
- returning such future is possible, it's just `newChild.mounted` -- but in most cases it won't be needed, and in those cases where users might want it, they can just access the `.mounted` future directly.

It is worth pointing out that the future returned from `Component.add()` also suffers from a similar problem: it creates an impression that the future resolves once the component was added, which is not true. Instead, `.add()` returns a future that resolves when the component finishes loading, and even that only sometimes.


> what if we are trying to replace something with itself?

This should probably be a no-op.

> what if the removal portion fails (component doesn't exist as child, etc)?
> what if the addition portion fails (component already child of someone else, etc)

This should produce assertion errors, similarly to how `.remove()` and `.add()` work right now.

luanpotter:
Those are good points @st-pasha, thanks for the thorough review

for the return type: agreed. I will change to void
for the replaceAll: agreed. we can have the Map<to, from> version too and call it something else if @erickzanardo feels passionate about that signature

about the "keeping the order": I agree it would be nice as a user, and I agree it is hard but doable, but I have fears that it might actually cause more confusion than it's worth.
while we could add a "bypass" to ordered_set, it would be kept in a "wrong" state. if anything re-set it, the ordering would suddenly shift.
also it would be weird for unaware users that might for example try to setup a specific priority and it "does't work" because some elements were replaced and had "phantom" priorities.

I think there is a question to be had about getting completely rid of priority, but mixing both priority + a bypass seems more confusing that its worth. wdyt? and also everyone else

st-pasha:
> it would be kept in a "wrong" state. if anything re-set it, the ordering would suddenly shift.

Ah, that's true. We could make a distinction here that when B replaces A and they have the same priority, then B is put exactly where A was. However, if their priorities are different, then we perform the regular `add()` and let the `ComponentSet` deal with the ordering. We could even have a special flag `replace(A, B, {bool copyPriority = false})` which would copy A's priority to B.

There are many cases where the user doesn't explicitly set the priority of components and thus they all remain at 0. At the same time, the relative order of those components still matters, and `replace` should honor it.

> I think there is a question to be had about getting completely rid of priority, but mixing both priority + a bypass seems more confusing that its worth. wdyt?

We had a discussion about this some time ago (#1457), but I don't think we figured out what to do there.

erickzanardo:
> for the replaceAll: agreed. we can have the Map<to, from> version too and call it something else if @erickzanardo feels passionate about that signature

That is a fair point, I agree that `replaceAll` would replace all the current components to the received set.

About having a version that receives a `Map<to, from>`, we could name it `replaceWhere`, wdyt?

spydon:
> About having a version that receives a `Map<to, from>`, we could name it `replaceWhere`, wdyt?

`replaceWhere` sounds more like it should replace if it fulfills some precondition.
Maybe this isn't needed? It would be pretty easy for the user to just iterate over that map and run replace instead?

erickzanardo:
I agree and disagree :P

> `replaceWhere` sounds more like it should replace if it fulfills some precondition. 

Exactly, the map that you pass is they one that sets that condition right, it is a one to one condition clause, but a clause nevertherless?

> Maybe this isn't needed? It would be pretty easy for the user to just iterate over that map and run `replace` instead?

I agree with that hahahha

luanpotter:
Going to close this for now since we never really reached a conclusion and I never followed up. I will open an issue and link this discussion, if anyone is interested to take this on.

PR #2307, opened 2023-01-30
fix: Camera no longer "sticks" to boundary with BoundedPositionBehavior

### LINKED ISSUE #1905: Camera should be able to slide along the world boundary
# What could be improved

As discussed in #1902, whenever the target moves outside of camera bounds, the camera would "stick" to the boundary and remain there until the target removes into the allowed region. A more natural behavior would be for the camera to "slide" along the camera boundary, so that it always remains as close as possible to the target within the allowed region.

# Why should this be improved
Currently it is possible for the target to leave the camera boundary, and then travel along the boundary sufficiently far away that the camera would stop seeing the target -- even if the camera can be placed closer to the target to be able to see it, while still remaining within the allowed region.

# Any risks?
The implementation needs to be smart enough to allow world boundaries of arbitrary shape.

st-pasha:
# Description

Previously, when using BoundedPositionBehavior, the camera would "stick" to the world boundary as soon as the target moves outside. With this PR, the camera will now try to keep as close to the target as possible, even if it moves outside of the world boundary. Thus, the camera will now behave as if the boundary became "slippery".

Also added method `nearestPoint()` in class `Shape`.

## Checklist

## Breaking Change?

## Related Issues

Closes #1905

### DIFF
--- a/packages/flame/lib/src/experimental/geometry/shapes/circle.dart
+++ b/packages/flame/lib/src/experimental/geometry/shapes/circle.dart
@@ -83,5 +83,15 @@ class Circle extends Shape {
   }
 
   @override
+  Vector2 nearestPoint(Vector2 point) {
+    if (_radius == 0) {
+      return _center;
+    }
+    final vectorToPoint = point - _center;
+    final distance = vectorToPoint.length;
+    return _center + vectorToPoint * (_radius/distance);
+  }
+
+  @override
   String toString() => 'Circle([${_center.x}, ${_center.y}], $_radius)';
 }
--- a/packages/flame/lib/src/experimental/geometry/shapes/circle.dart
+++ b/packages/flame/lib/src/experimental/geometry/shapes/circle.dart
@@ -89,7 +89,7 @@ class Circle extends Shape {
     }
     final vectorToPoint = point - _center;
     final distance = vectorToPoint.length;
-    return _center + vectorToPoint * (_radius/distance);
+    return _center + vectorToPoint * (_radius / distance);
   }
 
   @override
--- a/packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
+++ b/packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
@@ -217,7 +217,21 @@ class Polygon extends Shape {
 
   @override
   Vector2 nearestPoint(Vector2 point) {
-    throw UnimplementedError();
+    final result = Vector2.zero();
+    var shortestDistance2 = double.infinity;
+    for (var i = 0; i < _vertices.length; i++) {
+      final vertex = _vertices[i];
+      final edge = _edges[i];
+      final dotProduct = (point - vertex).dot(edge);
+      final t = (dotProduct / edge.length2).clamp(-1.0, 0.0);
+      final edgePoint = vertex + edge * t;
+      final distance2 = (edgePoint - point).length2;
+      if (distance2 < shortestDistance2) {
+        shortestDistance2 = distance2;
+        result.setFrom(edgePoint);
+      }
+    }
+    return result;
   }
 
   @override
--- a/packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
+++ b/packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
@@ -216,5 +216,10 @@ class Polygon extends Shape {
   }
 
   @override
+  Vector2 nearestPoint(Vector2 point) {
+    throw UnimplementedError();
+  }
+
+  @override
   String toString() => 'Polygon($vertices)';
 }
--- a/packages/flame/lib/src/experimental/geometry/shapes/rectangle.dart
+++ b/packages/flame/lib/src/experimental/geometry/shapes/rectangle.dart
@@ -124,5 +124,12 @@ class Rectangle extends Shape {
   }
 
   @override
+  Vector2 nearestPoint(Vector2 point) {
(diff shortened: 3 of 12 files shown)


packages/flame/lib/src/experimental/geometry/shapes/circle.dart
```diff
@@ -82,6 +82,16 @@ class Circle extends Shape {
...
   }
 
+  @override
+  Vector2 nearestPoint(Vector2 point) {
+    if (_radius == 0) {
+      return _center;
+    }
+    final vectorToPoint = point - _center;
+    final distance = vectorToPoint.length;
+    return _center + vectorToPoint * (_radius / distance);
```
spydon, line 92: This creates a lot of new `Vector2` objects, you should be able to create just one and use `setFrom` and then `sub` etc instead of the operators.

spydon: Could `length2` be used here instead? 🤔

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
```diff
@@ -215,6 +215,25 @@ class Polygon extends Shape {
...
+  @override
+  Vector2 nearestPoint(Vector2 point) {
+    final result = Vector2.zero();
+    var shortestDistance2 = double.infinity;
+    for (var i = 0; i < _vertices.length; i++) {
+      final vertex = _vertices[i];
+      final edge = _edges[i];
+      final dotProduct = (point - vertex).dot(edge);
+      final t = (dotProduct / edge.length2).clamp(-1.0, 0.0);
+      final edgePoint = vertex + edge * t;
```
spydon, line 227: This also creates new objects

packages/flame/lib/src/experimental/geometry/shapes/rectangle.dart
```diff
@@ -123,6 +123,13 @@ class Rectangle extends Shape {
     return Vector2(vx, vy);
   }
 
+  @override
+  Vector2 nearestPoint(Vector2 point) {
+    final xx = (point.x).clamp(_left, _right);
+    final yy = (point.y).clamp(_top, _bottom);
+    return Vector2(xx, yy);
```
spydon, line 130: We have a `Vector2.clamp` method, would it simplify this slightly maybe?

packages/flame/lib/src/experimental/geometry/shapes/rounded_rectangle.dart
```diff
@@ -171,6 +171,24 @@ class RoundedRectangle extends Shape {
...
+    if (containsPoint(result)) {
+      return result;
+    }
+    final center = result
+      ..setValues(
+        result.x <= _left + _radius ? _left + _radius : _right - _radius,
+        result.y <= _top + _radius ? _top + _radius : _bottom - _radius,
+      );
+    final vectorToPoint = point - center;
+    final distance = vectorToPoint.length;
```
spydon, line 188: It should be possible to use `distanceTo` here right?

spydon: Ah right, `vectorToPoint` is used later

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
st-pasha: (also refactored)

packages/flame/lib/src/experimental/geometry/shapes/rectangle.dart
st-pasha: not really, because `Vector2.clamp` takes 2 vectors as arguments, so we'd need to create those vectors first...

packages/flame/lib/src/experimental/geometry/shapes/rounded_rectangle.dart
st-pasha: (also refactored to use static temporary vectors)

packages/flame/lib/src/experimental/geometry/shapes/circle.dart
```diff
@@ -82,6 +82,20 @@ class Circle extends Shape {
...
   }
 
+  static final Vector2 _tmpResult = Vector2.zero();
+
+  @override
+  Vector2 nearestPoint(Vector2 point) {
+    if (_radius == 0) {
+      return _center;
+    }
+    return _tmpResult
```
spydon, line 92: I wonder if we don't want to at least return a new vector? Could be quite strange behavior if the user saves this away and then runs this function again for another circle and then both the vectors are the result of the last one, or well they are the same object. I know we have this pattern at some places, but not sure if it is good... At the very least we should warn the user in the dartdoc that they should clone the vector if they intend to save it.

st-pasha: Yeah, I have a note about it in the doc-comment.

We kinda don't have any consistency in how we approach this. There are methods that return new vectors, there are those that return existing vectors, and then those that sometimes do one thing and other times another.

It might be useful to do some benchmarking and figure out what's the actual cost of creating new vectors, and how would that compare to a "hand-crafted" Vector2 class which simply contain two doubles `x`, `y` as fields (thus, a bit like mutable `Offset`).

For now, though, we can go with this API, since it's easier to loosen up in the future if need be. Going in the other direction instead (i.e. we returned new objects, but then switched to reused objects) would be a breaking change.

spydon (approved):

PR #2310, opened 2023-02-01
fix:  Prevent atlas from overflowing the size limit of the device.

### LINKED ISSUE #2042: Performance & rendering issue came out after #1982
# Current bug behaviour
After bc2db2b17af997fd529800abe0124b0f8918295f, tiled map performance & rendering issue came out.
FPS drops to 0~10s and it renders inappropriate tile.

- On Android you can see the video below

### Before bc2db2b17af997fd529800abe0124b0f8918295f

https://user-images.githubusercontent.com/38072762/194514541-f87f3039-a1d3-4b01-a1a7-6dfb9a577a64.mp4

### After bc2db2b17af997fd529800abe0124b0f8918295f

https://user-images.githubusercontent.com/38072762/194514568-d22cb2bf-a890-4613-961f-70b642242430.mp4

- On IOS, it immediately crashes with following logs
```dart
libc++abi: terminating with uncaught exception of type std::bad_alloc: std::bad_alloc
* thread #10, name = 'io.flutter.1.raster', stop reason = signal SIGABRT
    frame #0: 0x00000001b7375b38 libsystem_kernel.dylib`__pthread_kill + 8
libsystem_kernel.dylib`__pthread_kill:
->  0x1b7375b38 <+8>:  b.lo   0x1b7375b58               ; <+40>
    0x1b7375b3c <+12>: pacibsp
    0x1b7375b40 <+16>: stp    x29, x30, [sp, #-0x10]!
    0x1b7375b44 <+20>: mov    x29, sp
Target 0: (Runner) stopped.
Lost connection to device.
```

# Expected behaviour
Renders right tile and fixes the performance.

# Flutter doctor output

```
Output of: flutter doctor -v
          
[✓] Flutter (Channel stable, 3.3.4, on macOS 12.6 21G115 darwin-arm, locale ko-KR)
    • Flutter version 3.3.4 on channel stable at /Users/lafity101/flutter
    • Framework revision eb6d86ee27 (2 days ago), 2022-10-04 22:31:45 -0700
    • Engine revision c08d7d5efc
    • Dart version 2.18.2
    • DevTools version 2.15.0

    • Platform android-33, build-tools 32.0.0
    • Java binary at: /Users/lafity101/Library/Application
      Support/JetBrains/Toolbox/apps/AndroidStudio/ch-0/212.5712.43.2112.8815526/Android
      Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.12+0-b1504.28-7817840)

    • Build 13F100
    • CocoaPods version 1.11.3

[✓] Android Studio (version 2021.2)
    • Android Studio at /Users/lafity101/Library/Application
      Support/JetBrains/Toolbox/apps/AndroidStudio/ch-0/212.5712.43.2112.8609683/Android Studio.app/Contents
    • Java version OpenJDK Runtime Environment (build 11.0.12+0-b1504.28-7817840)

[✓] Android Studio (version 2021.2)
    • Android Studio at /Users/lafity101/Library/Application
      Support/JetBrains/Toolbox/apps/AndroidStudio/ch-0/212.5712.43.2112.8815526/Android Studio.app/Contents
    • Java version OpenJDK Runtime Environment (build 11.0.12+0-b1504.28-7817840)

[✓] IntelliJ IDEA Ultimate Edition (version 2022.2.1)
    • IntelliJ at /Users/lafity101/Applications/JetBrains Toolbox/IntelliJ IDEA Ultimate.app

[✓] IntelliJ IDEA Ultimate Edition (version 2022.2.1)
    • IntelliJ at /Users/lafity101/Library/Application
      Support/JetBrains/Toolbox/apps/IDEA-U/ch-0/222.3739.54/IntelliJ IDEA.app

[✓] IntelliJ IDEA Ultimate Edition (version 2022.2)
    • IntelliJ at /Users/lafity101/Library/Application
      Support/JetBrains/Toolbox/apps/IDEA-U/ch-0/222.3345.118/IntelliJ IDEA.app

[✓] VS Code (version 1.71.2)
    • Flutter extension version 3.50.0

[✓] Connected device (4 available)
    • SM A325N (mobile) • RF9R40690PY               • android-arm64  • Android 11 (API 30)
    • iPad (mobile)     • 00008030-00022C403E63C02E • ios            • iOS 15.7 19H12
    • macOS (desktop)   • macos                     • darwin-arm64   • macOS 12.6 21G115 darwin-arm
    • Chrome (web)      • chrome                    • web-javascript • Google Chrome 105.0.5195.125

[✓] HTTP Host Availability

• No issues found!
```

Hwan-seok:
# Description
Devices have their own limits of texture size that are available by the graphical engine.
And the `RectangleBinPacker` is working great generating an atlas size of the device limit like 4k or 8k.

But when flip exists, `SpriteBatch` copies the atlas and doubles up its size. 
As a result, the atlas becomes bigger than the device limit.

After #1982, the individual atlas became bigger and its count decreased by composing all of the images into one texture.
It reduces lots of operations but flipped atlas made its regression.

## Checklist

## Breaking Change?

## Related Issues
Fixes #2042 


Hwan-seok:
Changed to draft to fix tests.

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -162,7 +187,11 @@ class SpriteBatch {
   }
 
   /// The atlas used by the [SpriteBatch].
-  Image atlas;
+  final Image atlas;
+
+  /// The horizontally flipped atlas generated by the [atlas].
+  /// This is available when there is flipped transform added by [addTransform].
+  Image? flippedAtlas;
```
spydon, line 194: Can't there be two sprite batch objects instead of duplicating all the fields in here?

Hwan-seok: That's a good idea but it narrows the use cases to tiled only and needs to remove `flip` from `addTransform` because the `SpriteBatch` does not have ability to render flipped `BatchItem`. 
https://github.com/flame-engine/flame/blob/e5345e62d0f1cdf96669141f5e22fe7bf7224a39/packages/flame/lib/src/sprite_batch.dart#L253-L258

Hwan-seok: @spydon 
I am going to switch implementation based on your opinion.

It would have a break changes to users who are using `SpriteBatch` directly with flipped transform.
1. No more `flip` property on `SpriteBatch.addTransform()`
2. No more rendering via drawImageRect before flipped atlas is ready because the atlas should be made before construction.

I agree with creating another SpriteBatch instead of internally supporting it by duplicating lots of properties. And IMHO, there are few users who use `SpriteBatch` directly with a flip.

Does breaking those changes are acceptable?

spydon: Sounds good to me!

Hwan-seok:
So sadly, I just realize that `drawAtlas` should be performed by one operation because of Isometric, staggered maps(as mentioned in #1982). Those maps are important to follow in rendering order. Dividing into two `drawAtlas` breaks rendering results as the following image.

pointy_hex_even_testImage


Back to square one(except for personal gain of knowledge), it will be very helpful to discuss about this.

As the max texture size is not public information, it is not good to prevent creating big texture maps strictly.
So textures should be made as smaller as possible. But when it comes to the web, it is way too small like 2k with flips.

So I am planning a few things to solve this.

1. Create texture to layer one by one instead of creating a big texture across layers.
 - Currently, The texture is copied into layers. But if the texture is created individually by layer, there are chances to optimize the texture size by dividing layers by tilesets.
2. Opt-out flipping behavior.
 - As it's inevitable to create an atlas that has 2x width from original one to support flip, it could bring some limitations. So it would be great to out-out the flip. The software `tiled` has shortcuts to go back by `cmd(ctrl)-z`, rotating by `z`, and flipping horizontally by `x`. Imagine that you clicked a tile to the wrong place, you may press `cmd-z` and so on. In this situation, you can also accidentally press `z` or `x`, and as a result, the tile flips! It is fine with some complicated or flip(rotate)-consiousable tiles. But some things like the floor or sky are hard to recognize. I even made a program for my design team to pick out coordinates of where the flipped tiles exist but always there are flips in every release😂. So, providing a way by simple boolean argument may prevent this.
3. Write docs to warn flame users that big tilesets can break some rendering.
4. Make some logging systems that warn flame users about what went wrong. Or throw errors strictly when needed.
 - This is a minor and personal thought or aspect. While I am debugging this issue, I encountered some methods that silently returned without warning or errors even though it goes wrong. Most of the methods are throws errors but a few are not. It would be helpful if those throw the error. But maybe this is mostly related to my knowledge I think😅 .

I will close this pr tomorrow as unfixable and these plans will be open as fresh issues or pr.

Any thoughts or advice are welcome.

spydon:
Thanks for the write-up of your conclusions @Hwan-seok!

PR #2314, opened 2023-02-02
feat: Add network assets package.

erickzanardo:
# Description

Adds some documentation and a package that can be used to fetch assets from the network.

## Checklist

## Breaking Change?

## Related Issues


packages/flame_network_image/pubspec.yaml
```diff
@@ -0,0 +1,27 @@
...
+  sdk: '>=2.18.0 <3.0.0'
+  flutter: ">=1.17.0"
+
+dependencies:
+  dev: ^1.0.0
+  flame: ^1.6.0
+  flutter:
+    sdk: flutter
+  http: ^0.13.5
+  mocktail: ^0.3.0
```
Hwan-seok, line 19: Maybe accidentally added to `dependencies` instead of `dev_dependencies`?

doc/flame/rendering/images.md
```diff
@@ -119,6 +119,45 @@ class MyGame extends Game {
 }
 ```
 
+## Loading images from the network
+
+Flame core package doesn't offer a built in method to loading images from the network.
+
+The reason for that is that Flutter/Dart does not have a built in http client, which required
+a package to be used and since there are a couple of packages available out there, we refrain
+from forcing, on even pushing the user in a specific package.
```
st-pasha, line 128: Isn't `http` the "official" package? Its publisher is `dart.dev`, and it's described as
> Official core packages published by the Dart Team; these are foundational packages that complement the core libraries.

packages/flame_network_image/example/analysis_options.yaml
```diff
@@ -0,0 +1,29 @@
+# This file configures the analyzer, which statically analyzes Dart code to
+# check for errors, warnings, and lints.
+#
+# The issues identified by the analyzer are surfaced in the UI of Dart-enabled
+# IDEs (https://dart.dev/tools#ides-and-editors). The analyzer can also be
+# invoked from the command line by running `flutter analyze`.
+
+# The following line activates a set of recommended lints for Flutter apps,
+# packages, and plugins designed to encourage good coding practices.
+include: package:flutter_lints/flutter.yaml
```
st-pasha, line 10: All lines except this one can be removed.

packages/flame_network_image/lib/src/flame_network_image.dart
```diff
@@ -0,0 +1,189 @@
...
+/// Another important note about the cache layers is that the first layer, is a
+/// per instance cache, while the local storage is a app global cache. This
+/// means that two different [FlameImageResponse] instances will have the same
+/// local storage cache, but not the memory cache.
+///
+/// Note that the local storage layer is not present when running on web since
+/// that platform doesn't really have a file system. The Browse caching will
+/// work as similar replacement for this layer, though that can't be controlled
+/// this package, make sure that the server where the images are being fetched
+/// returns the correct cache header to make the browser cache the assets.
```
st-pasha, line 72: All modern browsers have `localStorage`, aka Web Storage (https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage), which can be used to persist files indefinitely, or at least until the user manually purges the cache.

st-pasha (commented):
I feel like this could be a great package for fetching files from the Internet and then caching them locally -- regardless of the type of the file. Currently it focuses on images, but what about audio files, fire atlas bundles, tiled levels, yarn scripts, custom json, etc?

The task of converting a file into an `Image` is quite trivial -- and we can add a helper function `fromBytes` into the core Flame `Images` class:
```
Future<Image> fromBytes(String key, Uint8List bytes) {
    return (_assets[key] ??= _ImageAsset.future(decodeImageFromList(bytes)))
        .retrieveAsync();
  }
```

packages/flame_network_image/example/analysis_options.yaml
spydon: This should also be changed to the flame linter instead of flutter_lints

doc/flame/rendering/images.md
erickzanardo: It is, that is why that is the one that I choose for the default implementation on the package!

But it is not built in in flutter.

packages/flame_network_image/lib/src/flame_network_image.dart
erickzanardo: Yeah, I considered it, but I don't believe it is a good fit for what we are building here.

One of the biggest reason for that is that [`setItem`](https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem) values are strings, so we will need to serialize the assets into a base64 for example, something that can increase the size of the file when stored, and probably add some small CPU overhead.

In addition to that, local storage has a very limited storage capacity, I haven't any reference in the web with the correct value, but I remember from my old days as a JS developer that it was around 5mb.

So by relying on the browser cache, we have more storage capacity, code simplicity (since we don't need to write any JS interop code).

erickzanardo:
> I feel like this could be a great package for fetching files from the Internet and then caching them locally -- regardless of the type of the file. Currently it focuses on images, but what about audio files, fire atlas bundles, tiled levels, yarn scripts, custom json, etc?
> 
> The task of converting a file into an `Image` is quite trivial -- and we can add a helper function `fromBytes` into the core Flame `Images` class:
> 
> ```
> Future<Image> fromBytes(String key, Uint8List bytes) {
>     return (_assets[key] ??= _ImageAsset.future(decodeImageFromList(bytes)))
>         .retrieveAsync();
>   }
> ```

That is a good idea tbh, I don't think we need to add a fromBytes helper on Flame since the `decodeImageFromList` is just simple call :)

How should we call this then? `flame_network_assets`?

packages/flame_network_image/lib/src/flame_network_image.dart
st-pasha: Ah, you're right, it's only 5Mb. Though there's also [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) which has no size limit, but lasts only while the corresponding browser tab is open.

As for the string format - you're right again! So, some kind of encoding will be necessary. The base64 format uses 6 bits per byte, which means its overhead is 33%. However, the utf16 encoding (used by localStorage) has 63488 valid code points per 2-bytes, which means that theoretically there could be an encoding of bytes to utf16 with overhead of only 0.28%! But is there such a standard? Would we need to invent one?

However, all this is probably unnecessary for the initial version of the package, and perhaps never.

st-pasha:
> How should we call this then? flame_network_assets?

Some more alternatives: `flame_network_cache`, `flame_http_cache`.

packages/flame_network_image/LICENSE
```diff
@@ -0,0 +1 @@
+TODO: Add your license here.
```
st-pasha, line 1: Don't forget to add a license here before merging!

packages/flame_networks_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,223 @@
...
+
+  /// Flag indicating if files will be cached in memory.
+  final bool cacheInMemory;
+
+  /// Flag indicating if files will be cached in the local storage.
+  final bool cacheInStorage;
+
+  final bool _isWeb;
+
+  late final _memoryCache = MemoryCache<String, T>();
```
st-pasha, line 127: This doesn't have to be `late` either

packages/flame_networks_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,223 @@
+import 'dart:async';
```
st-pasha, line 1: The directory name has a typo: `flame_networks_assets` -> `flame_network_assets`

packages/flame_networks_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,223 @@
...
+import 'package:flutter/rendering.dart';
+import 'package:http/http.dart' as http;
+import 'package:path/path.dart' as path;
+import 'package:path_provider/path_provider.dart';
+
+/// {@template flame_assets_response}
+/// A class containing the relevant http attributes to
+/// Flame Assets Network package.
+/// {@endtemplate}
+class FlameAssetResponse {
```
st-pasha, line 17: It might be better to have each public class in its own file.

packages/flame_networks_assets/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 0.0.1
+
+* TODO: Describe initial release.
```
luanpotter, line 3: TODO?

erickzanardo: I believe melos will replace all this when we do the first release, @spydon can you confirm this? And that is not the case, what should go in here?

packages/flame_networks_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,223 @@
...
+        (
+          String url, {
+          Map<String, String>? headers,
+        }) =>
+            http.get(Uri.parse(url), headers: headers).then((response) {
+              return FlameAssetResponse(
+                statusCode: response.statusCode,
+                bytes: response.bodyBytes,
+              );
+            });
```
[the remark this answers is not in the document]
erickzanardo: Not sure what you mean about `statue` method

packages/flame_networks_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,223 @@
...
+
+    _getAppDirectory = getAppDirectory ?? getApplicationDocumentsDirectory;
+  }
+
+  late final GetAssetFunction _get;
+  late final GetAppDirectoryFunction _getAppDirectory;
+  late final DecodeAssetFunction<T> _decode;
+  late final EncodeAssetFunction<T> _encode;
```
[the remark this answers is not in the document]
erickzanardo: indeed they don't!

packages/flame_networks_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,223 @@
...
+
+  /// Flag indicating if files will be cached in memory.
+  final bool cacheInMemory;
+
+  /// Flag indicating if files will be cached in the local storage.
+  final bool cacheInStorage;
+
+  final bool _isWeb;
```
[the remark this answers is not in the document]
erickzanardo: It is used actually, if you look on the constructor, it is the default value for this attribute.

I wrapped it in this attribute to make it easier to mock on tests.

packages/flame_networks_assets/CHANGELOG.md
spydon: Melos only prepends to the file, so you can just leave it empty

packages/flame_network_assets/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 0.0.1
+
+* TODO: Describe initial release.
```
erickzanardo, line 3:
```suggestion
```

doc/flame/rendering/images.md
```diff
@@ -120,6 +120,32 @@ class MyGame extends Game {
 ```
 
 
+## Loading images from the network
+
+Flame core package doesn't offer a built in method to loading images from the network.
```
spydon, line 125:
```suggestion
The Flame core package doesn't offer a built in method to loading images from the network.
```

doc/flame/rendering/images.md
```diff
@@ -120,6 +120,32 @@ class MyGame extends Game {
...
+
+The reason for that is that Flutter/Dart does not have a built in http client, which required
```
spydon, line 127:
```suggestion
The reason for that is that Flutter/Dart does not have a built in http client, which requires
```

doc/flame/rendering/images.md
```diff
@@ -120,6 +120,32 @@ class MyGame extends Game {
...
+a package to be used and since there are a couple of packages available out there, we refrain
+from forcing, on even pushing the user in a specific package.
```
spydon, line 129:
```suggestion
from forcing the user to use a specific package.
```

doc/flame/rendering/images.md
```diff
@@ -120,6 +120,32 @@ class MyGame extends Game {
...
+import 'package:http/http.dart' as http;
+import 'package:flutter/painting.dart';
+
+final response = await http.get('https://url.com/image.png');
+final image = await decodeImageFromList(response.bytes);
+```
+
+```{note}
+Check [`flame_network_assets`](https://pub.dev/packages/flame_network_assets)
+for a ready to use network assets solution that provides built in cache.
```
spydon, line 145:
```suggestion
for a ready to use network assets solution that provides a built in cache.
```

doc/flame/rendering/images.md
```diff
@@ -120,6 +120,32 @@ class MyGame extends Game {
 ```
 
 
+## Loading images from the network
```
spydon, line 123: I think this section should be further down in the document? Feels like Sprite for example is of higher priority?

packages/flame_network_assets/README.md
```diff
@@ -0,0 +1,25 @@
...
+  <a title="Discord" href="https://discord.gg/pxrBmy4" ><img src="https://img.shields.io/discord/509714518008528896.svg" /></a>
+</p>
+<!-- markdownlint-enable MD013 -->
+
+---
+
+This package makes it easy to use and cache assets from the network inside a Flame game.
+
+For instructions on how to use this package to load images,
+check [Flame docs](https://docs.flame-engine.org/1.6.0/flame/rendering/images.html).
```
spydon, line 25: This has moved right?

packages/flame_network_assets/example/README.md
```diff
@@ -0,0 +1,3 @@
+# example
+
+`flame_network_images` example app.
```
spydon, line 3:
```suggestion
An example app that shows how to use the `flame_network_images` package.
```

packages/flame_network_assets/example/README.md
```diff
@@ -0,0 +1,3 @@
+# example
```
spydon, line 1:
```suggestion
# flame_network_images example
```

packages/flame_network_assets/example/lib/main.dart
```diff
@@ -0,0 +1,43 @@
...
+import 'package:flame_network_assets/flame_network_assets.dart';
+import 'package:flutter/material.dart' hide Image;
+
+void main() {
+  runApp(const GameWidget.controlled(gameFactory: MyGame.new));
+}
+
+class MyGame extends FlameGame with TapDetector {
+  final networkImages = FlameNetworkImages();
+  late Image charSprite;
```
spydon, line 16:
```suggestion
  late Image playerSprite;
```
Or something that isn't abbreviated.

packages/flame_network_assets/example/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
+name: example
```
spydon, line 1:
```suggestion
name: flame_network_assets_example
```
This would collide with any other package named example.

packages/flame_network_assets/example/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+version: 1.0.0+1
+
+environment:
+  sdk: '>=2.18.0 <3.0.0'
+
+dependencies:
+  flame: ^1.6.0
+  flame_network_assets:
+    path: ../
```
spydon, line 13: Is this still needed, can't you just write the version with a carret here and melos will take care of it even if it isn't released?

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,194 @@
...
+
+/// {@template flame_network_assets}
+///
+/// [FlameNetworkAssets] is a class similar to Flame's assets classes (like
+/// [Images] for example), but instead of loading assets from the assets bundle,
+/// it loads from networks urls.
+///
+/// By default, [FlameNetworkAssets] uses [http.get] method to make the requests
+/// ,that can be customized by passing a different [GetAssetFunction] to the get
+/// argument on the constructor.
```
spydon, line 38:
```suggestion
/// By default, [FlameNetworkAssets] uses the [http.get] method to make the
/// requests. It can be customized by passing a different [GetAssetFunction] to
/// the `get` argument on the constructor.
```

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,194 @@
...
+///
+/// [FlameNetworkAssets] also will also automatically cache files in a two layer
+/// system.
```
spydon, line 41:
```suggestion
/// [FlameNetworkAssets] also will automatically cache files in a two layer
/// system.
```

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,194 @@
...
+///
+/// The first layer is an in memory cache, handled by an internal [MemoryCache],
```
spydon, line 43:
```suggestion
/// The first layer is an in-memory cache, handled by an internal [MemoryCache],
```

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,194 @@
...
+/// while the second is the device own file system, where images are cached in
+/// the application document directory, which by default is provided by
+/// path_providers' [getApplicationDocumentsDirectory] method, and can be
+/// customized using the getAppDirectory argument in the constructor.
```
spydon, line 47:
```suggestion
/// while the second one is the device's own file system, where images are
/// cached in the application document directory, which by default is provided
/// by path_providers' [getApplicationDocumentsDirectory] method, and can
/// be customized using the `getAppDirectory` argument in the constructor.
```

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,194 @@
...
+///
+/// When an asset is requested, [FlameAssetResponse] will first check on its
+/// cache layers before making the http request, if both layer are cache miss,
+/// then the request is made and both layers set with the response.
+///
+/// Another important note about the cache layers is that the first layer, is a
+/// per instance cache, while the local storage is a app global cache. This
```
spydon, line 54:
```suggestion
/// per instance cache, while the local storage is an app global cache. This
```

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,194 @@
...
+/// means that two different [FlameAssetResponse] instances will have the same
+/// local storage cache, but not the memory cache.
```
spydon, line 56:
```suggestion
/// local storage cache, but not the same memory cache.
```

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,194 @@
...
+///
+/// Note that the local storage layer is not present when running on web since
+/// that platform doesn't really have a file system. The Browse caching will
```
spydon, line 59:
```suggestion
/// that platform doesn't really have a file system. The browser caching will
```

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,194 @@
...
+/// work as similar replacement for this layer, though that can't be controlled
+/// this package, make sure that the server where the images are being fetched
+/// returns the correct cache header to make the browser cache the assets.
```
spydon, line 62:
```suggestion
/// work as a similar replacement for this layer, though that can't be
/// controlled by this package, make sure that the server where the images
/// are being fetched returns the correct cache header to make the browser
/// cache the assets.
```

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,194 @@
...
+  /// - [encodeAsset] a [EncodeAssetFunction] responsible for encoding the asset
+  /// to its raw format.
+  /// - [get] is an optional [GetAssetFunction], if omitted [http.get] is used
+  /// by default.
+  /// - [getAppDirectory] is an optional [GetAppDirectoryFunction], if omitted
+  /// [getApplicationDocumentsDirectory] is used by default.
+  /// - [cacheInMemory] when false, will not cache assets in the memory,
+  /// (true by default).
+  /// - [cacheInStorage] when false, will not cache assets in the file system,
+  /// (true by default).
```
spydon, line 82:
```suggestion
  /// - [cacheInMemory] will not cache assets in the memory when false,
  /// (true by default).
  /// - [cacheInStorage] will not cache assets in the file system when false,
  /// (true by default).
```

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,194 @@
...
+  final DecodeAssetFunction<T> _decode;
+  final EncodeAssetFunction<T> _encode;
+
+  /// Flag indicating if files will be cached in memory.
+  final bool cacheInMemory;
+
+  /// Flag indicating if files will be cached in the local storage.
+  final bool cacheInStorage;
+
+  final bool _isWeb;
```
spydon, line 119: Why is this needed, can't `kIsWeb` be used directly?

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,194 @@
...
+
+  final _memoryCache = MemoryCache<String, T>();
+
+  String _urlToId(String url) {
+    final bytes = utf8.encode(url);
+    return base64.encode(bytes);
+  }
+
+  /// Loads the given asset from the given url.
```
spydon, line 128:
```suggestion
  /// Loads the asset from the given url.
```

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,194 @@
...
+      }
+
+      if (!_isWeb && cacheInStorage) {
+        unawaited(_saveAssetInLocalStorage(id, image));
+      }
+
+      return image;
+    } else {
+      throw Exception(
+        'Error fetching asset $url, response return status code '
```
spydon, line 165:
```suggestion
        'Error fetching asset from $url, response return status code '
```

packages/flame_network_assets/pubspec.yaml
```diff
@@ -0,0 +1,27 @@
+name: flame_network_assets
+description: Network assets support for Flame.
+version: 0.0.1
```
spydon, line 3:
```suggestion
version: 0.1.0
```

spydon (commented):
Lookin' good! Mostly just grammar stuff.

doc/bridge_packages/flame_network_assets/flame_network_assets.md
```diff
@@ -0,0 +1,34 @@
...
+  FlameNetworkImages({
+    super.get,
+    super.getAppDirectory,
+    super.cacheInMemory,
+    super.cacheInStorage,
+  }) : super(
+          decodeAsset: (bytes) => CustomAsset.decode(bytes),
+          encodeAsset: (CustomAsset asset) => asset.encode(),
+        );
+}
```
st-pasha, line 34: Missing closing "```" ?

st-pasha (approved):

packages/flame_network_assets/example/lib/main.dart
erickzanardo: Never stoped to think that char is an abbreviation...

packages/flame_network_assets/README.md
erickzanardo:
```suggestion
check [Flame docs](https://docs.flame-engine.org/1.6.0/bridge_packages/flame_network_assets/flame_network_assets.html).
```

packages/flame_network_assets/example/pubspec.yaml
erickzanardo: Yeah, definitely, this went in the automatic since most of projects that I work on doesn't have melos :P, changed!!

spydon (approved):
Lgtm!

packages/flame_network_assets/pubspec.yaml
```diff
@@ -0,0 +1,27 @@
+name: flame_network_assets
+description: Network assets support for Flame.
```
luanpotter, line 2: nit: this will trigger the "package description too short" pub penalty. I think the min is 60 characters. not endorsing unnecessary verbosity but losing points is also bad

packages/flame_network_assets/CHANGELOG.md
luanpotter, line 1: do we need an empty file at all?

packages/flame_network_assets/README.md
```diff
@@ -0,0 +1,25 @@
...
+  </a>
+</p>
+
+<p align="center">
+Adds network images support to <a href="https://github.com/flame-engine/flame">Flame</a>.
+</p>
+
+<p align="center">
+  <img src="https://github.com/flame-engine/flame_network_image/workflows/Lint/badge.svg?branch=master&event=push" alt="Test" />
+  <a title="Discord" href="https://discord.gg/pxrBmy4" ><img src="https://img.shields.io/discord/509714518008528896.svg" /></a>
```
luanpotter, line 16: why not the 4 standard badges? (I know the package hasnt been released yet if you add as soon as it is it will work)

packages/flame_network_assets/example/pubspec.yaml
```diff
@@ -0,0 +1,17 @@
...
+  sdk: '>=2.18.0 <3.0.0'
+
+dependencies:
+  flame: ^1.6.0
+  flame_network_assets: ^0.1.0
+  flutter:
+    sdk: flutter
+
+dev_dependencies:
+  flame_lint: ^0.1.3
```
luanpotter, line 17: use 0.2.0 ?

doc/bridge_packages/flame_network_assets/flame_network_assets.md
```diff
@@ -0,0 +1,35 @@
+# FlameNetworkAssets
+
+`FlameNetworkAssets` is a bridge package focused on providing a solution to fetch, and cache assets
```
luanpotter, line 3:
```suggestion
`FlameNetworkAssets` is a bridge package focused on providing a solution to fetch and cache assets
```

doc/bridge_packages/flame_network_assets/flame_network_assets.md
```diff
@@ -0,0 +1,35 @@
...
+from the network.
+
+The `FlameNetworkAssets` class provides an abstraction that should be extended in order to create
+asset specific handler.
+
+By default, the package relies on the `http` package to make http requests, and `path_provider`
+to get the place to store the local cache, to use a different approach for those, use the
+optional arguments in the constructor.
```
luanpotter, line 11:
```suggestion
By default, the package relies on the `http` package to make http requests, and `path_provider`
to get the place to store the local cache; to use a different approach for those, use the
optional arguments in the constructor.
```

doc/flame/rendering/images.md
```diff
@@ -120,6 +120,32 @@ class MyGame extends Game {
...
+import 'package:flutter/painting.dart';
+
+final response = await http.get('https://url.com/image.png');
+final image = await decodeImageFromList(response.bytes);
+```
+
+```{note}
+Check [`flame_network_assets`](https://pub.dev/packages/flame_network_assets)
+for a ready to use network assets solution that provides a built in cache.
+```
```
luanpotter, line 146: we could probably be even more persuasive about using the flame_network_assets, and frame importing http "by hand" as an alternative

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,195 @@
...
+  final DecodeAssetFunction<T> _decode;
+  final EncodeAssetFunction<T> _encode;
+
+  /// Flag indicating if files will be cached in memory.
+  final bool cacheInMemory;
+
+  /// Flag indicating if files will be cached in the local storage.
+  final bool cacheInStorage;
+
+  final bool _isWeb;
```
luanpotter, line 120: can we just set the default value of cacheInStorage to be kIsWeb instead of having two flags?

packages/flame_network_assets/lib/src/flame_network_assets.dart
```diff
@@ -0,0 +1,195 @@
...
+      if (storageAsset != null) {
+        if (cacheInMemory) {
+          _memoryCache.setValue(id, storageAsset);
+        }
+        return storageAsset;
+      }
+    }
+
+    final response = await _get(url, headers: headers);
+    if (response.statusCode >= 200 && response.statusCode < 400) {
```
luanpotter, line 152: nit: invert the `if` for an early-exit

Issue #1421, opened 2022-03-07
Add onParentResize lifecycle method


st-pasha:
Currently we have the `onGameResize()` lifecycle method which fires first before the component's `onLoad()`, and then every time the game canvas changes.

This FR is about creating an alternative lifecycle event `onParentResize()` that would be conceptually similar to `onGameResize`, but without the deficiencies of the latter. The `onGameResize` method will remain as-is, thus this FR will not introduce any breaking changes.

The logic of `onParentResize()` is as follows:
  - Whenever a component changes size, it will call `onParentResize()` for all its children.
  - If the child does NOT change its size, then `onParentResize()` is not propagated deeper into the tree.
  - This method is NOT invoked before `onLoad()`.
  - This method is invoked for the `FlameGame` class when the widget size changes.

spydon:
Isn't it the same for the user to just add a callback to the parent's `NotifyingVector2` size when they need this functionality?
Then we don't have to do lots of `onParentResize` calls where they aren't asked for.

st-pasha:
NotifyingVector is not available on Canvas or Camera, plus it's definitely not as convenient.

At the same time, resize events happen too rarely to worry about performance implications.

spydon:
> NotifyingVector is not available on Canvas or Camera, plus it's definitely not as convenient.

How can a `Canvas` be a parent of a component? 
`Camera` as in the `CameraComponent`?

 > NotifyingVector is not available on Canvas or Camera

True

> plus it's definitely not as convenient.

I think it is more convenient in some sense, since you can add a listener dynamically at runtime, but you can't do the same with overriding `onParentResize`. I see the value of it now though, let's see what the other say @flame-engine/flame-colab.

st-pasha:
> How can a Canvas be a parent of a component?

FlameGame is mounted into a GameWidget, thus for the FlameGame the "parent" is the Canvas.

st-pasha:
For example, one use case is this: suppose there is a game and you want to change the sizes/layout of the game elements when the canvas changes. Naturally you would assume that `onGameResize` would be a good fit here.

However, since `FlameGame.onGameResize` is called before the game was either loaded or mounted, we end up in a state where nothing can be properly sized because game elements hasn't been properly created yet. The only solution today is to manually create a flag to keep track when it is safe to resize. Ideally, though, this kind of logic should be handled by Flame itself.

spydon:
[...]

PR #2317, opened 2023-02-03
refactor!: Use ComponentTreeRoot for component removal

st-pasha:
# Description

Component removal is now handled via the `ComponentTreeRoot` class and its queue of lifecycle events.

- The `onGameResize` event now only fires before the component's `onMount`, but not before its `onLoad`.

## Checklist

## Breaking Change?

- [-] No, this PR is not a breaking change.

### Migration instructions

If you have components that rely on receiving `onGameResize` calls before they load, then 
you can retrieve the game's size in `onLoad` manually via `findGame()!.size`.

## Related Issues


packages/flame/lib/src/components/core/component.dart
```diff
@@ -198,6 +199,11 @@ class Component {
   bool get isLoaded => (_state & _loaded) != 0;
   void _setLoadedBit() => _state |= _loaded;
 
+  @internal
+  bool get isMounting => (_state & _mounting) != 0;
```
spydon, line 203: This is a breaking change if `onGameResize` doesn't fire before `onMount` any more.
If you size your components according to the game size and the game size changes when the component isn't in the tree your component's size will be outdated.
(Added this in here so that it has a discussion thread)

st-pasha: Ok, I'll add the Migration Instructions section for this.

st-pasha: I've also added the `OnGameResizeBeforeMount` mixin to make the transition easier.

spydon: What is the reason for not keeping the behavior like it is today?

st-pasha: Because we also call `onGameResize` before `onLoad`, and then if we also call it people would start complaining that it's been called too often. So, to avoid that, we need to have an extra status "justFinishedLoadingButHaventStartedMounting", which adds a lot of extra complexity with regards to what to do when this component is removed, or when an actual `onGameResize` is triggered at that time. (We dealt with this problem before by keeping the `isLoading` flag on after the component finished loading, which becomes problematic because if this component is suddenly removed, then it'll get stuck in this fake-loading state forever).

It's a whole lot easier to ditch all that complexity and let the users query the game size manually if they need it (which, I presume, is a tiny fraction of use cases).

spydon: Feels a bit inconsistent to even call the `onGameResize` before `onLoad` then, it doesn't really serve a purpose there then?
Maybe it could just be called before `onMount` instead, and remove the call before `onLoad`? I think you once suggested that, but we were hesitant to change the lifecycle.

st-pasha: Yes, we had a discussion about this in #1421. Ideally, `onGameResize()` would run only when the game actually resizes, as the name suggests. For all other purposes, the canvas size can be obtained manually via `findGame().size` both from `onLoad` and `onMount`. So the best course of action would be to remove `onGameResize` before component loads/mounts altogether.

But that would be a breaking change, so we might not want to do it just yet.

If there is a choice to call either before `onMount` or before `onLoad`, then the first option is easier (because then we don't need to worry about repeating the event if a game resize happens while the component is loading). So, if you're ok with that, I'll change the logic to only call game-resize before `onMount` but not before `onLoad`.

spydon: > If there is a choice to call either before onMount or before onLoad, then the first option is easier (because then we don't need to worry about repeating the event if a game resize happens while the component is loading). So, if you're ok with that, I'll change the logic to only call game-resize before onMount but not before onLoad.

Yeah, I know I didn't want that previously, but I think it makes more sense to do it that way now and we'll re-think completely the existence and purpose of `onGameResize` for v2.

spydon (approved):

doc/flame/components.md
```diff
@@ -36,8 +36,8 @@ Every `Component` has a few methods that you can optionally implement, which are
 ```{include} diagrams/component_life_cycle.md
 ```
 
-The `onGameResize` method is called whenever the screen is resized, and once in the beginning when
-the component is added to the game via the `add` method.
+The `onGameResize` method is called whenever the screen is resized, and each time a
```
wolfenrain, line 39: Reading this it implies that the `onGameResize` is called whenever ANY component is added, so my parent `onGameResize` is also called when a child is added to its tree?

I think we should word it more like `The method is called when screen resizes or when the component gets added to the tree`, something in that sense, making it more connected to `the component`

packages/flame/lib/src/components/core/component.dart
```diff
@@ -39,9 +39,9 @@ import 'package:vector_math/vector_math_64.dart';
 ///
 /// Each component goes through several lifecycle stages during its lifetime,
 /// at each stage invoking certain user-defined "lifecycle methods":
-///  - [onGameResize] when the component is first added into the tree;
-///  - [onLoad] immediately after;
-///  - [onMount] when done loading;
+///  - [onLoad] when the component is first added into the tree;
+///  - [onGameResize] + [onMount] when done loading, or when component is
+///    re-added into the component tree after having been removed;
```
wolfenrain, line 44:
```suggestion
///  - [onGameResize] + [onMount] when done loading, or when the component 
///    is re-added to the component tree after having been removed;
```

packages/flame/lib/src/components/core/component.dart
```diff
@@ -611,16 +590,38 @@ class Component {
 
   /// Removes all the children for which the [test] function returns true.
   void removeWhere(bool Function(Component component) test) {
-    children.forEach((component) {
+    for (final component in children) {
```
wolfenrain, line 593: You don't know how happy this change makes me, my IDE always complains about this line 😂

wolfenrain (approved):

packages/flame/lib/src/components/core/component.dart
erickzanardo: Just making sure I understood the discussion.

`onGameResize` will be called after `onLoad` and before `onMount`, but the game size will be available in both `onLoad` and `onMount` if the user tries to access it right?

spydon: It would indeed be able to get the game's size in `onLoad` since when changing a parent the first thing that happens is that the parent field is set and this is what `HasGameRef` uses to traverse up to the game where you can get the size.

doc/flame/components.md
st-pasha: done

spydon (approved):

PR #2325, opened 2023-02-06
docs: Update cspell github action and configuration

luanpotter:
# Description

I am on a crusade to improve our spellchecking. I have a big PR that I am now going to break down into very small chunks to for review.

[zwaldowski's cspell-action](https://github.com/zwaldowski/cspell-action) has been deprecated in favor of streetsidesoftware/cspell-action.

This PR:

* updates to use the new recommended cspell action
* update to use checkout v3, because why not
* fixes a few violations the new action caught

Note: I think we should make a distinction between things that should be in a dictionary (e.g. `AABB`, a standard industry acronym that we expect people to know what it is) and specific edge-cases/ignore rules for specific parts of specific files (e.g. something like `fccc` which is meaningless outside of the very specific context in which it is used). For the latter, I believe we should use case-by case exceptions.

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame/lib/src/extensions/color.dart
+++ b/packages/flame/lib/src/extensions/color.dart
@@ -37,6 +37,9 @@ extension ColorExtension on Color {
     );
   }
 
+  // used as an example hex color code on the documentation below
+  // cSpell:ignore fccc
+
   /// Parses an RGB color from a valid hex string (e.g. #1C1C1C).
   ///
   /// The `#` is optional.
--- a/packages/flame/lib/src/extensions/paint.dart
+++ b/packages/flame/lib/src/extensions/paint.dart
@@ -42,6 +42,9 @@ extension PaintExtension on Paint {
     return Paint()..color = color;
   }
 
+  // used as an example hex color code on the documentation below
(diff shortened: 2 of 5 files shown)


packages/flame/lib/src/extensions/color.dart
```diff
@@ -37,6 +37,9 @@ extension ColorExtension on Color {
     );
   }
 
+  // used as an example hex color code on the documentation below
```
spydon, line 40:
```suggestion
  // Used as an example hex color code on the documentation below.
```

packages/flame/lib/src/extensions/paint.dart
```diff
@@ -42,6 +42,9 @@ extension PaintExtension on Paint {
     return Paint()..color = color;
   }
 
+  // used as an example hex color code on the documentation below
```
spydon, line 45:
```suggestion
  // Used as an example hex color code on the documentation below.
```

spydon (approved):

wolfenrain (approved):

PR #2326, opened 2023-02-06
refactor: Divide TileLayer by its Layer type

Hwan-seok:
# Description
### As-is
1. The `TileLayer` became too big to maintain. 
2. Some tile related classes doesn't have factory method, their creation is done in other classes with big switch-case.

### To-be
1. Divided `TileLayer` into its `Layer` type. The `Layer` type is determined by `MapOrientaion` which has a member of `isometric`, `orthogonal`, `staggered`, and `hexagonal`.
2. Instantiation is done by factory method inside the class.

 

## Checklist

## Breaking Change?


erickzanardo (approved):

ufrshubham (approved):

packages/flame_tiled/lib/src/renderable_layers/tile_layers/tile_layer.dart
```diff
@@ -0,0 +1,200 @@
...
+
+    canvas.translate(offsetX, offsetY);
+
+    if (camera != null) {
+      applyParallaxOffset(canvas, camera);
+    }
+
+    tiledAtlas.batch!.render(canvas, paint: _layerPaint);
+
+    canvas.restore();
```
spydon, line 140:
```suggestion
    canvas.save();
    canvas.translate(offsetX, offsetY);
    if (camera != null) {
      applyParallaxOffset(canvas, camera);
    }
    tiledAtlas.batch!.render(canvas, paint: _layerPaint);
    canvas.restore();
```

spydon (approved):
Lgtm, just two minor comments

spydon:
@jtmcdole testing if I can assign you to review now.

packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
```diff
@@ -13,12 +18,66 @@ abstract class RenderableLayer<T extends Layer> {
...
+    } else if (layer is Group) {
+      final groupLayer = layer;
+      return GroupLayer(
+        layer: groupLayer,
+        parent: parent,
+        map: map,
+        destTileSize: destTileSize,
+      );
+    } else {
+      assert(false, '$layer layer is unsupported.');
```
jtmcdole, line 72: i don't think you want the assert, given that we have "unsupported layer" as the return type.

packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
```diff
@@ -13,12 +18,66 @@ abstract class RenderableLayer<T extends Layer> {
...
+      );
+    } else if (layer is Group) {
+      final groupLayer = layer;
+      return GroupLayer(
+        layer: groupLayer,
+        parent: parent,
+        map: map,
+        destTileSize: destTileSize,
+      );
+    } else {
```
jtmcdole, line 71: you can get rid of the `else {}` and just have it as the default fall through.

jtmcdole:
lgtm - just style comments really.  Not a fan of asserts when we don't currently provide a functional layer, but we supply a standin.

packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
Hwan-seok: I tried to decide whether I should touch this assertion or not.
I left this because the users can be conscious to some layers are not supported.

The assertion may be too strict but giving no warning is probably annoying to users like ...
 Why is it not working? I should report it to the issue tab.

As mentioned in [this comment](https://github.com/flame-engine/flame/pull/2310#issuecomment-1413876378),
> Make some logging systems that warn flame users about what went wrong.

It would be better if there is some way to warn users that doesn't break the code but can warn it.

What do you think?

spydon: Agree that the assert probably doesn't give anything here, especially since this is an internal class.

Issue #2319, opened 2023-02-04
[proposal] Flame Studio


st-pasha:
# Problem to solve

We need to have a UI to make developing games with Flame easier. All other game engines have it: Unity, Unreal, Godot, GDevelop, etc.

This has been asked many times before, but we never had a concrete proposal. So, here it goes:

# Proposal

We create a new package `flame_studio`, which will be responsible for the UI. The package exposes a single function `runFlameStudio()`, which is a replacement for `runApp()`:
```dart
import 'package:flame_studio/flame_studio.dart';

void main() {
  runFlameStudio(
    Blah(
      child: BlahBlah(
        child: GameWidget(game: MyGame()),
      ),
    ),
  );
}
```

In release mode, the function `runFlameStudio` collapses into simple `runApp()`. However, in debug mode it builds a FlameStudio UI framework around the provided widget. When you run this instrumented game in a browser, it displays the UI and connects it to the `GameWidget` found inside the widget tree.

The UI has multiple capabilities, among those:
- pause the game at any time;
- inspect the component tree;
- view/modify properties of components at runtime;
- add/delete/move components at runtime;
- inspect the game world outside of game canvas (i.e. a new camera that can look at the world independently).

In addition, the FlameStudio would still have access to the file system, which allows it to inspect the code, and maybe even allow scene modification from within the UI -- for example, the user adds an object to the scene, and the studio automatically writes code that would add that object in `onLoad()`. Similarly, one could permanently modify some properties of objects from within the studio.

There should also be a mechanism for writing extensions to Flame Studio -- say, `flame_jenny` could add a UI for inspecting dialogue trees, `flame_bloc` could show the state of all `Bloc`s, and `flame_tiled` work with Tiled maps directly.

ufrshubham:
Every time this topic came up in the past I imagined that it would be a standalone application. But given our tech stack, I think this instrumented approach will be perfect fit.


> for example, the user adds an object to the scene, and the studio automatically writes code that would add that object in `onLoad()`

About this☝🏻, would it be beneficial if Flame uses some sort of a meta file to store the component tree? The ideas is to use this file to load up the initial state of the scene, without needed to programmatically modify `onLoad`.

spydon:
Big undertaking, but would be a great tool!

It would be good to have some IDE support in it, maybe having either the whole studio inside of Android Studio or VSCode, or have some connection to the IDE outside of it.

spydon:
[...]

PR #2331, opened 2023-02-08
feat: Initial version of flame_studio package

st-pasha:
# Description

This is the beginning of work on Flame Studio. In this PR we add a simple UI, which shows the game itself, an info panel, and a toolbar with 2 buttons: start/pause. The UI is able to control the embedded game; in particular the toolbar buttons can pause or resume the game engine.

Screenshot 2023-02-08 at 12 32 06 AM

## Checklist

## Breaking Change?

## Related Issues

WIP for #2319


packages/flame_studio/lib/src/widgets/flame_studio.dart
```diff
@@ -0,0 +1,67 @@
...
+class FlameStudio extends StatefulWidget {
+  const FlameStudio(this.child, {super.key});
+
+  final Widget child;
+
+  @override
+  State<StatefulWidget> createState() => _FlameStudioState();
+
+  static final List<Widget> toolbarLeft = [];
+  static final List<Widget> toolbarMiddle = [
```
erickzanardo, line 17: These widget are used on the Toolbar component, so defining them there directly would make more sense IMO.

If I would add a new icon to the toolbar, that would be the first place that I would look for.

packages/flame_studio/lib/src/widgets/toolbar/toolbar.dart
```diff
@@ -0,0 +1,51 @@
...
+class Toolbar extends StatelessWidget {
+  const Toolbar({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    final settings = Settings.of(context);
+    final height = settings.toolbarHeight;
+    final gap = height * 0.10;
+
+    Widget _wrapButton(Widget button) {
```
erickzanardo, line 14: Ideally this would be an internal, stateless widget instead.

packages/flame_studio/lib/src/run_flame_studio.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:flame_studio/src/widgets/flame_studio.dart';
+import 'package:flutter/widgets.dart';
+
+void runFlameStudio(Widget app, {bool enabled = true}) {
+  var allowRun = true;
+  assert(() {
+    if (enabled) {
```
erickzanardo, line 7: I am trying to understand why this is needed, but I am failing to grasp why, could you go through this so I can get a better idea?

packages/flame_studio/lib/src/widgets/flame_studio.dart
```diff
@@ -0,0 +1,67 @@
...
+    const StartButton(),
+    const PauseButton(),
+  ];
+  static final List<Widget> toolbarRight = [];
+}
+
+class _FlameStudioState extends State<FlameStudio> {
+  Game? _game;
+  Future<void>? _findGameWidgetFuture;
+  void _scheduleGameWidgetSearch() {
```
erickzanardo, line 27: I wonder if we shouldn't simply ask the game as an attribute for the `runFlameStudio` method, so we don't need to go through all of this code, I imagine that if the game widget is too deep in the tree, this could cause potential perfomance issues?

We could make the `Widget` parameter optional for the cases where the user only wants to pass the game directly, like:

```dart
runFlameStudio(MyGame());
```

```dart
final game = MyGame();
runFlameStudio(game, child: Container(child: GameWidget(game: game)));
```

erickzanardo (commented):
This is a great idea! Loving to see how this will progress!

I have a few points though

 - If I understood correctly for the PR and the issue where this was discussed (and correct if I am wrong), this tool will be like a companion app that will help user to inspect the game at runtime. If so, I feel that the Studio word is a little bit misleading, when I read Studio, I get the idea that this will be something like Unity's editor or something like that. I think that to avoid confusion on the users, we could name this `flame_dev_tools`, just like `flutter_dev_tools`.
 - I believe that this has the potential to grow a lot, I noticed that a custom way of managing the state was created. I advise that we should choose an existent, popular solution available like bloc, riverpod, provider, etc. That will help set the tone of the project structure and will make it easier for the community to contribute.
  - I wonder if this deserves a repository of its own, since this is a stand alone application, it may help for it to live alone in a repo so we can do automation focused on it, like distributing it, when releasing a new version, we could have actions that build and upload the binaries (if we create desktop versions) of it on the release, just like we did on Fire Atlas Editor.

Let me know what you think, but I am looking forward to this!

packages/flame_studio/lib/src/run_flame_studio.dart
st-pasha: So, the idea here is that the user can simply replace `runApp()` with `runFlameStudio()` in their code.

Then, in release mode the `assert()` disappears, and what is left is simple `runApp()`. Thus, in release mode the game will behave exactly as before. Moreover, since this is the only hook into the `flame_studio` package, then the `FlameStudio` and all the internal widgets get eliminated via tree-shaking, so that the size of the final executable is likewise unaffected. 

However, in debug mode the code inside the assert is executed, and that code wraps the user's `app` into the `FlameStudio` widget, and builds the studio UI which controls the game.

erickzanardo: Ahhh got ya. So the user can always have a single entry point!!! This is very clever!

packages/flame_studio/lib/src/widgets/flame_studio.dart
st-pasha: This could be problematic in some cases. For example, if the user is using `GameWidget.controlled()`, then the game instance is only known to the GameWidget's State.

In addition, it might be interesting to explore in the future the possibility of restarting the game from within the studio.

st-pasha:
> If I understood correctly for the PR and the issue where this was discussed (and correct if I am wrong), this tool will be like a companion app that will help user to inspect the game at runtime. 

This would be one of the initial capabilities of the app. But it won't be the only one. Ideally, we would want to let the user modify the game's setup, add/remove components, and modify their properties -- allowing these changes to be persisted within the project.

In addition, this should be highly extendable, so that various bridge packages could integrate with Flame Studio and provide controls for their specific components.

> I noticed that a custom way of managing the state was created. I advise that we should choose an existent, popular solution available like bloc, riverpod, provider, etc.

Oh yes, I don't have much experience with Flutter, so the architecture might be a bit awkward. I'll try to switch to Bloc.

> I wonder if this deserves a repository of its own, since this is a stand alone application, it may help for it to live alone in a repo so we can do automation focused on it, like distributing it, when releasing a new version, we could have actions that build and upload the binaries (if we create desktop versions) of it on the release, just like we did on Fire Atlas Editor.

I envision Flame Studio being multi-platform, just like Flutter/Flame. So you should be able to run it as a desktop app, or on mobile (tablets, phones are probably too small), or as web app. The latter is particularly appealing, because it could potentially mean running FS on some internet site, with some cloud storage at the backend. This is especially important for those users who has no access to a desktop. For example, many schoolchildren have only a Chromebook which runs web apps only.

That being said, it might have sense to move Flame Studio into a separate repo, eventually. But not at the beginning, because at least initially we would need to build some hooks into the `flame` itself in order to better support FS. For example, adding some new controls into the GameLoop, or replacing a `game` instance within a `GameWidget`, etc.

It's a bit difficult to predict at this point, but I think it's better to start building it here in the main repo, and then move out when it's more mature.

erickzanardo:
> This would be one of the initial capabilities of the app. But it won't be the only one. Ideally, we would want to let the user modify the game's setup, add/remove components, and modify their properties -- allowing these changes to be persisted within the project.

> In addition, this should be highly extendable, so that various bridge packages could integrate with Flame Studio and provide controls for their specific components.

Still, I would be conservative about it for now, idk, maybe I am worrying too much, but I think we should avoid promising a studio for now. I would arguee that calling it dev_tools (or even editor?) would be safer until we are building its feature set, and when it gets bigger we can consider renaming it to something fancier.

> Oh yes, I don't have much experience with Flutter, so the architecture might be a bit awkward. I'll try to switch to Bloc.

Cool, I can help if you need, just ping me on discord!

> It's a bit difficult to predict at this point, but I think it's better to start building it here in the main repo, and then move out when it's more mature.

SGTM!

st-pasha:
> when it gets bigger we can consider renaming it to something fancier.

How do we go about renaming a published package though?

spydon:
I vote for keeping it named Flame Studio, aim high etc! :sunglasses: 
With proper explanations in the readme ofc.
I don't think naming it editor would make people's expectations any different and dev tools seem like the bar is too low instead for what is planned.

erickzanardo:
> How do we go about renaming a published package though?
> I vote for keeping it named Flame Studio, aim high etc! 😎

Sounds good then...I think I am just too traumatized because of Android Studio :P

packages/flame_studio/lib/src/widgets/toolbar/toolbar.dart
st-pasha: done

packages/flame_studio/lib/src/widgets/flame_studio.dart
st-pasha: done

st-pasha:
I've switched to using `riverpod` for state management, so it's a bit cleaner now.

packages/flame_studio/lib/src/widgets/toolbar/flame_studio_toolbar.dart
```diff
@@ -0,0 +1,74 @@
...
+    const StartButton(),
+    const PauseButton(),
+  ];
+
+  /// List of widgets to show on the right side of the toolbar.
+  static final List<Widget> right = [];
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final height = ref.watch(toolbarHeightProvider); // settings.toolbarHeight;
```
spydon, line 26:
```suggestion
    final height = ref.watch(toolbarHeightProvider);
```
Or was that intentionally left there?

packages/flame_studio/lib/src/widgets/toolbar/flame_studio_toolbar.dart
```diff
@@ -0,0 +1,74 @@
...
+  static final List<Widget> left = [];
+
+  /// List of widgets to show in the center of the toolbar.
+  static final List<Widget> middle = [
+    const StartButton(),
+    const PauseButton(),
+  ];
+
+  /// List of widgets to show on the right side of the toolbar.
+  static final List<Widget> right = [];
```
spydon, line 22: I think it is probably better to create widgets from these lists instead of having static lists here

spydon (commented):
This will be great!

packages/flame_studio/lib/src/widgets/toolbar/flame_studio_toolbar.dart
st-pasha: accident

packages/flame_studio/lib/src/widgets/toolbar/flame_studio_toolbar.dart
st-pasha: The idea here is that it should be possible for an external code to insert new functionality into Flame Studio. In this case, you could say `FlameStudioToolbar.right.add(SpydonButton())`, and that button would appear inside the toolbar.

We may figure out a better way of handling this; maybe these lists will become riverpod providers as well so that the UI can update on the fly -- but for now we can keep them as simple lists.

spydon: Hmm, this is recommended to do in Flutter, because the widget won't know that it should rebuild when these lists are modified, so depending on where the user edits these lists it might or might not work. If they should only be set from the beginning they could be passed in as arguments, or if it should be possible to change them during runtime they should indeed be providers.

st-pasha: Right now it's impossible to say what the future requirements will be. This is just a starting point, and I tried to make it as simple as possible.

erickzanardo: I am with Lukas here, I would rather have these directly on the toolbar build method than to introduce an anti pattern like that, even if it is an initial version.

Also, I bet that allowing external sources to add custom tool bar items isn't something we need to care about in this initial version

spydon: I don't think that we should use patterns that aren't recommended whatever the future intention is though.
For a simple starting point these widgets could just be put into the build method, since the current method isn't a feasible for modifications anyways? And then we can work out later how it should be structured for allowing users to add their own buttons here?

packages/flame_studio/lib/src/core/game_controller.dart
```diff
@@ -0,0 +1,64 @@
...
+import 'package:flutter/widgets.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+final gameControllerProvider =
+    StateNotifierProvider<_GameController, _GameState>(
+  (ref) => _GameController(),
+);
+
+Future<Game?> _findGame() async {
+  await null;
```
spydon, line 11: What does this do? Keeping it async? Maybe `FutureOr` could be used instead?

st-pasha: This adds an async delay before the future is actually executed. Within that delay the widget tree will be built, allowing us to explore it.

spydon: Ah, I see! Maybe better to trigger it from `WidgetsBinding.instance.addPostFrameCallback` instead?

packages/flame_studio/lib/src/widgets/toolbar/flame_studio_toolbar.dart
st-pasha: Ok, I removed those static lists.

packages/flame_studio/lib/src/core/game_controller.dart
st-pasha: Yep, this is better.

spydon (approved):
🔥

packages/flame_studio/lib/src/core/game_controller.dart
```diff
@@ -0,0 +1,67 @@
+import 'package:flame/game.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+final gameControllerProvider =
+    StateNotifierProvider<_GameController, _GameState>(
+  (ref) => _GameController(),
+);
+
+class _GameState {
```
erickzanardo, line 10: I am not sure how riverpod works, but should this class implement equals?

erickzanardo (approved):

erickzanardo (approved):

packages/flame_studio/lib/src/core/game_controller.dart
spydon: Good catch, it should indeed. It will still work most of the time now though since it always creates a new object, but if the new object would have the same data as the old object it would re-render unnecessarily.

spydon: Let's pull it in when we have a few more places that need it.
It's also very nice to stay away from code generation until it's needed. 😅

st-pasha: Added `operator==`

spydon (approved):

NashIlli:
It looks good and promising!

PR #2339, opened 2023-02-13
feat: Show component tree in Flame Studio

st-pasha:
# Description

Add ability to display game's component tree in the left panel.
The tree automatically re-syncs with the game every 300ms (eventually this should be a configurable value).
Nested parts of the tree can be expanded/collapsed, and one component can be selected.
In the future, the selected component can be shown in the Inspector, but that is not implemented yet.

Screenshot 2023-02-13 at 2 14 30 PM

## Checklist

## Breaking Change?

## Related Issues

WIP for #2319


packages/flame_studio/lib/src/core/component_tree.dart
```diff
@@ -0,0 +1,84 @@
...
+              for (final child in component.children)
+                ComponentTreeNode.fromComponent(child)
+            ]
+          : null,
+    );
+  }
+
+  final Component component;
+  final List<ComponentTreeNode>? children;
+  String get name => component.runtimeType.toString();
```
spydon, line 32:
```suggestion
  late final String name = component.runtimeType.toString();
```

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
```diff
@@ -0,0 +1,221 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    final componentTree = ref.watch(componentTreeProvider).root;
+    if (componentTree == null) {
+      return Container();
+    }
+
+    final listItems = <Widget>[];
+    _buildList(componentTree, 0, listItems);
```
spydon, line 25: It's not recommended to build widgets with functions like this afaik:
https://github.com/flutter/flutter/issues/19269

spydon: It would be cleaner to just create a `StatelessWidget` that takes in the the `componentTree` here instead.

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
```diff
@@ -0,0 +1,221 @@
...
+}
+
+class _ExpanderIcon extends StatelessWidget {
+  const _ExpanderIcon({
+    required this.hasChildren,
+    this.isExpanded = false,
+    this.isFirst = false,
+    this.isLast = false,
+  });
+  final bool hasChildren;
```
spydon, line 143:
```suggestion

  final bool hasChildren;
```

packages/flame_studio/lib/src/widgets/scaffold.dart
```diff
@@ -6,12 +6,13 @@ import 'package:flutter/widgets.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 
 class Scaffold extends ConsumerWidget {
```
spydon, line 8: We shouldn't re-use names for such basic Flutter widgets, that will be super confusing.

spydon: Must have missed this on the last PR.

packages/flame_studio/lib/src/core/component_tree.dart
st-pasha: `late final` fields don't work with const constructors. Of course, the same property can be initialized directly from the constructor, but it seems to be more complicated than a simple getter.

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
st-pasha: `_buildList` doesn't create a widget, it populates `listItems` with widgets for each line item in the list. And it has to be a function, because it is called recursively in order to traverse the tree (of course, recursion can be replaced with a stack of iterators, but it's much cleaner via the recursion).

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
st-pasha: This is something with GitHub's display of whitespace, the code itself looks fine (and the formatter would have complained if it wasn't the case).

packages/flame_studio/lib/src/widgets/scaffold.dart
st-pasha: Renamed into `UiScaffold`.

packages/flame_studio/lib/src/core/component_tree.dart
```diff
@@ -0,0 +1,84 @@
...
+  const ComponentTreeNode._(this.component, this.children);
+
+  factory ComponentTreeNode.fromComponent(Component component) {
+    return ComponentTreeNode._(
+      component,
+      component.hasChildren
+          ? [
+              for (final child in component.children)
+                ComponentTreeNode.fromComponent(child)
+            ]
```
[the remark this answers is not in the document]
st-pasha: done

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
spydon: I don't think there is a rule for this in the formatter? It looks like the same if I look at the raw file.

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
spydon: Right, gotcha! This could recursively build up a model that is then passed into a widget that creates the visual parts.
The current implementation doesn't feel very flutter-esque at least. Wdyt @flame-engine/flame-admin?

packages/flame_studio/lib/src/core/component_tree.dart
```diff
@@ -0,0 +1,81 @@
...
+
+class ComponentTreeObserver extends StateNotifier<ComponentTreeState> {
+  ComponentTreeObserver(Component? rootComponent)
+      : super(ComponentTreeState(rootComponent)) {
+    if (rootComponent != null) {
+      _refresh(Duration.zero);
+    }
+  }
+
+  static Duration refreshFrequency = const Duration(milliseconds: 300);
```
erickzanardo, line 67:
```suggestion
  static const Duration refreshFrequency = Duration(milliseconds: 300);
```

packages/flame_studio/lib/src/core/component_tree.dart
```diff
@@ -0,0 +1,81 @@
...
+  Duration _timeSinceLastRefresh = Duration.zero;
+
+  void _refresh(Duration timeDelta) {
+    _timeSinceLastRefresh += timeDelta;
```
erickzanardo, line 71: I think this could be simplified if we simply use `Future.delayed<void>(Duration(milliseconds: 300))`

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
erickzanardo: `_buildList` does creates widgets, I mean, it is populating a list with widgets that it creates.

It is not clear to me it needs to be a function, it could just be a Widget that once it identifies that it has to go into a node, it can just add "itself" as the part of its own children, forwarding the children of that node to the new instance, doing a recursion the same way

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
```diff
@@ -0,0 +1,221 @@
...
+  void paint(Canvas canvas, Size size) {
+    const iconSize = 10.0;
+    const halfSize = iconSize / 2;
+    final xExtent = size.width / 2;
+    final yExtent = size.height / 2;
+    canvas.save();
+    canvas.translate(xExtent, yExtent);
+    if (icon.hasChildren) {
+      if (!icon.isFirst) {
+        canvas.drawLine(
```
erickzanardo, line 180: We could use `..` cascading here on the consecutive calls.

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
st-pasha: done

packages/flame_studio/lib/src/core/component_tree.dart
st-pasha: This parameter is intended to be modifiable by the user, possibly even within the UI once we have a settings panel.

packages/flame_studio/lib/src/core/component_tree.dart
st-pasha: Right, this is indeed simpler.

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
st-pasha: Most of these calls have if conditions, so we wouldn't be able to cascade them.

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
st-pasha: I can make each item in this list a separate widget though.

spydon: So the plan is to make them widgets then? :)

st-pasha: Already did

packages/flame_studio/lib/src/core/component_tree.dart
spydon: This should probably be handled by a provider instead of a static variable if it should be modifiable.

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
```diff
@@ -0,0 +1,239 @@
...
+      } else {
+        expandedComponents.add(component);
+      }
+      selectedComponent = component;
+    });
+  }
+}
+
+class _ListItem extends StatelessWidget {
+  const _ListItem(this.state, this.node, this.indent, this.isFirst);
```
spydon, line 71: Since there will be quite a lot of these ones in a list they should have a `Key` so that it doesn't have to rebuild the whole list all the time. If there is something that a key can be created from?

spydon (approved):
Just some small suggestions, otherwise looks good!

packages/flame_studio/lib/src/core/component_tree.dart
st-pasha: Not really, since nothing needs to be rebuilt when this value changes. The new value will be used automatically on the next update tick.

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
st-pasha: Added a key

spydon (approved):

packages/flame_studio/lib/src/core/component_tree.dart
```diff
@@ -0,0 +1,79 @@
...
+@immutable
+class ComponentTreeNode {
+  const ComponentTreeNode._(this.component, this.children);
+
+  factory ComponentTreeNode.fromComponent(Component component) {
+    return ComponentTreeNode._(
+      component,
+      component.hasChildren
+          ? component.children.map(ComponentTreeNode.fromComponent).toList()
+          : null,
```
luanpotter, line 22: any reason to make this distinction between null and empty list?

packages/flame_studio/lib/src/core/component_tree.dart
luanpotter: we could make it const for now and refactor once it is possible to change?

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
```diff
@@ -0,0 +1,240 @@
...
+    final listItems = <Widget>[];
+    _buildList(componentTree, 0, listItems);
+
+    return ListView(
+      padding: const EdgeInsets.all(8),
+      children: listItems,
+    );
+  }
+
+  void _buildList(ComponentTreeNode node, int depth, List<Widget> out) {
```
luanpotter, line 33: I surprised there isn't already a "TreeComponent" on flutter that does this recursion for you

luanpotter (approved):
LGTM

luanpotter:
we should consider extracting this as a generic TreeViewComponent if there isn't one already in flutter, seems vastly useful for many projects, and could be a dep of flame_studio

packages/flame_studio/lib/src/core/component_tree.dart
st-pasha: If a component doesn't have children, then I don't want to instantiate its children ComponentSet unnecessarily.

st-pasha: The majority of components don't have any children, so why create empty lists unnecessarily?

packages/flame_studio/lib/src/core/component_tree.dart
st-pasha: It's already possible to change, only via code, not UI.

packages/flame_studio/lib/src/widgets/panels/hierarchy_view.dart
st-pasha: The `ListView` already does most of the job, and conversion of a tree into a list is too ambiguous to be supported by the framework.

PR #2344, opened 2023-02-14
docs: Refactor documentation for GameWidget

st-pasha:
# Description

Use the `dartdoc` directive for the `GameWidget` class, allowing us to keep the documentation in the code and on the site in sync.

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame/lib/src/game/game_widget/game_widget.dart
+++ b/packages/flame/lib/src/game/game_widget/game_widget.dart
@@ -8,122 +8,38 @@ import 'package:flame/src/game/game_widget/gesture_detector_builder.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 
-typedef GameLoadingWidgetBuilder = Widget Function(
-  BuildContext,
-);
-
-typedef GameErrorWidgetBuilder = Widget Function(
-  BuildContext,
-  Object error,
-);
-
-typedef OverlayWidgetBuilder<T extends Game> = Widget Function(
-  BuildContext context,
-  T game,
-);
-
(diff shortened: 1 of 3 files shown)


packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -8,122 +8,38 @@ import 'package:flame/src/game/game_widget/gesture_detector_builder.dart';
...
+/// }
+/// ```
+///
+/// At the same time, `GameWidget` is a regular Flutter widget, and can be
+/// inserted arbitrarily deep into the widget tree, including the possibility
+/// of having multiple `GameWidget`s within a single app.
+///
+/// The layout behavior of this widget is that it will expand to fill all
+/// available space. Thus, when used as a root widget it will make the app
+/// full-screen. Inside any other layout widget it will take as much space as
```
erickzanardo, line 31: One thing that I think we should also mention here, that `GameWidget` doesn't clip automatically. Meaning that is the user render something outside the boundaries of the parent of the game widget, the game content will "overflow", that if the user wants to clip, they should wrap the game widget into a ClipRect widget.

erickzanardo (commented):
Awesome docs improvement, left one small comment

packages/flame/lib/src/game/game_widget/game_widget.dart
st-pasha: Good point. Will need to add clipping to Flame Studio as well.

st-pasha: Done.

spydon (approved):

Issue #2337, opened 2023-02-12
Use the tileset image from the cache if it's already loaded


Hwan-seok:
# What could be improved

## Current behavior
TileAtlas has a different order of checking the cache.

If the tileset count

* is only one: checks if `atlas{$source}` exists 
   - if exists: get `atlas{$source}`
   - if not exists: read `$prefix$source` and set key as `atlas{$source}`
   - https://github.com/flame-engine/flame/blob/7920e2ba4d52a2461ae4631ffdaf8c52fbcd9dd3/packages/flame_tiled/lib/src/tile_atlas.dart#L99

* is more than one: checks if each item's `$prefix$source` exists
   - if and if not exists: get `$prefix$source`
   - set key as `atlas{$agreegated_sources}`
   - https://github.com/flame-engine/flame/blob/7920e2ba4d52a2461ae4631ffdaf8c52fbcd9dd3/packages/flame_tiled/lib/src/tile_atlas.dart#L124

The main difference here is the first check of each tileset.

### Example

Imagine loading maps in the following order: Map1 => Map2.
The Map2-set_1 could be loaded from the cache but it reloads the image because it checks "atlas{set_1}" instead of "set_1".

#### Map 1 - has 2 tilesets
1. tileset 1(source: "set_1")
2. tileset 2(source: "set_2")
Generates image cache 
  "set_1", "set_2", atlas{"set_1", "set_2"}

#### Map 2 - has a tileset
1. tileset 1(source: "set_1")
Generates
  atlas{"set_1"}

## Improved

If the tileset count is only one: check if **`$prefix$source`** exists then **`atlas{$source}`**.

# Why should this be improved
It could re-use the tileset which is loaded before.
It eliminates the unnecessary loading.

# Any risks?
No.

# More information

Are you interested in working on a PR for this?  Yes

PR #2348, opened 2023-02-17
feat: Use cached image when creating single source TiledAtlas if available

Hwan-seok:
# Description

Detailed use-case examples are described in #2337.

### As-is
- A single source TiledAtlas doesn't check if there is a cached image that is already loaded in the memory.
- It re-loads the image into the memory.

### To-be
- A single source TiledAtlas now uses the cached image if it exists.

## Checklist

## Breaking Change?

## Related Issues
Closes: #2337 


packages/flame_tiled/lib/src/tile_atlas.dart
```diff
@@ -95,9 +95,8 @@ class TiledAtlas {
       // The map contains one image, so its either an atlas already, or a
       // really boring map.
       final tiledImage = imageList.first;
-      final image =
-          (await Flame.images.load(tiledImage.source!, key: key)).clone();
```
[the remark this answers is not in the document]
Hwan-seok: You mean `Flame.images.add(key, image);` is not needed?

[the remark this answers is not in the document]
Hwan-seok: Actually yes, there is no need for adding `atlas{$source}` for the image key.
This is for backward compatibility for someone who is using this as a cache key.
But it can be eliminated of course.

packages/flame_tiled/lib/src/tile_atlas.dart
```diff
@@ -95,9 +95,8 @@ class TiledAtlas {
...
-
+      final image = (await Flame.images.load(tiledImage.source!)).clone();
+      Flame.images.add(key, image);
```
spydon, line 99: There will also be two image objects added to the cache now right?

Hwan-seok: Yes, you are right. But they share an image reference.
At present, `TiledAtlas` has a rule for storing keys.
 
 1. No image: `atlas{empty}`
 2. A image: `atlas{$source}`
 3. Images: `atlas{$source1,$source2}`

To keep this rule, I added this: `Flame.images.add(key, image);`.
But If there is no need to keep this rule, the line can be eliminated.

I prefer removing this but it's totally on your call.

spydon: Since the image is cloned before it is added to the cache again they don't share the reference.
I don't think there is a need to keep that rule, what do you think? :)

Hwan-seok: > Since the image is cloned before it is added to the cache again they don't share the reference.
Oh, I was totally missed it. Yes, you are right again!
> I don't think there is a need to keep that rule, what do you think? :)
I absolutely agree with you since there is a `key` property to `TiledAtlas` so that it could get a cached image if anyone wanted.

Hwan-seok: Also, do you think the `clone` is needed?

spydon: Not sure, feels like there must have been a reason for someone to put the clone there? Can you check who did it and we can ask? I'm currently on my phone.

Hwan-seok: Sure.

Hi @jtmcdole, can I ask a reason for the `clone` exists? I could roughly assume but curious about the exact reason.

Hwan-seok: @spydon Maybe he's unavailable now. We should leave the `clone` and ask him later :)

spydon (approved):

packages/flame_tiled/lib/src/tile_atlas.dart
jtmcdole: You can keep using the image so long as it is in the cache. We don't go back to the cache each time we want to reference the image and instead use ` final Image? atlas;`.  If the cache is cleared for any reason, it disposes of the image and we now have an invalid image.

I remember there being a crash due to this... there **should** have been a test as well :)

Hwan-seok: Probably some of your use cases are evicting the cache. Thanks for the answer 😄

Issue #1938, opened 2022-09-24
Flame v2 breaking changes


spydon:
# Flame v2 breaking changes

This issue doesn't mean that we are soon to release v2, it's not even on the horizon yet.
In here we'll track breaking changes that we would want to do for Flame v2 so that they are not forgotten.

- [ ] Rename `HasGameRef` to `HasGameReference`.
- [ ] Convert the use of `ShapeComponent` in the collision detection to use pure shapes instead.
- [ ] Remove the `Camera` from `FlameGame` in favor on the `CameraComponent` (and remove `BareFlameGame`).

Once we are ready to start working at v2 we'll convert this task list into issues.

MuZiLee:
Looking forward to the v2

DevinXiaoBai:
Can you consider building wechat small game running environment

spydon:
> Can you consider building wechat small game running environment

This is unrelated to V2 breaking changes, please open a separate issue for this.

DevinXiaoBai:
ok，Thank you for your reply




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "flame-engine/flame"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2023年2月9日(星期四) 下午3:06
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [flame-engine/flame] Flame v2 breaking changes (Issue #1938)





  
Can you consider building wechat small game running environment
  
This is unrelated to V2 breaking changes, please open a separate issue for this.

guyluz11:
> Once we are ready to start working at v2 we'll convert this task list into issues.

What holding us back?, there is only one task that got converted into an issue

NashIlli:
Hello, @spydon! Is Version 2 included in the route of the following months? Thaaanks! :)

spydon:
> Hello, @spydon! Is Version 2 included in the route of the following months? Thaaanks! :)

Hi, unfortunately not, we have not even decided on paper how the big architectural changes will look.

marzoukali:
Any updates on when to expect version 2 or do you need any help getting it back on track? We can share with the community the challenges and build separate team dedicated for it. I suggest if this is still in plan, we will need to have it's effort as separate from the current.

ufrshubham:
> we will need to have it's effort as separate from the current

[...]

PR #2350, opened 2023-02-19
feat: Added AlignComponent layout component

### LINKED ISSUE #1421: Add onParentResize lifecycle method

### LINKED ISSUE #2302: Create LayoutComponent classes
# Proposal

`LayoutComponent` will be the base class for a family of utility classes which handle the layout of their children. This would be quite similar to Flutter layout widgets such as `Align`, `Column`, `Center`, `Container`, `Padding`, `SizedBox`, etc.

The purpose of these new components is to assist with constructing UIs within a game: dialogue boxes, HUD, informational screens, in-game UIs such as dialogue or trade windows, etc.

PR #1971 already attempts to add two such layout classes: `Row` and `Column` -- this proposal is to have more of them, all under a single umbrella of `LayoutComponent`.

For example, suppose you're making a game and you want to display current score in the top right corner. Then, you should be able to declare it like this:
```dart
camera.viewport.add(
  Padding(
    padding: EdgeInsets.all(20),
    children: [
      Align(
        anchor: Anchor.topRight,
        child: ScoreComponent(),
      ),
    ],
  ),
);
```
After this, the game would take care to properly position the `ScoreComponent` and move it whenever the game resizes.

## Implementation

`LayoutComponent` would rely on the `onParentResize` lifecycle method (#1421).

The tree structure can be either normal (the target is added as a child of the LayoutComponent), or flat (the target is added to the parent directly, and layout component is a sibling). I'm not sure which option is better here.

st-pasha:
# Description

- This PR adds first **layout component**: `AlignComponent`, an equivalent of [Align](https://api.flutter.dev/flutter/widgets/Align-class.html) widget in Flutter.
- `AlignComponent` sizes itself to its parent, and then keeps its child aligned to the specified anchor within its own bounding box.
- Also adding `onParentResize()` lifecycle method, which is similar to `onGameResize`, but fires whenever the parent of the current component changes its size for any reason. (`FlameGame` is assumed to have the size `canvasSize`, and will invoke `onParentResize` whenever the canvas size changes).
- Additional layout components are planned to be added in future PRs.

## Checklist

## Breaking Change?

## Related Issues

Closes #2302
Closes #1421


examples/lib/stories/layout/align_component.dart
```diff
@@ -0,0 +1,94 @@
...
+    the shape of a diamond.
+    
+    The arrangement will remain intact if you change the window size.
+  ''';
+
+  @override
+  void onLoad() {
+    addAll([
+      AlignComponent(
+        child: CircleComponent(radius: 40)
```
spydon, line 19: I think this would be clearer if the `children` argument was used instead of the cascading adds

packages/flame/lib/src/components/position_component.dart
```diff
@@ -184,8 +185,16 @@ class PositionComponent extends Component
...
   /// recommended. Instead, in order to make the [PositionComponent] larger
   /// or smaller, change its [scale].
+  @override
   NotifyingVector2 get size => _size;
-  set size(Vector2 size) => _size.setFrom(size);
+
+  @override
+  set size(Vector2 size) {
+    _size.setFrom(size);
+    if (hasChildren) {
```
spydon, line 194: Does this check optimize anything? The foreach should also just do a check if it is empty right?

packages/flame/lib/src/camera/viewport.dart
```diff
@@ -66,6 +66,9 @@ abstract class Viewport extends Component
       camera.viewfinder.onViewportResize();
     }
     onViewportResize();
+    if (hasChildren) {
```
spydon, line 69: Does this check optimize anything? The foreach should also just do a check if it is empty right?

spydon, line 69: Does this check optimize anything? The foreach should also just do a check if it is empty right?

packages/flame/lib/src/layout/align_component.dart
```diff
@@ -0,0 +1,124 @@
...
+/// PlayerSprite().add(
+///   AlignComponent(
+///     child: HealthBar()..anchor = Anchor.bottomCenter,
+///     alignment: Anchor.topCenter,
+///     keepChildAnchor: true,
+///   ),
+/// );
+/// ```
+class AlignComponent extends PositionComponent {
+  /// Creates component that keeps its [child] positioned according to the
```
spydon, line 51:
```suggestion
  /// Creates a component that keeps its [child] positioned according to the
```

spydon (commented):
Very well written docs!

packages/flame/lib/src/components/position_component.dart
st-pasha: `hasChildren` avoids instantiating a `ComponentSet` object in case it is null.

packages/flame/lib/src/layout/align_component.dart
st-pasha: done

examples/lib/stories/layout/align_component.dart
st-pasha: done

spydon (approved):

examples/lib/stories/layout/align_component.dart
```diff
@@ -0,0 +1,88 @@
...
+              Vector2.all(25),
+              EffectController(
+                infinite: true,
+                duration: 0.75,
+                reverseDuration: 0.5,
+              ),
+            ),
+            AlignComponent(
+              alignment: Anchor.topCenter,
+              child: CircleComponent(radius: 10, anchor: Anchor.bottomCenter),
```
erickzanardo, line 32: Nit: this and the other occurrences could benefit from a trailing comma

packages/flame/lib/src/camera/viewport.dart
luanpotter: +1

examples/lib/stories/components/game_in_game_example.dart
```diff
@@ -36,9 +36,10 @@ class GameChangeTimer extends TimerComponent
...
     final child = gameRef.draggablesGame.square;
-    final newParent = child.parent == gameRef.draggablesGame
-        ? gameRef.composedGame.parentSquare
-        : gameRef.draggablesGame;
-    child.changeParent(newParent);
+    if (child.parent == gameRef.draggablesGame) {
+      child.parent = gameRef.composedGame.parentSquare;
+    } else {
+      child.parent = gameRef.draggablesGame;
+    }
```
luanpotter, line 43: is this related? we shouldn't use the method changeParent?

packages/flame/lib/src/components/position_component.dart
luanpotter: could we do that check inside the `.children` getter instead?
that way we avoid duplicating this all over & everyone gets the performance boost (even end users)

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -182,4 +184,9 @@ class FlameGame extends ComponentTreeRoot with Game {
       }
     }
   }
+
+  @override
+  set size(Vector2 value) {
+    throw UnsupportedError('The size of FlameGame cannot be set explicitly');
+  }
```
luanpotter, line 191: would it be beneficial to create another interface fot things that can only get the size?
like:
```dart
abstract class ReadableSizeProvider {
  Vector2 get size;
}
abstract class SizeProvider extends ReadableSizeProvider {
  // ...
}
```
then each class can implement specifically what they have, and things are a bit more type safe

packages/flame/lib/src/layout/align_component.dart
```diff
@@ -0,0 +1,124 @@
...
+/// The component requires a single [child], which will be the target of this
+/// component's alignment. Of course, other children can be added to this
+/// component too, but only the initial [child] will be aligned.
+///
+/// The [alignment] parameter describes where the child should be placed within
+/// the current component. For example, if the [alignment] is `Anchor.center`,
+/// then the child will be centered.
+///
+/// Normally, this component's size will match the size of its parent. However,
+/// if you provide properties [widthFactor] or [heightFactor], then the size of
```
luanpotter, line 19: should it be a factor: Vector2?

luanpotter: or scaleFactor

packages/flame/lib/src/layout/align_component.dart
```diff
@@ -0,0 +1,124 @@
...
+/// );
+/// ```
+///
+/// By default, the child's anchor is set equal to the [alignment] value. This
+/// achieves traditional alignment behavior: for example, the center of the
+/// child will be placed at the center of the current component, or bottom
+/// right corner of the child can be placed in the bottom right corner of the
+/// component. However, it is also possible to achieve more extravagant
+/// placement by giving the child a different anchor and setting
+/// [keepChildAnchor] to true. For example, if you set `alignment` to
```
luanpotter, line 38: any reason you would _not_ want to keepChildAnchor?
why would you set an anchor to the child if you want to ignore it?

st-pasha: Because when you receive a `child`, that child will have some non-null `anchor` value, and there is no way to know whether that anchor was set manually by the user or automatically by the framework.

packages/flame/lib/src/layout/align_component.dart
st-pasha: I tried to mimic here the api of the [Align](https://api.flutter.dev/flutter/widgets/Align-class.html) widget in Flutter.

Plus, having them as two separate parameters makes it easy to specify only one while leaving the other as `null`. For example, if you set `heightFactor: 1`, then it means a component that takes its height from the child, but the width from the parent.

examples/lib/stories/components/game_in_game_example.dart
st-pasha: There was something weird going on with this snippet: my Dart decided that the common superclass between these RHSs is `Object`. The `changeParent` method can still be used, I reverted this change to almost the same as it was before.

packages/flame/lib/src/layout/align_component.dart
```diff
@@ -0,0 +1,124 @@
+import 'package:flame/src/anchor.dart';
+import 'package:flame/src/components/position_component.dart';
+import 'package:flame/src/effects/provider_interfaces.dart';
+import 'package:flutter/widgets.dart';
+import 'package:vector_math/vector_math_64.dart';
+
+/// **AlignComponent** is a layout component that positions its child within
+/// itself using relative placement. It is similar to Flutter's [Align] widget.
+///
+/// The component requires a single [child], which will be the target of this
```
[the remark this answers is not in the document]
st-pasha: It's should still be possible to add other children to this component, though they wouldn't be aligned. For example, you can add an `Effect` that would move the `AlignComponent` across the screen, or even change the `alignment` value over time.

packages/flame/lib/src/layout/align_component.dart
```diff
@@ -0,0 +1,124 @@
...
+  AlignComponent({
+    required this.child,
+    required Anchor alignment,
+    this.widthFactor,
+    this.heightFactor,
+    this.keepChildAnchor = false,
+  }) {
+    this.alignment = alignment;
+    add(child);
+  }
```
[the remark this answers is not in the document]
st-pasha: `this.alignment = alignment` ensures that the setter for the property is called.

examples/lib/stories/layout/align_component.dart
st-pasha: done

packages/flame/lib/src/game/flame_game.dart
st-pasha: Good idea

spydon (approved):

examples/lib/stories/components/game_in_game_example.dart
luanpotter: sg, I was just curious if it was being removed or something

packages/flame/lib/src/layout/align_component.dart
luanpotter: ah! got it :)

packages/flame/lib/src/components/position_component.dart
luanpotter: (can be a separate PR of course)

luanpotter (approved):
LGTM, nice addition!

packages/flame/lib/src/components/position_component.dart
st-pasha: The `.children` property returns a `ComponentSet` right now, which means it must create a component set in case it is currently missing. We could change it to return a `ComponentSet?` instead, but that would be a major breaking change I'm afraid -- so perhaps we could add that to our 2.0 todo list.

spydon: Haha I think 3 people commented about this on this PR, maybe it would be better as nullable indeed, let's discuss that for V2 at least. :)

st-pasha: I added a note about this in #1938

rivella50:
Are there plans to implement additional layout components (mainly for row and column support)?
As you probably remember i did some [work](https://github.com/flame-engine/flame/pull/1971) on this topic last year and perhaps it could be used here.

spydon:
@rivella50 There is definitely some stuff in there that could be re-used, the main difference now is that the Row and Column components should be based on the `AlignComponent`. If you want to have a go at this I can assign you to the issue? :)

rivella50:
@spydon I pass this time, thanks. When i worked on it the last time it was quite frustrating and seemed to turn into a neverending story since all the time another guy appeared on the stage and suggested new things and changes.
This just took me too much time and there was no end foreseeable.

spydon:
@rivella50 makes sense, must have been frustrating!

PR #2352, opened 2023-02-20
refactor: Component rebalancing is now performed via a global queue

### LINKED ISSUE #2263: RouterComponent.pushOverlay() causes concurrent modification exception.

st-pasha:
# Description

This PR ensures that all component rebalancing operations are resolved from a single location, after the `update` stage but before the `render` stage (thus, components may get reordered during the update, and these changes will go into effect during the rendering step on the same game tick).

This also fixes the problem where the child changing the priorities of its parent would cause a ConcurrentModificationError.

A number of methods that were used to handle rebalancing are now marked as deprecated. From the user's perspective, the only API they should be using is the `.priority` setter.

## Checklist

## Breaking Change?

## Related Issues

Closes #2263

### DIFF
--- a/packages/flame/lib/src/components/core/component.dart
+++ b/packages/flame/lib/src/components/core/component.dart
@@ -489,7 +489,6 @@ class Component {
   /// children according to their [priority] order, relative to the
   /// priority of the direct siblings, not the children or the ancestors.
   void updateTree(double dt) {
-    _children?.updateComponentList();
     update(dt);
     _children?.forEach((c) => c.updateTree(dt));
   }
@@ -728,28 +727,26 @@ class Component {
   int get priority => _priority;
   int _priority;
   set priority(int newPriority) {
-    _priority = newPriority;
-    final game = findGame();
-    if (game != null && _parent != null) {
-      (game as FlameGame).enqueueRebalance(_parent!);
+    if (_priority != newPriority) {
+      _priority = newPriority;
+      final game = findGame();
+      if (game != null && _parent != null) {
+        (game as FlameGame).enqueueRebalance(_parent!);
+      }
     }
-
-    // if (parent == null) {
-    //   _priority = newPriority;
-    // } else {
-    //   parent!.children.changePriority(this, newPriority);
-    // }
   }
 
   /// Usually this is not something that the user would want to call since the
   /// component list isn't re-ordered when it is called.
   /// See FlameGame.changePriority instead.
+  @Deprecated('Will be removed in 1.8.0. Use priority setter instead.')
   void changePriorityWithoutResorting(int priority) => _priority = priority;
 
   /// Call this if any of this component's children priorities have changed
   /// at runtime.
   ///
   /// This will call [ComponentSet.rebalanceAll] on the [children] ordered set.
+  @Deprecated('Will be removed in 1.8.0.')
   void reorderChildren() => _children?.rebalanceAll();
 
   //#endregion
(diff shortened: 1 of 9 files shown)


packages/flame/lib/src/components/core/component.dart
```diff
@@ -728,22 +727,26 @@ class Component {
...
+      if (game != null && _parent != null) {
+        (game as FlameGame).enqueueRebalance(_parent!);
+      }
     }
   }
 
   /// Usually this is not something that the user would want to call since the
   /// component list isn't re-ordered when it is called.
   /// See FlameGame.changePriority instead.
+  @Deprecated('Will be removed in 1.8.0. Use priority setter instead.')
```
spydon, line 742: 🥳
Nice, so much cleaner without this!

spydon (approved):
Good stuff! This should make it quite easy to introduce global priorities too right?

packages/flame/test/components/priority_test.dart
```diff
@@ -1,27 +1,11 @@
...
-    super.reorderChildren();
-  }
-
-  void assertCalled(int n) {
-    expect(callCount, n);
-    callCount = 0;
-  }
-}
+import '../custom_component.dart';
+// ignore_for_file: deprecated_member_use_from_same_package
```
luanpotter, line 8: what does it use that is deprecated?

luanpotter (approved):
LGTM with one comment

packages/flame/test/components/priority_test.dart
st-pasha: I believe it's mostly calls like `game.children.changePriority(firstComponent, 11);`. Some time later we will refactor the test file to use more straightforward `firstComponent.priority = 11;`, but for now I wanted to leave everything mostly as-is to ensure that the change is non-breaking.

PR #2358, opened 2023-02-21
feat: Make `limit` field mutable in the `Timer` class

### LINKED ISSUE #2357: Being able to add more time to a `Timer`
# What could be improved

Currently, it isn't possible to change the elapsed time of a `Timer`. It's also not possible to change the limit.
I think this could easily be improved by removing the `final` keyword of the `limit` variable in the `Timer` class.

# Why should this be improved

I use timers in my game as a countdown before the player dies in a level. But some objects should act as bonus that give some time to the player.

What I've done so far is that I've created a `MutableTimer` class that allows me to change the limit :

<details>

```dart
/// A [Timer] that allows to change its limit.
class MutableTimer extends Timer {
  /// The current timer limit.
  double _limit;

  /// Creates a new mutable timer instance.
  MutableTimer(
    double limit, {
    super.onTick,
    super.repeat,
    super.autoStart,
  })  : _limit = limit,
        super(limit);

  @override
  double get limit => _limit;

  /// Changes the current timer limit.
  set limit(double limit) => _limit = limit;
}
```

</details>

but it really is a workaround.

# Any risks?

Not to my knowledge.

Skyost:
# Description

As stated in #2357, the field `limit` of the class `Timer` had no reason to be final. By removing this keyword, we make it possible to add more time to a given `Timer` object (so that they can be used as a countdown for example).

## Checklist

## Breaking Change?

## Related Issues

Closes #2357.

### DIFF
--- a/packages/flame/lib/src/timer.dart
+++ b/packages/flame/lib/src/timer.dart
@@ -5,6 +5,10 @@ import 'dart:ui';
 /// interval like events.
 ///
 /// Timer auto-starts by default.
+///
+/// _NOTE :_ You can change the [limit], but keep in mind that the timer
+/// won't start automatically if the limit is raised and the timer currently
+/// is stopped.
 class Timer {
   double limit;
   VoidCallback? onTick;
--- a/packages/flame/lib/src/timer.dart
+++ b/packages/flame/lib/src/timer.dart
@@ -6,7 +6,7 @@ import 'dart:ui';
 ///
 /// Timer auto-starts by default.
 ///
-/// _NOTE :_ You can change the [limit], but keep in mind that the timer
+/// NOTE: You can change the [limit], but keep in mind that the timer
 /// won't start automatically if the limit is raised and the timer currently
 /// is stopped.
 class Timer {
--- a/packages/flame/lib/src/timer.dart
+++ b/packages/flame/lib/src/timer.dart
@@ -6,7 +6,7 @@ import 'dart:ui';
 ///
 /// Timer auto-starts by default.
 class Timer {
-  final double limit;
+  double limit;
   VoidCallback? onTick;
   bool repeat;
   double _current = 0;
--- a/packages/flame/lib/src/timer.dart
+++ b/packages/flame/lib/src/timer.dart
@@ -19,7 +19,7 @@ class Timer {
     bool autoStart = true,
   }) : _running = autoStart;
 
-  /// The current amount of ms that has passed on this iteration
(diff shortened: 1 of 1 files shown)


Skyost:
I would like to add a little note here : in the description of the `current` getter, it is written :

> The current amount of ms that has passed on this iteration

Wouldn't it rather be ``the number of **seconds** passed on this iteration'' ?

spydon:
> Wouldn't it rather be ``the number of **seconds** passed on this iteration'' ?

Agree!

Skyost:
@spydon Should I open a separate PR ?

spydon:
> @spydon Should I open a separate PR ?

Nah, you can do it in this one.

spydon (approved):

luanpotter (commented):
Just mutating that value won't work if `_running` is set to false.
Wouldn't that be confusing?
Might be better to expose a method to "reconfigure" the timer and make all state variables private

spydon:
> Just mutating that value won't work if `_running` is set to false.
Wouldn't that be confusing?

Is it really confusing? That is what I would expect would happen if I used this at least, would you have expected the timer to start again?

> Might be better to expose a method to "reconfigure" the timer and make all state variables private

I would just update the `_running` value in the setter, if the expected behavior would be that it started again if it was lower than the limit.

luanpotter:
not sure what I would expect haha. just crossed my mind. if you think it's clear as is, no problem for me :)

spydon:
> not sure what I would expect haha. just crossed my mind. if you think it's clear as is, no problem for me :)

Maybe you can add a dartdoc line about it @Skyost? :)
Just describing that the timer won't start automatically if the `limit` is raised and the timer currently is stopped.

Skyost:
I've left a little note in the header of the `Timer` class. I've reused your words 🙂

packages/flame/lib/src/timer.dart
```diff
@@ -5,8 +5,12 @@ import 'dart:ui';
 /// interval like events.
 ///
 /// Timer auto-starts by default.
+///
+/// _NOTE :_ You can change the [limit], but keep in mind that the timer
```
spydon, line 9:
```suggestion
/// NOTE: You can change the [limit], but keep in mind that the timer
```

spydon (commented):
Just one final comment

packages/flame/lib/src/timer.dart
Skyost: 👍

PR #2359, opened 2023-02-21
docs: Fixing components sub-directory path

jonatasemidio:
At the *Other classes* session, when it is mentioned that we need to create the `components` sub-directory, the tutorial is guiding us to add it in the root, but on the structure below, it is possible to see that it should be created inside the `libs` folder instead.

It seems to be better to be more explicit in this paragraph.

# Description

## Checklist

- [ x] I have followed the [Contributor Guide] when preparing my PR.
- [ x] I have updated/added tests for ALL new/updated/fixed functionality.
- [ x] I have updated/added relevant documentation in `docs` and added dartdoc comments with `///`.
- [ x] I have updated/added relevant examples in `examples` or `docs`.

## Breaking Change?

## Related Issues

### DIFF
--- a/doc/tutorials/klondike/step2.md
+++ b/doc/tutorials/klondike/step2.md
@@ -85,8 +85,8 @@ As we already mentioned in the previous chapter, our game mainly consists of
 will defer implementation of that class to the next chapter.
 
 For now, let's create the container classes, as shown on the sketch. These are:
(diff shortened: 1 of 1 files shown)


spydon (approved):
Thanks for your contribution!
Fun fact; you are the 150th individual contributing to Flame. :)

Issue #2353, opened 2023-02-20
Official Riverpod support


devberkay:
# Problem to solve
Games need complex state management solutions and Riverpod is the core dependency for many flutter developers like me. 

# Proposal
It would be amazing if we could see a collabaration in the future as it was done with flame_bloc package.

# More information
@rrousselGit

spydon:
Lucky for you there is already a package for this, which we soon will be reviewing and bringing in to the monorepo.
https://github.com/markvideon/flame_riverpod
https://pub.dev/packages/flame_riverpod

spydon:
@markvideon do you know when you'll put up a PR for this? :)

rrousselGit:
Making an official Riverpod implementation involves more work.

As is, it's not going to work with any tools like riverpod_lint & upcoming.

spydon:
> Making an official Riverpod implementation involves more work.
> 
> As is, it's not going to work with any tools like riverpod_lint & upcoming. 

Well depends on what is meant with official here I guess, if he means an official bridge package from the Flame side (like flame_bloc) or official from the Riverpod side.

rrousselGit:
I assume it should be official for both sides at the same time. Otherwise it's not truly official right? :)
Where the project sits doesn't really matter. I'm fine with it being in the flame repository.

But it should be compatible with riverpod_lint, which is a core Riverpod package.  Nothing blocking, but I'm sure it'd be inconvenient for Riverpod users of flame's consumer broke the various refactors & lints

spydon:
> I assume it should be official for both sides at the same time. Otherwise it's not truly official right? :)

True true. :)

> But it should be compatible with riverpod_lint, which is a core Riverpod package. Nothing blocking, but I'm sure it'd be inconvenient for Riverpod users if flame's consumer broke the various refactors & lints

Alright, let's create an issue for that once the initial version is merged into the monorepo.
I ran the linter on the the example that was provided in the repository and there were no warnings reported in there, but you're probably right that refactors wouldn't work like they should.

markvideon:
> @markvideon do you know when you'll put up a PR for this? :)

I'll try to get to it this week

markvideon:
@spydon can probably close this one?

PR #2367, opened 2023-02-25
feat: Integration of flame_riverpod

markvideon:
# Description

Previously discussed with @spydon and referenced in issue #2353, flame_riverpod is to be integrated into the monorepo

## Checklist

## Breaking Change?

## Related Issues
Contribution toward #2353


markvideon:
The example Flutter project contains a widget test that accesses lifecycle methods (eg processQueue), which is currently annotated as `protected`. Perhaps it could be `visibleForTesting`? 

The widget test in question is a standard Flutter widget test that compares the contents of a Flutter Text Widget against a Flame Text Component for equality. This is to validate that both widgets are reading from the StreamProvider defined in the project.

st-pasha (dismissed):
Could you please remove all the files that are not part of the library? Such as `linux/`, `macos/`, `web/`, and `windows/` folders.

spydon:
> The example Flutter project contains a widget test that accesses lifecycle methods (eg processQueue), which is currently annotated as `protected`. Perhaps it could be `visibleForTesting`? 

It can be both, so you can just add that meta tag there and it should be fine.

packages/flame_riverpod/example/.gitignore
```diff
@@ -0,0 +1,44 @@
...
+# Symbolication related
+app.*.symbols
+
+# Obfuscation related
+app.*.map.json
+
+# Android Studio will place build artifacts here
+/android/app/debug
+/android/app/profile
+/android/app/release
```
spydon, line 44: We have a global `.gitignore` so this file shouldn't be needed

packages/flame_riverpod/example/README.md
```diff
@@ -0,0 +1,16 @@
+# example
```
spydon, line 1: Update example readme with just a short comment about what it is showing, one or two sentences is enough.

packages/flame_riverpod/example/lib/main.dart
```diff
@@ -0,0 +1,156 @@
...
+
+class RiverpodGameWidget extends ConsumerStatefulWidget {
+  const RiverpodGameWidget.readFromProvider({super.key})
+      : uninitialisedGame = null;
+  const RiverpodGameWidget.initialiseWithGame({
+    super.key,
+    required this.uninitialisedGame,
+  });
+
+  final FlameGame Function(WidgetRef ref)? uninitialisedGame;
```
spydon, line 102:
```suggestion
  final FlameGame Function(WidgetRef ref)? uninitializedGame;
```
We're trying to keep everything to American english for consistency

packages/flame_riverpod/example/pubspec.yaml
```diff
@@ -0,0 +1,85 @@
+name: example
```
spydon, line 1:
```suggestion
name: flame_riverpod_example
```

packages/flame_riverpod/example/pubspec.yaml
```diff
@@ -0,0 +1,85 @@
...
+description: A new Flutter project.
```
spydon, line 2:
```suggestion
description: Showcasing the flame_riverpod functionality.
```

packages/flame_riverpod/example/pubspec.yaml
```diff
@@ -0,0 +1,85 @@
...
+# followed by an optional build number separated by a +.
+# Both the version and the builder number may be overridden in flutter
+# build by specifying --build-name and --build-number, respectively.
+# In Android, build-name is used as versionName while build-number used as versionCode.
+# Read more about Android versioning at https://developer.android.com/studio/publish/versioning
+# In iOS, build-name is used as CFBundleShortVersionString while build-number is used as CFBundleVersion.
+# Read more about iOS versioning at
+# https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html
+# In Windows, build-name is used as the major, minor, and patch parts
+# of the product and file versions while build-number is used as the build suffix.
```
spydon, line 19: You can remove all the comments in this file

packages/flame_riverpod/example/pubspec.yaml
```diff
@@ -0,0 +1,85 @@
...
+  sdk: '>=2.18.5 <3.0.0'
+
+# Dependencies specify other packages that your package needs in order to work.
+# To automatically upgrade your package dependencies to the latest versions
+# consider running `flutter pub upgrade --major-versions`. Alternatively,
+# dependencies can be manually updated by changing the version numbers below to
+# the latest version available on pub.dev. To see which dependencies have newer
+# versions available, run `flutter pub outdated`.
+dependencies:
+  flame: ^1.5.0
```
spydon, line 32:
```suggestion
  flame: ^1.6.0
```

packages/flame_riverpod/example/pubspec.yaml
```diff
@@ -0,0 +1,85 @@
...
+  flame_riverpod:
+    path: ..
```
spydon, line 34:
```suggestion
  flame_riverpod: ^4.0.0+2
```
This will be automatically updated when we release with melos

packages/flame_riverpod/example/pubspec.yaml
```diff
@@ -0,0 +1,85 @@
...
+  #       - asset: fonts/Schyler-Italic.ttf
+  #         style: italic
+  #   - family: Trajan Pro
+  #     fonts:
+  #       - asset: fonts/TrajanPro.ttf
+  #       - asset: fonts/TrajanPro_Bold.ttf
+  #         weight: 700
+  #
+  # For details regarding fonts from package dependencies,
+  # see https://flutter.dev/custom-fonts/#from-packages
```
spydon, line 85:
```suggestion
```

packages/flame_riverpod/example/test/widget_test.dart
```diff
@@ -0,0 +1,67 @@
...
+    // Find the game widget
+    final gameWidgetFinder = find.descendant(
+      of: riverpodGameWidgetFinder,
+      matching: find.byType(GameWidget<FlameGame>),
+    );
+    expect(gameWidgetFinder, findsOneWidget);
+
+    final gameWidget = widgetTester.widget(gameWidgetFinder);
+
+    // Initialise the game lifecycle
```
spydon, line 48:
```suggestion
    // Initialize the game lifecycle
```

packages/flame_riverpod/example/test/widget_test.dart
```diff
@@ -0,0 +1,67 @@
...
+    (gameWidget as GameWidget<FlameGame>).game?.lifecycle.processQueues();
+    expect(gameWidget.game?.lifecycle.hasPendingEvents, false);
+    expect(gameWidget.game?.children.isNotEmpty ?? false, true);
```
spydon, line 51: @st-pasha can you give some pointers how to migrate this to the new lifecycle code?

packages/flame_riverpod/lib/src/has_component_ref_mixin.dart
```diff
@@ -0,0 +1,41 @@
...
+import 'package:flame_riverpod/src/component_ref.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+mixin HasComponentRef on Component {
+  ComponentRef get ref => _reference;
+  static late ComponentRef _reference;
+  static set widgetRef(WidgetRef value) => _reference = ComponentRef(value);
+
+  // ignore: prefer_final_fields
+  List<ProviderSubscription> _subscriptions = [];
```
spydon, line 11: Why not keep it final?

packages/flame_riverpod/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
+name: flame_riverpod
+description: Model classes and an example project for using Riverpod in conjunction with Flame.
```
spydon, line 2:
```suggestion
description: Helpers for using Riverpod in conjunction with Flame.
```
Taken from the barrel file, since the example project has it's own description field it doesn't need to be mentioned here.

packages/flame_riverpod/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
...
+version: 4.0.0+2
+homepage: https://markvideon.dev
```
spydon, line 4:
```suggestion
homepage: https://github.com/flame-engine/flame/tree/main/packages/flame_riverpod
```
You can put your own page in the readme if you want though :)

packages/flame_riverpod/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+environment:
+  sdk: '>=2.18.5 <3.0.0'
+  flutter: ">=1.17.0"
+
+dependencies:
+  flame: ^1.5.0
```
spydon, line 11:
```suggestion
  flame: ^1.6.0
```

packages/flame_riverpod/README.md
```diff
@@ -0,0 +1,59 @@
...
+    listen(countingStreamProvider, (p0, p1) {
+      if (p1.hasValue) {
+        currentValue = p1.value!;
+        textComponent.text = '$currentValue';
+      }
+    });
+  }
+}
+
+```
```
spydon, line 59:
```suggestion
```

# Credits

[Mark Videon](https://markvideon.dev) for the initial groundwork and implementation.

```

Something like that maybe? :)

Comment (spydon) [reply packages/flame_riverpod/README.md:59]: Haha aaah, can't suggest changes that contains three back ticks since it will exit the change suggestion block.

This was the suggestion to append to the bottom:
```

# Credits

[Mark Videon](https://markvideon.dev) for the initial groundwork and implementation.

```

File: packages/flame_riverpod/example/.metadata [hunk 17]
```diff
@@ -0,0 +1,45 @@
...
+
+  # User provided section
+
+  # List of Local paths (relative to this file) that should be
+  # ignored by the migrate tool.
+  #
+  # Files that are not part of the templates will be ignored by default.
+  unmanaged_files:
+    - 'lib/main.dart'
+    - 'ios/Runner.xcodeproj/project.pbxproj'
```
st-pasha, line 45: This file can be removed as well.

packages/flame_riverpod/README.md
```diff
@@ -0,0 +1,59 @@
+Helpers for using Riverpod in conjunction with Flame - to share state from
+the game into other parts of your application, or from other parts of your
+application into your game.
```
st-pasha, line 3: It would be better to start with a brief description of what `riverpod` is (not all readers will be familiar with it), and then explain how it would fit into the Flame universe.

AFAIU, the purpose of riverpod is to maintain user-defined state, and force the widgets that depend on that state to automatically rebuild when the state changes. It is not clear from the description how this interoperates with Flame which (1) has components instead of widgets, (2) components don't need to be rebuilt.

packages/flame_riverpod/example/test/widget_test.dart
st-pasha: There should be no need to access the lifecycle queue directly - as long as the game was properly added to a GameWidget, and there was a `tester.pump()`, it should resolve all its lifecycle events automatically.

packages/flame_riverpod/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
+name: flame_riverpod
+description: Helpers for using Riverpod - a reactive caching and data-binding framework, 
+  in conjunction with Flame. The ComponentRef is used as an intermediary between Flame and
+  the Widget tree in order to provide state to Flame Components.
+version: 4.0.0+2
```
luanpotter, line 5: why start on this specific version?

packages/flame_riverpod/lib/src/has_component_ref_mixin.dart
```diff
@@ -0,0 +1,44 @@
...
+    ProviderListenable<T> provider,
+    void Function(T?, T) onChange, {
+    void Function(Object error, StackTrace stackTrace)? onError,
+  }) {
+    _subscriptions.add(
+      ref.listenManual<T>(
+        provider,
+        (p0, p1) {
+          onChange(p0, p1);
+        },
```
luanpotter, line 31: can this just be:
```suggestion
        onChange,
```

packages/flame_riverpod/lib/src/has_component_ref_mixin.dart
```diff
@@ -0,0 +1,44 @@
...
+  ComponentRef get ref => _reference;
+  static late ComponentRef _reference;
+  static set widgetRef(WidgetRef value) => _reference = ComponentRef(value);
+
+  final List<ProviderSubscription> _subscriptions = [];
+
+  @override
+  Future<void> onLoad() async {
+    await super.onLoad();
+  }
```
luanpotter, line 19: this overload is not necessary?

packages/flame_riverpod/lib/flame_riverpod.dart
```diff
@@ -0,0 +1,7 @@
+/// Helpers for using Riverpod in conjunction with Flame, to share state from
+/// the game into other parts of your application, or from other parts of your
+/// application into your game.
```
luanpotter, line 3: do we have such library/package-level comments elsewhere? is this supported by dartdoc?

packages/flame_riverpod/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
...
+homepage: https://github.com/flame-engine/flame/tree/main/packages/flame_riverpod
+environment:
+  sdk: '>=2.18.5 <3.0.0'
```
luanpotter, line 8: nit:
```suggestion
  sdk: ">=2.18.5 <3.0.0"
```
(see https://github.com/flame-engine/flame/pull/2374)

packages/flame_riverpod/example/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
+name: flame_riverpod_example
+description: Showcasing the flame_riverpod functionality.
+publish_to: 'none' # Remove this line if you wish to publish to pub.dev
```
luanpotter, line 3:
```suggestion
publish_to: 'none'
```

packages/flame_riverpod/example/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
...
+version: 1.0.0+1
+environment:
+  sdk: '>=2.18.5 <3.0.0'
```
luanpotter, line 6:
```suggestion
  sdk: ">=2.18.5 <3.0.0"
```

packages/flame_riverpod/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+When using Flame, we are interacting with components, which are *not* Widgets.
+
+`flame_riverpod` provides the `ComponentRef` and `HasComponentRef` to
+facilitate managing state from Providers in your Flame Game.
+
+
+## Usage
+
+Your Widget that extends `FlameGame` should use the HasComponentRef mixin,
```
luanpotter, line 16: technically the game (that extends FlameGame) is not itself a widget -- it must be wrapped by a GameWidget for that

packages/flame_riverpod/pubspec.yaml
markvideon: This is an existing package, and that version is the current live version

packages/flame_riverpod/README.md
```diff
@@ -0,0 +1,76 @@
+# flame_riverpod
+
+Riverpod is a reactive caching and data-binding framework for Dart & Flutter.
```
st-pasha, line 3: "Riverbed" could be an external link here.

packages/flame_riverpod/example/lib/main.dart
```diff
@@ -0,0 +1,158 @@
...
+  /// which is only called if the [Component] was mounted.
+  @override
+  void onMount() {
+    super.onMount();
+    add(textComponent = TextComponent(position: position + Vector2(0, 27)));
+
+    // "Watch" a provider using [listen] from the [HasComponentRef] mixin.
+    // Watch is not exposed directly as this would rebuild the ancestor that
+    // exposes the [WidgetRef] unnecessarily.
+    listen(countingStreamProvider, (p0, p1) {
```
rrousselGit, line 151: The syntax should be `ref.listen`

The idea is to have the same syntax everywhere

packages/flame_riverpod/example/lib/main.dart
```diff
@@ -0,0 +1,158 @@
...
+            error: (error, stackTrace) => Text('$error', style: textStyle),
+            loading: () => Text('Loading...', style: textStyle),
+          )
+        ],
+      ),
+    );
+  }
+}
+
+class RefExampleGame extends FlameGame with HasComponentRef {
```
rrousselGit, line 81: I'd suggest making a custom base class (like ConsumerFlameGame) instead of a mix-in. That's more in line with what Riverpod does.

packages/flame_riverpod/example/lib/main.dart
```diff
@@ -0,0 +1,158 @@
...
+  RefExampleGame(WidgetRef ref) {
+    // Note: WidgetRef is not stored and is used to build a [ComponentRef],
+    // a wrapper around it.
+    HasComponentRef.widgetRef = ref;
+  }
```
rrousselGit, line 86: This bit looks strange. Why is this needed?

packages/flame_riverpod/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+`flame_riverpod` provides the `ComponentRef` and `HasComponentRef` to
+facilitate managing state from Providers in your Flame Game.
+
+
+## Usage
+
+Your Widget that extends `FlameGame` should use the HasComponentRef mixin,
+and accept a WidgetRef inside its constructor, and should be wrapped in a
+ConsumerStatefulWidget, as per the example.
```
st-pasha, line 18: What should be wrapped in ConsumerStatefulWidget?

packages/flame_riverpod/example/lib/main.dart
```diff
@@ -0,0 +1,158 @@
...
+    with HasComponentRef {
+  late TextComponent textComponent;
+  int currentValue = 0;
+
+  /// [onMount] should be used over [onLoad] to initialize subscriptions,
+  /// cancellation is handled for the user inside [onRemove],
+  /// which is only called if the [Component] was mounted.
+  @override
+  void onMount() {
+    super.onMount();
```
rrousselGit, line 145: Is there no equivalent to a "build" method in components?

We should be able to use ref.watch instead of ref.listen

What about:

```dart
class Example extends MyConsumerComponent {
  final component = TextComponent();

  @override
  void update(ComponentRef ref) {
    component.text = ref.watch(countingStreamProvider)?.value.toString() ?? '<loading>";
  }
}
```

packages/flame_riverpod/README.md
```diff
@@ -0,0 +1,76 @@
...
+When using Flame, we are interacting with components, which are *not* Widgets.
+
+`flame_riverpod` provides the `ComponentRef` and `HasComponentRef` to
+facilitate managing state from Providers in your Flame Game.
+
+
+## Usage
+
+Your Widget that extends `FlameGame` should use the HasComponentRef mixin,
+and accept a WidgetRef inside its constructor, and should be wrapped in a
```
st-pasha, line 17: Is the requirement that the `MyGame` class derived from `FlameGame` must have a single parameter in the constructor which is a `WidgetRef`? This seems overly restrictive, what if the user needs to add some other parameters into the constructor?

packages/flame_riverpod/example/lib/main.dart
```diff
@@ -0,0 +1,158 @@
...
+
+  @override
+  ConsumerState<RiverpodGameWidget> createState() => _RiverpodGameWidgetState();
+}
+
+class _RiverpodGameWidgetState extends ConsumerState<RiverpodGameWidget> {
+  @override
+  void initState() {
+    WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
+      if (widget.uninitializedGame is FlameGame Function(WidgetRef ref)) {
```
rrousselGit, line 114: instead of `is`, should it be `!= null`?

packages/flame_riverpod/example/lib/main.dart
```diff
@@ -0,0 +1,158 @@
...
+        ref
+            .read(riverpodAwareGameProvider.notifier)
+            .set(widget.uninitializedGame!(ref));
```
rrousselGit, line 117: Any chance this logic should be moved to a provider instead of in a widget?
To remove the addPostFrameCallback which will likely be problematic

Like:

```dart
final gameProvider = Provider((ref) => RefExampleGame());
```

packages/flame_riverpod/lib/src/component_ref.dart
```diff
@@ -0,0 +1,43 @@
+import 'package:flutter/cupertino.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+/// A wrapper around [WidgetRef]. Methods that correspond to interactions with
+/// the widget tree, such as [WidgetRef.watch] are not exposed.
```
st-pasha, line 5: Is there a way to explain what this class is without referring to `WidgetRef`? And what does the name mean, "component reference"?

packages/flame_riverpod/lib/src/component_ref.dart
```diff
@@ -0,0 +1,43 @@
...
+class ComponentRef {
+  ComponentRef(this._widgetRef);
+  final WidgetRef _widgetRef;
+
+  BuildContext get context => _widgetRef.context;
```
st-pasha, line 10: What is this context? In Flutter, every widget gets its own build context, so which build context is this?

packages/flame_riverpod/lib/src/has_component_ref_mixin.dart
```diff
@@ -0,0 +1,44 @@
...
+    );
+  }
+
+  @override
+  void onRemove() {
+    for (final listener in _subscriptions) {
+      listener.close();
+    }
+    super.onRemove();
+  }
```
st-pasha, line 43: What happens if the component is later re-added into the component tree?

packages/flame_riverpod/example/lib/main.dart
spydon: The problem with that is that we would have to do that with all potential game classes, so we would get double the amount of game-classes. For example we would need `ConsumerForge2DGame` too then and that class wouldn't really fit in neither `flame_forge2d` or `flame_riverpod`.

rrousselGit: Make a flame_forg2d_riverpod package.
That's what hooks_riverpod is

It's not a huge deal though. But it's likely necessary if you wish to support `ref.watch` (cf https://github.com/flame-engine/flame/pull/2367#discussion_r1121339874)

packages/flame_riverpod/example/lib/main.dart
spydon: Flame works very differently from Flutter since it uses a game-loop, that's why it doesn't have a `build` method. It constantly renders and runs `update` -> `render` 60/120 times a second, so it doesn't need to re-execute anything when a provider updates, so you could pretty much use `ref.read` directly in `update` and it would work, but I suspect that `ref.listen` becomes cleaner and a bit more efficient?

rrousselGit: That's not about the 60/120 times a second. It's about having the rendering logic be more declarative and independent from when the component is mounted.

Using ref.listen you might need a `fireImmediately: true` for instance, and it's very imperative.


There's also the concern of keeping autoDispose providers alive if they are used (which ref.read wouldn't handle).

packages/flame_riverpod/README.md
markvideon: The expectation is that a WidgetRef is provided within the constructor. It need not be the only parameter. IMO a user consuming the package will see this from the usage, but open to suggestions on rewording this description.

packages/flame_riverpod/lib/src/component_ref.dart
markvideon: The name Component Reference exists only to suggest that it is distinct in some sense from a Widget Reference - that the APIs between the two may differ, and to prompt a consuming user to investigate further. 

Specifically, at this time, **watch**, the typically recommended API for subscribing to a value exposed by a provider in Riverpod is not exposed directly. The reason for this is to prevent unnecessary rebuilds

packages/flame_riverpod/lib/src/component_ref.dart
markvideon: Some widget that extends ConsumerWidget or ConsumerStatefulWidget provides its context to a FlameGame.

packages/flame_riverpod/lib/src/has_component_ref_mixin.dart
markvideon: My understanding is that in that event, onMount is called once again.

packages/flame_riverpod/example/lib/main.dart
markvideon: In the initial version of the package I did make a decision along those lines (among others, e.g. including a widget that extended ConsumerStatefulWidget as part of the package), but I became convinced that it may be better left to application developers to define these. I'm not especially committed to one position or another

packages/flame_riverpod/example/lib/main.dart
markvideon: Do you see a use case for interacting with providers within a Flame game component when it is not considered "alive"?

packages/flame_riverpod/example/lib/main.dart
markvideon: The thinking here is that consumers of Riverpod will go straight for watch - if the widget that provides a WidgetRef is a simple shell, would we not be rebuilding that Widget unnecessarily by exposing watch?

packages/flame_riverpod/example/lib/main.dart
markvideon: I think this is a better general solution

packages/flame_riverpod/example/lib/main.dart
rrousselGit: I'm not sure we're talking about the same thing. I'm not talking about having such a provider in the package itself

I meant in this example specifically. I don't think a StateNotifierProvider is needed here

packages/flame_riverpod/example/lib/main.dart
rrousselGit: I do not understand that question sorry

packages/flame_riverpod/example/lib/main.dart
rrousselGit: I have nothing against the abstraction, it makes sense.

My issue is more about:
- Why does RefExampleGame take a WidgetRef has parameter when it use the ref mixin?
- Why are we setting a static `HasComponentRef.widgetRef` variable? I would expect the mixin to create a custom ref on its own rather than rely on WidgetRef

packages/flame_riverpod/README.md
st-pasha: Is `WidgetRef` even necessary? Why can't `ComponentRef` talk to the `riverpod` backend directly?

spydon: > Why can't ComponentRef talk to the riverpod backend directly?

Do you mean by using a `ProviderContainer`? Since most people use `ProviderScope` in Flutter I don't think that is a good idea.

packages/flame_riverpod/example/lib/main.dart
spydon: > Make a flame_forg2d_riverpod package.

That would end up being way too many extra packages, and we'd have to think about other people creating packages on top of `Flame` too, like Bonfire. So we would need it to be more generic.

> It's not a huge deal though. But it's likely necessary if you wish to support ref.watch (cf https://github.com/flame-engine/flame/pull/2367#discussion_r1121339874)

Couldn't that be done by a mixin? :thinking:

packages/flame_riverpod/README.md
spydon: Ah, or maybe you are correct, maybe that is the second point that @rrousselGit is mentioning [here](https://github.com/flame-engine/flame/pull/2367#discussion_r1121528187).
I don't know enough about the internals of Riverpod and how the `WidgetRef` works, but I assumed that we'd need some connection to the `WidgetRef` to be able to listen to the same providers as are used in the Flutter tree.

packages/flame_riverpod/example/lib/main.dart
markvideon: Not sure we're on the same page about the component lifecycle so I want to ensure there's a shared understanding of a key concept:

Components are mounted once, and are removed / unmounted at some point during a game (e.g. enemy dies) - in this way they are distinct from Widgets that we understand as being unmounted / re-mounted as part of a re-render

I don't understand the case for decoupling the interactions with providers from when a component is mounted

packages/flame_riverpod/example/lib/main.dart
rrousselGit: It may need to be life-cycle aware, so it might have to be more specific than a mixin reusable everywhere

And in general I'm not sure about the current implementation of said mixin (relying on a static widgetRef field)

packages/flame_riverpod/lib/src/has_component_ref_mixin.dart
st-pasha: Yes, indeed. The question was more about what would happen to riverpod listener subscriptions -- will they get re-subscribed automatically? If not, then why are they being kept inside the `listeners` list?

packages/flame_riverpod/example/lib/main.dart
rrousselGit: I don't think there's a real difference here. You could most definitely make a game engine controlled by widgets. Some people have done this in React.  
Currently Flame components seems more similar to a RenderObject than a Widget though. But it's likely possible to use a widget abstraction to manipulate them.

That's out of the scope of this PR of course. But Riverpod part of the point of Riverpod is to have business logic be more "widget-like" instead of the old imperative syntax.   
You're not meant to use `ref.listen` for too much.

packages/flame_riverpod/README.md
rrousselGit: Connecting to the ProviderContainer directly (though `ProviderScope.containerOf(context)`) and making your own ref from scratch is the most logical solution

spydon: That sounds like a great solution, then there would be less boilerplate för the user too right?

garysm:
@markvideon Can we expect an update for this?

markvideon:
> @markvideon Can we expect an update for this?

Yes

packages/flame_riverpod/lib/flame_riverpod.dart
markvideon: Yes (although it is not regularly used), and yes

Comments in this part of a library are shown underneath the name of the library in the 'Libraries' list of the generated docs. It is currently used elsewhere in:

- experimental in https://pub.dev/documentation/flame/latest/
- flame_lint https://pub.dev/documentation/flame_lint/latest/

.github/.cspell/words_dictionary.txt
```diff
@@ -38,3 +39,4 @@ toolset
 underutilize
 unproject
 unscale
+videon
```
spydon, line 42: We have a `people_usernames.txt` dictionary where you could add this

packages/flame_riverpod/example/.gitignore
```diff
@@ -0,0 +1,44 @@
+# Miscellaneous
```
spydon, line 1: We have a global .gitignore file, so I don't think this one is needed

packages/flame_riverpod/example/lib/main.dart
```diff
@@ -0,0 +1,109 @@
...
+}
+
+class RiverpodAwareTextComponent extends PositionComponent
+    with RiverpodComponentMixin {
+  late TextComponent textComponent;
+  int currentValue = 0;
+
+  /// [onMount] should be used over [onLoad] to initialize subscriptions,
+  /// cancellation is handled for the user inside [onRemove],
+  /// which is only called if the [Component] was mounted.
```
spydon, line 87: This sounds like `onRemove` is called when the component is mounted? :sweat_smile:

packages/flame_riverpod/example/pubspec.yaml
```diff
@@ -0,0 +1,21 @@
...
+dependencies:
+  flame: ^1.10.1
+  flame_riverpod:
+    path: ../
```
spydon, line 10: Since this is within the monorepo now it should just say the latest version here and Melos will connect it with path through bootstrap (and update it when we run `melos version`)

packages/flame_riverpod/lib/src/widget.dart
```diff
@@ -0,0 +1,206 @@
+import 'package:flame/game.dart';
+import 'package:flame_riverpod/flame_riverpod.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+class RiverpodAwareGameWidget<T extends Game> extends GameWidget<T> {
```
spydon, line 6: Some dartdocs here would be good

spydon (commented):
Looks great! :heart_eyes: 
Just a few minor comments, nothing on the implementation.

packages/flame_riverpod/example/lib/main.dart
spydon, line 87:
```suggestion
  /// [onMount] should be used over [onLoad] to initialize subscriptions,
  /// which is only called if the [Component] was mounted.
  /// Cancellation is handled for the user automatically inside [onRemove].
```
Something like this maybe?

packages/flame_riverpod/lib/src/widget.dart
markvideon: Added some comments, I'll leave it to you to resolve if this is satisfactory

spydon: Looks great!

spydon (approved):
Lgtm! :partying_face:

spydon:
@markvideon I missed one thing in my review, docs are missing.
Similar to these ones: https://github.com/flame-engine/flame/tree/main/doc/bridge_packages/flame_bloc

doc/bridge_packages/flame_riverpod/flame_riverpod.md
```diff
@@ -0,0 +1,9 @@
+# flame_riverpod
+
+```{toctree}
+Overview    <riverpod.md>
+Component    <component.md>
+Widget    <widget.md>
+
```
spydon, line 7:
```suggestion
Overview   <riverpod.md>
Component  <component.md>
Widget     <widget.md>
```

doc/bridge_packages/flame_riverpod/riverpod.md
```diff
@@ -0,0 +1,67 @@
+# flame_riverpod
+
+
+## Riverpod
+
+[Riverpod](https://pub.dev/packages/flutter_riverpod) is a reactive caching and data-binding
```
spydon, line 6:
```suggestion
[Riverpod](https://riverpod.dev/) is a reactive caching and data-binding
```
Maybe linking to their page instead since there is more info there?

spydon (approved):
Two minor comments, otherwise I think we're good to go! :heart_eyes:

PR #2374, opened 2023-03-01
chore: Standardize and unify SDK versions across packages

luanpotter:
# Description

This does two things:

## Use double quotes for SDK constraints

Standardize the usage of single or double quotes to specify sdk constraints across pubspecs
I see no reason this should not be kept consistent
I also see no reason to prefer one over the other, so I searched the code base and there are 7 instances of single quote vs 32 of double quotes, so I favored the later

## Update all SDK constraints to 2.18

Let me know if there are any issues with it, but I believe we should keep this consistent across all packages.
Also there is a pubspec on root which imply all should be on 2.18 anyway.

## Checklist

## Breaking Change?

### DIFF
--- a/doc/flame/examples/pubspec.yaml
+++ b/doc/flame/examples/pubspec.yaml
@@ -4,7 +4,7 @@ version: 1.0.0
 publish_to: none
 
 environment:
-  sdk: ">=2.17.0 <3.0.0"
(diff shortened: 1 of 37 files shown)


spydon (approved):
Is double quotes what it generates when running flutter create?

luanpotter:
It is double quotes. At least for me

Issue #2308, opened 2023-01-30
Unused code report


incendial:
Hi! I'm testing new `check-unused-code` command implementation that can find separate methods, fields, etc.

In this issue I'll list all the reported files and then we can decide how (if should) they will be removed (by me, by you, etc).

If you're not interested, feel free to close this issue 🙂.

Report output for `packages/flame/` folder:
```
✔ Analysis is completed. Preparing the results: 16.3s

packages/flame/test/collisions/collision_test_helpers.dart:
    ⚠ unused method intersections
      at flame/test/collisions/collision_test_helpers.dart:128:3

packages/flame/test/components/component_test.dart:
    ⚠ unused field name
      at flame/test/components/component_test.dart:1041:3
    ⚠ unused field name
      at flame/test/components/component_test.dart:1167:3

packages/flame/test/events/game_mixins/multi_touch_tap_detector_test.dart:
    ⚠ unused field tapTimes
      at flame/test/events/game_mixins/multi_touch_tap_detector_test.dart:91:3

packages/flame/test/fixtures/fixture_reader.dart:
    ⚠ unused function fixtureForString
      at flame/test/fixtures/fixture_reader.dart:11:1
```

Report output for `packages/flame_bloc/` folder:
```
✔ Analysis is completed. Preparing the results: 3.6s

packages/flame_bloc/example/lib/src/game_stats/bloc/game_stats_event.dart:
    ⚠ unused constructor ScoreEventCleared
      at flame_bloc/example/lib/src/game_stats/bloc/game_stats_event.dart:17:3

packages/flame_bloc/test/inventory_cubit.dart:
    ⚠ unused method selectSword
      at flame_bloc/test/inventory_cubit.dart:15:3

packages/flame_bloc/test/player_cubit.dart:
    ⚠ unused method riseFromTheDead
      at flame_bloc/test/player_cubit.dart:16:3
```

Report output for `packages/flame_jenny/jenny/` folder:
```
✔ Analysis is completed. Preparing the results: 2.5s

packages/flame_jenny/jenny/lib/src/parse/parse.dart:
    ⚠ unused type alias FunctionBuilder
      at flame_jenny/jenny/lib/src/parse/parse.dart:968:1

✖ total unused code (classes, functions, variables, extensions, enums, mixins and type aliases) - 1
```

incendial:
That's it! I'm truly impressed, considered the amount of code you have 🚀

spydon:
[...]

PR #2377, opened 2023-03-01
refactor: Use variable name on toString in component_test.dart

luanpotter:
# Description

This is a cleanup identified on this issue: https://github.com/flame-engine/flame/issues/2308
Using an amazing unused-code tooling
Now, since Flame is a public API, unused code might not be trivial - it might just mean untested code.

In this case, the unused code is in the tests, which likely means it is indeed necessary.
I tried to infer the intent of the author here, and even considered adding the names of the components to the events, but the events are already always per component, so that was meaningless. I also thought it might be a missing toString, that they would wish to show the component name for debugging purposes, but I didn't see any asserts/logs on the component objects themselves (just the events).
So I concluded there is no reason to keep this field, but lmk if anyone disagrees and I can bring it back and add some function to it.

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame/test/components/component_test.dart
+++ b/packages/flame/test/components/component_test.dart
@@ -62,8 +62,8 @@ void main() {
       testWithFlameGame(
         'component removed completes after changing parent',
         (game) async {
-          final parent = LifecycleComponent()..addToParent(game);
-          final child = LifecycleComponent()..addToParent(parent);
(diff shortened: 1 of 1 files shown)


st-pasha:
The `name` field was for debugging purposes: if some test breaks, you can set up a breakpoint and see clearly which component is which. Perhaps a `toString()` method would be helpful here?

luanpotter:
Updated!

PR #2382, opened 2023-03-01
feat: Add listener for initial state on flame_bloc

luanpotter:
# Description

While analyzing the tests for flame_bloc, I stumbled upon an interesting conundrum.
It seems that the `initialState` provided does not trigger a state change callback through `onNewState`. While that is in line with bloc, I think for most game use cases it is important to keep track of the "last" state of stuff - and in fact most of our existing examples do use the library in that way, to keep track of the last (i.e., the current) state.
However, doing so generically is not simple with the current library because of how the initial state is not provided. While you could initialize your last manually, if there is a dynamic initial state, it would not be trivial to wire it up.
This, therefore, adds a new listener of `onInitialState`, maintaining the behaviour of `onNewState`.

Note: this solves the last unused reference [from this issue](https://github.com/flame-engine/flame/issues/2308) by using it on the new tests.

## Checklist

## Breaking Change?


packages/flame_bloc/lib/src/flame_bloc_listenable.dart
```diff
@@ -50,6 +50,7 @@ mixin FlameBlocListenable<B extends BlocBase<S>, S> on Component {
     }
     _bloc = bloc;
     _state = bloc.state;
+    onNewState(_state);
```
luanpotter, line 53: this is the only production change
the rest is fixing & adding tests

KyleKun:
> However, if we want to avoid the breaking change, the other option is to add a second callback hook, analogous to onNewState, but called onInitialState that of course is called only once.

I prefer this approach, makes more sense than triggering `onNewState` imo.

luanpotter:
I agree, let me try out that approach.

spydon (approved):

packages/flame_bloc/test/src/flame_bloc_listenable_test.dart
```diff
@@ -78,7 +92,7 @@ void main() {
 
         bloc.makeSad();
         await Future.microtask(() {});
-        expect(component.last, isNull);
+        expect(component.last, PlayerState.alive);
```
erickzanardo, line 95:
```suggestion
        expect(component.last, equals(PlayerState.alive));
```

The same suggestion a couple of line bellow

luanpotter: what's the difference?

erickzanardo: Matchers give better error message when a test fails

packages/flame_bloc/test/src/flame_bloc_listenable_test.dart
```diff
@@ -96,7 +110,7 @@ void main() {
 
         bloc.kill();
         await Future.microtask(() {});
-        expect(component.last, isNull);
+        expect(component.last, PlayerState.alive);
```
luanpotter, line 113:
```suggestion
        expect(component.last, equals(PlayerState.alive));
```

erickzanardo (approved):

joaosilva2095:
Hi everyone 👋

Great work on adding this feature! It is something definitely useful 😊
While checking the changes, I have noticed that the `FlameBlocListener` was not updated with the new `onInitialState` parameter.

Was this the goal? Or should that parameter be added and passed to the `FlameBlocListenable`?

erickzanardo:
Good callout @joaosilva2095 that totally escaped our attention, here is the PR addressing it: https://github.com/flame-engine/flame/pull/2565

Issue #2005, opened 2022-10-03
Documentation: Forge2D joints


spydon:
# What could be improved

All the different joint types in Forge2D should be explained and have examples, currently there is no pure `other moduler/forge2d` section so either create that section or put these docs under `flame_forge2d` meanwhile.

https://docs.flame-engine.org/main/flame_forge2d/forge2d.html

# Good to know
Make sure that you comment on the issue that you will be working on this before starting to program.

To set up the Flame environment you need to follow the instructions in our [contributing.md](https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md#environment-setup) doc, if you have any problems setting up your environment (or other issues of course) feel free to ask for help in the Flame channel on our [Discord server](https://discord.com/invite/pxrBmy4).

Since this is a documentation change, you need to be able to build the docs. You can do that by running melos sphinxdoc after you have followed the instructions in the contributing.md doc and also [installed Sphinx on your system](https://www.sphinx-doc.org/en/master/usage/installation.html). After you have built the docs you can open the doc/_build/html/index.html file in your web browser to see the changes.

Make sure to include a video of the example when you upload the PR so that it will be easier for us to review it. (You can attach videos directly in the PR, so no need to upload them somewhere else)

**Happy hacking!**

AbhishekTiwari23:
Please assign to me

spydon:
> Please assign to me

Super! I've assigned you to the issue, remember to join our Discord server if you have any questions etc. :)

AbhishekTiwari23:
> > Please assign to me
> 
> Super! I've assigned you to the issue, remember to join our Discord server if you have any questions etc. :)

Thanks, @spydon, for assigning me. I will respond as soon as possible.

spydon:
> > Hi @AbhishekTiwari23, any progress on this? If you don't plan to do this, please let us know so we can open up this issue for others. :)
> 
> I will make a pr by tomorrow. Actually, there is a fest going on, and I am getting less time for this, but surely I will do it.

Good to hear!

spydon:
Hi @AbhishekTiwari23, there is less than a week left of Hacktoberfest now, do you think that you'll be able to complete this issue in time?

munsterlander:
Less than 48 hours remaining.

spydon:
Since we haven't had any response from the assignee, this is again up for grabs!

eukleshnin:
@spydon Happy to pick it up

spydon:
[...]

PR #2383, opened 2023-03-01
docs: FrictionJoint documentation and example

eukleshnin:
# Description

Added documentation and example for `FrictionJoint`.
Also modified `main.dart` to be able to run individual examples by providing page param.

## Checklist

## Breaking Change?

## Related Issues

PR 3/X for #2005


examples/lib/main.dart
```diff
@@ -19,9 +23,29 @@ import 'package:examples/stories/system/system.dart';
 import 'package:examples/stories/tiled/tiled.dart';
 import 'package:examples/stories/utils/utils.dart';
 import 'package:examples/stories/widgets/widgets.dart';
+import 'package:flame/game.dart';
 import 'package:flutter/material.dart';
 
 void main() {
+  var page = window.location.search ?? '';
```
eukleshnin, line 30: I'm trying to make embedded `flutter-app` examples work. Not sure if it helps though, I don't know how to test it locally.

st-pasha: See the documentation here: https://docs.flame-engine.org/main/development/documentation.html#building-documentation-locally

doc/bridge_packages/flame_forge2d/joints.md
```diff
@@ -112,3 +112,45 @@ to 0, the higher the value, the less springy the joint becomes.
...
+
+### `FrictionJoint`
+
+A `FrictionJoint` is used for top-down friction. It provides 2D translational friction and angular
+friction.
+
+The `initialize` method requires two bodies and an anchor. The first body is the body the friction
+will be applied to. The second body can be any body in the world, like a floor or an invisible
+body. The reason is that the friction force will be applied to the first body when it collides with
+any other bodies, not only the second body of the joint.
```
luanpotter, line 125: why do we need to specify a second body then? can we make it more clear, this seems to imply that the second body is irrelevant

doc/bridge_packages/flame_forge2d/joints.md
```diff
@@ -112,3 +112,45 @@ to 0, the higher the value, the less springy the joint becomes.
...
+
+The third parameter is the anchor point in the world where the friction force will be applied.
```
luanpotter, line 127: could you comment how this evolves as the body moves?
for example you are providing the center of the ball in world coordinates as the third parameter in the example; when the ballBody moves, how does that affect the friction?

doc/bridge_packages/flame_forge2d/joints.md
eukleshnin: Good catch. I completely misunderstood that this joint is designed for top-down games, and wrote this piece based on that assumption. Fixed now

doc/bridge_packages/flame_forge2d/joints.md
eukleshnin: It gets converted to a local coordinate inside `FrictionJointDef`. 

```
void initialize(A bodyA, B bodyB, Vector2 anchor) {
  this.bodyA = bodyA;
  this.bodyB = bodyB;
  localAnchorA.setFrom(bodyA.localPoint(anchor));
  localAnchorB.setFrom(bodyB.localPoint(anchor));
}
```

It's really hard to find any documentation in box2d on this parameter. From tests, it seems like the friction force is applied along the axis between two local anchors of the Joint.

doc/bridge_packages/flame_forge2d/joints.md
```diff
@@ -112,3 +112,51 @@ to 0, the higher the value, the less springy the joint becomes.
 ```{warning}
 Do not use a zero or short length.
 ```
+
+
+### `FrictionJoint`
+
+A `FrictionJoint` is used for simulating friction in a top-down game. It provides 2D translational
```
spydon, line 119: It didn't have to be a top-down game, right?

spydon (approved):

doc/bridge_packages/flame_forge2d/joints.md
eukleshnin: It certainly works with non-zero gravity, but it was designed specifically for top-down games. From [Box2D documentation](https://box2d.org/documentation/md__d_1__git_hub_box2d_docs_dynamics.html#autotoc_md94):

> The friction joint is used for top-down friction

I also confirmed it in the Box2D discord:

> Erin — 03/02/2023 11:12 PM
The friction joint is designed for a top down game. Imagine you have a top down game where a box is sliding across the floor (which is the plane of the screen). This joint applies friction between the box and the floor. Now suppose you have a top down game where the box is on the floor of a moving vehicle. The floor is now moving, so you need a second body. The friction joint is not related the friction that you get when shapes collide in the x-y space of the screen plane, instead it is related to friction across the z direction, perpendicular to the screen.

luanpotter (approved):
LGTM, really nice!

examples/lib/main.dart
spydon: Just realized that this breaks the examples app for all platforms except web, so we'll have to come up with a way to be able to run this on all platforms.

eukleshnin: But for all other platforms `window.location.search` would be null, so it will be run as dashbook? Or am I missing something?

spydon: The use of dart:html will not allow it to compile to other platforms than web

spydon: Maybe something like this could work instead of pulling in a new package?
https://gpalma.pt/blog/conditional-importing/

eukleshnin: Please check it out https://github.com/flame-engine/flame/pull/2411

PR #2391, opened 2023-03-08
fix: TiledComponent now can be safely loaded regardless of the order

### LINKED ISSUE #2390: TiledComponent renders wrong map if some condition is met
# Current bug behavior
If there are several maps with the same tileset but different tile alignments and also they have only one tileset, the latter one is not rendered correctly.

Let's say there are two Tiled maps `A` and `B`
Each of them has the same tileset named `image`.
But they have different tile alignments respectively.

When we load map A and then load B, The latter one not renders correctly but renders tile alignments of A.

This is because TiledAtlas caches its map based on its tilesets' names.
Please see the following section, there is operator `??=` which makes this issue.
https://github.com/flame-engine/flame/blob/f125593aaaaef160395b772180b1514f6be3ac4f/packages/flame_tiled/lib/src/tile_atlas.dart#L98-L108

# Expected behavior
Regardless of the `TiledComponent` loading order, they will render correctly.

* Flame version: main

# More information

Are you interested in working on a PR for this? Yes

Hwan-seok:
# Description
Basically, the `TiledAtlas` must not be reused without `clone()` because it includes `batch` which is different for each `TiledAtlas`.
 But the following snippets(before this PR) can lead to the reuse of the TiledAtlas if I have loaded the map with the same tilesets(same atlasKey).
```dart
   return atlasMap[key] ??= TiledAtlas._(...); 
```

Also, this PR includes the change of the `TestAssetBundle`.
I don't know about the exact intention of it.
But it can load only a single map and also it ignored the actual file name to the bundle's first file name.
This makes tests to be hard to debug.

This change is quite far from the original one. So please feel free to tell me to divide it to the separate PR.

## Checklist

## Breaking Change?

## Related Issues
Fixes #2390 


packages/flame_tiled/lib/src/tile_atlas.dart
```diff
@@ -100,7 +100,7 @@ class TiledAtlas {
       // really boring map.
       final image = (await Flame.images.load(key)).clone();
 
-      return atlasMap[key] ??= TiledAtlas._(
+      return atlasMap[key] = TiledAtlas._(
```
Hwan-seok, line 103: This is the only production change.

ufrshubham, line 103: A question, if the atlas keys are same for 2 maps, wouldn't that case be handled by existing `atlasMap.containsKey(key)` few lines above?

Hwan-seok: You are absolutely right, my test case was terrible. Thanks for pointing it out!

Also, my explanation of this issue also has some flaws.
As you seen, maps are handled by `if(atlasMap.containsKey(key))` if they are loaded sequentially😂 

But if they are loaded in the race condition, both of them would be stopped at: 
`final image = (await Flame.images.load(key)).clone();`.
Then they keep going...
As a result, they reach `??=` without `if(atlasMap.containsKey(key))` check.

So the test case should be changed from:
```dart
final component1 = await TiledComponent.load(
  'single_tile_map_1.tmx',
  Vector2(16, 16),
);
final component2 = await TiledComponent.load(
  'single_tile_map_2.tmx',
  Vector2(16, 16),
);
```
to: 
```dart
final components = await Future.wait([
  TiledComponent.load(
    'single_tile_map_1.tmx',
    Vector2(16, 16),
  ),
  TiledComponent.load(
    'single_tile_map_2.tmx',
    Vector2(16, 16),
  )
]);
```
The difference between these is racing or not.

**But even after changing the test case, it passes with and without this fix.**

This is because of here:
https://github.com/flame-engine/flame/blob/f125593aaaaef160395b772180b1514f6be3ac4f/packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart#L46
This line clones `TiledAtlas` one more time. So it was safe indeed.

You can remove this `clone()` and then see the (updated) test would be failed with `??=`.

But still, I think it should be changed to `??=` to `=` because of 
1) Consistency with others like more than one tileset
2) It's safe to use without a thorough understanding of this class (maybe extended from 1 )


> Probably, you have been curious about why I had an issue without any problem in the code.
That is because I am using the forked flame version with optimized TiledLayer without `clone()` one more time.
But regardless of my forked version, this would be better to be fixed.
I am thoroughly testing my patch in production and will show up to you guys

ufrshubham: Okay, that makes sense. So in that case, does something following be better?
```suggestion
      return atlasMap.contains(key) ? atlasMap[key]!.clone() : atlasMap[key] = TiledAtlas._(
```
It is a bit verbose, but it avoids replacing the atlas in the map if it is already loaded by some other async call.

In any case, consider adding at least a comment at the return about this special case 🙂.

ufrshubham (approved):
LGTM!

packages/flame_tiled/lib/src/tile_atlas.dart
Hwan-seok: Sounds great!

But this is another option.
```dart
if (atlasMap.containsKey(key)) {
  return atlasMap[key]!.clone();
}

return TiledAtlas._(
  atlas: image,
  offsets: {key: Offset.zero},
  key: key,
);
```
Which one do you prefer? I will follow your decision.

ufrshubham: Go ahead with your option. It looks more readable👍🏼

Hwan-seok: Final option:
just overwrite atlasMap[key] without checking if exists.
IMHO, it has less code but leads to the same result.

ufrshubham: From the looks of it, the whole point of having `atlasMap` seems to be to avoid atlas recreations. But even after caching, we are anyways returning a new or cloned version. So I guess even overwriting without checking should technically be fine. Also, this part of code is not in critical path, so I'd say prefer whichever option is more readable.🙂

Hwan-seok:
I don't know the reason why the re-request is for unique 😂 
Anyway, I added a comment for it. Thanks to everyone.

ufrshubham (approved):

spydon (approved):

Issue #2343, opened 2023-02-14
flame_noise bridge package to fast_noise


spydon:
# What could be improved

Since `SimplexNoise` is now [deprecated](https://github.com/google/vector_math.dart/issues/270) we want to deprecate `NoiseEffectController` and move it into a bridge package that bridges to fast_noise. Then the `NoiseEffectController` can use different types of noise and I'm sure there will also be more applications for where we want to use noise which we can put in that package.

# Why should this be improved

To avoid putting in another dependency in the core of Flame, but also open up the possibility for more applications within Flame using noises.

# Any risks?

It will be a breaking change.

# More information

https://github.com/frankpepermans/fast_noise
https://github.com/google/vector_math.dart/issues/270

Issue #2329, opened 2023-02-07
`SimplexNoise` is now deprecated in `vector_math`


spydon:
# What could be improved
We are using `SimplexNoise` in `noise_effect_controller.dart` line 29 and 34, we have to migrate to another noise package since it will be removed. Maybe it would be worth it to pull in `fast_noise`, we talked to them a long time ago and even though there hasn't been much activity in their repository it seemed like they would be willing to handle any PRs that we might come up with.

# Why should this be improved
Because the file will be removed in the future and we are currently losing pub points on this.

# Any risks?
Yes, that the fast_noise library isn't well maintained.

# More information
https://github.com/google/vector_math.dart/issues/270

PR #2393, opened 2023-03-10
feat: Introduce flame_noise, deprecate NoiseEffectController

luanpotter:
# Description

* Deprecate the NoiseEffectController that is based on the deprecated vector2 impl
* Add a new bridge package `flame_noise` to bridge fast noise

Note: the *goal* of this PR is to allow the immediate deprecation of the old NoiseEffectController by providing an experimental, suitable replacement. I believe the package flame_noise will look nothing like this at all if I am able to land [some improvements to the underlying fast_noise lib](https://github.com/frankpepermans/fast_noise/pull/5). We will be able to have for example a generic NoiseEffectController.

However I believe we should merge this as an experimental package for now to unblock removing the current implementation from flame which is negatively affecting our scores on pub. Therefore I would advise we don't spend any time discussing specifics of how the API/impl provided for the flame_noise, and just go with _something_ instead of nothing.

## Checklist

## Breaking Change?

- [x] No, this PR is not a breaking change (yet).

Not yet. Will remove deprecated noise class later.

## Related Issues

https://github.com/flame-engine/flame/issues/2343
https://github.com/flame-engine/flame/issues/2329


packages/flame/lib/src/effects/controllers/noise_effect_controller.dart
```diff
@@ -21,7 +18,9 @@ import 'package:vector_math/vector_math_64.dart';
 /// example, putting into a `MoveEffect.by` will create a shake motion, where
 /// the magnitude and the direction of shaking is controlled by the effect's
 /// `offset`.
+@Deprecated('Use flame_noise instead')
```
spydon, line 21: Add which version that it will be removed in, which is v1.8.0

packages/flame/lib/src/effects/controllers/noise_effect_controller.dart
```diff
@@ -21,7 +18,9 @@ import 'package:vector_math/vector_math_64.dart';
...
 class NoiseEffectController extends DurationEffectController {
+  @Deprecated('Use flame_noise instead')
```
spydon, line 23: I don't think you have to deprecate both the constructor and the class? I think the class should be enough.

packages/flame_noise/README.md
```diff
@@ -0,0 +1,6 @@
+# flame_noise
```
spydon, line 1: Should this readme look the same as the other ones that you consolidated?

spydon (approved):

packages/flame/lib/src/effects/controllers/noise_effect_controller.dart
luanpotter: I recall vscode complaining if the constructor wasn't. Let me try again

luanpotter: image

spydon (approved):

PR #2405, opened 2023-03-13
docs: Do not use shell=True in dart_domain.py

st-pasha:
# Description

This fixes a problem that we accidentally introduced in a recent PR: the `shell=True` setting requires that the arguments were passed as a single string instead of a list. Passing as a string, in turn, is somewhat problematic because the file names need to be escaped, and there is no utility function to make this escaping cross-platform.

So, instead we'll use the standard `shell=False`, and invoke `dartdoc_json.bat` on Windows instead of simply `dartdoc_json`. Hope this would work.

## Checklist

## Breaking Change?

### DIFF
--- a/doc/_sphinx/extensions/dart_domain.py
+++ b/doc/_sphinx/extensions/dart_domain.py
@@ -2,7 +2,6 @@ import json
 import os
 import re
 import subprocess
-import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


munsterlander:
It needs to be .bat and this is what I first did prior to our discussing shell=True.  I can confirm this works.

PR #2408, opened 2023-03-15
chore: Update version of flutter_svg on flame_svg

luanpotter:
# Description

The `flutter_svg` version used on `flame_svg` was outdated, which causes us to lose precious pub points:

image

This updates the library by migrating a bit of the code (it was def a breaking change on their side, but shouldn't be on ours).

## Checklist

## Breaking Change?

- [] Yes, this PR is a breaking change.


examples/.metadata
```diff
@@ -4,7 +4,7 @@
 # This file should be version controlled.
 
 version:
-  revision: 85684f9300908116a78138ea4c6036c35c9a1236
+  revision: 2ad6cd72c040113b47ee9055e722606a490ef0da
```
luanpotter, line 7: not sure if I should be updating this file or not

packages/flame_svg/example/.gitignore
```diff
@@ -0,0 +1,49 @@
+linux/
+macos/
+web/
+windows/
```
luanpotter, line 4: this gitignore was missing, causing issues after a `flutter create`

packages/flame_svg/example/.metadata
```diff
@@ -1,10 +1,45 @@
 # This file tracks properties of this Flutter project.
 # Used by Flutter tool to assess capabilities and perform upgrades etc.
 #
-# This file should be version controlled and should not be manually edited.
+# This file should be version controlled.
```
luanpotter, line 4: but should it be updated? :thinking:

packages/flame_svg/lib/svg.dart
```diff
@@ -96,8 +98,17 @@ class Svg {
...
+
+    final scale = math.min(
+      size.width / pictureInfo.size.width,
+      size.height / pictureInfo.size.height,
+    );
+    canvas.translate(
+      (size.width - pictureInfo.size.width * scale) / 2,
+      (size.height - pictureInfo.size.height * scale) / 2,
+    );
+    canvas.scale(scale);
```
luanpotter, line 110: sadly scaleCanvasToViewBox seems to no longer exist.
this could be vastly simpler with Vector2's, but that could introduce latency due to object allocation.

luanpotter: dart really need inline classes/records/something like that

examples/.metadata
spydon: Doesn't matter much, it would be replaced when someone runs it with a different version of Flutter

pubspec.lock
```diff
@@ -61,10 +61,10 @@ packages:
     dependency: transitive
     description:
       name: conventional_commit
-      sha256: "40321e55c2416c0940a846bf938c0a70efaf9eb72da042d5ea76b79b6f5045ee"
```
spydon, line 64: This file you can add to gitignore since it's only for the workspace

spydon (approved):

PR #2401, opened 2023-03-12
docs: Fix melos doc setup for windows

munsterlander:
# Description

Closes: https://github.com/flame-engine/flame/issues/2398

Finally got the script all figured out, so `melos doc-setup` works on windows.  It should work on Unix as well, but I can't verify.  Its simply scalar fold, color encoding removed, python string literal wrapped in double quotes, and false converted to a bash exit.

## Checklist

- [X] I have followed the [Contributor Guide] when preparing my PR.
- [ ] I have updated/added tests for ALL new/updated/fixed functionality.
- [X] I have updated/added relevant documentation in `docs` and added dartdoc comments with `///`.
- [ ] I have updated/added relevant examples in `examples` or `docs`.

## Breaking Change?

- [ ] Yes, this PR is a breaking change.
- [X] No, this PR is not a breaking change.

## Related Issues

[Contributor Guide]: https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md
[Conventional Commit]: https://conventionalcommits.org
[CHANGELOG]: https://github.com/flame-engine/flame/blob/main/CHANGELOG.md

PR #2416, opened 2023-03-17
docs: Fix for missing jQuery

munsterlander:
# Description

As mentioned in discord, the current build of the docs loses jQuery functionality.  This is due to sphinx being upgraded to version 6 in which they removed jQuery and created a stand-alone package for it.

Additionally, in the process of doing that, I noticed the previously mentioned fix (in the comments) for the `melos doc-setup` about removing the single quotes to keep it clean didn't make into the final code.  I have updated that as it was previously confirmed to work on all platforms in: https://github.com/flame-engine/flame/pull/2401

I noticed that the copyright that was in the `conf.py` was locked at 2021 and so I made it go to 2023 as well.  This can be removed if not desired, but thought maybe this had just been overlooked.

Finally, there was an error regarding a duplicate instantiation of a constant.  This had to do with the fact the basic theme provided with sphinx, the next version after which the docs were based, had a bug fix implemented: https://www.sphinx-doc.org/en/master/changes.html#release-5-1-0-released-jul-24-2022.  In their fix, it also broke the way the flames theme works.  On a side note, the docs are using a very outdated version of the basic theme and should likely be overhauled, but that is outside the scope of this pr.  None the less, after I finally figured out how the basic theme is inherited and where that lived on my machine, I was able to resolve the discrepancies and eliminate the conflicting names between the two themes.

In short, all errors have been resolved and I have confirmed that mobile, web, search, highlighting, menus, etc. all work - at least on my machine.

## Checklist

## Breaking Change?

## Related Issues


doc/_sphinx/conf.py
```diff
@@ -17,7 +17,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'Flame'
-copyright = '2021, Blue Fire Team'
+copyright = '2021-2023, Blue Fire Team'
```
spydon, line 20: Maybe the year could be skipped here entirely? We don't really care about the copyright anyways since it's MIT licenced.

doc/_sphinx/theme/doctools.js
```diff
@@ -21,11 +21,18 @@ const _ready = (callback) => {
...
+  "SELECT",
+  "BUTTON",
+]);
+
 /**
  * highlight a given string on a node by wrapping it in
  * span elements with the given class name.
  */
-const _highlight = (node, addItems, text, className, index) => {
+const _highlight_flames = (node, addItems, text, className, index) => {
```
spydon, line 35: How come it needed a rename? And shouldn't this be singular if the rename is needed?

doc/_sphinx/theme/doctools.js
```diff
@@ -86,11 +93,11 @@ const _highlightText = (thisNode, text, className, index) => {
 /**
  * Small JavaScript module for the documentation.
  */
-const Documentation = {
+const Documentation_flames = {
```
spydon, line 96: How come it needed a rename? And shouldn't this be singular if the rename is needed?

spydon: It also looks weird with a starting capital letter and then snake-case?

doc/_sphinx/conf.py
munsterlander: Removed.

doc/_sphinx/theme/doctools.js
munsterlander: Same as above.  It is the name of the custom theme the docs uses, so it stays consistent.

doc/_sphinx/theme/doctools.js
munsterlander: In short, the docs are in this hybrid theme land where they are using some features from one version but need a later version for other features that were then upgraded in the latest version.  The hybrid naming convention keeps the functions separate so they don't conflict and keep the functionality for both old and new.

If there is a desire to completely overhaul the theme, I can go do that as well and then we some of these files would not be needed anymore.

doc/_sphinx/theme/doctools.js
spydon: But shouldn't it be CamelCase?

munsterlander: The theme name is "flames".  I was keeping everything consistent.  If you want it all changed, I can do that, it just seems inconsistent.

spydon: I understood that, what I meant is, shouldn't the variable name be `DescriptionFlames` and not with underscore?

munsterlander: What I'm saying is,  "Documentationflames" was worse than "Documentation_flames" which is more consistent than "DocumentationFlames".  I don't care and it has no impact. If you want the latter, fine by me.

munsterlander: Updated and pushed.

doc/_sphinx/theme/doctools.js
```diff
@@ -72,12 +79,12 @@ const _highlight = (node, addItems, text, className, index) => {
       }
     }
   } else if (node.matches && !node.matches("button, select, textarea")) {
-    node.childNodes.forEach((el) => _highlight(el, addItems, text, className, index));
+    node.childNodes.forEach((el) => _highlight_flame(el, addItems, text, className, index));
   }
 };
-const _highlightText = (thisNode, text, className, index) => {
+const _highlightText_flame = (thisNode, text, className, index) => {
```
spydon, line 85: This should be changed too right?
```suggestion
const _highlightTextFlame = (thisNode, text, className, index) => {
```

munsterlander: Sure.  Now that it doesn't reference the theme anymore, I don't think it matters.

doc/_sphinx/conf.py
st-pasha: General suggestion ([see here](https://www.copyrightlaws.com/copyright-symbol-notice-year/#:~:text=Which%20Year%20Should%20You%20Use,to%20the%20public%20without%20restriction.)) is to only write the year of first publication, not the range.

doc/_sphinx/theme/doctools.js
```diff
@@ -21,11 +21,18 @@ const _ready = (callback) => {
...
+  "SELECT",
+  "BUTTON",
+]);
+
 /**
  * highlight a given string on a node by wrapping it in
  * span elements with the given class name.
  */
-const _highlight = (node, addItems, text, className, index) => {
+const _highlight_flame = (node, addItems, text, className, index) => {
```
st-pasha, line 35: I guess this should be `_highlightFlame` now, to keep everything consistent :)

st-pasha (approved):
LGTM

doc/_sphinx/theme/doctools.js
munsterlander: Done.  Now we can't parse any of the modified variables by theme name, so if that situation arises down the road until a full theme overhaul, it will just need to be considered then.

spydon (approved):
Good job :)

PR #2429, opened 2023-03-24
feat: Add reusable vector to the Vector2 extension

### LINKED ISSUE #2427: Improve moveToTarget performance by reusing the diff vector
# What could be improved
Add optional reusable diff argument which is `Vector2` to the `Vector2.moveToTarget`.

https://github.com/flame-engine/flame/blob/295cd72422ee068635057fe9e6684edd5021a9e4/packages/flame/lib/src/extensions/vector2.dart#L126-L139

# Why should this be improved

At present, if there is a continuous call of the `moveToTarget()` inside for example `update()`, it could impact the performance since it has a `clone()`.
Adding a reusable vector can prevent creating a new Vector2 object in every tick.

# Any risks?

No

# More information

 Are you interested in working on a PR for this?  **Yes**

Hwan-seok:
# Description

Add a reusable static member which is Vector for reducing object(Vector2) creation overheads if available.

## Checklist

## Breaking Change?

## Related Issues

Closes #2427 


packages/flame/lib/src/extensions/vector2.dart
```diff
@@ -123,12 +123,21 @@ extension Vector2Extension on Vector2 {
...
   /// **not** a percentage (relative value).
+  ///
+  /// The [efficientDiff] is an optional parameter that can be used to avoid
+  /// creating a new [Vector2] when calling this method multiple times.
   void moveToTarget(
     Vector2 target,
-    double ds,
-  ) {
+    double ds, {
+    Vector2? efficientDiff,
```
spydon, line 132: Maybe we could just have a private static vector in here instead so that the user doesn't have to pass in this vector?

Hwan-seok: That's a great idea!
What do you think about sharing it inside the extension by naming it generally?
It might be useful when there is some point to save performance same as `moveToTarget`.

spydon: Sounds good :)

packages/flame/lib/src/extensions/vector2.dart
```diff
@@ -6,6 +6,13 @@ import 'package:vector_math/vector_math_64.dart';
 export 'package:vector_math/vector_math_64.dart' hide Colors;
 
 extension Vector2Extension on Vector2 {
+  /// This is a reusable vector that can be used with the [Vector2Extension]
```
spydon, line 9:
```suggestion
  /// This is a reusable vector that can be used within the [Vector2Extension]
```

spydon (approved):

spydon (approved):

PR #2430, opened 2023-03-25
fix: Auto-resize `SpriteComponent` on sprite change

### LINKED ISSUE #2419: `SpriteComponent` doesn't auto-resize
# Current bug behavior

If sprite of a `SpriteComponent` is set or updated after construction of the component, the size of the component does not update.

# Expected behavior

Ideally the size of `SpriteComponent` should be automatically updated by default. Additionally a an option can be provided to turn this resizing off by setting a flag in `SpriteComponent`.

# Steps to reproduce

Setting up a sprite component like this will make it render with zero size:
```dart
final spriteComponent = SpriteComponent();
spriteComponent.sprite = await Sprite.load('arrow_up.png');
add(spriteComponent);
```

# Flutter doctor output

NA

Flame version: 1.6.0
Platform affected: All

# Log information

```
```
NA

# More information

I can work on this.

ufrshubham:
# Description

This PR adds an `autoResize` parameter to `SpriteComponent` using which users will be able to control if the component should auto-resize on sprite change to match the sprite's `srcSize`. Some important design decisions about `autoResize`:

- It is mutually exclusive with `size` while construction: This means, if `size` is non-null while construction of `SpriteComponent`, `autoResize` cannot be true. This allows use-cases where one might want to render a sprite at a different size than its original size. If an user is setting the size explicitly, we can leave it up to them to handle resizing in such cases. 

- It is optional: If an user does not set this parameter while construction, it will be auto-populated depending upon if a `size` is provided or not.

## Checklist

## Breaking Change?

## Related Issues

Closes #2419

### DIFF
--- a/packages/flame/lib/src/components/sprite_component.dart
+++ b/packages/flame/lib/src/components/sprite_component.dart
@@ -22,8 +22,8 @@ class SpriteComponent extends PositionComponent
   /// Returns current value of auto resize flag.
   bool get autoResize => _autoResize;
 
-  /// Sets the given value of autoResize flag.
-  /// Will update the size if new value is true.
+  /// Sets the given value of autoResize flag. Will update the [size]
+  /// to fit srcSize of [sprite] if set to  true.
   set autoResize(bool value) {
     _autoResize = value;
     if (value) {
@@ -38,7 +38,7 @@ class SpriteComponent extends PositionComponent
   Sprite? get sprite => _sprite;
 
   /// Sets the given sprite as the new [sprite] of this component.
-  /// Will update the size is [autoResize] is set to true.
+  /// Will update the size if [autoResize] is set to true.
   set sprite(Sprite? value) {
     _sprite = value;
 
@@ -65,10 +65,9 @@ class SpriteComponent extends PositionComponent
     super.children,
     super.priority,
   })  : assert(
-          ((size == null) && (autoResize ?? true)) ||
-              ((size != null) && !(autoResize ?? false)),
-          '''If size is set, autoResize should be false and vice versa or
-          size should be null when autoResize is true''',
+          (size == null && (autoResize ?? true)) ||
+              (size != null && !(autoResize ?? false)),
+          '''If size is set, autoResize should be false or size should be null when autoResize is true.''',
         ),
         _autoResize = autoResize ?? size == null,
         _sprite = sprite,
--- a/packages/flame/lib/src/components/sprite_component.dart
+++ b/packages/flame/lib/src/components/sprite_component.dart
@@ -27,7 +27,11 @@ class SpriteComponent extends PositionComponent
   set autoResize(bool value) {
     _autoResize = value;
     if (value) {
-      size.setFrom(sprite?.srcSize ?? Vector2.zero());
+      if (_sprite != null) {
+        size.setFrom(_sprite!.srcSize);
+      } else {
+        size.setZero();
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/flame/lib/src/components/sprite_component.dart
```diff
@@ -15,12 +15,46 @@ export '../sprite.dart';
...
+
+  /// Returns current value of auto resize flag.
+  bool get autoResize => _autoResize;
+
+  /// Sets the given value of autoResize flag. Will update the [size]
+  /// to fit srcSize of [sprite] if set to  true.
+  set autoResize(bool value) {
+    _autoResize = value;
+    if (value) {
+      size.setFrom(sprite?.srcSize ?? Vector2.zero());
```
spydon, line 30: It should probably do a null check outside here instead and use `setZero` if it's null so that no new vector is created

spydon: Ah, just like you do further down

packages/flame/lib/src/components/sprite_component.dart
```diff
@@ -30,9 +64,14 @@ class SpriteComponent extends PositionComponent
     super.anchor,
     super.children,
     super.priority,
-  }) : super(
-          size: size ?? sprite?.srcSize,
-        ) {
+  })  : assert(
+          (size == null && (autoResize ?? true)) ||
+              (size != null && !(autoResize ?? false)),
```
spydon, line 69:
```suggestion
          (size == null && (autoResize ?? true)) ||
              (size != null && !(autoResize ?? false)),
```
Hmm, shouldn't this be possible to write as a single XOR expression?

spydon (approved):
Lgtm, just some small nits :)

packages/flame/lib/src/components/sprite_component.dart
ufrshubham: Done.

packages/flame/lib/src/components/sprite_component.dart
ufrshubham: Simplified as suggested.

luanpotter (approved):
Nice!

packages/flame/lib/src/components/sprite_component.dart
```diff
@@ -15,12 +15,50 @@ export '../sprite.dart';
...
+
+  /// Sets the given value of autoResize flag. Will update the [size]
+  /// to fit srcSize of [sprite] if set to  true.
+  set autoResize(bool value) {
+    _autoResize = value;
+    if (value) {
+      if (_sprite != null) {
+        size.setFrom(_sprite!.srcSize);
+      } else {
+        size.setZero();
```
erickzanardo, line 33: Shouldn't we just do nothing when it is false? I would expect the previous size to still be there if setting auto size to false

packages/flame/lib/src/components/sprite_component.dart
```diff
@@ -15,12 +15,50 @@ export '../sprite.dart';
...
+  /// Sets the given sprite as the new [sprite] of this component.
+  /// Will update the size if [autoResize] is set to true.
+  set sprite(Sprite? value) {
+    _sprite = value;
+
+    if (_autoResize) {
+      if (value != null) {
+        size.setFrom(value.srcSize);
+      } else {
+        size.setZero();
```
erickzanardo, line 53: Same suggestions

erickzanardo (commented):
Awesome idea! Addded a couple of comments before approving

packages/flame/lib/src/components/sprite_component.dart
ufrshubham: Wouldn't that break the contract of this setter? If `autoResize` is on and there is no sprite, wouldn't you expect the size of component to be zero as well?

erickzanardo: Ah, got ya, you are only setting the zero when there is no sprite. That makes sense!!!

erickzanardo (approved):

PR #2431, opened 2023-03-26
feat: Add `HasTimeScale` mixin

### LINKED ISSUE #2392: Add a variable to change the universal game speed
# Problem to solve
Sometimes we need to increase the game speed (to fast-forward a boring and long process or implement a slow-motion effect).
* For example in the [Ballz game](https://play.google.com/store/apps/details?id=com.ketchapp.ballz&hl=en&gl=US), there is a fast-forward button that you can click to speed up the game for a short time to see the final result quickly. (You see everything but with something like 3x or more speed)
* Or maybe you need to have a slow-motion and zoom effect when the user couldn't hit the target using his shot.

# Proposal
In Unity, there is a variable called [timeScale](https://docs.unity3d.com/ScriptReference/Time-timeScale.html), we need to have a variable like that in the Flame.
Simply, `timeScale` is a double (float in C#) variable that determines the speed of the game,
* When it is 1.0 -> Game runs at a normal speed
* When it is 0.5 -> Game runs in 2x slower than normal speed
* When it is 2.0 -> Game runs in 2x faster than normal speed
* And when it is 0.0 -> Game is paused

# More information
I checked some game engines and it seems they have exactly this concept with different names:
* In [Godot](https://godotengine.org/) this variable is called [Engine.time_scale](https://docs.godotengine.org/en/stable/classes/class_engine.html#class-engine-property-time-scale)
* In [Cocos2d](https://www.cocos.com/en) this variable is called [Scheduler.setTimeScale](https://docs.cocos.com/creator/2.2/api/en/classes/Scheduler.html#settimescale)

ufrshubham:
# Description

This PR adds a new mixin on `Component`. When attached to a component, it allows scaling the delta time of that component as well as all its children by a non-negative factor. The idea is to allows slowing down or speeding up the gameplay by change the scaling factor.

Note: This approach works only for framerate independent game logic. Code in `update()` that is not dependent on delta time will remain unaffected by time scale.

Here are few examples:

https://user-images.githubusercontent.com/33748002/227756636-ee00405a-2d6a-4751-91a3-2b78a522e723.mp4

https://user-images.githubusercontent.com/33748002/227756645-bc570e6b-4985-4d51-8f1a-5f2cbaef99c6.mp4

## Checklist

## Breaking Change?

## Related Issues

Closes #2392

### DIFF
--- a/doc/flame/other/util.md
+++ b/doc/flame/other/util.md
@@ -146,6 +146,37 @@ class MyFlameGame extends FlameGame {
 ```
 
 
+## Time Scale
+
+In many game it is often desirable to create slow-motion or fast-forward effect based on some in  
+game events. A very common apporach to achieve these results is to manipulate the in game time or
+tick rate.
+
+To make this manipulation easier, Flame provides a `HasTimeScale` mixin. This mixin can be attached
+to any Flame `Component` and exposes a simple get/set API for `timeScale`. The default value of
+`timeScale` is `1`, implying in-game time of the component is running at the same speed as real life
+time. Setting it to `2` will make the component tick twice as fast and setting it to `0.5` will make
+it tick at half the speed as compared to real life time.
+
+Since `FlameGame` is a `Component` too, this mixin can be attached to the `FlameGame` as well. Doing
+so will allow controlling time scale for all the component of the game from a single place.
+
+```dart
+import 'package:flame/components.dart';
+import 'package:flame/game.dart';
+
+class MyFlameGame extends FlameGame with HasTimeScale {
+  void speedUp(){
+    timeScale = 2.0;
+  }
+
+  void slowDown(){
+    timeScale = 1.0;
+  }
+}
+```
+
+
 ## Extensions
 
 Flame bundles a collection of utility extensions, these extensions are meant to help the developer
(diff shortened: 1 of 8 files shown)


munsterlander:
This is the smarter implementation of the whole game timescale.  I was also working on this but @ufrshubham had a great idea to make this component specific so you can allow somethings to be changed and not others kind of like matrix bullet time and stuff.  Overall, I think it's a brilliantly simple and effective solution.

spydon:
Nice! Needs some docs and tests otherwise it looks good to me. :)

ufrshubham:
> Nice! Needs some docs and tests otherwise it looks good to me. :)

Cool! Just wanted a go-ahead before jumping onto tests and docs. I'll get started with that now 🙂

renancaraujo (approved):
Love/needed this idea!

doc/flame/other/util.md
```diff
@@ -146,6 +146,37 @@ class MyFlameGame extends FlameGame {
 ```
 
 
+## Time Scale
+
+In many game it is often desirable to create slow-motion or fast-forward effect based on some in
```
spydon, line 151:
```suggestion
In many games it is often desirable to create  slow-motion or fast-forward effects based on some in  
```

ufrshubham:
> Lgtm, would be really nice if the example ran directly in the docs too!

Added a simple example in docs too.

PR #2442, opened 2023-03-29
feat: Add `autoResize` for `SpriteGroupComponent`

### LINKED ISSUE #2441: Auto-resize `SpriteGroupComponent`
# What could be improved

As of today, `SpriteGroupComponent` doesn't support auto-resizing the component when its `current` state is changed. So if the sprites in the group are of varying sizes, users are required to manually updates the size as well. For example, if car sprites looking east and north are rendered via `SpriteGroupComponent`, the output will look something like this, if size is not manually updated.

image

# Why should this be improved

It might help in reducing additional efforts from Flame user.

# Any risks?

None that I can think of.

# More information

I'll work on this.

ufrshubham:
# Description

This PR adds an `autoResize` parameter to `SpriteGroupComponent` which will allow Flame users to control if the size of the component should be re-calculated to fit the size of current active sprite in the group.

## Checklist

## Breaking Change?

## Related Issues

Closes #2441

### DIFF
--- a/packages/flame/lib/src/components/sprite_group_component.dart
+++ b/packages/flame/lib/src/components/sprite_group_component.dart
@@ -11,32 +11,63 @@ export '../sprite_animation.dart';
 class SpriteGroupComponent<T> extends PositionComponent
     with HasPaint
     implements SizeProvider {
-  /// Key with the current playing animation
-  T? current;
+  T? _current;
 
   /// Map with the available states for this sprite group
   Map<T, Sprite>? sprites;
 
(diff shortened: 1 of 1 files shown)


spydon (approved):
Lgtm! This should be done for `SpriteAnimationGroupComponent` too right?

ufrshubham:
> Lgtm! This should be done for `SpriteAnimationGroupComponent` too right?

Yes and probably `SpriteAnimationComponent` too. Should it be fine to create separate issues for them?

PR #2457, opened 2023-04-02
feat!: Add `SpriteAnimationTicker`

### LINKED ISSUE #1827: Sharing SpriteAnimation
# What could be improved

Would be better to warn users when same `SpriteAnimation` is shared between multiple `SpriteAnimationComponent`s or `SpriteAnimationGroupComponent`s.

# Why should this be improved

Right now, if a `SpriteAnimation` is shared between multiple components, it gets ticked by all the components that are using it, causing the animation to essentially speed up.

# Any risks?

None that I can think of.

# More information

ufrshubham:
# Description

This PR moves out the ticking logic from `SpriteAnimation` into a new class `SpriteAnimationTicker`. With this change, users will be able to share the same sprite animation between multiple component without the animation speed getting increased.

## Checklist

## Breaking Change?

### Migration instructions

All timing and frame related APIs from `SpriteAnimation` are moved to `SpriteAnimationTicker`. Tickers can be accessed from the `SpriteAnimation` based component.

## Related Issues

Closes #1827


doc/flame/components.md
```diff
@@ -498,47 +498,61 @@ Future<void> onLoad() async {
 }
 ```
 
-If you are not using `FlameGame`, don't forget this component needs to be updated, because the
-animation object needs to be ticked to move the frames.
+If you are not using `FlameGame`, don't forget this component needs to be updated, because a ticker
```
spydon, line 501: Not new for this PR, but I don't think we need to mention cases not using it as a component for every separate component and then we can just mention it once that in the overview for components.

spydon (approved):
Lgtm! Sliiiightly more cumbersome to use, but that's a tradeoff worth to take I think.

doc/flame/components.md
ufrshubham: Agreed. Probably it can even be a separate section altogether. Something like "Things to consider when not using `FlameGame`"

ufrshubham: Oh, just to be clear, is this change expected in this PR?😅

spydon: Whatever you feel like, it would be faster in this PR, but it has nothing to do with this PR so we'll definitely merge it anyways. 😅

erickzanardo (approved):

mosentest:
？

PR #2469, opened 2023-04-05
fix: It should be possible to re-add `ColorEffect`

spydon:
# Description

Currently you'll get an error that the `late final _original` field is already initialized if you re-add effect to a component, since that field is set in `onMount`. This PR simply removes final from that field so that it can be updated in `onMount`.

## Checklist

## Breaking Change?

## Related Issues


spydon:
> LGTM! Strange how no one discovered this until now 😄

Indeed, it must have been like this for ages!

PR #2498, opened 2023-04-14
fix: Modify size only if changed while auto-resizing

### LINKED ISSUE #2488: bug: with enabled autoResize transform listeners callbacks are called constantly
# Current bug behavior
If to subscribe to component's transform events and enable autoResize, callbacks will be called even there are no any changes.

# Expected behavior
Transform callbacks should be called only when something really changed

# More information
We should to check equality of current component's size and sprite's size before calling `size.setFrom` here: https://github.com/flame-engine/flame/blob/95e34515dfbeed09f7cbfaace5e3659321dcf659/packages/flame/lib/src/components/sprite_animation_group_component.dart#L172

ufrshubham:
# Description

Checks the size before modifying and skips if it hasn't changed.

## Checklist

## Breaking Change?

## Related Issues

Closes #2488

### DIFF
--- a/packages/flame/lib/src/components/sprite_animation_component.dart
+++ b/packages/flame/lib/src/components/sprite_animation_component.dart
@@ -139,11 +139,15 @@ class SpriteAnimationComponent extends PositionComponent
   void _resizeToSprite() {
     if (_autoResize) {
       _isAutoResizing = true;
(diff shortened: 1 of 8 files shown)


spydon (approved):
Thanks for fixing this up!

PR #2512, opened 2023-04-19
fix: Prevent memory leak in FlameBlocListenable when adoption

Hwan-seok:
# Description
FlameBlocListenable has a memory leak between adoptions because the subscription didn't cancel.

Added canceling the StreamSubscription within the lifecycle: `onMount`.

## Checklist

## Breaking Change?

## Related Issues


erickzanardo:
Hey, thanks for the PR!

Just wanted to understand this better, wdym about adoptions?

It would also help understanding the leak if you could add a test case?

Hwan-seok:
Hey @erickzanardo,
I added the test code that only shows the subscription is changed.

I wish I could write a test code that `StreamSubscription` is canceled when `_subscription.cancel()` is called but the cancel() could only be caught by `StreamController.onDone()` and also, `FlameBlocListenable._subscription` is originally from the `BlocBase._StateController` which is private. So we cannot catch the subscription canceled directly.
The only one I can show temporally is the code snippet I wrote in the test code which indicates the subscription is changed while adoption(parent changed) and the previous one can be dangling if there is no `cancel()` call.

Initially, I thought this PR is very simple because the following two are known things
1. `onMount()` can be called multiple times 
2. `StreamSubscription` would make a memory leak if `cancel()` is omitted

munsterlander:
@Hwan-seok If my pr for dispose gets merged, there is a new event that can be hooked into for this as well. I have been looking at the bloc listeners that are not closing like they are supposed to and I figure if they are closing in dispose it may finally function correctly.

Hwan-seok:
@spydon 
Sorry for the late response.

The `_subscription?.cancel();` in the `onMount` is definitely still needed because `onMount` executes multiple times.

But I will close this issue due to this PR cannot provide a way to test if it actually works because the bloc library doesn't provide some method for it.

Hwan-seok:
```
Short review of this PR:

Current behavior:
1. First mount of the `FlameBlocListenable`, `StreamSubscription` is created within the `onMount`, and let's tag this subscription as (A).
2. Change the parent of the `FlameBlocListenable` => without canceling the (A), another `StreamSubscription` (B) is created as same as before within the `onMount`.

Here comes what I was trying to do.
- So I am adding `StreamSubscription(A).cancel` before creating (B).
``` 



> Is cancel a no-op if it has already been cancelled?

As far as I know, Yes. You can see that at:
https://github.com/dart-lang/sdk/blob/8493e98bc42444af10cef30fe128b7bf2c3f9ea3/sdk/lib/async/stream_impl.dart#L192


>  Is it not possible to see if a subscription has been canceled?

We can see whether the subscription has been canceled or not from only `StreamController.onCancel` but unfortunately `StreamController` in the bloc library is private.


> Maybe the subscription could be mocked to add a check for that in the test?
I am not sure... I still lack so much. do you have an idea?

PR #2515, opened 2023-04-20
feat: Add onComplete callback to `AnimationWidget`

willhlas:
# Description

Adds optional onComplete callback to `SpriteAnimationWidget` for `asset` constructor.

## Checklist

## Breaking Change?

## Related Issues


packages/flame/lib/src/widgets/animation_widget.dart
```diff
@@ -36,7 +39,8 @@ class SpriteAnimationWidget extends StatelessWidget {
   })  : _animationFuture = animation,
         _animationTicker = animationTicker,
         errorBuilder = null,
-        loadingBuilder = null;
+        loadingBuilder = null,
+        _onComplete = null;
```
spydon, line 43: Why not expose onComplete from the default constructor too?

packages/flame/lib/src/widgets/animation_widget.dart
```diff
@@ -48,22 +52,27 @@ class SpriteAnimationWidget extends StatelessWidget {
     required String path,
     required SpriteAnimationData data,
     Images? images,
+    VoidCallback? onComplete,
```
spydon, line 55:
```suggestion
    this.onComplete,
```
I don't think onComplete needs to be private, and we can set it directly here.

spydon (commented):
Just some small nits, the idea is good!

packages/flame/lib/src/widgets/animation_widget.dart
erickzanardo: The default constructor receives a concrete instance of a `SpriteAnimation`, which can have the callback in it. So I believe we don't need that callback there?

But at the same time, I agree that there is no harm in having it there as well

ufrshubham (approved):
Not sure how useful the other callbacks will be in this widget, but it sure would look more complete if `onStart` and `onFrame` are exposed too. Anyways, current changes look fine to me 👍🏼

packages/flame/lib/src/widgets/animation_widget.dart
spydon: True, I'm fine with either way then

erickzanardo:
> Not sure how useful the other callbacks will be in this widget, but it sure would look more complete if `onStart` and `onFrame` are exposed too. Anyways, current changes look fine to me 👍🏼

Yeah, good call. I think we should expose either way, there may always have a use case that we can't see right away.

I would argue that we should do each callbacks in a separate PR, wdyt?

wolfenrain (approved):

packages/flame/lib/src/widgets/animation_widget.dart
erickzanardo: I would keep as is then :D

st-pasha (approved):

packages/flame/lib/src/widgets/animation_widget.dart
spydon: @willhlas, did you see this?

willhlas: @spydon just pushed the update, sorry for the delay!

spydon (approved):

PR #2540, opened 2023-05-17
fix: Fix disappearing text on TextBoxComponent for larger pixelRatios

luanpotter:
# Description

When updating to the new Flutter & Flame version I was forced to revisit the occurences of `window.pixelRatio` and realized there is actually a bug on TextBoxComponent for certain pixelRatios (depending on the size of the Canvas).

The bug only manifests after a certain pixelRatio for each base size because while the canvas is created with the unscaled size, it does appear to be able to "render" outside of it's bounds for the most times. However I imagine there must be some optimization under the hood, and after a certain size increment, the rendering is void.

## Checklist

## Breaking Change?


packages/flame/lib/src/components/text_box_component.dart
```diff
@@ -241,12 +241,13 @@ class TextBoxComponent<T extends TextRenderer> extends TextComponent {
...
-    final c = Canvas(recorder, size.toRect());
+    final scaledSize = size * pixelRatio;
+    final c = Canvas(recorder, scaledSize.toRect());
     c.scale(pixelRatio);
     _fullRender(c);
     return recorder.endRecording().toImageSafe(
-          (width * pixelRatio).ceil(),
-          (height * pixelRatio).ceil(),
+          scaledSize.x.ceil(),
+          scaledSize.y.ceil(),
```
luanpotter, line 250: this was also wrong - `width` and `height` would be referring to the previous `this.size` and not the parameter size witch is the new size. but this was not the source of the issue in my case

packages/flame/test/components/text_box_component_test.dart
```diff
@@ -148,6 +148,20 @@ void main() {
...
     );
   });
+
+  testGolden(
+    'Big upscale',
+    (game) async {
+      game.addAll([
+        TextBoxComponent(
+          text: 'quickly',
+          pixelRatio: 8,
```
luanpotter, line 158: this was more than enough to cause issues to me.
on my application anything more than 2.51 was causing issues.
the test seems to be a little more lenient - it appears to be dependent on the actual size of the canvas

spydon (approved):
Lgtm!

PR #2538, opened 2023-05-14
chore: Remove 1.8.0 deprecations

spydon:
# Description

Removes all the deprecated methods before 1.8.0 release.

## Checklist

- [x] I have followed the [Contributor Guide] when preparing my PR.
- [x] I have updated/added tests for ALL new/updated/fixed functionality.
- [x] I have updated/added relevant documentation in `docs` and added dartdoc comments with `///`.
- [x] I have updated/added relevant examples in `examples` or `docs`.

## Breaking Change?

- [x] No, this PR is not a breaking change.

## Related Issues

[Contributor Guide]: https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md
[Conventional Commit]: https://conventionalcommits.org
[CHANGELOG]: https://github.com/flame-engine/flame/blob/main/CHANGELOG.md

PR #2549, opened 2023-05-23
fix: Tiled component orthogonal test

### LINKED ISSUE #2548: flame_tiled is failing the goldens in Flutter 3.10
# Current bug behavior
flame_tiled is failing the goldens for creating an orthogonal map.

# Expected behavior
The golden test should pass.

# Steps to reproduce
Run the tiled_test suite.

* Flame version: main
* Flutter version: 3.10

erickzanardo:
# Description

## Checklist

## Breaking Change?

## Related Issues

Closes #2548

### DIFF
--- a/packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
+++ b/packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
@@ -106,19 +106,10 @@ abstract class RenderableLayer<T extends Layer> {
   /// position.
   /// https://doc.mapeditor.org/en/latest/manual/layers/#parallax-scrolling-factor
   void applyParallaxOffset(Canvas canvas, CameraComponent camera) {
-    final anchor = camera.viewfinder.anchor;
     final cameraX = camera.viewfinder.position.x;
     final cameraY = camera.viewfinder.position.y;
     final viewportCenterX = camera.viewport.size.x / 2;
     final viewportCenterY = camera.viewport.size.y / 2;
-
-    if (parent is GroupLayer) {
-      print(layer.name);
-      print(parallaxX);
-      print(parallaxY);
-      print(viewportCenterX);
-      print(viewportCenterY);
-    }
 
     // Due to how Tiled treats the center of the view as the reference
     // point for parallax positioning (see Tiled docs), we need to offset the
--- a/packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
+++ b/packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
@@ -106,10 +106,11 @@ abstract class RenderableLayer<T extends Layer> {
   /// position.
   /// https://doc.mapeditor.org/en/latest/manual/layers/#parallax-scrolling-factor
   void applyParallaxOffset(Canvas canvas, CameraComponent camera) {
+    final anchor = camera.viewfinder.anchor;
     final cameraX = camera.viewfinder.position.x;
     final cameraY = camera.viewfinder.position.y;
-    final viewportCenterX = camera.viewport.size.x / 2;
-    final viewportCenterY = camera.viewport.size.y / 2;
+    final viewportCenterX = camera.viewport.size.x * anchor.x;
+    final viewportCenterY = camera.viewport.size.y * anchor.y;
 
     // Due to how Tiled treats the center of the view as the reference
     // point for parallax positioning (see Tiled docs), we need to offset the
--- a/packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
+++ b/packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
@@ -109,8 +109,16 @@ abstract class RenderableLayer<T extends Layer> {
     final anchor = camera.viewfinder.anchor;
     final cameraX = camera.viewfinder.position.x;
     final cameraY = camera.viewfinder.position.y;
-    final viewportCenterX = camera.viewport.size.x * anchor.x;
-    final viewportCenterY = camera.viewport.size.y * anchor.y;
+    final viewportCenterX = camera.viewport.size.x / 2;
+    final viewportCenterY = camera.viewport.size.y / 2;
+
+    if (parent is GroupLayer) {
+      print(layer.name);
+      print(parallaxX);
+      print(parallaxY);
+      print(viewportCenterX);
+      print(viewportCenterY);
+    }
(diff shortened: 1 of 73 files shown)


spydon (approved):
Good job! 💙

packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
```diff
@@ -109,8 +109,16 @@ abstract class RenderableLayer<T extends Layer> {
     final anchor = camera.viewfinder.anchor;
     final cameraX = camera.viewfinder.position.x;
     final cameraY = camera.viewfinder.position.y;
-    final viewportCenterX = camera.viewport.size.x * anchor.x;
-    final viewportCenterY = camera.viewport.size.y * anchor.y;
+    final viewportCenterX = camera.viewport.size.x / 2;
+    final viewportCenterY = camera.viewport.size.y / 2;
+
+    if (parent is GroupLayer) {
```
erickzanardo, line 115: @spydon check this out. I am pretty sure this isn't right, but I got the test working by doing two changes.

The column was getting out of place because it have a parallax factor
Screenshot 2023-05-25 at 22 27 46

And the viewport size was zero, so the parallax was being calculated wrongly.

I believe that the viewport size should never be zero right? Even if it isn't a fixed one, it should resort to the canvas, for the test to pass, I hard coded the viewport size on the test here: https://github.com/flame-engine/flame/pull/2549/files#diff-92d1b7e7c2fa039cbccdba852c50d50aab7c0db8f1a206c03d8597ab097cb3c9R415

But we probably want that the viewport fallback to the canvas size when a fixed one isn't defined?

I also reverted this [change](https://github.com/flame-engine/flame/pull/2538/files#diff-461c0b715937055337c5a06eee3cfd5e3b3768525e098d217c8964eda5c8a0aaR112), because a top left anchor is x = 0, y =0. If we want the center of the viewport, shouldn't we always divide by two? Or maybe I am missing something, we need to keep the anchor into consideration but we should change the anchor to center on the test?

Idk, it is late and my head can't think straight anymore. But hopefully when you read this, some of it will make sense, or maybe the Erick of tomorrow, well rested will also be able to make sense of all this.

spydon: I think that we have to take the viewfinder anchor into consideration? But otherwise it looks correct.

erickzanardo: @spydon I am unsure about that, I feel that the camera anchor sounds affect the tiled parallax. The camera should only affect how the map is seen in the game, not how it is rendered. If my argument makes sense to yo, feel free to merge this tomorrow, otherwise, I am no tiled expert, so I am not really sure how to take it in account, so maybe we should seek for help here?

spydon (approved):

packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
kurtome: > The camera should only affect how the map is seen in the game, not how it is rendered

@erickzanardo I'm not sure I follow, the camera is passed in here so that it can affect how the map is seen by the camera, the parallax effect by definition can't be statically rendered up front since it's causes the layers to change relative to one another based on the camera's current position and zoom

packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
```diff
@@ -106,11 +106,10 @@ abstract class RenderableLayer<T extends Layer> {
   /// position.
   /// https://doc.mapeditor.org/en/latest/manual/layers/#parallax-scrolling-factor
   void applyParallaxOffset(Canvas canvas, CameraComponent camera) {
-    final anchor = camera.viewfinder.anchor;
     final cameraX = camera.viewfinder.position.x;
     final cameraY = camera.viewfinder.position.y;
-    final viewportCenterX = camera.viewport.size.x * anchor.x;
-    final viewportCenterY = camera.viewport.size.y * anchor.y;
+    final viewportCenterX = camera.viewport.size.x / 2;
+    final viewportCenterY = camera.viewport.size.y / 2;
```
kurtome, line 112: Did you test this in Flame and compare it to Tiled? I originally verified everything lined up by hand

packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
erickzanardo: Humm, yeah, I think that makes sense, I don't have an advanced knowledge on Tiled so I still get confused about those "more advanced" features from it.

But in fact a parallax needs to be affected by the camera, otherwise the parallax effect will not happen!

Either way, if we dismiss that statement of mine, the camera position and zoom is been taken into account, the only missing key here is about how we could apply the anchor too, and if it should be applied.

packages/flame_tiled/test/tiled_test.dart
```diff
@@ -406,29 +406,31 @@ void main() {
 
       final world = World(children: [component]);
       final cameraComponent = CameraComponent(world: world);
-      cameraComponent.viewfinder.anchor = Anchor.topLeft;
 
       // Need to initialize a game and call `onLoad` and `onGameResize` to
       // get the camera and canvas sizes all initialized
       final game = FlameGame(children: [world, cameraComponent]);
-      await game.ready();
+      cameraComponent.viewfinder.anchor = Anchor.center;
```
erickzanardo, line 413: @spydon @kurtome @ufrshubham 

So, following up the logic that we need to apply the camera's anchor to the parallax, I thought that, hey, maybe the code previously was already correct, but with the previous camera, tiled was always considering a center camera, hence why it was doing a `/ 2`, which would make the generated snapshots be based on a `Anchor.center` camera.

So for the test to pass, I just changed the camera's anchor on the test to center, to match the previous behaviour, which seems to make sense to me.

Let me know if this also make sense to you all.

packages/flame_tiled/lib/src/renderable_layers/renderable_layer.dart
erickzanardo: move the discussion to another comment https://github.com/flame-engine/flame/pull/2549#discussion_r1219505747

packages/flame_tiled/test/tiled_test.dart
spydon: Seems to make sense!

PR #2554, opened 2023-05-28
fix: Update sdk constraints to >=3.0.0

spydon:
# Description

Update the SDK constraint to use >= 3.0.0

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/doc/flame/examples/pubspec.yaml
+++ b/doc/flame/examples/pubspec.yaml
@@ -4,7 +4,7 @@ version: 1.0.0
 publish_to: none
 
 environment:
-  sdk: ">=2.18.0 <4.0.0"
+  sdk: ">=3.0.0 <4.0.0"
   flutter: ">=3.3.0"
 
 dependencies:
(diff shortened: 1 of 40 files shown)


luanpotter (approved):
LGTM

Gustl22:
@spydon was there a special reason to do this? I don't find any docs which propose to use constraints for dart sdk >=3.0.0. I mean every depending app needs the same constraints now, so I think it's kind of breaking (?).
Flutter still uses older constraints, I guess. Does it make sense to revert this?
https://github.com/flutter/packages/blob/4a5cc3e6947b2b9840d7d810fe310e00726803fb/packages/video_player/video_player/pubspec.yaml#L9

spydon:
> @spydon was there a special reason to do this? I don't find any docs which propose to use constraints for dart sdk >=3.0.0. I mean every depending app needs the same constraints now, so I think it's kind of breaking (?).

If we don't do this we can't use the new language features. And indeed, it's a bit breaking, but the migration should be trivial in almost all cases.

> Does it make sense to bump flutter min version to `3.10.0` as well as dart sdk 3.0.0 only comes with that version?

Yeah, that totally makes sense, I just forgot. 😅

PR #2561, opened 2023-05-31
fix: Remove `mustCallSuper` from `onComponentTypeCheck`

ASGAlex:
# Description

@mustCallSuper is very strict requirement. This is true only for hitboxes and internal engine's logic, but unnecessary for other, "client-side" game components.
When overriding `onComponentTypeCheck` for component you can simple return true or false and this would not break anything.

## Checklist

## Breaking Change?

### DIFF
--- a/doc/flame/collision_detection.md
+++ b/doc/flame/collision_detection.md
@@ -302,8 +302,9 @@ class Bullet extends PositionComponent with CollisionCallbacks {
       // do NOT collide with Player or Water
       return false;
     }
-    // Just return true if you not interested in component parent's type check result.
-    // Or call super it component's parent should be able to override the result.
+    // Just return true if you're not interested in the parent's type check result.
+    // Or call super and you will be able to override the result with the parent's
+    // result.
     return super.onComponentTypeCheck(other);
   }
 
--- a/doc/flame/collision_detection.md
+++ b/doc/flame/collision_detection.md
@@ -302,6 +302,8 @@ class Bullet extends PositionComponent with CollisionCallbacks {
       // do NOT collide with Player or Water
       return false;
     }
+    // Just return true if you not interested in component parent's type check result.
+    // Or call super it component's parent should be able to override the result.
     return super.onComponentTypeCheck(other);
   }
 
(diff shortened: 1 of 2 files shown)


spydon (commented):
There should be some dartdocs added for when one should call super at least then

spydon:
Yes, the decision is up to the game developer, but we should inform them of what is happening when super is called and with an example for when they'd want to do it so that the developer can make an informed decision on whether they should call super or not.

spydon:
Because there are cases when the developer would want to call super right?

doc/flame/collision_detection.md
```diff
@@ -302,6 +302,8 @@ class Bullet extends PositionComponent with CollisionCallbacks {
       // do NOT collide with Player or Water
       return false;
     }
+    // Just return true if you not interested in component parent's type check result.
+    // Or call super it component's parent should be able to override the result.
```
spydon, line 306:
```suggestion
    // Just return true if you're not interested in the parent's type check result.
    // Or call super and you will be able to override the result with the parent's
    // result.
```

packages/flame/lib/src/collisions/collision_callbacks.dart
```diff
@@ -79,6 +79,8 @@ mixin GenericCollisionCallbacks<T> {
   /// reimplement [onComponentTypeCheck]. The result of calculation is cached
   /// so you should not check any dynamical parameters here, the function
   /// intended to be used as pure type checker.
+  /// Call super.onComponentTypeCheck to get component parent's result of
```
spydon, line 82:
```suggestion
  /// Call super.onComponentTypeCheck to get the parent's result of the
```

packages/flame/lib/src/collisions/collision_callbacks.dart
```diff
@@ -79,6 +79,8 @@ mixin GenericCollisionCallbacks<T> {
...
+  /// type check if needed. In other causes this call is redundant in game code
```
spydon, line 83:
```suggestion
  /// type check if needed. In other causes this call is redundant in game code.
```

spydon (approved):
Thanks for the contribution!

PR #2566, opened 2023-06-07
feat: ComponentKey API

### LINKED ISSUE #2443: Proposal: Component Keys
# What could be improved
Finding a specific component in the component tree.

# Why should this be improved
Right now, we have a couple of ways of finding components in the component tree in a `FlameGame`, like the `firstChild`, `descendants().where(...)`, `where` and so on.

But, if we need to find a specific component in the tree, and do not worry about possible perfomance issues like iterating of the descendants of the game of a component node, developers usually have to keep a reference of it manually somewhere in their code.

Although this is not complicated, I feel this could be improved.

# Any risks?
None that I can think of.

# More information
My proposal is to introduce a similar API of the Flutter keys on widgets. In this proposal, components could receive an optional `Key` instance (or probably `ComponentKey` so we don't clash with Flutter's class), and when that is provided, that component could be retrieved directly from the game instance in an API like this:

```
final player = gameRef.getComponentById(const ComponentKey('player'));
```
We could also make that method generic for better type safeness

```
final player = gameRef.getComponentById<Player>(const ComponentKey('player'));
```

I haven't thought much on the implementation, but my (possible naive) suggestion is that we could simply keep a map in the game, and on the `Component.onMount` method, if a key was supplied, that component would put a reference of itself on the map using that key, and on its `onRemove`, it would be removed from that map.

That way, getting a component by key would always be a very cheap operation and I also believe that this API may have the bonus of being similar to Flutter's one, which may help Flutter developers that aren't familiar with Flame yet.

Looking forward for feedbacks on this.

erickzanardo:
# Description

This PR adds a new key api on FCS, which will allow users to get a component from the tree, without needing to iterate over all the children or a parent descendants.

## Checklist

## Breaking Change?

## Related Issues


packages/flame/lib/src/components/core/component_tree_root.dart
```diff
@@ -127,6 +127,23 @@ class ComponentTreeRoot extends Component {
       }
     });
   }
+
+  @mustCallSuper
+  @internal
+  void registerKey(ComponentKey key, Component component) {
+    _index[key] = component;
```
spydon, line 134: Maybe put an assertion here if the key is already in the map?

packages/flame/lib/src/components/core/component_tree_root.dart
```diff
@@ -127,6 +127,23 @@ class ComponentTreeRoot extends Component {
...
+  }
+
+  @mustCallSuper
+  @internal
+  void unregisterKey(ComponentKey key) {
+    _index.remove(key);
+  }
+
+  T? findByKey<T extends Component>(ComponentKey key) {
```
spydon, line 143: Maybe there could be a method that finds it by name too?

packages/flame/lib/src/components/core/component.dart
```diff
@@ -918,6 +932,11 @@ class Component {
   /// the output.
   int? get debugCoordinatesPrecision => 0;
 
+  /// A key that can be used to identify this component in the tree.
+  ///
+  /// It can be used to retrive this component from anywhere in the tree.
```
spydon, line 937:
```suggestion
  /// It can be used to retrieve this component from anywhere in the tree.
```

packages/flame/lib/src/components/core/component_tree_root.dart
erickzanardo: Done!

packages/flame/lib/src/components/core/component_tree_root.dart
erickzanardo: Done!

erickzanardo:
@spydon added a more specific example, let me know what you think

https://github.com/flame-engine/flame/assets/835641/ee2c08ff-e62d-4bff-9682-a8907a8ce2e7

spydon:
> @spydon added a more specific example, let me know what you think
> 
> https://github.com/flame-engine/flame/assets/835641/ee2c08ff-e62d-4bff-9682-a8907a8ce2e7
> 
> 

Cute! Love it, and it should be clear what is happening too 😊

spydon (approved):

PR #2569, opened 2023-06-09
feat: Add option for a custom image and asset loader

### LINKED ISSUE #2568: Tiled asset loader
# Problem to solve

I want to give users in my app the option to create their own maps. For this I want to sandbox the asset loading process.

# Proposal
Create an asset loader argument, for example with `(String path) => Uint8List` to get the image.

# More information

This feature can also has the feature to create multiple asset directories for different levels.

CodeDoctorDE:
# Description

Add images and bundle to the named parameters for the creation of a tiled map.

## Checklist

## Breaking Change?

No

### Migration instructions

\-

## Related Issues

Closes #2568

### DIFF
--- a/packages/flame_tiled/lib/src/flame_tsx_provider.dart
+++ b/packages/flame_tiled/lib/src/flame_tsx_provider.dart
@@ -1,10 +1,11 @@
 import 'package:flame/flame.dart';
+import 'package:flutter/services.dart';
 import 'package:tiled/tiled.dart';
 import 'package:xml/xml.dart';
 
 /// A implementation of [TsxProvider] use by RenderableTileMap.
 ///
-/// It uses [Flame.bundle] and has a built-in cache for the file read.
+/// It uses [Flame.bundle] or a custom asset bundle and has a built-in cache for the file read.
 class FlameTsxProvider implements TsxProvider {
   /// Parsed data for this tsx file.
   final String data;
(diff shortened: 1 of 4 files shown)


CodeDoctorDE:
I don't know why the ci fails
> Test run was interrupted.
Error: The operation was canceled.

spydon:
Sometimes the test stage gets cancelled by the CI for unknown reasons, this is what happened here.
What is strange is that the analyze step passed, because there are a few analyze issues, one line that is too long and some missing trailing commas (you can see the issues annotated in the `files` tab in GitHub).

packages/flame_tiled/lib/src/flame_tsx_provider.dart
```diff
@@ -5,7 +5,8 @@ import 'package:xml/xml.dart';
 
 /// A implementation of [TsxProvider] use by RenderableTileMap.
 ///
-/// It uses [Flame.bundle] or a custom asset bundle and has a built-in cache for the file read.
+/// It uses [Flame.bundle] or a custom asset bundle
+/// and has a built-in cache for the file read.
```
spydon, line 9: This is fine, but ideally it should only break once it reaches 80 and not before.

CodeDoctorDE: Hmm, okay. Should I change that?
For me it makes sense to break if a new part of the sentence has started

spydon: It's a bit like using a line break as a comma, something one wouldn't do in a book for example.
Anyhow this is fine, but now you know until the next PR. :)

spydon (approved):
Thanks for the contribution!

PR #2579, opened 2023-06-18
fix: Adds a check to confirm the component is not loaded

### LINKED ISSUE #2563: Failed assertion: line 768 pos 12: '!isMounted'
# Current bug behavior
When game widget is loaded and unloaded and reloaded to a page, it throws this
`'package:flame/src/components/core/component.dart': Failed assertion: line 768 pos 12: '!isMounted': is not true. `

# Expected behavior
The game should load when game widget is reloaded to a page

# Steps to reproduce
Below is a simplified example. Tap on Load Game button and then Home Page button. Finally reload the game by tapping on LoadGame button again.  

flame: ^1.7.3
Flutter 3.10.0 
Framework • revision 84a1e904f4 (4 weeks ago) • 2023-05-09 07:41:44 -0700
Engine • revision d44b5a94c9
Tools • Dart 3.0.0 • DevTools 2.23.1

```
void main() {
  runApp(const MyGameApp());
}

class MyGameApp extends StatelessWidget {
  const MyGameApp({super.key});

  @override
    final myGame = MyGame();
    return MaterialApp(
      title: "MyGame",
      theme: ThemeData(
        brightness: Brightness.dark,
        primarySwatch: Colors.red,
      ),
      onGenerateRoute: RouteGenerator.generateRoute,
      home: HomePage(myGame),
    );
  }
}

import 'package:quick_flame/game.dart';
import 'package:quick_flame/gamepage.dart';
import 'package:quick_flame/home.dart';

class RouteGenerator {
  static const mainView = '/home';
  static const gameView = '/game';

  static Route<dynamic> generateRoute(RouteSettings settings) {
    final args = settings.arguments;

    switch (settings.name) {
      case mainView:
        if (args is MyGame) {
          return MaterialPageRoute(builder: (_) => HomePage(args));
        } else {
          return _errorRoute();
        }

      case gameView:
        if (args is MyGame) {
          return MaterialPageRoute(builder: (_) => GamePage(args));
        } else {
          return _errorRoute();
        }
      default:
        // If there is no such named route in the switch statement, e.g. /third
        return _errorRoute();
    }
  }

  static Route<dynamic> _errorRoute() {
    return MaterialPageRoute(builder: (_) {
      return Scaffold(
        appBar: AppBar(
          title: const Text('Error'),
        ),
        body: const Center(
          child: Text('ERROR'),
        ),
      );
    });
  }
}


import 'package:quick_flame/route_generator.dart';

import 'game.dart';

class HomePage extends StatefulWidget {
  MyGame myGame;
  HomePage(this.myGame, {super.key});

  @override
  State createState() {
    return _HomePageState();
  }
}

class _HomePageState extends State<HomePage> with TickerProviderStateMixin {
  @override
    return SafeArea(
        child: Column(
      children: [
        TextButton(
          onPressed: () => {
            Navigator.of(context).pushReplacementNamed(RouteGenerator.gameView,
                arguments: widget.myGame)
          },
          child: const Text(
            "Load Page",
            style: TextStyle(fontSize: 20),
          ),
        )
      ],
    ));
  }
}


import 'package:quick_flame/game.dart';
import 'package:quick_flame/route_generator.dart';

class GamePage extends StatefulWidget {
  MyGame myGame;

  GamePage(this.myGame, {super.key});

  @override
  State createState() {
    return _GamePageState();
  }
}

class _GamePageState extends State<GamePage> {
  @override
    return SafeArea(
        child: Column(
      mainAxisSize: MainAxisSize.min,
      children: [
        Expanded(
            child: GameWidget(
          game: widget.myGame,
        )),
        TextButton(
            onPressed: () => {
                  Navigator.of(context).pushReplacementNamed(
                      RouteGenerator.mainView,
                      arguments: widget.myGame)
                },
            child: const Text(
              "Home Page",
              style: TextStyle(fontSize: 20),
            ))
      ],
    ));
  }
}


import 'dart:async';

  late RectangleComponent component;
  bool addVal = true;
  @override
  FutureOr<void> onLoad() {
    component = RectangleComponent.square(
        size: size.x / 2, position: size / 2, anchor: Anchor.center)
      ..paint.color = Colors.white;
    add(component);

    return super.onLoad();
  }

  @override
  void update(double dt) {
    component.position.x =
        addVal ? component.position.x + 10 : component.position.x - 10;
    component.paint.color = addVal ? Colors.lightBlue : Colors.red;
    addVal = !addVal;

    return super.update(dt);
  }
}


# More information
If I comment out  `return super.update(dt);`, this issue disappears.

Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel stable, 3.10.0, on Microsoft Windows [Version 10.0.19044.2965], locale en-US)
[√] Windows Version (Installed version of Windows is version 10 or higher)
[√] Android toolchain - develop for Android devices (Android SDK version 33.0.0) 
[√] Visual Studio - develop for Windows (Visual Studio Build Tools 2019 16.11.18)
[√] Android Studio (version 2021.2)
[√] VS Code, 64-bit edition (version 1.78.2)
[√] Connected device (4 available)
[√] Network resources

### PR DESCRIPTION
**Pull request by munsterlander**
# Description

This fixes https://github.com/flame-engine/flame/issues/2563 by validating that a component is mounted and the child component is not mounted before adding it to the queue.

## Checklist

## Breaking Change?

## Related Issues

Closes #2563 

### CODE REVIEW

**Comment by munsterlander** [comment]
@spydon I am going to need some help on this one.  The code works fine except for its breaking the `DoubleTapCallbacks`,`HasDraggableComponents`, and `HasTappableComponents` tests.  Some of their tests fail and others do not.  None of these tests use `onDetach`, so I am at a bit of a loss as to why it would be doing that.

**Comment by spydon** [comment]
Like I wrote on discord (so others can discuss too):
> The components shouldn't be removed, when the game is attached again `onLoad` shouldn't run. So when the game instance is used again it is in exactly the same state as when it was detached.

**Comment by munsterlander** [comment]
> Like I wrote on discord (so others can discuss too):
> 
> > The components shouldn't be removed, when the game is attached again `onLoad` shouldn't run. So when the game instance is used again it is in exactly the same state as when it was detached.

I realized late last night that I was being dumb in my implementation and forgot to focus on processLifecycleEvents onDetach.  I actually think there is a larger, albeit different issue though and will look at it now.  Regardless, it is very interesting that these tests call FlameGame onDetach - I don't think that is expected behaviour.

**Review by spydon — APPROVED** [review]
This looks like i! Could you add some test for it? :)

**Comment by munsterlander** [comment]
> This looks like i! Could you add some test for it? :)
Will do, but first I have to solve why those 3 tests fail.

@st-pasha @Hwan-seok The 3 test suites that are failing were written by you guys, so I am hoping you can fill me in on what I am missing.  

Here is a very simple test that is now failing.  In all of these, `FlameGame` class `onDetach / onRemove` gets called.

```dart
testWithFlameGame(
      'DoubleTapDispatcher is added to game when the callback is mounted',
      (game) async {
        final component = _DoubleTapCallbacksComponent();
        await game.add(component);
        await game.ready();

        expect(game.firstChild<DoubleTapDispatcher>(), isNotNull);
      },
    );
    ```
    And the component:
    ```dart
    class _DoubleTapCallbacksComponent extends PositionComponent
    with DoubleTapCallbacks {
  _DoubleTapCallbacksComponent() {
    anchor = Anchor.center;
    size = Vector2.all(10);
  }

  int doubleTapDown = 0;
  int doubleTapCancel = 0;
  int doubleTap = 0;

  @override
  void onDoubleTapUp(DoubleTapEvent event) {
    doubleTap++;
  }

  @override
  void onDoubleTapCancel(DoubleTapCancelEvent event) {
    doubleTapCancel++;
  }

  @override
  void onDoubleTapDown(DoubleTapDownEvent event) {
    doubleTapDown++;
  }
}
    ```

**Comment by munsterlander** [comment]
Ok everyone, I finally have this solved.  I have learned way more about the Flame engine internals than I had ever planned on, lol.  To summarise:

The issue has nothing to do with what it seemed like on the surface.  In short, the `_reAddChildren()` method is the culprit, because it recalls `_addChild` onto itself.  No big deal, but after the game and all components have been loaded and mounted, the component class that is called from the `FlameGame.mount`, is already mounted, thus `isMounted` AND `isLoaded` are true, BUT because `_reAddChildren()` clears the children, `child.isMounted` is false.  Thus we simply need to check for `if (isMounted && !child.isMounted) {` and everything works including the tests.

**Comment by munsterlander** [comment]
> This looks like it! Could you add some test for it? :)

I don't even know how to begin testing these scenarios.  I would think quite the contrary that the existing tests, caught when it was wrong.  To test this, you would have to have access to the internals that aren't exposed and you can't assert because its a valid scenario, so honestly, I am at a loss as to where to start.

File: packages/flame/lib/src/components/core/component.dart [hunk 1]
```diff
@@ -580,7 +580,7 @@ class Component {
     );
     child._parent = this;
     final game = findGame();
-    if (isMounted) {
+    if (isMounted && !child.isMounted) {
```
spydon, line 583: I don't fully get this, if the `child` is mounted already we will run `children.add(child);` from the `else`, that doesn't sound correct?

munsterlander: Ah yes,  the confusing part.  If you put a print statement in each section numerically (1 for isMount, 2 for else, and 3 for the bottom if), the order of load is 2,1,3,1,3. 

So if you use the flow I sent you in discord, it helps to follow, but: 

2 = add() is called in onload
1 = called again for onMount
3 = future returns which was call for 1
1 = reAddChildren call to reorder parent->child relationships (also where I think global priority could be added) 
3 = future returns which was call for 1 from reAddChildren
Done.

So in short, the _addChild gets called 5 times for every component to progress through the life cycle. 

That's what had me so confused all this time. That method is 1 incorrect line away from an infinite loop - ask me how I know, lol.

spydon: So to clarify for myself when reading this again; this case is for when it comes in here from `detach`

munsterlander: It's actually for any case in which the components have been loaded and mounted but now are not mounted. 

This error could come up where you detached a component but never tripped their remove events and you then add a new component (without ever leaving FlameGame) which would trip reAddChildren which then would throw the error.

munsterlander: Loading Diagram drawio

Hopefully this helps others map out the loading process....

munsterlander: So this is what clued me in:

Initial adding of the Rectangle:
```dart
flutter: _reAddChildren this: Instance of 'MyGame' and isLoaded: true and isMounted: true
flutter: _reAddChildren child: Instance of 'RectangleComponent' and isLoaded: true and isMounted: false
```
And this is when it was being reloaded:
```
flutter: _reAddChildren this: Instance of 'MyGame' and isLoaded: true and isMounted: true
flutter: _reAddChildren child: Instance of 'RectangleComponent' and isLoaded: true and isMounted: true
```

So originally it was passing an already mounted component to the queue for `updateTree` which only knows to mount, remove, or move parents for components.  As the rectangle was already mounted, it threw the error.

munsterlander: > I don't fully get this, if the `child` is mounted already we will run `children.add(child);` from the `else`, that doesn't sound correct?

You are correct BUT `children.add` in this instance is just the `ComponentSet` and because `children` gets wiped every `reAddChildren` cycle, it never grows. as evidenced by this dump:

```dart
flutter: Load Page clicked
flutter: _addChild this: Instance of 'MyGame' and isLoaded: false and isMounted: false
flutter: _addChild child: Instance of 'RectangleComponent' and isLoaded: false and isMounted: false
flutter: Adding to Children and the length before is: 0
flutter: Adding to Children and the length after is: 1

flutter: Home Page clicked

flutter: Load Page clicked
flutter: _addChild this: Instance of 'MyGame' and isLoaded: true and isMounted: true
flutter: _addChild child: Instance of 'RectangleComponent' and isLoaded: true and isMounted: true
flutter: Adding to Children and the length before is: 0
flutter: Adding to Children and the length after is: 1
```

spydon:
> > This looks like it! Could you add some test for it? :)
> 
> I don't even know how to begin testing these scenarios. I would think quite the contrary that the existing tests, caught when it was wrong. To test this, you would have to have access to the internals that aren't exposed and you can't assert because its a valid scenario, so honestly, I am at a loss as to where to start.

You should only have to make sure that `onLoad` only runs once when a component is re-added right, you could just create a component with a counter in `onLoad` for that?
Something like this:
https://github.com/flame-engine/flame/blob/main/packages/flame/test/components/component_test.dart#L1319

munsterlander:
Ok, during the testing of the induced failing tests for Tap and Drag callbacks, I discovered that `onRemove` was not being called, so that has been fixed here, BUT more importantly, I learned that when you remove the game from the Flutter widget tree, you aren't removing (unmounting / detaching) the `FlameGame` or `Game` class themselves, but you are detaching the `GameRenderBox`.  Just like components, you would expect the `FlameGame` to remember its state, so it should not dismount itself as I previously committed and thought.  It's a component, it should stay exactly like the other children.  Thus the code was fine with my original fix, but my test was wrong.  We can now confirm the right state by simply checking to see if the game is attached or not.

spydon (approved):

PR #2580, opened 2023-06-20
docs: Update step4.md to use `!isFaceUp` instead of nonexistent `isFaceDown`

JosephCarrington:
# Description
`isFaceDown` does not exist on `Card`. Instead, use `!isFaceUp`

## Checklist

### Migration instructions

-->


spydon (commented):
`isFaceDown` does exist in the final code at least, so then it should just be added to `Card` if it is missing there in the tutorial:
https://github.com/flame-engine/flame/blob/main/doc/tutorials/klondike/app/lib/step4/components/card.dart#L27

PR #2593, opened 2023-06-28
feat: Option to use toImageSync in ImageComposition class

### LINKED ISSUE #2315: Option to use toImageSync in ImageComposition class
# What could be improved

With latest Flutter update `toImageSync` function now rasterize images on GPU and also keeps image in GPU for future rendering.
So it would be useful to allow uset to choose between `toImage` variants when calling `ImageComposition.compose()` function. For example, by adding function `composeSync` which will be the same as `compose` but with `toImageSync` call at the end. 

# Why should this be improved

This allows user to choose between 'GPU-accelerated' version and less efficient but asynchronous version of image composition.

# Any risks?

This should not affect any of excisting Flame project.

# More information

Actually I have no idea, how faster this should work, but if Flutter team promizes faster rendering - it worth to try then!

ASGAlex:
# Description

Added `ImageComposition.composeSync()` function

## Checklist

## Breaking Change?

## Related Issues

Closes #2315

### DIFF
--- a/doc/flame/rendering/images.md
+++ b/doc/flame/rendering/images.md
@@ -237,7 +237,7 @@ Image imageSync = composition.composeSync();
 
 As you can see, two versions of composing image are available. Use `ImageComposition.compose()` for
 the async approach. Or use the new `ImageComposition.composeSync()` function to rasterize the
-image into GPU context using the benefits of `Picture.toImageSync` function.
+image into GPU context using the benefits of the `Picture.toImageSync` function.
 
 **Note:** Composing images is expensive, we do not recommend you run this every tick as it affect
 the performance badly. Instead we recommend to have your compositions pre-rendered so you can just
--- a/doc/flame/rendering/images.md
+++ b/doc/flame/rendering/images.md
@@ -236,7 +236,7 @@ Image imageSync = composition.composeSync();
 ```
 
 As you can see, two versions of composing image are available. Use `ImageComposition.compose()` for
-old async approach. Or switch to new `ImageComposition.composeSync()` function to rasterize the
+the async approach. Or use the new `ImageComposition.composeSync()` function to rasterize the
 image into GPU context using benefits of `Picture.toImageSync` function.
 
 **Note:** Composing images is expensive, we do not recommend you run this every tick as it affect
--- a/doc/flame/rendering/images.md
+++ b/doc/flame/rendering/images.md
@@ -230,9 +230,14 @@ final composition = ImageComposition()
     Vector2(128, 0),
     source: Rect.fromLTWH(32, 32, 64, 64),
   );
-
+  
 Image image = await composition.compose();
+Image imageSync = composition.composeSync();
 ```
+
+As you can see, two versions of composing image are available. Use `ImageComposition.compose()` for
+old async approach. Or switch to new `ImageComposition.composeSync()` function to rasterize the
+image into GPU context using benefits of `Picture.toImageSync` function. 
 
 **Note:** Composing images is expensive, we do not recommend you run this every tick as it affect
 the performance badly. Instead we recommend to have your compositions pre-rendered so you can just
--- a/doc/flame/rendering/images.md
+++ b/doc/flame/rendering/images.md
@@ -237,7 +237,7 @@ Image imageSync = composition.composeSync();
 
 As you can see, two versions of composing image are available. Use `ImageComposition.compose()` for
 old async approach. Or switch to new `ImageComposition.composeSync()` function to rasterize the
-image into GPU context using benefits of `Picture.toImageSync` function. 
+image into GPU context using benefits of `Picture.toImageSync` function.
 
 **Note:** Composing images is expensive, we do not recommend you run this every tick as it affect
 the performance badly. Instead we recommend to have your compositions pre-rendered so you can just
--- a/doc/flame/rendering/images.md
+++ b/doc/flame/rendering/images.md
@@ -237,7 +237,7 @@ Image imageSync = composition.composeSync();
 
 As you can see, two versions of composing image are available. Use `ImageComposition.compose()` for
 the async approach. Or use the new `ImageComposition.composeSync()` function to rasterize the
-image into GPU context using benefits of `Picture.toImageSync` function.
+image into GPU context using the benefits of `Picture.toImageSync` function.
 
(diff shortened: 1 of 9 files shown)


doc/flame/rendering/images.md
```diff
@@ -230,10 +230,15 @@ final composition = ImageComposition()
...
     source: Rect.fromLTWH(32, 32, 64, 64),
   );
-
+  
 Image image = await composition.compose();
+Image imageSync = composition.composeSync();
 ```
 
+As you can see, two versions of composing image are available. Use `ImageComposition.compose()` for
+old async approach. Or switch to new `ImageComposition.composeSync()` function to rasterize the
```
spydon, line 239:
```suggestion
the async approach. Or use the new `ImageComposition.composeSync()` function to rasterize the
```

doc/flame/rendering/images.md
```diff
@@ -230,10 +230,15 @@ final composition = ImageComposition()
...
+image into GPU context using benefits of `Picture.toImageSync` function.
```
spydon, line 240:
```suggestion
image into GPU context using the benefits of `Picture.toImageSync` function.
```

packages/flame/lib/src/parallax.dart
```diff
@@ -196,9 +197,8 @@ class ParallaxAnimation extends ParallaxRenderer {
 
     final animation =
         await SpriteAnimation.load(path, animationData, images: images);
-    final prerenderedFrames = await Future.wait(
-      animation.frames.map((frame) => frame.sprite.toImage()).toList(),
-    );
+    final prerenderedFrames =
+        animation.frames.map((frame) => frame.sprite.toImageSync()).toList();
```
spydon, line 201: When we are in an async context, why would we want to use the sync method? This also loads them in a serial manner instead of parallelized like the old method did.

packages/flame/lib/src/sprite.dart
```diff
@@ -130,4 +130,14 @@ class Sprite {
       ..add(image, Vector2.zero(), source: src);
     return composition.compose();
   }
+
+  /// Return a new [Image] based on the [src] of the Sprite.
+  ///
+  /// A sync version of [toImage] function. Read [Picture.toImageSync] for
+  /// detailed description of possible benefits in performance
```
spydon, line 137:
```suggestion
  /// A sync version of the [toImage] function. Read [Picture.toImageSync] for a
  /// detailed description of possible benefits in performance.
```

packages/flame/lib/src/parallax.dart
ASGAlex: Hmm, I can also create here additional sync constructor. 
According to docs, there is no problem with 'serial manner' because rasterizing going async under hood. But all exceptions will be thrown at the render time, not in toImageSync function.

ASGAlex: Done.

packages/flame/lib/src/parallax.dart
```diff
@@ -55,6 +55,7 @@ extension ParallaxExtension on Game {
     ImageRepeat repeat = ImageRepeat.repeatX,
     Alignment alignment = Alignment.bottomLeft,
     LayerFill fill = LayerFill.height,
+    bool composeSync = false,
```
spydon, line 58: I don't think we need to expose this option to the user, or what would be the reason for that?
Can we get a runtime crash if using the sync method when it starts rendering and the rasterization isn't done?

ASGAlex: In theory - yes, we can. But I used it everywhere since this feature introduction and never had any problems. 
From perfectionists point of view - if we decide to use Sync everywhere, draw exception also should be catched everywhere in render methods, but I'm in doubts how to catch a thing I had never seen and have no idea how to reproduce..

spydon: Let's skip this Boolean and just use sync then. So the sync method is pretty much fire and forget then, if it has no impact on parallelization like you say?

ASGAlex: I haven't noticed that, maybe you have launched my demo projects with alternative collision system recently - there is a lot of `toImageSync` operations per update, and everything works fine even in browser.
Also Flutter team promizes that everything should be async under hood: https://api.flutter.dev/flutter/dart-ui/Picture/toImageSync.html. I hope we can beleive them :-) 
Of course I'd prefer to use more performant implementation instead old async version everywhere.

ASGAlex: Maybe we can switch this boolean to true by default in case if somebody will have an issue with that. After some time without bugreports we could to remove option completely. I see your point that unnecessary code is not good thing, but on other hand I am not a representative enougth, if somebody will have any problems with Sync on specific devices - he could just change the flag in constructor, without needing to fork whole framework to change just one line of code...

spydon: Let's do the other way around instead, let's skip this code for now and if people report bugs regarding it we'll introduce the boolean again.

doc/flame/rendering/images.md
```diff
@@ -230,10 +230,15 @@ final composition = ImageComposition()
...
   );
-
+  
 Image image = await composition.compose();
+Image imageSync = composition.composeSync();
 ```
 
+As you can see, two versions of composing image are available. Use `ImageComposition.compose()` for
+the async approach. Or use the new `ImageComposition.composeSync()` function to rasterize the
+image into GPU context using the benefits of `Picture.toImageSync` function.
```
spydon, line 240:
```suggestion
image into GPU context using the benefits of the `Picture.toImageSync` function.
```

packages/flame/lib/src/sprite.dart
```diff
@@ -130,4 +130,14 @@ class Sprite {
...
     return composition.compose();
   }
+
+  /// Return a new [Image] based on the [src] of the Sprite.
+  ///
+  /// A sync version of the [toImage] function. Read [Picture.toImageSync] for a
+  /// detailed description of possible benefits in performance.
+  Image toImageSync() {
+    final composition = ImageComposition()
+      ..add(image, Vector2.zero(), source: src);
```
spydon, line 140: The `Vector2.zero` should be possible to add as a static on this class instead, so that it doesn't need to be recreated for every call.

ASGAlex: Done.

spydon (approved):

PR #2602, opened 2023-07-06
fix: FlameGame onRemove fix to prevent memory leak

munsterlander:
# Description

In addressing https://github.com/flame-engine/flame/issues/2403, this PR removes the game widget `onRemove` call which was technically locked into an internal method and places the call in `FlameGame` that can then be triggered by `Game` which can be triggered by `GameRenderBox.detach`. 

So why not use `dispose` as was discussed in the issue?  I would refer to the Flutter API docs: https://api.flutter.dev/flutter/widgets/State/dispose.html

_Called when this object is removed from the tree permanently._

_The framework calls this method when this [State](https://api.flutter.dev/flutter/widgets/State-class.html) object will never build again. After the framework calls [dispose](https://api.flutter.dev/flutter/widgets/State/dispose.html), the [State](https://api.flutter.dev/flutter/widgets/State-class.html) object is considered un[mounted] (https://api.flutter.dev/flutter/widgets/State/mounted.html) and the mounted property is false. It is an error to call [setState](https://api.flutter.dev/flutter/widgets/State/setState.html) at this point. This stage of the lifecycle is terminal: there is no way to remount a [State](https://api.flutter.dev/flutter/widgets/State-class.html) object that has been disposed._

So, the issue comes down to you may be removing a `FlameGame` but not want it permanently destroyed.  Since you need the state maintained, then by naming the method `dispose()`, users would think it permanently destroys the game.  So, this keeps it with `onRemove` and thus inline with user expectations imo.

A user simply needs to override `onRemove` and set `FlameGame.disposeAll = true` then call `super.onRemove()` which did not exist previously for `FlameGame`; thus, I do not think this is a breaking change.

```dart
  @override
  void onRemove() {
    disposeAll = true;
    super.onRemove();
  }
  ```
Currently, due to my lack of knowledge of bloc, the test is not working as the game is never mounted, so I am posting the PR in order to get some assistance with that.  I tried to use the exact code that was originally posted in the issue, so we can validate the fix resolves that specific leak.  This code currently passes all tests and the behavior when stepping through the code, is as would be expected for the life cycle.

I am hoping @imaNNeo can take a look at the test and provide some pointers.

## Checklist

## Breaking Change?

## Related Issues

### CODE REVIEW

**Comment by munsterlander** [comment]
I will look at the analyzer error. I guess technically your don't have to call super and simply if a user overrides it without calling super than it's on them to dispose.

File: packages/flame/lib/src/game/game.dart [hunk 1]
```diff
@@ -202,6 +202,7 @@ abstract mixin class Game {
   ///
   /// Should not be called manually.
   void detach() {
+    onRemove();
     _gameRenderBox = null;
 
     onDetach();
```
spydon, line 208: Sorry, can you remind me again why this is not done in `onDetech` instead of `onRemove`?

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -86,6 +86,24 @@ class FlameGame extends ComponentTreeRoot
     setMounted();
   }
 
+  /// Setting to true will remove all children and process any remaining life
+  /// cycle events.
+  bool disposeAll = false;
```
spydon, line 91: This name should reflect when it is doing the disposal, since there is no natural way of doing this.

spydon (commented):
I'm thinking that maybe it is better to just add a `dispose` method instead, just like `Image.dispose`, which we provide as a convenience method for the user, but the user actually has to call it themselves. We should also empty the caches within this method.

packages/flame/lib/src/game/flame_game.dart
ufrshubham: Maybe `disposeAllOnRemove`?

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -86,6 +86,24 @@ class FlameGame extends ComponentTreeRoot
...
+  /// You can override this for additional behavior, but for [disposeAll] to be
+  /// triggered, you must call super.
+  @override
+  void onRemove() {
+    if (disposeAll) {
+      removeAll(children);
+      processLifecycleEvents();
+    }
+    super.onRemove();
+  }
```
ufrshubham, line 105: I was wondering if this behavior can be moved to `Component` and `ComponentTreeRoot`? Basically one handles removal of children and the other handle lifecycle events.

```dart
class Component {
  bool disposeChildrenOnRemove = false;
  void onRemove() {
    if (disposeChildrenOnRemove) {
      removeAll(children);
    }
  }
}

class ComponentTreeRoot extends Component {
  @override
  void onRemove() {
    super.onRemove();
    if (disposeChildrenOnRemove) {
      processLifecycleEvents();
    }
  }
}
```

spydon: I think that it probably would be better to keep disposal at a top level for now and empty the global caches in there. Possibly create a new lifecycle event which is `onDispose` for the components, but I wouldn't go that far already until we really see a need for it. I don't think disposal should be this tightly coupled with `onRemove`.

ufrshubham (approved):

packages/flame/lib/src/game/game.dart
munsterlander: Sorry for taking so long to respond to all of this.  I have / am still out on work travel and have been slammed.  The issue is that components had onRemove that got called BUT game.dart's onRemove never gets called - ever.  So, the concept of the game level event method BEFORE it is removed from the tree, even though it says it does this in the docs, it never makes that call.  This gets the onRemove method called for game now before it is detached.  This needs to be called when game is actually being detached.  I created this graphic just now, because if you chase the events all the way up into the Flutter widget tree, you will see this event was being missed by Flame.

flame_tree

packages/flame/lib/src/game/flame_game.dart
munsterlander: What I really wanted was a named optional parameter for onRemove but that would then be a breaking change, so I was trying to come up with a way to make it accessible and clear.  Because google says `dispose` is permanent and there is no coming back, I wanted to keep it like that because if set to true, then everything gets wiped.  It could be `disposeAllOnRemove` but you still have to call super to invoke it.  

I didn't make it `@mustcallsuper` because after talking with @spydon, that would be breaking and again, this is really optional and in their opinion, cleaning up your game state is really the users decision and not something Flame should be deciding.  This gives them the option, but does not force it on the user.

packages/flame/lib/src/game/flame_game.dart
munsterlander: No no.  You are falling in the trap that I got into in the beginning with this.  `FlameGame` likes to think its a component (my chance at being funny), but its really like a quasi component because it isn't treated like other components.  Its the root of all components and as such, its lifecycle events are very different from other components.

Components already have `onMount`,`onRemove` and I would caution against using the term dispose in the component because, as stated above, that should be permanent and never coming back, so the user would have to re-instantiate the component.  Of which, I think that is far worse to do to a user.

munsterlander:
I would add, it took my a long time to really realize how the Flutter widget tree and Flame component tree exist together.  Although the lifecycle events have the same names, they are not invoking each other across the spectrum.  Only at the entry points that I specify in the above graphic, do they interact; otherwise, they are completely isolated.

I must give tons of credit to @Holofox https://stackoverflow.com/questions/49966808/how-to-get-the-name-of-the-current-and-calling-function-in-dart for their `LoggerStackTrace` function as I put those calls everywhere and finally deciphered the lifecycle across Flutter -> Game -> Components.  It was the main source to solutions in https://github.com/flame-engine/flame/pull/2579

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -86,6 +86,24 @@ class FlameGame extends ComponentTreeRoot
...
+
+  /// Called when the game is about to be removed from the Flutter widget tree,
+  /// but before it is actually removed.
+  ///
+  /// You can override this for additional behavior, but for
+  ///  [disposeAllOnRemove] to be triggered, you must call super.
+  @override
+  void onRemove() {
+    if (disposeAllOnRemove) {
+      removeAll(children);
```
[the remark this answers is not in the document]
spydon: Those will already be ignored, a component can't be removed twice without being re-added again.
Me and Munsterlander is going to have a chat about the current implementation, so it might come to change soon.

doc/flame/game.md
```diff
@@ -93,6 +93,20 @@ in order: `onLoad`, `onGameResize` and `onMount`. After that, it goes on to call
...
 
+```{note}
+The `onRemove` can be used to clean up potential memory leaks by doing the following:
+```
+
+```dart
+  @override
+  void onRemove() {
+    removeAll(children);
+    processLifecycleEvents();
```
spydon, line 104:
```suggestion
    processLifecycleEvents();
    Flame.images.clearCache();
    Flame.assets.clearCache();
```

doc/flame/game.md
```diff
@@ -93,6 +93,20 @@ in order: `onLoad`, `onGameResize` and `onMount`. After that, it goes on to call
...
+    // Any other code necessary to remove references that will be orphaned when
+    // your game is removed.
```
spydon, line 106:
```suggestion
    // Any other code that you want to run when the game is removed.
```
I think this can be simplified to this, because these references should already be handled by the GC if you just make sure that you don't keep a reference to the game class itself.

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -86,6 +86,24 @@ class FlameGame extends ComponentTreeRoot
     setMounted();
   }
 
+  /// Setting to true will remove all children and process any remaining life
+  /// cycle events.
+  bool disposeAllOnRemove = false;
```
spydon, line 91: We agreed that this should be removed right, and `onRemove` should be left empty?

packages/flame/lib/src/game/game.dart
```diff
@@ -202,13 +202,15 @@ abstract mixin class Game {
...
   }
 
   /// Called when the game is about to be removed from the Flutter widget tree,
-  /// but before it is actually removed.
+  /// but before it is actually removed.  See the docs for an example on how to
+  /// cleanup children to avoid memory leaks.
```
spydon, line 213:
```suggestion
  /// do cleanups to avoid memory leaks.
```

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -256,7 +256,6 @@ class GameWidgetState<T extends Game> extends State<GameWidget<T>> {
 
   void disposeCurrentGame() {
     currentGame.removeGameStateListener(_onGameStateChange);
-    currentGame.onRemove();
```
spydon, line 259: This was the one you thought never could be called right?

packages/flame/lib/src/game/flame_game.dart
munsterlander: OOPS!  I forgot to remove that entirely.  `onRemove` was added by me to `FlameGame` so that whole section can be removed.

packages/flame/lib/src/game/game_widget/game_widget.dart
munsterlander: Yes, that is not showing in my file.  Let me see why the file diff is not being picked up.

spydon: It is showing as removed in GitHub too, just checking :)

munsterlander: > This was the one you thought never could be called right?

I also just want to add for others reading this, it does get called for many situations, but the bug was, it does NOT get called when a route is destroyed that contains the game.

spydon (approved):
Great job! Few lines in the end, but massive underlying work 👏

imaNNeo:
Sorry, I was not able to test it yet. But I am amazed to see how actively you worked on this issue.

PR #2616, opened 2023-07-21
feat: Add optional world input to `CameraComponent.canSee`

ufrshubham:
# Description

`CameraComponent.canSee` wasn't performing any kind of sanity checks on the given components world or mounted-ness. This PR adds these checks to correctly return false if the component is not mounted or if the optional world is not the same as camera's current target world.

## Checklist

## Breaking Change?

## Related Issues

NA


spydon (commented):
Hmm, I wonder if it is worth looking up the component's world, since this method is used for culling it needs to be as efficient as possible. Maybe just allow for an optional world to be sent in that the component belongs to? Or just write in the dartdoc that the user should check that it belongs to the camera component's world first, it's probably a very niche use-case when that check needs to be performed anyways.

The mounted check is good though!

ufrshubham:
> Maybe just allow for an optional world to be sent in that the component belongs to? 

Decided to go with this.

> The mounted check is good though!

Also added a `.isMounted` check for the world.

Issue #2011, opened 2022-10-04
Add scan distance input for raycast


ufrshubham:
# Problem to solve

Current raycast API allows for ignoring hitboxes, but it does not allow limiting the hit scan distance for the cast. As a result users have to perform additional check on distance in their code.

Use case: Implementing an entity with ranged sight. For example, an enemy that can see objects within a certain range.

![image](https://user-images.githubusercontent.com/33748002/193758796-3b907860-0027-4187-b53f-7fd92f88a2df.png)
![image](https://user-images.githubusercontent.com/33748002/193758826-92208125-0dd0-46ef-bcdc-325b97b7ae3a.png)

# Proposal

Add an optional input to raycast using which users can specify the max distance for scanning ray hits.

# More information

I will work on this.

spydon:
Haha, love the images!

PR #2617, opened 2023-07-21
perf: Improve performance of raycasts

filiph:
# Description
Creates an axis-aligned bounding box (AABB) around the ray to short-circuit checking.

In the general case, we skip ~75% of the checks because the ray originates at some point in the 2D cartesian space and goes towards infinity. The AABB in that case is a quadrant of space, and so only entities intersecting with this quadrant of space need to be checked.

In the case where the raycast is called with a `maxDistance`, the savings are much higher. A relatively short ray will generate a small AABB around its origin and end point. In a typical game world, there will be many entities that do not intersect with this AABB, and can be skipped.

## Results

Toy example when run without the optimization:
Screenshot 2023-07-21 at 12 27 33

Toy example when run with optimization but unbounded `maxDistance`:
Screenshot 2023-07-21 at 13 28 19

Toy example when run with optimization and a reasonable `maxDistance`:
Screenshot 2023-07-21 at 13 54 59

That's 200 entities all sending 3 raycasts each.

I didn't add any tests since raycasts are tested, including maxDistance.

## Checklist

## Breaking Change?

## Related Issues

The original maxDistance issue: #2011

### DIFF
--- a/packages/flame/lib/src/collisions/standard_collision_detection.dart
+++ b/packages/flame/lib/src/collisions/standard_collision_detection.dart
@@ -89,8 +89,8 @@ class StandardCollisionDetection<B extends Broadphase<ShapeHitbox>>
       final possiblyFirstResult = !(finalResult?.isActive ?? false);
       if (currentResult != null &&
           (possiblyFirstResult ||
-              currentResult.distance! < finalResult!.distance!)) {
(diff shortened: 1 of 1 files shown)


spydon (approved):
Lgtm!

PR #2624, opened 2023-07-24
docs: Add more guidance to collision detection algorithm choices

### LINKED ISSUE #2622: Address performance issues of QuadTreeBroadphase collision detection
# What could be improved

I noticed that when having a lot of moving components with collision detection, the `HasQuadTreeCollisionDetection` algorithm is significantly _slower_ than the standard `HasCollisionDetection` (using `Sweep`). For something like 800 entities, quad tree is about 3 times slower.

Screenshot 2023-07-23 at 19 09 31

Screenshot 2023-07-23 at 19 08 43

(Sidenote: the "benchmark" code isn't trying to be particularly _correct_ — you can see that many objects that are clearly colliding are white instead of red. This is because we're just blindly switching colors on collision start and on end, and don't care whether there are other collisions active. My point is to stress the collision detection code, not to build a simulation.)

I realize no optimization is a good fit for all cases, but this was a surprise to me, as I thought that QuadTree would deal with this exact case (lots of moving objects, some of which are far away from each other) better than the default.

I looked into it in profile mode, and it seems that a _lot_ of the slowness comes from adding things into the `HashSet` of `CollisionProspect`s and doing other operations on various Sets. 

Screenshot 2023-07-24 at 13 59 19

Screenshot 2023-07-24 at 13 59 31

Screenshot 2023-07-24 at 14 00 14

I tried different low-effort optimizations (such as caching `CollisionProspect`'s `hashCode` or trying different ways to implement `CollisionProspect.==`) but nothing yielded significant improvements. It's of course possible that I'm missing something obvious, but I'm thinking that maybe the whole approach of using Sets of prospects might need to be rethought. 

One idea I had was to try to implement a much simpler broadphase algorithm (based on a simple grid of static proportions). But I think the current API would still force me to use Sets of prospects, which could undermine any performance improvements over `Sweep`.

I'm seeking opinions and feedback. Maybe I'm just holding it wrong?

Code of my simple "benchmark" in [this repo](https://github.com/filiph/flame-collision-benchmark). DartPad here:
https://dartpad.dev/?id=a36320487f5a18fd6418f1db1e21a1da

# Why should this be improved

Movement of a lot of collidable objects on a large map is a prerequisite for quite a lot of 2D game genres (shoot 'em ups, RTS, physics games, etc.).

It should be easy to make this relatively fast, and potential pitfalls should be documented.

# Any risks?

- The options I see could lead to breaking changes of the CollisionDetection API (especially for the "internals" — when someone is providing their own `Broadphase` — other users might not even notice).
- This could be a lot of finicky work that ultimately only serves a small minority of Flame users.

# More information

I'd be interested in taking this on but I have trouble gauging how big of a problem this is for Flame as a whole. My game is pretty special in that it has a big world with lots of entities — but that might be an outlier. I'd hate to work on a scalable solution to a problem that only a few people actually have. In such cases, my policy is to solve only the particular problem at hand (my own game) and document it, instead of throwing hours of work at an "elegant" solution that isn't needed (YAGNI).

filiph:
# Description

This is adding a bit more precise language and more context to API docs of `HasCollisionDetection`, `Broadphase` and `HasQuadTreeCollisionDetection`.

## Checklist

## Breaking Change?

## Related Issues

Closes #2622 (imho)

### DIFF
--- a/doc/flame/collision_detection.md
+++ b/doc/flame/collision_detection.md
@@ -268,22 +268,22 @@ collisions to the whole hat, instead of for just each hitbox separately.
 
 ## Broad phase
 
-If your game field is small and do not have a lot of collidable components - you don't have to
+If your game field isn't huge and does not have a lot of collidable components - you don't have to
 worry about the broad phase system that is used, so if the standard implementation is performant
 enough for you, you probably don't have to read this section.
 
 A broad phase is the first step of collision detection where potential collisions are calculated.
-To calculate these potential collisions are a lot cheaper to calculate than to check the exact
-intersections directly and it removes the need to check all hitboxes against each other and
-therefore avoiding O(n²). The broad phase produces a set of potential collisions (a set of
-`CollisionProspect`s), this set is then used to check the exact intersections between hitboxes, this
-is sometimes called narrow phase.
+Calculating these potential collisions is faster than to checking the intersections exactly, and it removes the need to check all hitboxes against each other and
+therefore avoiding O(n²).
 
-By default Flame's collision detection is using a sweep and prune broadphase step, if your game
+The broad phase produces a set of potential collisions (a set of
+`CollisionProspect`s). This set is then used to check the exact intersections between hitboxes (sometimes called "narrow phase").
+
+By default, Flame's collision detection is using a sweep and prune broadphase step. If your game
 requires another type of broadphase you can write your own broadphase by extending `Broadphase` and
 manually setting the collision detection system that should be used.
 
-For example if you have implemented a broadphase built on a magic algorithm instead of the standard
+For example, if you have implemented a broadphase built on a magic algorithm instead of the standard
 sweep and prune, then you would do the following:
 
 ```dart
@@ -301,6 +301,7 @@ class MyGame extends FlameGame with HasCollisionDetection {
 If your game field is large and the game contains a lot of collidable
 components (more than a hundred), standard sweep and prune can
 become inefficient. If it does, you can try to use the quad tree broad phase.
+
 To do this, add the `HasQuadTreeCollisionDetection` mixin to your game instead of
 `HasCollisionDetection` and call the `initializeCollisionDetection` function on game load:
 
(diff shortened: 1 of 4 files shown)


packages/flame/lib/src/collisions/broadphase/quadtree/has_quadtree_collision_detection.dart
```diff
@@ -5,6 +5,12 @@ import 'package:flame/game.dart';
 /// This should be applied to a [FlameGame] to bring QuadTree collision
 /// support.
 ///
+/// Use [HasQuadTreeCollisionDetection] if you have lots of collidable entities
+/// in your game, but most of them are static (such as platforms, walls, trees,
+/// buildings). Always experiment before deciding which collision detection
```
spydon, line 10:
```suggestion
/// buildings).
///
/// Always experiment before deciding which collision detection
```

spydon (approved):
Would be great with some comment about this in the normal docs too if you're up for it. :)
https://github.com/flame-engine/flame/blob/main/doc/flame/collision_detection.md

filiph:
Good point. I made some improvements to the `.md` file.

doc/flame/collision_detection.md
```diff
@@ -268,22 +268,22 @@ collisions to the whole hat, instead of for just each hitbox separately.
...
 worry about the broad phase system that is used, so if the standard implementation is performant
 enough for you, you probably don't have to read this section.
 
 A broad phase is the first step of collision detection where potential collisions are calculated.
-To calculate these potential collisions are a lot cheaper to calculate than to check the exact
-intersections directly and it removes the need to check all hitboxes against each other and
-therefore avoiding O(n²). The broad phase produces a set of potential collisions (a set of
-`CollisionProspect`s), this set is then used to check the exact intersections between hitboxes, this
-is sometimes called narrow phase.
+Calculating these potential collisions is faster than to checking the intersections exactly, and it removes the need to check all hitboxes against each other and
```
spydon, line 276: There is a 100 char line length limit on the docs, so some of these lines will fail the pipeline

filiph: I tried to address this throughout the doc with my latest commit.

spydon (approved):

PR #2625, opened 2023-07-24
perf!: Pool `CollisionProspect`s and remove some list creations from the collision detection

spydon:
# Description
This change introduces a very simple pool for `CollisionProspect`s so that those objects don't have to be re-created each tick.
It means that the `CollisionProspect` needs to be mutable though, so the code becomes a little bit harder to read since sets can't be used anymore.

## Checklist

## Breaking Change?

### Migration instructions
If you are depending on that the query method returns a set instead of an iterable you have to accommodate for that change, or do `toSet`, but the latter might not be reliable depending on how you use the set afterwards, since the objects are now mutable.

## Related Issues

### DIFF
--- a/packages/flame/lib/src/collisions/broadphase/quadtree/quad_tree_broadphase.dart
+++ b/packages/flame/lib/src/collisions/broadphase/quadtree/quad_tree_broadphase.dart
@@ -17,44 +17,43 @@ typedef ExternalMinDistanceCheck = bool Function(
 ///
 /// See [HasQuadTreeCollisionDetection.initializeCollisionDetection] for a
 /// detailed description of its initialization parameters.
-class QuadTreeBroadphase<T extends Hitbox<T>> extends Broadphase<T> {
+class QuadTreeBroadphase extends Broadphase<ShapeHitbox> {
   QuadTreeBroadphase({
     required Rect mainBoxSize,
     required this.broadphaseCheck,
     required this.minimumDistanceCheck,
     int maxObjects = 25,
     int maxDepth = 10,
-  }) : tree = QuadTree<T>(
+  }) : tree = QuadTree<ShapeHitbox>(
           mainBoxSize: mainBoxSize,
           maxObjects: maxObjects,
           maxDepth: maxDepth,
         );
 
-  final QuadTree<T> tree;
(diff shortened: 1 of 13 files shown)


filiph:
Have you tried to profile this, and what were the results? Just curious. My (very limited!) tests didn't point to object creation being the obvious main culprit. Then again, saving thousands of newly created objects every frame will probably be huge, and it would be cool to see the impact.

spydon:
> Have you tried to profile this, and what were the results? Just curious. My (very limited!) tests didn't point to object creation being the obvious main culprit. Then again, saving thousands of newly created objects every frame will probably be huge, and it would be cool to see the impact.

Yeah, I actually see no difference at all!
In the beginning it was even worse with the pool, but then I made it a bit easier for the branch prediction and it went down on par, but still, since there is no difference with the pool I'll remove that since the code was simpler before. It must be really efficient to create const objects in Dart, compared to like `Vector2` and `Paint` where we see performance impact directly when they are created in tight loops.

spydon:
It gives a few percentage increase for android and web, Linux seems unaffected, and I didn't try any other platforms.
I honestly thought that it would have a bigger impact, the GC must have been greatly improved for const classes since when I was last playing around with this.

packages/flame/lib/src/collisions/broadphase/quadtree/quad_tree_broadphase.dart
```diff
@@ -2,6 +2,7 @@ import 'dart:collection';
 
 import 'package:flame/collisions.dart';
 import 'package:flame/extensions.dart';
+import 'package:flame/src/collisions/broadphase/prospect_pool.dart';
```
erickzanardo, line 5: Maybe this file could be exporters from the collision barrel?

erickzanardo (approved):

filiph:
I just want to say: even if the improvement is small:

A) every nanosecond counts
B) the exercise helps us gain experience. We now _know_ what would happen if CollisionProspect was optimized, because now it is. I, for one, can move on to look at other parts of the game because I'm confident that there's not much else that could be done here.

I'm other words, thanks @spydon!

PR #2634, opened 2023-08-01
fix: Compute scale in TileLayers based on native map tile size rather than image sizes to support oversized/undersized tiles.

### LINKED ISSUE #2633: flame_tiled: Isometric tiles not scaled properly when image size does not match the map tile size.
# Current bug behavior
Tiles larger than the map's grid size are resized so they fit into the map's grid:
image

# Expected behavior
Tiles should overflow the map's grid when they are naturally larger than the cells:
image

# Steps to reproduce
- Create a tileset in Tiled (eg with 96x48 tiles)
- Create an Isometric map with a smaller grid size (eg 48x24)
- Place some tiles and load the tmx file into a RenderableTileMap
- Notice the tiles are spaced twice as far apart as they appeared in the Tiled editor

* Flame version: 1.8.1
* flame_tiled version: 1.12.0

# More information
It appears that `IsometricTileLayer` is computing the `scale` value incorrectly. Scale is currently calculated by dividing the destination tile size by the natural size of the tile image. Instead, the scale should be computed by diving the destination tile size by the map's native tile size. If the images are the same size as the map's tiles then these result in the same scale factor, but if larger images are placed on a smaller grid the resulting scale factor is incorrect (compared to the preview in the Tiled editor).

I can PR this fix in `isometric_tile_layer.dart`:
```
-        final scale = size.x / src.width;
+        final scale = size.x / map.tileWidth;
```

# Fixed behavior
image
Notice no gaps and the tiles are twice as large at 100% scale, matching what appears in the Tiled editor.

chippydip:
# Description
Compute isometric tile scale based on the map's tile size rather than the tile image's default size. This fixes an issue where scale was incorrect when those two sizes differed.

## Checklist

## Breaking Change?

## Related Issues
Closes #2633 


packages/flame_tiled/lib/src/renderable_layers/tile_layers/isometric_tile_layer.dart
```diff
@@ -60,7 +60,7 @@ class IsometricTileLayer extends FlameTileLayer {
               .translate(spriteOffset.dx, spriteOffset.dy),
         );
         final flips = SimpleFlips.fromFlips(tileGid.flips);
-        final scale = size.x / src.width;
+        final scale = size.x / map.tileWidth;
```
jtmcdole, line 63: Can you update the tests to include a larger tile? Does this still pack well for faster rendering?

spydon (approved):
Good find, thanks for the PR!

packages/flame_tiled/lib/src/renderable_layers/tile_layers/isometric_tile_layer.dart
chippydip: I'll be honest, I didn't dig into the inner working of flame_tiled much beyond this little bit of the code to figure out why my sample tiles weren't rendering as expected. I threw up a PR with the fix since I'd found it, but if there's other things this might impact I don't know enough to comment other than to say that this should be no different for cases where the native image size matches the map tile size (which I assume is all current uses).

I took another quick look just now and realized that all the other types of tile_layers seems to actually have this same bug, so maybe those should be updated as well? I don't have test cases for them, though, and while I took a quick look at the existing tests I couldn't figure out enough to know where/how was the right way to add a test for this.

jtmcdole: Here's my take: if none of the other tests exploded, you're good!  I'm just asking for a simple test that shows your expected results to prevent regressions. That could just be using a small map with your one oversized tile and making a golden image.

chippydip: Tests added and included fixes for the other layer types as well. Hopefully I did that right!

jtmcdole:
LGTM

Issue #2613, opened 2023-07-19
flame_forge2d: Use a meter-to-pixels constant for rendering instead of zoom


spydon:
# What could be improved
We should be using a meter-to-pixels constant for rendering instead of using the zoom of the camera.
The zoom is needed since there is a speed limit in Forge2D which we currently easily hit when we calculate one Forge2D unit as one pixel, instead of something larger.

# Why should this be improved
To make the coding more natural, you shouldn't have to zoom in just to not hit the speed limit.

# Any risks?
Yes, this will be a breaking change.

# More information
https://www.iforce2d.net/b2dtut/gotchas#speedlimit

PR #2642, opened 2023-08-08
feat!: Pause game when backgrounded

adil192:
# Description

When the game is backgrounded and then resumed, the `update` function is called with a very large `dt` time step.

This might cause the physics engine to hit the maximum translation per timestep, causing moving bodies to slow down to almost-zero velocities.

This PR makes the game pause if the game is backgrounded. If you'd like to change this behaviour, set the `pauseWhenBackgrounded` flag to `true` (default) or `false`.
```dart
  MyGame() {
    pauseWhenBackgrounded = false;
  }
}
```

This is currently not working on the web, see https://github.com/flutter/flutter/issues/53107.

You can view previous discussion about this here: https://github.com/flame-engine/forge2d/pull/84.

## Checklist

## Breaking Change?

- [-] No, this PR is not a breaking change.

### Migration instructions
If you override the `lifecycleStateChange` method in your `Game` class, you must now call `super.lifecycleStateChange`.

If you'd like the game not to pause when backgrounded, set the `pauseWhenBackgrounded` flag in your `Game` class to `false`.

## Related Issues

Closes https://github.com/flame-engine/forge2d/pull/84


spydon (commented):
This should not be done like this directly in the engine (it's also a breaking change), the user should instead handle pausing the engine on events when the app is backgrounded. Not sure if there are any such events to react to when switching browser tabs though.

adil192:
Should this not be the default out-of-the-box behaviour though?
If the physics engine can't handle time steps of multiple seconds, then the sensible default is to avoid such time steps

spydon:
> Should this not be the default out-of-the-box behaviour though?
> If the physics engine can't handle time steps of multiple seconds, then the sensible default is to avoid such time steps

That depends on the rest of your settings though, and with a zoom it should very rarely need this except when backgrounded or having massive lags. We should follow how box2d handles it.

The better solution that solves most of these problems is: https://github.com/flame-engine/flame/issues/2613

adil192:
> with a zoom it should very rarely need this except when backgrounded or having massive lags

But an app being backgrounded is not rare at all. The flutter package should handle a common event in a flutter app.

> The better solution that solves most of these problems is: https://github.com/flame-engine/flame/issues/2613

I would argue that this doesn't solve this problem with backgrounding. Our previous discussion in https://github.com/flame-engine/forge2d/pull/84 was about hitting the speed limit with fast bodies, but this PR is only concerned with backgrounding. 

Even if we raise the speed limit with some unit conversion like zoom or a meter-to-pixels constant, the app being backgrounded for a few seconds will still hit any speed limit and break expected behaviour.

adil192:
Perhaps we could use a getter for `maxDt` instead of a constant, so when devs extend the `Forge2DGame` they can opt-in to pause the simulation when backgrounded like this:
```dart
class MyGame extends Forge2DGame {
  @override
  double get maxDt => 0.5;
}
```

spydon:
Something like this should be used: https://api.flutter.dev/flutter/widgets/WidgetsBindingObserver/didChangeAppLifecycleState.html

I'm not sure if that would work for web though.

adil192:
I'm unsure if the AppLifecycleState would work in this scenario, since I think the `update` method is called after the app is foregrounded again. I need to test it to make sure.
If I'm correct about that, I'll need to modify the code to only restore `AppLifecycleState.resumed` after the first discarded update

spydon:
> I'm unsure if the AppLifecycleState would work in this scenario, since I think the `update` method is called after the app is foregrounded again. I need to test it to make sure. If I'm correct about that, I'll need to modify the code to only restore `AppLifecycleState.resumed` after the first discarded update

I'm not sure if I understand you correctly here, but you are not the one setting the `AppLifecycleState`, you get that from overriding `didChangeAppLifecycleState`.

packages/flame/lib/src/game/game.dart
```diff
@@ -35,6 +35,16 @@ abstract mixin class Game {
...
 
+  /// Whether the app is currently in the foreground.
+  ///
+  /// This is used to discard the first [update] call after the app was in the
+  /// background to avoid a massive `dt` value.
+  ///
+  /// If you make use of this value,
+  /// remember to set this back to `true` when the app is resumed.
+  /// If you are using `Forge2DGame`, this is handled automatically.
+  bool isAppInForeground = true;
```
adil192, line 46: So far we've only discussed making this change for the flame_forge2d plugin so I've only implemented it in the `Forge2DGame` class.
Would you like me to implement it in any `FlameGame` class?

spydon: Yeah, sounds like a good idea, it could be done in `FlameGame`, but the default should be that it doesn't pause the game loop. :) Did it work in the browser?

adil192:
> Did it work in the browser?

Nope :(
Running this code
```
  @override
  void didChangeAppLifecycleState(AppLifecycleState state) {
    print('AppLifecycleState: $state');
    super.didChangeAppLifecycleState(state);
  }
```
tells me that `didChangeAppLifecycleState` isn't called at all when switching tabs.

It runs as expected on Android though, so I'm guessing this is a Flutter bug since it's easy to [detect tab focus with js](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState).

spydon:
> Edit: Just realised that this means isAppInForeground is true when it shouldn't be. I'm thinking I should make an enum like

I didn't follow, in which case is the app in the background but with the boolean set to true?

adil192:
> I didn't follow, in which case is the app in the background but with the boolean set to true?

Yes. What happened was:

1. I backgrounded the app. This sets `isAppInForeground` to false.
2. An update was ran while the app was still backgrounded, which incorrectly set `isAppInForeground` back to true.
3. I reopened the app.
4. An update ran with a dt of `2.482243` seconds because the app thinks it has been in the foreground.

packages/flame_forge2d/lib/forge2d_game.dart
```diff
@@ -23,6 +23,20 @@ class Forge2DGame extends FlameGame {
 
   @override
   void update(double dt) {
+    switch (gameLifecycleState) {
```
spydon, line 26: I think this could move to a separate private method in `FlameGame` and then have a flag `pauseWhenBackgrounded` or something (open to other name suggestions, not a big fan of that name). And that flag should default to false, so that this won't be a breaking change.
In this private method should use `pauseEngine` and `resumeEngine` instead of using an early `return`.
Maybe `foregroundedFirstUpdate` won't be a problem if those two methods are used instead either?

packages/flame/lib/src/game/game.dart
```diff
@@ -363,3 +390,19 @@ abstract mixin class Game {
...
+
+enum GameLifecycleState {
+  /// The game is backgrounded.
+  ///
+  /// This corresponds to [AppLifecycleState.paused] and [AppLifecycleState.detached].
+  backgrounded,
+
+  /// The game has been foregrounded but the first update should be
+  /// disregarded.
+  foregroundedFirstUpdate,
```
spydon, line 402: Maybe just `foregrounding`? Should indicate clear enough that it is on its way to be in the foreground.

spydon:
> which means I need to add `foregroundedFirstUpdate` (`foregrounding`) back in

Probably better to modify how pause/resume works instead, that behaviour doesn't look like it would be intentional

spydon:
> Probably better to modify how pause/resume works instead, that behaviour doesn't look like it would be intentional

Did you have any look at this @adil192?

spydon:
Added a test in #2731 that makes sure that `dt` isn't accumulating meanwhile a game is paused with the `pauseEngine` method.

adil192:
No sorry I've been quite busy recently

adil192:
With some extra logging, it seems everything is in order.
In the previous log, I must have forgotten to set `pauseWhenBackgrounded` to `true` in the constructor.

```
I/flutter (22999): update dt: 0.008343
I/flutter (22999): update dt: 0.008343
I/flutter (22999): update dt: 0.008344
I/flutter (22999): update dt: 0.00834
I/flutter (22999): lifecycleStateChange: AppLifecycleState.hidden
I/flutter (22999): Pausing engine
I/flutter (22999): lifecycleStateChange: AppLifecycleState.paused
I/flutter (22999): Pausing engine
I/FA-Ads  (22999): Application backgrounded at: timestamp_millis: 1694538751731
I/flutter (22999): lifecycleStateChange: AppLifecycleState.hidden
I/flutter (22999): Pausing engine
I/flutter (22999): lifecycleStateChange: AppLifecycleState.inactive
I/flutter (22999): Resuming engine
I/flutter (22999): update dt: 0.0
I/flutter (22999): update dt: 0.008329
I/flutter (22999): update dt: 0.008342
I/flutter (22999): update dt: 0.008331
I/flutter (22999): update dt: 0.008316
I/flutter (22999): update dt: 0.008332
I/flutter (22999): lifecycleStateChange: AppLifecycleState.resumed
I/flutter (22999): update dt: 0.058368
I/flutter (22999): update dt: 0.008339
```

packages/flame_forge2d/lib/forge2d_game.dart
```diff
@@ -38,4 +48,35 @@ class Forge2DGame extends FlameGame {
   Vector2 screenToFlameWorld(Vector2 position) {
     return screenToWorld(position)..y *= -1;
   }
+
+  @override
+  @mustCallSuper
+  void lifecycleStateChange(AppLifecycleState state) {
```
adil192, line 54: Does this count as a breaking change (adding the `@mustCallSuper` decorator)?

spydon: Good with the migration instructions, but since it is so minor breaking you don't have to add an exclamation mark, so that the more breaking changes will be more visible. :)

packages/flame_forge2d/lib/forge2d_game.dart
```diff
@@ -21,6 +23,14 @@ class Forge2DGame extends FlameGame {
 
   final World world;
 
+  /// Whether the game should pause when the app is backgrounded.
+  ///
+  /// If true, the first update after the app is foregrounded will be skipped.
```
spydon, line 28: Not true anymore right?

spydon (commented):
Are you moving this to `FlameGame` instead of `Forge2DGame`?

packages/flame_forge2d/lib/forge2d_game.dart
```diff
@@ -8,6 +8,7 @@ class Forge2DGame extends FlameGame {
     double zoom = defaultZoom,
     Camera? camera,
     ContactListener? contactListener,
+    super.pauseWhenBackgrounded,
```
adil192, line 11: Do we need to add this to other subclasses?

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -24,6 +24,7 @@ class FlameGame extends ComponentTreeRoot
   FlameGame({
     super.children,
     Camera? camera,
+    this.pauseWhenBackgrounded = false,
```
spydon, line 27: I don't think this needs to be in the constructor

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -38,6 +39,12 @@ class FlameGame extends ComponentTreeRoot
   @internal
   late final List<ComponentsNotifier> notifiers = [];
 
+  /// Whether the game should pause when the app is backgrounded.
+  ///
+  /// Defaults to false.
+  bool pauseWhenBackgrounded;
+  bool _pausedBecauseBackgrounded = false;
```
spydon, line 46: Move these down closer to where they are used

packages/flame_forge2d/lib/forge2d_game.dart
spydon: I'd say don't have in the constructor at all and just default the boolean directly in the variable

spydon (approved):
Looks good, just a few nits.
Maybe this would be nicer as a mixin, what do you think?

adil192:
> Looks good, just a few nits. Maybe this would be nicer as a mixin, what do you think?

Yeah sure, what should it be called? `PauseWhenBackgrounded`?

```dart
class RicochlimeGame extends Forge2DGame with PauseWhenBackgrounded {
````

**Comment by spydon** [comment]
> Yeah sure, what should it be called? `PauseWhenBackgrounded`?
> 
> ```dart
> class RicochlimeGame extends Forge2DGame with PauseWhenBackgrounded {
> ```

I changed my mind again. :sweat_smile: 
It is probably better to keep it like it is now, it is quite a crucial feature that maybe we actually should even make quite a breaking change and default to true directly, I'll have a a chat with the rest of the Flame team.

**Comment by spydon** [comment]
@adil192 did you try this on iOS?

**Comment by spydon** [comment]
> > @adil192 did you try this on iOS?
> 
> I haven't, no. I don't have access to a macOS/iOS device currently

Alright, we'll have a test on those then. Which platforms have you tried?

**Comment by adil192** [comment]
> Alright, we'll have a test on those then. Which platforms have you tried?

Just Android at the moment. I can test other platforms if you want me to

**Comment by spydon** [comment]
> Just Android at the moment. I can test other platforms if you want me to

Please do, I guess you at least have access to one desktop platform and web too?
Would be interesting to see if this also would work if you switch tabs for example, or minimize the application on desktop.

**Comment by spydon** [comment]
Alright we had a chat about this, let's make it a breaking change and set it to `true` by default. :)

**Comment by adil192** [comment]
This doesn't work on Windows when you minimise an application, it still updates as if it was foregounded. I assume that's another implementation that Flutter needs.

The web implementation is also missing, see https://github.com/flutter/flutter/issues/53107.

**Comment by spydon** [comment]
> I'm guessing that AppLifecycleState is only supported on Android and iOS. Do you still want to make `pauseWhenBackgrounded` `true` by default?

Yeah, we should mention it in the docs though.

**Comment by spydon** [comment]
Some small issues that needs to be fixed:
1. Add Backgrounding to the dictionary (probably update the branch first, it might already be added)
2. Run `dart format .`
3. Make sure no lines in the markdown are longer than 100 chars.

**Comment by adil192** [comment]
> probably update the branch first, it might already be added 

I will do this shortly!

**Comment by spydon** [comment]
It seems to somehow break the lifecycle according to one of the lifecycle tests :thinking:

**Comment by adil192** [comment]
> It seems to somehow break the lifecycle according to one of the lifecycle tests 🤔

With some added events:
```
Expected: [
            'onGameResize',
            'onLoad',
            'onMount',
            'update',
            'render',
            'update',
            'onRemove',
            'onDispose',
            'onGameResize',
            'onMount',
            'update', // <-- this is missing
            'render'
          ]
  Actual: [
            'onGameResize',
            'onLoad',
            'onMount',
            'update',
            'lifecycleStateChange(AppLifecycleState.resumed)',
            'render',
            'update',
            'lifecycleStateChange(AppLifecycleState.paused)',
            'onRemove',
            'onDispose',
            'onGameResize',
            'onMount',
            'lifecycleStateChange(AppLifecycleState.resumed)',
            'render'
          ]
```

It looks like the first update is discarded because of the use of `pauseEngine` and `resumeEngine`. Are you happy for me to just remove the expected `'update',` line?

**Comment by spydon** [comment]
> It looks like the first update is discarded because of the use of pauseEngine and resumeEngine. Are you happy for me to just remove the expected 'update', line?

Unfortunately not, we need to be calling `update` before `render` when coming back from being unmounted.

File: packages/flame/lib/src/game/game_widget/game_widget.dart [hunk 9]
```diff
@@ -254,6 +254,9 @@ class GameWidgetState<T extends Game> extends State<GameWidget<T>> {
       currentGame = widget.game!;
     }
     currentGame.addGameStateListener(_onGameStateChange);
+    currentGame.lifecycleStateChange(
```
spydon, line 257: Hmm, it's pretty weird to call these manually since these state changes should be coming from the framework.

adil192: I think it makes more sense for the game to act identically to if the app is backgrounded when really it was just unmounted.
This makes the link between "backgrounding the app" and "backgrounding the game (widget)"

adil192: Plus as an added bonus, it makes the behaviour more consistent on non-mobile platforms since the game will be paused when the widget is disposed of

spydon: I see what you mean, it feels like the state machine could become inconsistent with the one on Flutter's side though, which might confuse developers and be bug prone.

packages/flame/test/game/flame_game_test.dart
```diff
@@ -698,6 +698,28 @@ void main() {
         });
       });
     });
+
+    group('pauseWhenBackgrounded:', () {
```
spydon, line 702: Can you add some tests where it is using a widget and not calling `lifecycleStateChange` manually? There should be some other widget tests that you can take inspiration from.

spydon: I'll add some tests, so that we can get this PR into the release.

spydon (approved):

PR #2649, opened 2023-08-13
feat(fire_atlas): Encoded option to load json instead of .fa

Arenukvern:
# Description

Hello.
I would like to suggest to add option to load atlas without compression as simple json file.
I found it useful to use for example in itch.io, where gzip encoder causes an error during decoding.
Please notice is just a draft, if it is a case that this change can be useful - I will add all relevant documentation and tests if needed.

Thank you for your time!

## Checklist

## Breaking Change?

## Related Issues


erickzanardo (commented):
I think this is a great idea! The changes in place right now also look good, I think we should just address this github warning about path not being a dependency of the project, and we should maybe also add a method that saves the atlas as json?

Arenukvern:
@erickzanardo understood, thank you for your review!
Will make the changes (path & save the atlas as json) this week

Arenukvern:
@erickzanardo 
made changes, could you please take a look?

packages/flame_fire_atlas/lib/flame_fire_atlas.dart
```diff
@@ -270,24 +271,43 @@ class FireAtlas {
...
+    bool encoded = true,
   }) async {
     final assetsCache = assets ?? Flame.assets;
-
-    final bytes = await assetsCache.readBinaryFile(fileName);
-    final atlas = FireAtlas.deserialize(bytes);
+    final FireAtlas atlas;
+    final fileExtension = path.extension(fileName);
+    if (encoded) {
+      assert(fileExtension == '.fa');
```
luanpotter, line 286: I feel it is a bit conflicting to require both the specification of the `encoded` parameter *and* at the same time enforcing the file extension.
if we are enforcing the file extension anyway, can't we just use that to determine the format instead of having to specify the parameter?
or, alternatively: if we want the option for the user to fine-control behaviour with the parameter, then shouldn't we allow them to use whatever file name they prefer?

luanpotter (approved):
LGTM with one consideration

packages/flame_fire_atlas/lib/flame_fire_atlas.dart
erickzanardo: I feel like Luan in here, Maybe we don't really need to have that enforcement? Or did you do that for some reason that we are missing @Arenukvern ?

spydon:
@Arenukvern do you want to finish this up with that last comment? If not we'll take it over and fix it. :)

packages/flame_fire_atlas/lib/flame_fire_atlas.dart
spydon: I removed the assertions, is there anything more that should be done?

erickzanardo (approved):

Arenukvern:
> @Arenukvern do you want to finish this up with that last comment? If not we'll take it over and fix it. :)

sorry for late answer, got sick and completely lost free time

spydon:
> sorry for late answer, got sick and completely lost free time

No worries, hope you're feeling better now!
The PR is merged and will be included in the next release. :)

PR #2652, opened 2023-08-17
fix: Change DialogueView to a mixin class

projectitis:
# Description

The DialogueView is responsible for handling and (almost always) displaying the dialog to the user. A very common use case requires the `DialogueView` to be a `PositionComponent` or other visual component in the game. This would require extending from both `PositionComponent` and `DialogueView`.

**Problem:** The current `DialogueView` implementation is an abstract class. Dart does not support multiple inheritance. This means that the case above is difficult to implement.

**Solution**: Changing the `DialogueView` from an abstract class to a mixin class allows the developer to extend `PositionComponent` and mix in `DialogueView`. The class can now be used as a base class OR as a mixin. All existing tests still pass. An additional test has been added that validates the use as a mixin (extending one class and mixing in `DialogueView`).

## Checklist

_Notes on docs:_ Updated class docs. Checked doc pages. These still read correctly.

_Notes on examples:_ Currently no jenny examples.

## Breaking Change?

## Related Issues

NA


packages/flame_jenny/jenny/lib/src/dialogue_view.dart
```diff
@@ -27,7 +32,7 @@ import 'package:meta/meta.dart';
 /// be implemented either synchronously or asynchronously. In the latter case
 /// the dialogue runner will wait for the future to resolve before proceeding
 /// (futures from several dialogue views will be awaited simultaneously).
-abstract class DialogueView {
+mixin class DialogueView {
```
spydon, line 35: I guess this should still be abstract since it isn't much use instantiating it?
```suggestion
abstract mixin class DialogueView {
```

projectitis:
Ah, didn't know that was possible, thanks @spydon. Tests still pass, too.

spydon (approved):
Lgtm, thanks for your contribution! :)

Issue #2647, opened 2023-08-12
The onRemove of game is be called when game is running.


MiaoShichang:
# Current bug behavior
The onRemove of game is be called when game is running.

# Expected behavior
Don't be called;

# Steps to reproduce
this is test code
```dart
import 'dart:async';

import 'package:flame/components.dart';
import 'package:flame/events.dart';
import 'package:flame/game.dart';
import 'package:flutter/material.dart';

///  flame:^1.8.2

void main() {
  runApp(const GameWidget.controlled(gameFactory: TestWidthGame.new));
}

// with TapCallbacks
class TestFirstZone extends PositionComponent with TapCallbacks {
  TestFirstZone({super.position, super.size}) : super(anchor: Anchor.center);
  @override
  void render(Canvas canvas) {
    super.render(canvas);
    canvas.drawRect(Rect.fromLTWH(0, 0, width, height), Paint()..color = const Color(0xff9999cc));
  }
}

// with null
class TestSecondZone extends PositionComponent {
  TestSecondZone({super.position, super.size}) : super(anchor: Anchor.center);
  @override
  void render(Canvas canvas) {
    super.render(canvas);
    canvas.drawRect(Rect.fromLTWH(0, 0, width, height), Paint()..color = const Color(0xffff66cc));
  }
}

class TestWidthGame extends FlameGame {
  @override
  Color backgroundColor() {
    return const Color(0xff9966cc);
  }

  @override
  FutureOr<void> onLoad() {
    // first
    add(TestFirstZone(position: size / 2, size: Vector2(160, 160)));

    // second
    // add(TestSecondZone(position: size / 2, size: Vector2(160, 160)));

    super.onLoad();
  }

  @override
  void onMount() {
    super.onMount();
    print("onMount");
  }

  @override
  void onRemove() {
    super.onRemove();
    // why this method is be called when only add TestFirstZone in onLoad method,
    // but not when only add TestSecondZone ?
    print("onRemove");
  }
}
```

# this is output
--
Launching lib/main.dart on POT AL00a in debug mode...
Running Gradle task 'assembleDebug'...
✓  Built build/app/outputs/flutter-apk/app-debug.apk.
Debug service listening on ws://127.0.0.1:59555/2h5VAvqcp8c=/ws
Syncing files to device POT AL00a...
I/flutter ( 9136): onMount
W/Gralloc3( 9136): mapper 3.x is not supported
I/HwViewRootImpl( 9136): removeInvalidNode jank list is null
[...]

PR #2650, opened 2023-08-13
docs: Updated docs for gesture detector lifecycle

munsterlander:
# Description

As discussed in https://github.com/flame-engine/flame/issues/2647, this PR just updates the docs to clarify that `onRemove` will get called when you first add a gesture based component.

## Checklist

- [X] I have followed the [Contributor Guide] when preparing my PR.
- [ ] I have updated/added tests for ALL new/updated/fixed functionality.
- [X] I have updated/added relevant documentation in `docs` and added dartdoc comments with `///`.
- [ ] I have updated/added relevant examples in `examples` or `docs`.

## Breaking Change?

- [ ] Yes, this PR is a breaking change.
- [X] No, this PR is not a breaking change.

## Related Issues

[Contributor Guide]: https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md
[Conventional Commit]: https://conventionalcommits.org
[CHANGELOG]: https://github.com/flame-engine/flame/blob/main/CHANGELOG.md

PR #2653, opened 2023-08-18
fix: Prevent `onRemove`/`onDetach` being called for initial Gesture Detector addition

munsterlander:
# Description

This fixes: https://github.com/flame-engine/flame/issues/2647
Reverts: https://github.com/flame-engine/flame/pull/2602
OBE: https://github.com/flame-engine/flame/pull/2650

To be clear, the memory leak issue is not addressed because it is not a Flame issue.  Flame properly calls the `onRemove` event and if it is needed to pass that on to children, users should do as documented: https://docs.flame-engine.org/latest/flame/game.html#lifecycle.  

So if you are using `GameWidget.controlled`, add a tappable component and then remove the game, the lifecycle looks like:

```
flutter: onLoad
flutter: mount
flutter: onMount
flutter: attach
flutter: add has_tappable_components
flutter: refreshWidget 
// Remove Game
flutter: detach
flutter: removeGameStateListener
flutter: onRemove
flutter: remove
```

If you do not use `GameWidget.controlled`, add a tappable component and then remove the game, the lifecycle looks like:

```
flutter: onLoad
flutter: mount
flutter: onMount
flutter: attach
flutter: add has_tappable_components
flutter: refreshWidget
// Remove Game
flutter: detach
flutter: removeGameStateListener
flutter: onRemove
flutter: remove
```

If you do not use `GameWidget.controlled`, do not add a tappable component and then remove the game, the lifecycle looks like:

```
flutter: onLoadFuture
flutter: mount
flutter: onMount
flutter: attach
// Remove Game
flutter: detach
flutter: removeGameStateListener
flutter: onRemove
flutter: remove
```

Previously, the lifecycle would look like:

```
flutter: onLoad
flutter: mount
flutter: onMount
flutter: attach
flutter: add has_tappable_components
flutter: refreshWidget 
flutter: detach // These unnecessary calls have been eliminated
flutter: removeGameStateListener // These unnecessary calls have been eliminated
flutter: onRemove // These unnecessary calls have been eliminated
flutter: attach
// Remove Game
flutter: detach
flutter: removeGameStateListener
flutter: onRemove
flutter: remove
```

I have updated the below diagram for those who may need it in the future:

Flame and Flutter Events Fixed

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/game/game.dart
+++ b/packages/flame/lib/src/game/game.dart
@@ -347,7 +347,7 @@ abstract mixin class Game {
 
   set mouseCursor(MouseCursor value) {
     _mouseCursor = value;
-    refreshWidget(internalRefresh: false);
+    refreshWidget();
   }
 
   @visibleForTesting
--- a/packages/flame/lib/src/game/game.dart
+++ b/packages/flame/lib/src/game/game.dart
@@ -370,8 +370,8 @@ abstract mixin class Game {
   /// When [isInternalRefresh] is passed as false it will trigger the `onAttach`
   /// and `onDetach` events; otherwise, those events will not be called.
   @internal
-  void refreshWidget({bool internalRefresh = true}) {
-    _isInternalRefresh = internalRefresh;
+  void refreshWidget({bool isInternalRefresh = true}) {
+    _isInternalRefresh = isInternalRefresh;
     gameStateListeners.forEach((callback) => callback());
   }
 }
--- a/packages/flame/lib/src/game/game.dart
+++ b/packages/flame/lib/src/game/game.dart
@@ -366,9 +366,9 @@ abstract mixin class Game {
 
   /// When a Game is attached to a `GameWidget`, this method will force that
   /// widget to be rebuilt. This can be used when updating any property which is
-  /// implemented within the Flutter tree.  [internalRefresh] as a named
-  /// parameter, when passed as false, will trigger the `onAttach` and
-  /// `onDetach` events; otherwise, those events will not be called.
+  /// implemented within the Flutter tree.  [internalRefresh] when passed as
+  /// false, will trigger the `onAttach` and `onDetach` events; otherwise, those
+  /// events will not be called.
   @internal
   void refreshWidget({bool internalRefresh = true}) {
     isInternalRefresh = internalRefresh;
--- a/packages/flame/lib/src/game/game.dart
+++ b/packages/flame/lib/src/game/game.dart
@@ -202,10 +202,10 @@ abstract mixin class Game {
   ///
   /// Should not be called manually.
   void detach() {
-    onRemove();
+    if (_gameRenderBox?.isGestureDetectorAdding == false) {
+      onDetach();
+    }
     _gameRenderBox = null;
-
-    onDetach();
   }
 
   /// Called when the game is about to be removed from the Flutter widget tree,
--- a/packages/flame/lib/src/game/game.dart
+++ b/packages/flame/lib/src/game/game.dart
@@ -213,6 +213,9 @@ abstract mixin class Game {
   /// do cleanups to avoid memory leaks.
   void onRemove() {}
 
+  /// Called when the GameWidget is disposed by Flutter.
+  void onDispose() {}
+
   /// Called after the game has left the widget tree.
   /// This can be overridden to add logic that requires the game
   /// not be on the flutter widget tree anymore.
--- a/packages/flame/lib/src/game/game.dart
+++ b/packages/flame/lib/src/game/game.dart
@@ -365,9 +365,10 @@ abstract mixin class Game {
 
   /// When a Game is attached to a `GameWidget`, this method will force that
   /// widget to be rebuilt. This can be used when updating any property which is
-  /// implemented within the Flutter tree.  [internalRefresh] when passed as
-  /// false, will trigger the `onAttach` and `onDetach` events; otherwise, those
-  /// events will not be called.
+  /// implemented within the Flutter tree.
+  ///
+  /// When [isInternalRefresh] is passed as false it will trigger the `onAttach`
+  /// and `onDetach` events; otherwise, those events will not be called.
   @internal
   void refreshWidget({bool internalRefresh = true}) {
     _isInternalRefresh = internalRefresh;
--- a/packages/flame/lib/src/game/game.dart
+++ b/packages/flame/lib/src/game/game.dart
@@ -190,7 +190,10 @@ abstract mixin class Game {
       );
     }
     _gameRenderBox = gameRenderBox;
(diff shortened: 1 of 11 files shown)


spydon:
I think this can be done in a simpler manner by using the `refreshWidget` method that is passed in as a callback to `GestureDetectorBuilder` in `Game` and add a flag that says that it is an internal refresh and then when that flag is set just skip calling the methods that you are skipping now too:
```dart
  @internal
  bool isInternalRefresh = false;

  @internal
  void refreshWidget() {
    isInternalRefresh = true;
    gameStateListeners.forEach((callback) => callback());
    isInternalRefresh = false;
  }
```
What do you think?

>  flutter: onLoad
flutter: mount
flutter: onMount
flutter: attach
flutter: add has_tappable_components
flutter: refreshWidget 
flutter: detach // These unnecessary calls have been eliminated
flutter: removeGameStateListener // These unnecessary calls have been eliminated
flutter: onRemove // These unnecessary calls have been eliminated
flutter: attach

Shouldn't the last attach call here also be removed?

munsterlander:
@spydon This was my original plan of attach and I even went a step further by trying to identify the callback to check and see if it was a gesture detector, but the callback name is an internal id to Flutter (or maybe I don't know enough to get a simpleName or such).

`refreshWidget` is also called by the overlay manager and the mouseCursor, so not knowing the full lifecycle of those items, I didn't want to do a blanket override.  I also wanted to do refreshWidget with an optional named parameter, but didn't know how to pass true for an invocation such as: `late final GestureDetectorBuilder gestureDetectors =
      GestureDetectorBuilder(refreshWidget)..initializeGestures(this);`
      
I am open to this way so any pointers are greatly appreciated!  

As far as the last attach, I thought about that but with what I implemented, GRB was being set to null, so then it wouldn't be available to read the flag in onAttach.  Having said that, if we can set the flag in Game and base it off of the refreshWidget, this can be overcome.

packages/flame/lib/src/game/game.dart
```diff
@@ -341,7 +344,7 @@ abstract mixin class Game {
 
   set mouseCursor(MouseCursor value) {
     _mouseCursor = value;
-    refreshWidget();
+    refreshWidget(internalRefresh: false);
```
spydon, line 347: This one is the same as adding a gesture detector so this should be true

munsterlander: Implemented, although I looked at the code and it doesn't implement the injection of a widget the others do, but I can see how its going to attach the cursor to the nearest node in the tree, so I presume the event lifecycle becomes the same and thus we would not want attach / detach to occur.

packages/flame/lib/src/game/game.dart
```diff
@@ -190,7 +190,10 @@ abstract mixin class Game {
       );
     }
     _gameRenderBox = gameRenderBox;
-    onAttach();
+    if (!isInternalRefresh) {
+      onAttach();
+    }
+    isInternalRefresh = false;
```
spydon, line 196: How come this is set to false here? 🤔

packages/flame/lib/src/game/game.dart
```diff
@@ -355,11 +358,17 @@ abstract mixin class Game {
     gameStateListeners.remove(callback);
   }
 
+  @internal
+  bool isInternalRefresh = false;
```
spydon, line 362: This can be private instead of `@internal` now I think?

munsterlander: Changed.

packages/flame/lib/src/game/game.dart
munsterlander: The default state is false.  The order is detach then attach.  If you set it to false after callback in the `refreshWidget` or in `detach` where it originally was, then attach will still get called because it would be true by then.  No matter what, attach is always going to be called so its putting it back to its initial state.

spydon: Right, that makes sense, I hope there won't be any events in-between detach and attach that could be affected in the future.

spydon (approved):

packages/flame/lib/src/game/game.dart
```diff
@@ -355,11 +358,16 @@ abstract mixin class Game {
...
   }
 
+  bool _isInternalRefresh = false;
+
   /// When a Game is attached to a `GameWidget`, this method will force that
   /// widget to be rebuilt. This can be used when updating any property which is
-  /// implemented within the Flutter tree.
+  /// implemented within the Flutter tree.  [internalRefresh] when passed as
+  /// false, will trigger the `onAttach` and `onDetach` events; otherwise, those
+  /// events will not be called.
```
spydon, line 367:
```suggestion
  /// implemented within the Flutter tree.
  ///
  /// When [isInternalRefresh] is passed as false it will trigger the `onAttach`
  /// and `onDetach` events; otherwise, those events will not be called.
```

packages/flame/lib/src/game/game.dart
```diff
@@ -355,11 +358,16 @@ abstract mixin class Game {
...
   @internal
-  void refreshWidget() {
+  void refreshWidget({bool internalRefresh = true}) {
+    _isInternalRefresh = internalRefresh;
```
spydon, line 370:
```suggestion
  void refreshWidget({bool isInternalRefresh = true}) {
    _isInternalRefresh = isInternalRefresh;
```

munsterlander:
Analyzer failed after direct committing your changes from GitHub.  When pulled, its passing local.  Let me update branch and see what happens.

spydon:
> Analyzer failed after direct committing your changes from GitHub. When pulled, its passing local. Let me update branch and see what happens.

How strange, the warning wasn't even about any code that you have touched. :man_shrugging:

munsterlander:
> > Analyzer failed after direct committing your changes from GitHub. When pulled, its passing local. Let me update branch and see what happens.
> 
> How strange, the warning wasn't even about any code that you have touched. 🤷‍♂️

Yeah, I pulled the last updates down, but I can't figure out what is going on.  I can't see the logs or I don't know how to see the Analyzer logs in GitHub.  What am I missing on this?

spydon:
> OH! The little triangle. Ok, I see the issue. So with your suggested name change, it just needed to be updated in the other file.

Yeah that was one thing, but there also was a warning about a file in Jenny. Let's see if that warning pops up again now.

munsterlander:
```dart
class _RecordingDialogueViewAsMixin extends _SomeOtherBaseClass
    with DialogueView {
  _RecordingDialogueViewAsMixin([this.waitDuration = Duration.zero]);
  final List<String> events = [];
  final Duration waitDuration;
  ```
  
  This is the warning.

PR #2659, opened 2023-08-19
feat: Add `onDispose` to `game.dart` called from `game_widget.dart`

munsterlander:
# Description

This adds, imo, the missing piece in the event lifecycle based on all the other lifecycle work I have been doing.  

Currently, when dispose is called in `game_widget.dart`, it calls `disposeCurrentGame` which has the call for `onRemove`, at issue though is, `disposeCurrentGame` is also called by `didUpdateWidget` if the `oldWidget.game != widget.game`.  It may not be needed, but I feel like because `FlameGame` is inherently a stateless set of widgets (children of the LeafRenderObjectWidget), the hook for `onRemove` is muddled with what is NOT actually a dispose event.  So my thought is, introduce the `onDispose` method to game which gets called by making `disposeCurrentGame({isDispose = false})` then in the actual overridden `dispose` method, pass that as true, so we can then call `currentGame.onDispose()` after the call to `currentGame.onRemove()`.

Flame and Flutter Events onDipose

## Checklist

## Breaking Change?

## Related Issues


packages/flame/lib/src/game/game.dart
```diff
@@ -213,6 +213,10 @@ abstract mixin class Game {
   /// do cleanups to avoid memory leaks.
   void onRemove() {}
 
+  /// Called when the game has been removed from the Flutter widget tree when
+  /// it was disposed by Flutter.
```
spydon, line 217: What do you mean with "when it was disposed by Flutter" here?

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -254,8 +254,11 @@ class GameWidgetState<T extends Game> extends State<GameWidget<T>> {
     _loaderFuture = null;
   }
 
-  void disposeCurrentGame() {
+  void disposeCurrentGame({bool isDispose = false}) {
```
spydon, line 257: `isDispose` needs a clearer name for what it does. In which cases will this method be called with this variable set to false?

packages/flame/lib/src/game/game.dart
munsterlander: Meaning it only comes from the dispose event on Game Widget and not via didUpdateWidget event.  What do you think is a better way of clarifying that?

spydon: Simply "Called when the `GameWidget` is disposed by Flutter"  maybe? :thinking:

packages/flame/lib/src/game/game_widget/game_widget.dart
munsterlander: Changed to `callGameOnDispose`.  I think that is pretty straightforward now.

spydon: I still think it is a bit unclear, especially since this method is called `disposeCurrentGame`, the reader of the code would have to guess in what cases such a variable would be false.
But maybe it's fine if we just get some dartdocs for this method explaining how that flag is used.

packages/flame/lib/src/game/game.dart
munsterlander: /// Called when the game has been removed from the Flutter widget tree via
  /// the Game Widget's Flutter dispose event.

packages/flame/lib/src/game/game_widget/game_widget.dart
munsterlander: I updated the diagram and the dart doc in the comments above.

packages/flame/lib/src/game/game.dart
spydon: The game never really is in the widget tree, only the widget is, so I would go with the even shorter version I provided anyways.

packages/flame/lib/src/game/game_widget/game_widget.dart
spydon: I can't see the dartdoc being updated? :thinking: 
image

packages/flame/lib/src/game/game.dart
munsterlander: Done.

packages/flame/lib/src/game/game_widget/game_widget.dart
munsterlander: Look again, lol.  Forgot that part.

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -254,8 +254,14 @@ class GameWidgetState<T extends Game> extends State<GameWidget<T>> {
     _loaderFuture = null;
   }
 
-  void disposeCurrentGame() {
+  /// [disposeCurrentGame] is called by two flutter events - `didUpdateWidget`
+  /// and `dispose`.  When the parameter `callGameOnDispose` = true, the
```
spydon, line 258:
```suggestion
  /// and `dispose`.  When the parameter [callGameOnDispose] is true, the
```

spydon (approved):

PR #2667, opened 2023-08-20
refactor: Enable DCM linting

luanpotter:
This is a first step towards enabling DCM for Flame. Though I have tested with all rules, and am working on selecting, assessing, and fixing violations, as a first step, we can merge the infrastructural changes to flame_lint to and GitHub actions.
As a proof of concept, I am enabling two rules for which we have no violations.
I will followup with enabling more rules, adding discussions for controversial changes, and fixing non-controversial violations.


.github/workflows/cicd.yml
```diff
@@ -42,6 +42,24 @@ jobs:
...
+          channel: 'stable'
+      - uses: bluefireteam/melos-action@v2
+
+      - name: Install DCM
+        uses: CQLabs/setup-dcm@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run DCM
+        run: dcm analyze --ci-key="${{ secrets.DCM_CI_KEY }}" --email="${{ secrets.DCM_EMAIL }}" .
```
incendial, line 62: @luanpotter hey, have you tried completely removing both `--ci-key` and `--email`? It should work that way, here is an example https://github.com/espresso-cash/espresso-cash-public/pull/1045/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R29

incendial: If this does not work, can you please make sure you have dcm 1.8.4? And what's the output?

luanpotter: hey @incendial !
it actually is working as is! (but I can try removing them for simplicity)

I did have a couple questions, though, regarding the `analysis_options.yaml` file:

1. do I need to specify the rules for each project under `packages`? is there a way to define global rules for the entire repo?
2. is there a way to enable all rules, except rules x and y? or a way to "inherit" some base configuration? or do I need to specify each rule I want explicitly

spydon: > do I need to specify the rules for each project under packages? is there a way to define global rules for the entire repo?

Since we are pulling in the analysis_options file from flame_lint I guess it would be enough to just specify it in there?

incendial: The ones that I'm not intended to change are `recommended` and `metrics_recommended`.

luanpotter: > Since we are pulling in the analysis_options file from flame_lint I guess it would be enough to just specify it in there?

I consider this but I felt it would be better not to include this on the base flame_lint because not everyone who uses that on their projects has the license to use dcm. we could make it a separate package, or a separate file within the base flame_lint

> Ah, so you prefer that way. Then sure, there is no "right" way.

I don't, I prefer your way if it works too haha. let me change

spydon: > I consider this but I felt it would be better not to include this on the base flame_lint because not everyone who uses that on their projects has the license to use dcm. we could make it a separate package, or a separate file within the base flame_lint

We could have it as a separate file in flame_lint that can be included in other analysis_options files but is named like internal or something?

luanpotter: let me try that!

luanpotter: I think I got it. now I need to choose the rules and fix the violations.
@incendial is it possible to run it locally for us?

incendial: Yes, it should work as usual. Do you have any problem with that?

Also noticed that you get a "preset-not-found" error. Please make sure that each package that uses the `dart_code_metrics_presets` directly or indirectly has `dart_code_metrics_preset` in its dependencies.

incendial: The only way not to install the presets package is to list the rules manually. Which you can also achieve by just copying all of them from the presets (dart and flutter ones).

luanpotter: I will probably do that! I just used the preset for now. I will copy all rules and then assess one by one and remove each one that we can't/won't fix right now

I tried running locally with the `dcm` cli, but it says that it is not activated

.github/workflows/cicd.yml
```diff
@@ -42,6 +42,24 @@ jobs:
           node-version: 18
       - run: npm install -g markdownlint-cli
       - run: markdownlint . -p .markdownlintignore -c .markdownlint.yaml
+
+  dcm:
```
luanpotter, line 46: I can make it mandatory once this merges

packages/flame_lint/lib/analysis_options_with_dcm.yaml
```diff
@@ -0,0 +1,10 @@
+include: analysis_options.yaml
+
+# DCM rules defined here: https://dcm.dev/docs/rules/
+
+dart_code_metrics:
+  rules:
+  # dart rules
+    - avoid-banned-imports
+  # flutter rules
+    - prefer-define-hero-tag
```
luanpotter, line 10: just the first for each from the `all` preset

spydon (approved):

erickzanardo (approved):

PR #2678, opened 2023-08-25
fix: Absolute angle takes into account BodyComponent ancestors too

maurovanetti:
# Description
Currently, the implementation of the `absoluteAngle` getter in `PositionComponent` only takes into account `PositionComponent` ancestors. The implementation simply runs through the whole hierarchy summing up the relative angles.

This gives a wrong result with Forge2D because `BodyComponent`s have angles too: in the common situation where a `PositionComponent` (e.g. a `SpriteComponent`) is the child of a `BodyComponent`, the `BodyComponent`'s rotation is ignored when computing the `absoluteAngle`.

The fix simply introduces a new minimal interface `HasAngle` that both `PositionComponent` and `BodyComponent` implement (`PositionComponent` implements it automatically via `AngleProvider`), and reimplements `absoluteAngle` so that it considers all ancestors with an angle and not just the `PositionComponent` ones.

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame/lib/effects.dart
+++ b/packages/flame/lib/effects.dart
@@ -30,6 +30,7 @@ export 'src/effects/provider_interfaces.dart'
     show
         AnchorProvider,
         AngleProvider,
+        HasAngle,
         PositionProvider,
         ScaleProvider,
         SizeProvider,
--- a/packages/flame/lib/effects.dart
+++ b/packages/flame/lib/effects.dart
@@ -34,6 +34,7 @@ export 'src/effects/provider_interfaces.dart'
         PositionProvider,
         ScaleProvider,
         SizeProvider,
+        ReadOnlySizeProvider,
         OpacityProvider;
 export 'src/effects/remove_effect.dart';
 export 'src/effects/rotate_effect.dart';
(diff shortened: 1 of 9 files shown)


spydon:
Good idea! `HasAngle` should be named `ReadOnlyAngleProvider` though, to follow the new conventions set by `ReadOnlySizeProvider`.

maurovanetti:
> Good idea! `HasAngle` should be named `ReadOnlyAngleProvider` though, to follow the new conventions set by `ReadOnlySizeProvider`.

Oops, I missed it. But it's got a lower-case O for some reason.

Updated.

spydon:
> But it's got a lower-case O for some reason.

Since it hasn't been exported previously, do you mind renaming it to `ReadOnlySizeProvider` and same with `ReadOnlyAngleProvider`? :)

maurovanetti:
There we go.

packages/flame/lib/effects.dart
```diff
@@ -30,6 +30,7 @@ export 'src/effects/provider_interfaces.dart'
     show
         AnchorProvider,
         AngleProvider,
+        ReadOnlyAngleProvider,
```
spydon, line 33: Can you export the `ReadOnlySizeProvider` too? 😇

spydon (approved):
Lgtm!

packages/flame/lib/effects.dart
maurovanetti: Symmetry… so satisfactory.

PR #2663, opened 2023-08-20
refactor!: Simplify text rendering pipeline

luanpotter:
# Description

While studying the existing text rendering pipeline and infrastructure, in order to better document it and add some easier user-facing ways to render rich text, I found it to be on a slightly overcomplicated state that seems to me to be an artifact of a migration.

Let me first briefly describe how it works with some diagrams (I am writing docs for everything and will re-use this content but wanted to propose this PR first as it will change how things look like).

We have 3 hierarchies related to rendering:
* renderers: children of TextRenderer, these define how to render text (style). do not include text
* formatters: children of TextFormatter, this also define how to render text (style), but a bit more higher level than TextRenderer.
* elements: these are laid-out, styled and ready-render pieces of text (style + string), created by formatters

Note that the renderers and formatters kinda serve the same purpose, but the renderers are a more low-level API. Here is how it it looks like today:

```mermaid
classDiagram
    %% renderers
    class TextRenderer {
        Vector2 measureText(String text)
        void render(Canvas canvas, String text, ...)
    }
    class FormatterTextRenderer {
        TextFormatter formatter
        TextElement format(String text)
    }
    note for FormatterTextRenderer "This is just connecting the Formatter to comply with the TextRenderer interface.\nDoes not add anything new."
    class TextPaint
    class SpriteFontRenderer

    %% formatters
    class TextFormatter
    class SpriteFontTextFormatter
    class TextPainterTextFormatter
    class DebugTextFormatter

    %% elements
    class TextElement {
        LineMetrics metrics
        render(Canvas canvas)
    }
    class TextPainterTextElement

    TextRenderer --> FormatterTextRenderer
    FormatterTextRenderer --> TextPaint
    FormatterTextRenderer --> SpriteFontRenderer

    TextRenderer *-- TextFormatter
    TextPaint *-- TextPainterTextFormatter
    SpriteFontRenderer *-- SpriteFontTextFormatter

    note for TextFormatter "This includes the style (how to render), not the text."
    TextFormatter : TextElement format(String text)
    TextFormatter *-- TextElement
    TextFormatter --> SpriteFontTextFormatter
    TextFormatter --> TextPainterTextFormatter
    TextFormatter --> DebugTextFormatter

    TextFormatter *-- TextElement
    TextPainterTextFormatter *-- TextPainterTextElement
    SpriteFontTextFormatter *-- SpriteFontTextElement

    note for TextElement "This includes the text and the style, ready to render."
    TextElement --> TextPainterTextElement
[...]

PR #2679, opened 2023-08-25
refactor!: Make TextElement more usable on its own

luanpotter:
# Description

This is part of my ongoing effort to simplify the text rendering pipeline.
My ultimate goal is to:
* get rid of renders
* rename formatters to renderers
* make the interface complies to both
All details are specified here: https://github.com/flame-engine/flame/pull/2663

As a first step to break down that huge PR, this makes a small change to TextElements to make them more useful

This PR will:
### rename render -> draw

draw becomes the "internal", underlying impl, raw method, that just draws the element w/ any custom options

### add a new render method that takes in more options

this does not need to be extended by every impl.
this is for end users and accepts parameters like position and anchor to be more in line with the renderer interface

## Checklist

## Breaking Change?

### Migration instructions

This is technically a breaking change but should have no effect for users, unless you are creating your own custom `TextElement`s. In that case, to migrate:

* rename your `render` method to `draw`


spydon (approved):
Lgtm!

PR #2681, opened 2023-08-26
feat: Component visibility (HasVisibility mixin)

### LINKED ISSUE #2669: feat(component): Visibility support
# Problem to solve

Often a component needs to stay on the tree, but just be invisible for a while.

The [suggested method](https://stackoverflow.com/questions/71971021/show-hide-a-spritecomponent-using-flame-in-flutter) (?) is to remove them from the tree. Adding and removing them has several problems:
- It's an async operation, and a remove followed by an immediate add again can cause issues
- Lifecycle events are fired, when all you want to do is not render the component

# Proposal

~~Mixin~~ A change (and `visibilty` flag) that adds visibility to components while still on the tree.

# More information

With suggestions from maintainers on how this could best be implemented, I'd be willing to contribute it.

projectitis:
# Description

This PR introduces a new `HasVisibility` mixin on `Component`. It prevents the renderTree method from progressing if the `isVisible` property is `false`. It therefore prevents the component and all it's children from rendering.

The purpose of this mixin is to allow showing and hiding a component _without_ removing it from the tree.

An important note is that the component (and all it's children) will still be on the tree. They will continue to receive `update` events, and all other lifecycle events. If the user has implemented input such as tap events, or collision detection, or any other interactivity, this will continue to operate without affect (unless it relies on the render step - such as per-pixel collision detection).

This is expected behaviour. If it is not desired, the user needs to account for it (by checking `isVisible` in their code) _or_ someone needs to create another mixin for `HasEnabled` or `HasActive` which would prevent these other actions 😁 

I am very happy to make changes, take suggestions, etc!

## Checklist

## Breaking Change?

## Related Issues

Closes #2669 


projectitis:
Apologies for all the commits 🙄

spydon:
> Apologies for all the commits 🙄 

Don't worry, they will all be squashed on merge

doc/flame/components.md
```diff
@@ -363,10 +363,84 @@ Do note that this setting is only respected if the component is added directly t
...
+
+```{note}
+Important! Even when the component is not visible, it is still in the tree and
+will continue to receive calls to 'update' and all other lifecycle events. It
+will still respond to input events, and will still interact with other
+components, such as collision detection for example.
+```
+
+The mixin works by preventing the `renderTree` method, therefore if `renderTree` is being
+implemented, a manual check for `isVisible` should be included to retain this functionality.
```
spydon, line 420:
```suggestion
overridden, a manual check for `isVisible` should be included to retain this functionality.
```

spydon (approved):
Lgtm, just one minor comment

PR #2688, opened 2023-08-27
fix: Allow AssetCache to load json files with array at root

### LINKED ISSUE #2682: bug: AssetCache.readJson fails for json files where the root is an array
# Current bug behavior

Reading a JSON file using `AssetCache.readJson` results in an error if the JSON file root element is an array.

An example JSON file that meets this criteria is

```json
[
    "The root of this JSON file is an array",
    "Which is perfectly legal JSON"
]
```

The error is 

```
_TypeError (type 'List<dynamic>' is not a subtype of type 'Map<String, dynamic>' in type cast)
```

This is because it attempts to cast the result to a `Map<String, dynamic>`, however, a JSON file can also be a `List<dynamic>`.

# Expected behavior

`AssetCache.readJson` should return a `Future<dynamic>` (It uses `jsonDecode` under the hood, which returns `dynamic`).

# Steps to reproduce

1. Include the JSON file above in the assets (and your pubspec)
2. ensure you `import 'package:flame/cache.dart';`
3. call `AssetsCache().readJson('test.json')`

- Flame version latest
- Affects all platforms

# More information

I can provide the fix. It's very simple and I _think_ it will be backwards compatible in most cases?

projectitis:
# Description

- Fix `AssetCache.readJson` to allow reading json files with either an object or an array at the root. 
- Added simple tests for both scenarios.

## Checklist

## Breaking Change?

## Related Issues

Closes #2682 

### DIFF
--- a/packages/flame/lib/src/cache/assets_cache.dart
+++ b/packages/flame/lib/src/cache/assets_cache.dart
@@ -47,9 +47,9 @@ class AssetsCache {
   }
 
   /// Reads a json file from the assets folder.
-  Future<Map<String, dynamic>> readJson(String fileName) async {
+  Future<dynamic> readJson(String fileName) async {
     final content = await readFile(fileName);
-    return jsonDecode(content) as Map<String, dynamic>;
+    return jsonDecode(content);
   }
 
   Future<_StringAsset> _readFile(String fileName) async {
--- a/packages/flame/test/_resources/test_1.json
+++ b/packages/flame/test/_resources/test_1.json
@@ -1,0 +1,16 @@
+[
+    1,
+    2.0,
+    true,
+    null,
+    "Hello",
+    {
+        "key": "value"
+    },
+    [
+        1,
+        2.0,
+        "three",
+        false
+    ]
+]
--- a/packages/flame/test/_resources/test_2.json
+++ b/packages/flame/test/_resources/test_2.json
@@ -1,0 +1,16 @@
+{
+    "int": 1,
+    "double": 2.0,
+    "bool": true,
+    "null": null,
+    "string": "Hello",
+    "object": {
+        "key": "value"
+    },
+    "array": [
+        1,
+        2.0,
+        "three",
+        false
+    ]
+}
--- a/packages/flame/test/cache/assets_cache_test.dart
+++ b/packages/flame/test/cache/assets_cache_test.dart
@@ -26,10 +26,16 @@ void main() {
       );
     });
 
-    test('readJson', () async {
+    test('readJson with object root', () async {
       final assetsCache = AssetsCache(prefix: '');
-      final file = await assetsCache.readJson(fixture('chopper.json').path);
+      final file = await assetsCache.readJson(fixture('test_1.json').path);
       expect(file, isA<Map<String, dynamic>>());
+    });
+
+    test('readJson with array root', () async {
+      final assetsCache = AssetsCache(prefix: '');
+      final file = await assetsCache.readJson(fixture('test_2.json').path);
+      expect(file, isA<List<dynamic>>());
     });
 
     test('readBinaryFile', () async {


packages/flame/lib/src/cache/assets_cache.dart
```diff
@@ -47,9 +47,9 @@ class AssetsCache {
   }
 
   /// Reads a json file from the assets folder.
-  Future<Map<String, dynamic>> readJson(String fileName) async {
+  Future<dynamic> readJson(String fileName) async {
```
spydon, line 50: Hmm... This is a clear degradation with having a completely dynamic type here (and a breaking change), maybe we should just have another method to load lists.

projectitis: I would strongly argue against that. For a few reasons:

1. JSON is allowed to be both an array or object at the root
2. The native json implementation returns `dynamic` for this reason
3. A user may not know what the JSON looks like before they load it, and will then not know which method to choose ahead of time (this is what I personally ran into)
4. Implementations that use JSON already expect to cast the internal JSON structures to the correct format - see below

This example is from the "aseprite" animation example that I 'fixed' as part of this PR. In the PR, I cast the dyanmic JSON to a `Map<String, dynamic>` before passing to `fromAsepriteData`. I believe this is the correct approach. As soon as you go into the method you can see that the internal `dynamic` structures are cast to `Map` as appropriate (and would be cast to `List` as well if there were any). So this is already something that consumers of JSON expect to have to do.

```
factory SpriteAnimation.fromAsepriteData(
    Image image,
    Map<String, dynamic> jsonData,
  ) {
    final jsonFrames = jsonData['frames'] as Map<String, dynamic>; // Projectitis: here
    return SpriteAnimation(
      jsonFrames.values.map((dynamic value) {
        final map = value as Map;
        final frameData = map['frame'] as Map<String, dynamic>; // Projectitis: here
        final x = frameData['x'] as int;
        final y = frameData['y'] as int;
        final width = frameData['w'] as int;
        final height = frameData['h'] as int;
        final stepTime = (map['duration'] as int) / 1000;
        final sprite = Sprite(
          image,
          srcPosition: Vector2Extension.fromInts(x, y),
          srcSize: Vector2Extension.fromInts(width, height),
        );
        return SpriteAnimationFrame(sprite, stepTime);
      }).toList(),
    );
  }
```

If you did want to go ahead with separate methods for object and node based JSON files, I would suggest:

1. Keep `dynamic readJson` as it is
2. Add `Map<String, dynmaic> readJsonObject`, and
3. Add `List<dynamic> readJsonArray`

Thoughts?

spydon: Even though it obviously is the preferred way to accept all valid json structures, we have to take into consideration that many have already done their implementations on top of the current function definition so we should try to really avoid creating a breaking change for this.

> If you did want to go ahead with separate methods for object and array based JSON files, I would suggest:

This suggestion keeps the breaking change. I think only step 3 should be done here, and then the breaking change can be done for v2.

spydon: Yeah, that could be a possibility.
What do you think @flame-engine/flame-admin?
I would introduce both 2 and 3 too, so that the user can do one less cast.

projectitis: I propose that we leave this one until v2 or their could be a bunch of confusion. JSON with an array at the root does not seem very common (or it would have come up before now). I am using `jsonDecode` directly to get around this issue myself. 

But how do we mark something so it's not forgotten in v2?

A workaround for those who run into this issue and still want to use AssetsCache to take advantage of caching is:

```dart
AssetsCache _assets = AssetsCache():
 
@override
Future<void> onLoad() async {
    /// Instead of 
    final json = await _assets.readJson('data.json');

    // Do one of these
    final json = jsonDecode(await _assets.readFile('data.json'));
    final json = jsonDecode(await _assets.readFile('data.json')) as Map<String, dynamic>;
    final json = jsonDecode(await _assets.readFile('data.json')) as List<dynamic>;
}
```

spydon: > But how do we mark something so it's not forgotten in v2?

We add it to the #1938 ticket, I added it now :)
Should we close this for now then?

projectitis: Yup, go ahead :)

spydon:
Closing since we came to the conclusion that this should wait until v2.

PR #2694, opened 2023-08-31
feat: Add TextElementComponent

luanpotter:
# Description

Add `TextElementComponent`

Example:

## Checklist

## Breaking Change?


doc/flame/rendering/text_rendering.md
```diff
@@ -113,6 +113,63 @@ You can find all the options under [TextBoxComponent's
 API](https://pub.dev/documentation/flame/latest/components/TextBoxComponent-class.html).
 
 
+### ElementComponent
+
+If you want to render an arbitrary Element, ranging from a single TextElement to a formatted
```
spydon, line 118: Is it arbitrary, or is it only related to text? Because if it is only regarding text I think it could have a more explanatory name?

spydon (approved):
Looks good, except possibly the `ElementComponent` name

doc/flame/rendering/text_rendering.md
luanpotter: By arbitrary I mean any child of Element, not only TextElement. That is how the classes are already named, even though they all have to do with the rendering of text. I didn't like calling it ElementComponent either but TextElementComponent would definitely be misnomer.

renancaraujo (approved):

doc/flame/rendering/text_rendering.md
spydon: Something completely different then? `TextRenderingComponent` or `RichTextComponent` or something? `ElementComponent` doesn't tell the user much about what it does. 😅

luanpotter: Honestly I feel `Element` is already a bad name for those reasons exactly; could be named TextElement and what is currently TextElement could be TextLineElement. But might be too much of a breaking change to change that now, probably?

For the component, I am over the fence because while I see just calling it ElementComponent can cause confusion it is actually clearly describing what it is (the confusing being on the other class, Element). `TextRenderingComponent` or `RichTextComponent` can also be confusing because are hiding what it actually does. Like unclear what the difference between `TextRenderingComponent` and `TextComponent` would be just from the name. So maybe `RichTextComponent` then?

luanpotter: actually never mind all that. I think it is worth it and not that bad to rename all these classes. let me put out a proposal.

luanpotter: https://github.com/flame-engine/flame/pull/2700

luanpotter: then this will become `TextElementComponent` and bliss will befall upon us both 😌

luanpotter: done!

spydon (approved):
Lgtm!

PR #2709, opened 2023-09-07
feat: SpawnComponent

spydon:
# Description
This PR introduces the `SpawnComponent`, which randomly spawns components within a set area.

## Checklist

## Breaking Change?


packages/flame/lib/src/components/core/component.dart
```diff
@@ -302,7 +302,7 @@ class Component {
   ///
   /// If [includeSelf] is set to true (default is false) then the component
   /// which the call is made for is also included in the search.
-  T? findParent<T extends Component>({bool includeSelf = false}) {
```
luanpotter, line 305: how can your ancestor not be a component?

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
```diff
@@ -241,4 +243,35 @@ class Polygon extends Shape {
...
+        final randomX = min.x + randomGenerator.nextDouble() * (max.x - min.x);
+        final randomY = min.y + randomGenerator.nextDouble() * (max.y - min.y);
+        result.setValues(randomX, randomY);
+
+        if (containsPoint(result)) {
+          return result;
+        }
+      }
+    } else {
+      final edgeIndex = randomGenerator.nextInt(vertices.length);
```
luanpotter, line 265: not uniformly distributed bc the edges can have different lengths and thus should have different weights.
how about adding up the perimeter, generate a random number and find the edge and then find the point

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
```diff
@@ -241,4 +243,35 @@ class Polygon extends Shape {
...
+
+  @override
+  Vector2 randomPoint({Random? random, bool within = true}) {
+    final randomGenerator = random ?? randomFallback;
+    if (within) {
+      final result = Vector2.zero();
+      final min = aabb.min;
+      final max = aabb.max;
+
+      while (true) {
```
luanpotter, line 255: I think there are ways to do this, at least for convex polygons
https://blogs.sas.com/content/iml/2020/10/21/random-points-in-polygon.html
but this is probably good enough, I wouldn't expect it to be slow for "normal" shapes

packages/flame/lib/src/experimental/geometry/shapes/rectangle.dart
```diff
@@ -161,6 +162,44 @@ class Rectangle extends Shape {
...
+  @override
+  Vector2 randomPoint({Random? random, bool within = true}) {
+    final randomGenerator = random ?? randomFallback;
+    if (within) {
+      return Vector2(
+        left + randomGenerator.nextDouble() * width,
+        top + randomGenerator.nextDouble() * height,
+      );
+    } else {
+      final edge = randomGenerator.nextInt(4);
```
luanpotter, line 174: same issue

packages/flame/lib/src/experimental/geometry/shapes/rounded_rectangle.dart
```diff
@@ -197,7 +197,9 @@ class RoundedRectangle extends Shape {
   @override
   String toString() =>
       'RoundedRectangle([$_left, $_top], [$_right, $_bottom], $_radius)';
-}
 
-@internal
-const tau = Transform2D.tau; // 2π
+  @override
+  Vector2 randomPoint({Random? random, bool within = true}) {
+    throw UnimplementedError();
```
luanpotter, line 203: you could use the rectangle algo and do the contains check on a while loop, like the arbitrary poly

packages/flame/lib/src/components/core/component.dart
spydon: If you check against a mixin for example, which I think is a pretty common usecase

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
spydon: True true! That will be quite expensive though, I guess if we use length2 it shouldn't be a problem.

packages/flame/lib/src/components/core/component.dart
luanpotter: wouldn't the mixin be `on Component` or `on PositionComponent`?

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
spydon: Yeah, I wanted it to work for concave ones too, and there doesn't seem to be any good efficient algorithm for that.

packages/flame/lib/src/components/core/component.dart
spydon: Not necessarily, see for example `PositionProvider` and all the other "provider" classes in that file,

packages/flame/lib/src/components/spawn_component.dart
```diff
@@ -0,0 +1,107 @@
...
+    this.area,
+    this.within = true,
+    double? period,
+    double? minPeriod,
+    double? maxPeriod,
+    Random? random,
+    super.key,
+  })  : _period = period ?? 0,
+        _random = random ?? _randomFallback,
+        assert(
```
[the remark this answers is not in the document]
spydon: How do you mean? They either have to specify `period` or both `minPeriod` and `maxPeriod`, so all of them can't be required?

packages/flame/lib/src/components/core/component.dart
spydon: On the other hand, then that type is not really a `parent`, so I guess it is better to go through the `ancestors` method instead.

doc/flame/components.md
```diff
@@ -340,6 +340,12 @@ void onDragUpdate(DragUpdateInfo info) {
 
 ### PositionType
 
+```{note}
+If you are using the `CameraComponent` you should not use `PositionType`, but
```
erickzanardo, line 344: Is this related to this new component?

packages/flame/lib/src/components/spawn_component.dart
```diff
@@ -0,0 +1,107 @@
...
+          period != null || (minPeriod != null && maxPeriod != null),
+          'Either the static period must be defined, or both minPeriod and '
+          'maxPeriod must be defined',
+        );
+
+  /// The function used to create new components to spawn.
+  PositionComponent Function() factory;
```
erickzanardo, line 34: maybe would be cool to receive the index of the created component, maybe people could do some logic on top of it, like, spawn X is even and Y if odd, or every 10 spawnmed, spawn a boss.

doc/flame/components.md
spydon: It is not, I just saw that the docs missed it since I added this component to the section underneath.

packages/flame/lib/src/components/spawn_component.dart
spydon: It loses the nice way of being able to do `factory: MyComponent.new`, but it is true that the index could be useful, so I'll add it!

doc/flame/components.md
erickzanardo: Got ya!

packages/flame/lib/src/components/spawn_component.dart
erickzanardo: yeah, it is a shame that we don't have unions yet :(

packages/flame/lib/src/components/spawn_component.dart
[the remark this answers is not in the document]
luanpotter: that is why I was thinking of two named constructors, `.period()` and `.random()` (or something), so for each the parameters for that type of setup are clearly defined in a type safe way.
the can both all call this, private version where everything is nullable and the check is at runtime

packages/flame/lib/src/experimental/geometry/shapes/rounded_rectangle.dart
spydon: Done! It won't work for `within = false` though.

packages/flame/lib/src/components/spawn_component.dart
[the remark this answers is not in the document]
spydon: Aaah of course, I read named arguments in my head for some reason. Fixed!

packages/flame/lib/src/components/spawn_component.dart
spydon: For this use-case unions would be really dirty since it then would have to check what the function signature is in runtime to be able to either send or not send the argument

erickzanardo (approved):

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
```diff
@@ -241,4 +243,58 @@ class Polygon extends Shape {
...
+    while (vertexIndex < verticesLengths2.length) {
+      currentEndPoint += verticesLengths2[vertexIndex];
+      if (currentEndPoint >= pointOnEdges) {
+        break;
+      }
+      vertexIndex++;
+    }
+    final startPoint = vertices[vertexIndex];
+    final endPoint = vertices[(vertexIndex + 1) % vertices.length];
+    final t = randomGenerator.nextDouble();
```
luanpotter, line 296: you don't need to generate another random double. just decrease pointOnEdges as you iterate the edges and t is the "remainder" for this edge

spydon: Ah, good idea!

packages/flame/lib/src/math/random_fallback.dart
```diff
@@ -0,0 +1,5 @@
+import 'dart:math';
+
+/// When you don't care about what [Random] object you have and don't want to
+/// create an unnecessary object you can use this pre-created object.
+final Random randomFallback = Random();
```
luanpotter, line 5: is this just to avoid instantiating a random? I feel `?? Random()` is actually shorter than `?? randomFallback`

luanpotter (approved):
LGTM. Thanks for unifying tau, it has been bothering me!

packages/flame/lib/src/math/random_fallback.dart
spydon: Exactly, so that there aren't any new `Random` objects created in tight loops

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
spydon: Hmm, then we'll have to do a square root though, I think it probably is cheaper to generate another double?

PR #2727, opened 2023-09-11
fix: Avoid creating new `Vector2` in `globalToLocal` and `localToGlobal`

spydon:
# Description
Since `globalToLocal` and `localToGlobal` are called a lot when gestures are being triggered we should try to minimize the amount of new Vector2 objects created here.

## Checklist

## Breaking Change?


luanpotter (approved):
I really really really want us to just use tuples. We need to take a look at this.

PR #2730, opened 2023-09-11
fix: Adding scale parameter to RectangleComponent constructors

erickzanardo:
# Description

While working with RectangleComponent I noticed that it doesn't expose its super `scale` parameter, this PR changes that.

## Checklist

## Breaking Change?

## Related Issues


packages/flame/lib/src/geometry/rectangle_component.dart
```diff
@@ -58,6 +60,7 @@ class RectangleComponent extends PolygonComponent {
     super.paint,
     super.paintLayers,
     super.shrinkToBounds,
+    super.scale,
```
spydon, line 63: scale already exists in this constructor, I must have accidentally only added it to one
```suggestion
```

packages/flame/lib/src/geometry/rectangle_component.dart
```diff
@@ -33,6 +34,7 @@ class RectangleComponent extends PolygonComponent {
     super.paint,
     super.paintLayers,
     super.children,
+    super.scale,
```
spydon, line 37: To keep consistent with the other components add it after `position` (usually it is after `size`, but that is not used in this constructor), and same with the default constructor.

packages/flame/lib/src/geometry/rectangle_component.dart
```diff
@@ -81,6 +84,7 @@ class RectangleComponent extends PolygonComponent {
     int? priority,
     Paint? paint,
     List<Paint>? paintLayers,
+    Vector2? scale,
```
spydon, line 87: Here it should be before `angle`

spydon (approved):
Lgtm, just some nitpicks.
It seems like it is missing for the `CircleComponent` too.

spydon:
@erickzanardo fixed up those small nits so that you don't have to, hope your don't mind :sweat_smile:

Issue #2583, opened 2023-06-21
Laggy player movement with ne camera component


KurtLa:
# Current bug behavior
The movement is laggy. I the camera itself moves smooth, but not the player.

![movement_issue](https://github.com/flame-engine/flame/assets/9027932/bfbe8cf1-bd6a-4c46-80de-a17462f8ed2b)

# Expected behavior
Since moving to the new camera component creates this, the same smooth movement like the old system.

# Steps to reproduce
You can use my repository to reproduce it:
https://bitbucket.org/Jinairu/islandsurvivor_bugs/src/master/
it seems to be at least ok in chrome, but its hard on native mac

clone, flutter pub get, run 
use W, A, S, D to navigate

# Flutter doctor output
flutter doctor
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.10.2, on macOS 13.2.1 22D68 darwin-arm64, locale en-DE)
[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 14.2)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2021.2)
[✓] VS Code (version 1.79.2)
[✓] Connected device (2 available)
[✓] Network resources

• No issues found!

# More environment information
Flame version: 1.8.0
Platform affected: web, osx
Platform version affected: osx 13.2

ufrshubham:
Possibly a duplicate of #2574

KurtLa:
> Reference

Could be the same underlying problem, but my camera runs smoothly and moves flawlessly, only the player jitters.

ufrshubham:
It is very likely that this is same as the one I mentioned earlier. Both the example code were identical in terms of the camera setup.

Can you try testing your code by depending on the `main` branch of flame from GitHub? The original issue is fixed in main 🙂

To depend on `main`, replace your flame entry with the following entry in your `pubspec.yaml` file:
```yaml
dependencies:
  flame:
    git:
      url: https://github.com/flame-engine/flame.git
      ref: main
      path: packages/flame
```

KurtLa:
I have done flutter clean before, but unfortunately the problem is persistent.
![movement_issue2](https://github.com/flame-engine/flame/assets/9027932/2ce44470-1670-423c-99d5-aa1a18734a93)

spydon:
@KurtLa are you still experiencing this?

[...]

PR #2732, opened 2023-09-12
fix: Set `CameraComponent.priority` to max

spydon:
# Description
The priority is set to the max here to avoid some bugs for the users,
if they for example would add any components that modify positions
before the CameraComponent, since it then will render the positions
of the last tick each tick.

## Checklist

## Breaking Change?

## Related Issues

Found in #2583 


luanpotter (approved):
LGTM

PR #2734, opened 2023-09-12
chore: Enable DCM rules avoid-duplicate-exports and avoid-unnecessary-type-casts

luanpotter:
# Description

Enable DCM rules avoid-duplicate-exports and avoid-unnecessary-type-casts.

I don't think there is any controversy with these. There is also no current violations.

## avoid-duplicate-exports

Forbid duplicate exports (same file) on a module file

Ref: https://dcm.dev/docs/rules/common/avoid-duplicate-exports

## avoid-unnecessary-type-casts

Forbid unnecessary casts

Ref: https://dcm.dev/docs/rules/common/avoid-unnecessary-type-casts

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame_lint/lib/analysis_options_with_dcm.yaml
+++ b/packages/flame_lint/lib/analysis_options_with_dcm.yaml
@@ -8,5 +8,7 @@ dart_code_metrics:
     - avoid-banned-imports
     - avoid-cascade-after-if-null
     - avoid-collection-methods-with-unrelated-types
+    - avoid-duplicate-exports
+    - avoid-unnecessary-type-casts
   # flutter rules
     - prefer-define-hero-tag


spydon:
Don't both of these rules already exist from the normal analyzer? I think we should avoid overlap between them as much as we can to improve maintainability.

luanpotter:
Oh, I just assumed all rules did not exist on the normal analyzer.
I guess that explains why we don't have any violations then!
@incendial is there anything these would add on top of the ones we already have?

incendial:
There are some rules that are present in both DCM and the analyzer, but usually the analyzer adds them after DCM. By my estimations there are around 10 such rules.

As for these particular rules, AFAIK "avoid-unnecessary-type-casts" does not exist in the analyzer.

spydon:
> There are some rules that are present in both DCM and the analyzer, but usually the analyzer adds them after DCM. By my estimations there are around 10 such rules.
> 
> As for these particular rules, AFAIK "avoid-unnecessary-type-casts" does not exist in the analyzer.

It does: https://dart.dev/tools/diagnostic-messages#unnecessary_cast

incendial:
It does not work with class hierarchies

incendial:
And it's a huge part of any codebase.
So if you treat it as the same rule, then it adds additional checks.

spydon:
> It does not work with class hierarchies

Interesting, can you give an example of such an occurrence? I'm guessing it is for upcasting?

incendial:
I'd say casting in general since you can make a mistake in any casting.

```dart
class Animal {}

class HomeAnimal extends Animal {}

void patterns() {
  final animal = Animal();
  
  animal as HomeAnimal; // Correct
  animal as Animal; // Wrong 

  if (animal case Animal() as HomeAnimal) {} // Correct
  if (animal case Animal() as Animal) {} // Wrong

  HomeAnimal() as HomeAnimal; // Wrong
}
```

incendial:
Then yeah, sorry for misleading, these two are equal to what the analyzer has.

spydon:
> Then yeah, sorry for misleading, these two are equal to what the analyzer has.

It probably was updated, I have a vague memory of it not being able to catch all the examples I just tried previously. :)

spydon:
Closing this since we already have both.

PR #2746, opened 2023-09-18
feat: Add `HasWorldReference` mixin

spydon:
# Description

Adds a mixin for components similar to  `HasAncestor`, `HasParent` and `HasGameReference` but which provides access to the `World` which the component has as an ancestor.

## Checklist

## Breaking Change?

## Related Issues


doc/flame/components.md
```diff
@@ -189,6 +189,28 @@ that they will appear in the children list in the same order as they were
 scheduled for addition.
 
 
+### Access to the World from a Component
+
+When a component requires access to the `World` that it is attached to one can
```
erickzanardo, line 194: Unsure how to improve but the "that is attached to one" seems a bit confusing to me, maybe you meant when the user need to attach and have access to a world instance?

ufrshubham (commented):
Not sure why a separate mixin is needed for this. Wouldn't `HasAncestor<MyWorld>` achieve the same result?

spydon:
> Not sure why a separate mixin is needed for this. Wouldn't `HasAncestor<MyWorld>` achieve the same result?

It would, but the world will be very used in the upcoming release and when you use `HasAncestor` you probably want to use it for another ancestor than the root world a lot of the time and you can't use the mixin twice.

ufrshubham:
Okay, in that case, should we call it `HasAncestorWorld` or `HasWorldAncestor`? Or maybe `BelongsToWorld`, `DescendantOfWorld`. `HasWorld` makes me think that the component is saying it has a world within it 😅

ufrshubham (approved):

spydon:
> Okay, in that case, should we call it `HasAncestorWorld` or `HasWorldAncestor`? Or maybe `BelongsToWorld`, `DescendantOfWorld`. `HasWorld` makes me think that the component is saying it has a world within it 😅

Haha we actually discussed exactly that in the chat yesterday! It was previously called `HasWorldReference` to be in line with `HasGameReference`, Luan also had the same argument as you, so time to change back I think (or to `HasWorldAncestor` maybe).

wolfenrain (approved):

PR #2749, opened 2023-09-18
chore: Remove deprecations for v1.9.0

spydon:
# Description

Removes everything that was marked as deprecated and that should be removed for v1.9.0.

## Checklist

## Breaking Change?

## Related Issues


packages/flame/lib/src/camera/viewfinder.dart
```diff
@@ -169,8 +169,8 @@ class Viewfinder extends Component
 
   @override
   void onGameResize(Vector2 size) {
-    super.onGameResize(size);
     _initZoom();
+    super.onGameResize(size);
```
erickzanardo, line 173: I am curious on why the change in the order?

erickzanardo (approved):

packages/flame/lib/src/camera/viewfinder.dart
spydon: If you have children of the viewfinder, you might want to know the new zoom from your parent when the child is running `onGameResize`. (Same thing with viewport, where it actually caused some bugs)

erickzanardo: Maybe we should document that?

spydon: Doesn't the user expect that the parent has the correct value? Or how do you mean document it? :sweat_smile:

erickzanardo: I mean adding some kind of docs that if you want the change the zoom on the on game resize, that you might want to call the super after calculating the new values, I mean, now that you explained to me it is pretty obvious, but at first I didn't realized it haha

spydon: This is the core viewfinder, on a normal component you can do what you want and it will still work, and if you subclass the viewfinder you might want to handle zoom differently and you definitely know what you're doing if you've gone that deep. So I don't think there is anything obvious to document. 😅

PR #2753, opened 2023-09-20
chore: Remove CSpell regex bypasses (part 3)

luanpotter:
# Description

Remove CSpell regex bypasses (part 3) and fix violations accordingly

## Checklist

## Breaking Change?


.github/cspell.json
```diff
@@ -11,8 +11,6 @@
     "**/CHANGELOG.md"
   ],
   "ignoreRegExpList": [
-    "\\:\\w+",
-    "\\/\\w+",
     "\\#[\\w\\-]+"
```
luanpotter, line 14: this last one is probably for colors. I will do some digging to confirm and replace with a more specific regex, and also add one for the `0xFF....` format, and remove the ignores for those.

I think other than specific categories of things (like color codes), we should have no generic regex bypass like these I am removing.

spydon (approved):

PR #2767, opened 2023-09-24
feat: Align component refactoring

denisgl7:
I suggest changing the child element of the Align component. In this case, the child element is not a required parameter.
Motivation:
Thus, the Align component can be created as a class field.
We can replace the child at any time.
Hello:
I can create a label field on a button and change the label depending on the state of the button.


packages/flame/lib/src/layout/align_component.dart
```diff
@@ -48,41 +48,61 @@ import 'package:vector_math/vector_math_64.dart';
...
-    add(child);
+    this.child = child;
   }
 
-  late Anchor _alignment;
-
-  /// The component that will be positioned by this component. The [child] will
+  /// The component that will be positioned by this component. The [_child] will
   /// be automatically mounted to the current component.
-  final PositionComponent child;
```
spydon, line 72: I would add a getter for `child` too and then refer to that getter in the dartdocs.

packages/flame/lib/src/layout/align_component.dart
```diff
@@ -48,41 +48,61 @@ import 'package:vector_math/vector_math_64.dart';
...
+  PositionComponent? _child;
+
+  set child(PositionComponent? value) {
+    if (_child == value) {
+      return;
+    }
+    final previousChild = _child;
+    if (previousChild?.parent == this) {
```
spydon, line 77: I think we can connect a listener to when the child gets its parent updated if I remember correctly and then we can unassign `_child` if it is added to another parent.

denisgl7: Yes, I understand the problem. Perhaps the child will change parents outside of our class?

denisgl7: How was this problem solved before?

denisgl7: Снимок экрана 2023-09-24 в 13 27 27
I failed the golden test by 0.18%
How is that?)))

spydon: > I failed the golden test by 0.18% How is that?))) 
It's because you're generating the goldens on an OS that is not Linux, I can generate them for you :)

denisgl7: Great! What else needs to be changed in the code?

spydon: You can override this class and set the child to null when it has been assigned to another parent:
```dart
  /// This method will be invoked from lifecycle if [child] has been added
  /// to or removed from its parent children list.
  void onChildrenChanged(Component child, ChildrenChangeType type) {}
  ```

Comment (denisgl7) [reply packages/flame/lib/src/layout/align_component.dart]: I understand correctly?
``` 
@override
  void onChildrenChanged(Component child, ChildrenChangeType type) {
    if (_child?.parent != this) {
      this.child = null;
    }
  }
```

Comment (spydon) [reply packages/flame/lib/src/layout/align_component.dart]: Exactly!

Comment (denisgl7) [reply packages/flame/lib/src/layout/align_component.dart]: I added everything to the commit
I want to use this component for a button and make a system of labels on the button depending on the states.

File: packages/flame/lib/src/layout/align_component.dart [hunk 3]
```diff
@@ -1,15 +1,13 @@
...
+import 'package:flame/components.dart';
 import 'package:flame/src/effects/provider_interfaces.dart';
 import 'package:flutter/widgets.dart';
-import 'package:vector_math/vector_math_64.dart';
 
 /// **AlignComponent** is a layout component that positions its child within
 /// itself using relative placement. It is similar to Flutter's [Align] widget.
 ///
-/// The component requires a single [child], which will be the target of this
+/// The component requires a single [_child], which will be the target of this
```
spydon, line 8: Change all these reference back to `[child]`

packages/flame/lib/src/layout/align_component.dart
```diff
@@ -115,10 +127,29 @@ class AlignComponent extends PositionComponent {
...
+      super.size = Vector2(
+        widthFactor == null ? maxSize.x : _child!.size.x * widthFactor!,
+        heightFactor == null ? maxSize.y : _child!.size.y * heightFactor!,
+      );
+    }
+    _updateChildPosition();
+  }
+
+  @override
+  void onChildrenChanged(Component child, ChildrenChangeType type) {
```
spydon, line 140: Add @mustCallSuper here

spydon (commented):
Looks good! Have you checked if there are any docs that need to be updated?

packages/flame/lib/src/layout/align_component.dart
```diff
@@ -55,34 +53,48 @@ class AlignComponent extends PositionComponent {
...
   }
 
-  late Anchor _alignment;
-
   /// The component that will be positioned by this component. The [child] will
   /// be automatically mounted to the current component.
-  final PositionComponent child;
+  PositionComponent? _child;
+
+  PositionComponent? get child => _child;
```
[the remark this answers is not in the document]
denisgl7:
```  PositionComponent? _child;

  /// The component that will be positioned by this component. The [child] will
  /// be automatically mounted to the current component.
  PositionComponent? get child => _child;

  set child(PositionComponent? value) {
    if (_child == value) {
      return;
    }
    if (_child?.parent == this) {
      _child?.removeFromParent();
    }
    _child = value;
    _child?.parent = this;
    _updateChildAnchor();
    _updateChildPosition();
  }
```  
This is true?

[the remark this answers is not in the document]
spydon: Correct indeed! :)

spydon (approved):

PR #2775, opened 2023-09-26
fix: Hardcode initCurrentGame lifecycle state as resumed

### LINKED ISSUE #2771: Flame isn't Rendering Sprites
# Current bug behavior
Flame is not rendering the sprites sometimes. It appears for a moment after pressing "back" (closing the app), or after going "home" and go back to the app. But after closing, and running the app again, it doesn't render again too. See this video: https://youtu.be/gIdxXewLPNU

# Expected behavior
Flame is always rendering the sprites.

# Steps to reproduce
I build this as an APK and install it on my phone:
```
import 'package:flame/flame.dart';

void main() {
  Flame.device.setLandscape();
  runApp(const MaterialApp(home: SafeArea(child: Scaffold(body:
    GameWidget<Engine>.controlled(gameFactory: Engine.new),
  ))));
}

class Engine extends FlameGame {
  late Scene scene;
  late CameraComponent cameraComponent;

  @override
  Future<void> onLoad() async {
    await images.load("cube.png");
    scene = Scene();
    cameraComponent = CameraComponent(world: scene);
    addAll([scene, cameraComponent]);
  }
}

class Scene extends World with HasGameRef<Engine> {
  late List<SpriteComponent> sprites;

  @override
  Future<void> onLoad() async {
    sprites = List.empty(growable: true);
    for (int x = -1; x < 2; x++) {
      for (int y = -1; y < 2; y++) {
        sprites.add(SpriteComponent(
          anchor: Anchor.center,
          position: Vector2(x * 96, y * 96),
          size: Vector2(64, 64),
          sprite: Sprite(game.images.fromCache("cube.png")),
        ));
      }
    }
    addAll(sprites);
  }
}
```

# Flutter doctor output

```
[√] Flutter (Channel stable, 3.13.5, on Microsoft Windows [Version 10.0.22621.2283], locale en-US)
    • Flutter version 3.13.5 on channel stable at C:\Users\MSI\Documents\Flutter
    • Framework revision 12fccda598 (5 days ago), 2023-09-19 13:56:11 -0700
    • Engine revision bd986c5ed2
    • Dart version 3.1.2
    • DevTools version 2.25.0

[√] Windows Version (Installed version of Windows is version 10 or higher)

[√] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at C:\Users\MSI\AppData\Local\Android\sdk
    • Platform android-34, build-tools 34.0.0
    • Java binary at: C:\Program Files\Android\Android Studio\jbr\bin\java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-b2043.56-10027231)

[X] Chrome - develop for the web (Cannot find Chrome executable at .\Google\Chrome\Application\chrome.exe)
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.

[√] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.6.5)
    • Visual Studio at C:\Program Files\Microsoft Visual Studio\2022\Community
    • Visual Studio Community 2022 version 17.6.33829.357
    • Windows 10 SDK version 10.0.22000.0

[√] Android Studio (version 2022.3)
    • Android Studio at C:\Program Files\Android\Android Studio
       https://plugins.jetbrains.com/plugin/9212-flutter
       https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-b2043.56-10027231)

    • CPH1937 (mobile)  • 7410004d • android-arm64  • Android 11 (API 30)
    • Windows (desktop) • windows  • windows-x64    • Microsoft Windows [Version 10.0.22621.2283]
    • Edge (web)        • edge     • web-javascript • Microsoft Edge 117.0.2045.36

[√] Network resources

```

* Flame version: 1.9.1
* Platform affected: android
* Platform version affected: android 11

# Log information
No error on the log

# More information

adil192:
# Description

When the init game method is called straight after the app has launched, the lifecycle state from `WidgetsBinding.instance.lifecycleState` often returns `disposed` or `inactive` which causes the engine to stay paused even though the app is foregrounded.
This PR fixes the lifecycle state as `resumed` when the init method is called.

## Checklist

## Breaking Change?

## Related Issues

Fixes #2771

### DIFF
--- a/packages/flame/lib/src/game/game_widget/game_widget.dart
+++ b/packages/flame/lib/src/game/game_widget/game_widget.dart
@@ -253,9 +253,22 @@ class GameWidgetState<T extends Game> extends State<GameWidget<T>> {
     } else {
       currentGame = widget.game!;
     }
-    currentGame.addGameStateListener(_onGameStateChange);
-    currentGame.initLifecycleState();
+    initGameStateListener(currentGame, _onGameStateChange);
     _loaderFuture = null;
+  }
+
+  /// Visible for testing for
+  /// https://github.com/flame-engine/flame/issues/2771.
+  @visibleForTesting
+  static void initGameStateListener(
+    Game currentGame,
+    void Function() onGameStateChange,
+  ) {
+    currentGame.addGameStateListener(onGameStateChange);
+
+    // See https://github.com/flame-engine/flame/issues/2771
+    // for why we aren't using [WidgetsBinding.instance.lifecycleState].
+    currentGame.lifecycleStateChange(AppLifecycleState.resumed);
   }
 
   /// [disposeCurrentGame] is called by two flutter events - `didUpdateWidget`
@@ -263,7 +276,7 @@ class GameWidgetState<T extends Game> extends State<GameWidget<T>> {
   /// `currentGame`'s `onDispose` method will be called; otherwise, it will not.
   void disposeCurrentGame({bool callGameOnDispose = false}) {
     currentGame.removeGameStateListener(_onGameStateChange);
-    currentGame.disposeLifecycleState();
+    currentGame.lifecycleStateChange(AppLifecycleState.paused);
     currentGame.onRemove();
     if (callGameOnDispose) {
       currentGame.onDispose();
--- a/packages/flame/lib/src/game/game_widget/game_widget.dart
+++ b/packages/flame/lib/src/game/game_widget/game_widget.dart
@@ -254,9 +254,7 @@ class GameWidgetState<T extends Game> extends State<GameWidget<T>> {
       currentGame = widget.game!;
     }
     currentGame.addGameStateListener(_onGameStateChange);
-    currentGame.lifecycleStateChange(
-      WidgetsBinding.instance.lifecycleState ?? AppLifecycleState.resumed,
-    );
+    currentGame.lifecycleStateChange(AppLifecycleState.resumed);
     _loaderFuture = null;
   }
 
--- a/packages/flame/lib/src/game/game_widget/game_widget.dart
+++ b/packages/flame/lib/src/game/game_widget/game_widget.dart
@@ -254,7 +254,7 @@ class GameWidgetState<T extends Game> extends State<GameWidget<T>> {
       currentGame = widget.game!;
     }
     currentGame.addGameStateListener(_onGameStateChange);
-    currentGame.lifecycleStateChange(AppLifecycleState.resumed);
+    currentGame.initLifecycleState();
     _loaderFuture = null;
   }
 
@@ -263,7 +263,7 @@ class GameWidgetState<T extends Game> extends State<GameWidget<T>> {
   /// `currentGame`'s `onDispose` method will be called; otherwise, it will not.
   void disposeCurrentGame({bool callGameOnDispose = false}) {
     currentGame.removeGameStateListener(_onGameStateChange);
-    currentGame.lifecycleStateChange(AppLifecycleState.paused);
+    currentGame.disposeLifecycleState();
     currentGame.onRemove();
     if (callGameOnDispose) {
       currentGame.onDispose();
--- a/packages/flame/lib/src/game/game.dart
+++ b/packages/flame/lib/src/game/game.dart
@@ -157,6 +157,7 @@ abstract mixin class Game {
     // for why we aren't using [WidgetsBinding.instance.lifecycleState].
     lifecycleStateChange(AppLifecycleState.resumed);
   }
+
   @internal
   void disposeLifecycleState() {
     lifecycleStateChange(AppLifecycleState.paused);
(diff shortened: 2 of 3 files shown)


adil192:
@spydon Do I need to add/edit tests for this?

spydon:
> @spydon Do I need to add/edit tests for this?

Yeah, we try to add tests whenever we fix a bug to make sure we don't have regressions later.

adil192:
I'm finding this pretty hard to discriminate with tests.
If the lifecycle state is set to e.g. `detached` then the tester doesn't call `initState` on the game widget (or even create a state at all).

spydon:
> Update: I ended up adding some `@internal` `initLifecycleState()` and `disposeLifecycleState()` methods in the `Game` class and using those in the test. Let me know what you think. I also don't know if `@internal` is the right annotation here

Since the method that those methods are using isn't internal or private, I don't see the point in creating them?

The correct annotation would be `@visibleForTesting`.

adil192:
> Since the method that those methods are using isn't internal or private, I don't see the point in creating them?
> 
> The correct annotation would be `@visibleForTesting`.

We can't use `State` objects in tests when we have backgrounded lifecycle states.
Having them as separate methods makes them testable.

`@visibleForTesting` wouldn't work since they're in `Game` so `GameWidget` wouldn't be able to access them (unless we added an ignore lint comment)

spydon:
> We can't use `State` objects in tests when we have backgrounded lifecycle states.

Why not? That was how it was done previously right?

adil192:
The existing tests change the lifecycle state _after_ creating the widget, e.g.
```dart
(tester) async {
  await tester.pumpWidget(GameWidget(game: game));

  expect(game.paused, isFalse);
  WidgetsBinding.instance.handleAppLifecycleStateChanged(
    AppLifecycleState.paused,
  );
  expect(game.paused, isTrue);
},
```

If you set it to `AppLifecycleState.paused` _before_ creating the `GameWidget`,
its `GameWidgetState` object is never created (which is what contains the `initCurrentGame()` method).

I tried manually instantiating the State object but it threw a bunch of errors because some of its private fields weren't populated.

I'm going to try another way to test this which I'll commit in a second

adil192:
This new test also correctly succeeds and fails with and without this patch

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -253,13 +253,24 @@ class GameWidgetState<T extends Game> extends State<GameWidget<T>> {
...
-      WidgetsBinding.instance.lifecycleState ?? AppLifecycleState.resumed,
-    );
+    initGameStateListener(currentGame, _onGameStateChange);
     _loaderFuture = null;
   }
 
+  /// Visible for testing for
+  /// https://github.com/flame-engine/flame/issues/2771.
+  @visibleForTesting
+  static void initGameStateListener(
```
spydon, line 263: I still don't understand the need for a new function here, you already have access to do all that the function does from the test?
But even so, it feels like cheating the test by calling this manually?

adil192: We can't use the `initCurrentGame` function since it's in a State object which isn't accessible with a backgrounded lifecycle state.
You're welcome to try some other solution but it's not as simple as the other tests.

spydon: What I mean is that you can call this straight from the tests, instead of creating a new method:
```dart
currentGame.addGameStateListener(onGameStateChange); currentGame.lifecycleStateChange(AppLifecycleState.resumed);
```

adil192: Oh that wouldn't do anything to prevent regressions though. It's also basically a duplicate of this existing test:
```dart
testWidgets(
  'game is paused when app is backgrounded',
  (tester) async {
    final game = FlameGame();

    await tester.pumpWidget(GameWidget(game: game));

    await game.toBeLoaded();
    await tester.pump();

    expect(game.paused, isFalse);
    WidgetsBinding.instance.handleAppLifecycleStateChanged(
      AppLifecycleState.paused,
    );
    expect(game.paused, isTrue);
    WidgetsBinding.instance.handleAppLifecycleStateChanged(
      AppLifecycleState.resumed,
    );
    expect(game.paused, isFalse);
  },
);
```

spydon: I don't see how adding the new method would prevent regressions?

adil192: Adding the new method allows us to test that we're using `AppLifecycleState.resumed` _not_ `WidgetsBinding.instance.lifecycleState` when the game is initialized.

spydon: Ah alright, I missed that you were actually calling this from within the widget when reviewing on the phone. I guess without doing some much more complicated test we won't be able to test this better. Approved!

spydon (approved):

PR #2784, opened 2023-10-01
refactor: Avoid nested conditional expressions whenever possible [DCM]

luanpotter:
# Description

I was playing around with the rule [avoid-nested-conditional-expressions](https://dcm.dev/docs/rules/common/avoid-nested-conditional-expressions/).
While I philosophically agree with the rule, sadly Dart's new switch-as-an-expression, while a much-needed improvement, still pales in comparison to Kotlin's powerful `when`, which makes it impossible to convert all cases of nested ternaries into something actually better. In some cases, given the tools that Dart provides, I think nested ternary is the best option.
Notwithstanding, I used the rule to catch many cases where the new switch expression can be used to vastly simplify our codebase, making it much more readable.

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame_jenny/jenny/lib/src/structure/expressions/expression.dart
+++ b/packages/flame_jenny/jenny/lib/src/structure/expressions/expression.dart
@@ -3,18 +3,17 @@ abstract class Expression {
 
   dynamic get value;
 
-  bool get isNumeric => false;
-  bool get isBoolean => false;
-  bool get isString => false;
+  bool get isNumeric => type == ExpressionType.numeric;
+  bool get isBoolean => type == ExpressionType.boolean;
+  bool get isString => type == ExpressionType.string;
 
   ExpressionType get type {
-    return isNumeric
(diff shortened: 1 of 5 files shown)


packages/flame_jenny/jenny/lib/src/structure/expressions/expression.dart
```diff
@@ -3,18 +3,17 @@ abstract class Expression {
...
 
   ExpressionType get type {
-    return isNumeric
-        ? ExpressionType.numeric
-        : isBoolean
-            ? ExpressionType.boolean
-            : isString
-                ? ExpressionType.string
-                : ExpressionType.unknown;
+    return switch (this) {
```
spydon, line 11: I think this new way will be less performant if the getters will be used a lot when going through a deep tree for example, but it's way more readable. I guess it's not worth optimizing for performance here already though!

spydon (approved):

PR #2787, opened 2023-10-02
fix!: Viewfinder children statically on top of the world

spydon:
# Description

This PR changes the order in which the `CameraComponent` renders its internal components; the viewport and the viewfinder.
With this change the rendering order looks like this:
```
viewport - Good for adding static background components to.
world
viewfinder - Good for adding HUD components to.
```

In the following example you can see a `ParallaxComponent` attached to the viewport, a player and some other components attached to the world, and some texts attached to the viewfinder and viewport:
image

## Checklist

## Breaking Change?

### Migration instructions

If you were previously adding your HUDs directly to the viewport, add them to the viewfinder instead.
If you added your HUDs through the `hudComponents` argument in the `CameraComponent` constructor the change has already been made for you.

## Related Issues

### DIFF
--- a/doc/flame/camera_component.md
+++ b/doc/flame/camera_component.md
@@ -28,6 +28,10 @@ Then, a [](#cameracomponent) class that "looks at" the `World`. The
 flexibility of rendering the world at any place on the screen, and also control
 the viewing location and angle.
 
+If you add children to the `Viewfinder` they will appear as static HUDs in
+front of the world and if you add children to the `Viewport` they will appear
+statically behind the world.
+
 
 ## World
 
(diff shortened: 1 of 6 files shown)


ufrshubham:
@spydon, just a thought, what if let viewfinder + viewport both render on top of the world and use a separate parent for the background? Something like a world background component?

Just thinking in terms of a real-life camera, where UI elements of viewfinder are always displayed on top of the image that is being captured.

spydon:
> @spydon, just a thought, what if let viewfinder + viewport both render on top of the world and use a separate parent for the background? Something like a world background component?
> 
> Just thinking in terms of a real-life camera, where UI elements of viewfinder are always displayed on top of the image that is being captured.

Yeah, that might make more sense. Do you think the viewfinder should be affected by the transformations or not then?

ufrshubham:
> Do you think the viewfinder should be affected by the transformations or not then?

Making it behave similar to viewport should be fine I think.

spydon:
@ufrshubham what do you think about this solution? :)

ufrshubham:
Looks perfect! (until someone says otherwise 😉)

ufrshubham (approved):

erickzanardo (approved):

PR #2791, opened 2023-10-03
feat: Prevent looping over full components tree on every tap event

ASGAlex:
# Description

In current implementation every touch event produces at least one full-scanning of whole components tree. This is cause when every simple HUD button can hang the game for a moment or more...

There are several approaches to fix that, but many of them requires significant changes and tracking component's state changes, parent changes and so on. This can produce additional errors, so I chosed more simple way.

Game developer now can choose a component, which will contain all interactable components. All tappable, draggable components (and so on) should be placed inside this root component. After taht, components lookup on touch event will be limited by this root component and it's descendants. 

This optimisation requires user do not place components ingo game directly, but everithing is optional. Developer can skip components grouping if performance is ok.

## Checklist

## Breaking Change?

### DIFF
--- a/examples/lib/stories/input/draggables_multiple_example.dart
+++ b/examples/lib/stories/input/draggables_multiple_example.dart
@@ -1,0 +1,62 @@
+import 'package:examples/commons/ember.dart';
+import 'package:flame/components.dart';
+import 'package:flame/events.dart';
+import 'package:flame/game.dart';
+import 'package:flutter/material.dart' show Colors;
+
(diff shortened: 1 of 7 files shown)


spydon:
I think we should be able to come up with a better solution than this, where the user shouldn't have to do anything to get the performance boost. The idea we were talking about before was building a separate tree that is tracking components that can be part of an interaction.

spydon:
Closing in favor of #2811 (after taking to Alex), and in the long run we have to create a better event propagation system.

PR #2792, opened 2023-10-03
fix: Un-register component keys down the component tree

ufrshubham:
# Description

ComponentKey for children components were not getting unregistered when the sub-tree was removed from the game. As a result:

- same key couldn't be reused
- components with such keys couldn't be readded once removed
- component with such keys were not being garbage collected (didn't actually verify this one though)

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/components/core/component.dart
+++ b/packages/flame/lib/src/components/core/component.dart
@@ -896,17 +896,13 @@ class Component {
   void _remove() {
     assert(_parent != null, 'Trying to remove a component with no parent');
 
-    if (_key != null) {
-      final game = findGame();
-      if (game is FlameGame) {
-        game.unregisterKey(_key!);
-      }
-    }
+    _nukeKey();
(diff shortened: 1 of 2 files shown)


packages/flame/lib/src/components/core/component.dart
```diff
@@ -920,6 +915,15 @@ class Component {
     );
   }
 
+  void _nukeKey() {
```
spydon, line 918:
```suggestion
  void _unregisterKey() {
```

spydon (approved):
Good find!

luanpotter (approved):
LGTM

packages/flame/lib/src/components/core/component.dart
ufrshubham: Renamed as suggested 👍🏻

spydon (approved):

PR #2807, opened 2023-10-09
feat: AssetsBundle can be customized in Images and AssetsCache.

erickzanardo:
# Description

Previously, `Images` and `AssetsCache` would always try to fetch assets from the `rootBundle`.

With this PR, we can have instances of those classes that uses a custom `AssetsBundle`. When this gets merged, we could consider refactoring flame_network_assets to  use this new feature and allow Flame users to use the same API
to load both local and network assets.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/doc/flame/structure.md
+++ b/doc/flame/structure.md
@@ -44,5 +44,5 @@ global ones provided by Flame.
 
 Additionally, `AssetsCache` and `Images` can receive a custom
 [`AssetBundle`](https://api.flutter.dev/flutter/services/AssetBundle-class.html).
-This can be used to make Flame look for assets in different location other the `rootBundle`,
+This can be used to make Flame look for assets in a different location other the `rootBundle`,
 like the file system for example.
--- a/doc/flame/structure.md
+++ b/doc/flame/structure.md
@@ -39,7 +39,7 @@ flutter:
 ```
 
 If you want to change this structure, this is possible by using the `prefix` parameter and creating
-your instances of `AssetsCache`, `Images`, `AudioCache`, and `SoundPool`s, instead of using the
+your instances of `AssetsCache`, `Images`, and `AudioCache`, instead of using the
 global ones provided by Flame.
 
 Additionally, `AssetsCache` and `Images` can receive a custom
--- a/doc/flame/structure.md
+++ b/doc/flame/structure.md
@@ -39,5 +39,10 @@ flutter:
 ```
 
 If you want to change this structure, this is possible by using the `prefix` parameter and creating
-your instances of `AssetsCache`, `ImagesCache`, `AudioCache`, and `SoundPool`s, instead of using the
+your instances of `AssetsCache`, `Images`, `AudioCache`, and `SoundPool`s, instead of using the
 global ones provided by Flame.
+
+Additionally, `AssetsCache` and `Images` can receive a custom
+[`AssetBundle`](https://api.flutter.dev/flutter/services/AssetBundle-class.html).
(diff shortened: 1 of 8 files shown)


doc/flame/structure.md
```diff
@@ -39,5 +39,10 @@ flutter:
 ```
 
 If you want to change this structure, this is possible by using the `prefix` parameter and creating
-your instances of `AssetsCache`, `ImagesCache`, `AudioCache`, and `SoundPool`s, instead of using the
+your instances of `AssetsCache`, `Images`, `AudioCache`, and `SoundPool`s, instead of using the
 global ones provided by Flame.
+
+Additionally, `AssetsCache` and `Images` can receive a custom
+[`AssetBundle`](https://api.flutter.dev/flutter/services/AssetBundle-class.html).
+This can be used to make Flame look for assets in different location other the `rootBundle`,
```
spydon, line 47:
```suggestion
This can be used to make Flame look for assets in a different location other the `rootBundle`,
```

packages/flame/lib/src/cache/assets_cache.dart
```diff
@@ -1,16 +1,26 @@
...
 /// It automatically looks for files in the `assets` directory.
 class AssetsCache {
-  final String prefix;
-  final Map<String, _Asset<dynamic>> _files = {};
+  AssetsCache({
+    this.prefix = 'assets/',
+    AssetBundle? bundle,
+  }) {
+    this.bundle = bundle ?? Flame.bundle;
+  }
```
spydon, line 16:
```suggestion
  }) : bundle = bundle ?? Flame.bundle;
```
Like this the `bundle` doesn't have to be late :)

packages/flame/lib/src/cache/images.dart
```diff
@@ -7,12 +7,20 @@ import 'package:flutter/painting.dart';
 import 'package:flutter/services.dart';
 
 class Images {
-  Images({String prefix = 'assets/images/'}) {
+  Images({
+    String prefix = 'assets/images/',
+    AssetBundle? bundle,
+  }) {
     this.prefix = prefix;
+    this.bundle = bundle ?? Flame.bundle;
```
spydon, line 15: Same here as the other one, and the prefix can be set directly in the constructor right?

doc/flame/structure.md
```diff
@@ -39,5 +39,10 @@ flutter:
 ```
 
 If you want to change this structure, this is possible by using the `prefix` parameter and creating
-your instances of `AssetsCache`, `ImagesCache`, `AudioCache`, and `SoundPool`s, instead of using the
+your instances of `AssetsCache`, `Images`, `AudioCache`, and `SoundPool`s, instead of using the
```
spydon, line 42: Not your change, but `SoundPool` here should be `AudioPool` right? :thinking:

spydon (approved):
Lgtm!

doc/flame/structure.md
erickzanardo: I had the same doubt, but since I wasn't sure, I just left as is hahah, but I believe so, lets change?

spydon: Let's do eeet!

spydon: Or just remove it? Since there isn't any static `AudioPool` in the same sense as the others? :thinking:

erickzanardo: Agreed!

```suggestion
your instances of `AssetsCache`, `Images`,  and `AudioCache`, instead of using the
```

spydon (approved):

PR #2814, opened 2023-10-11
feat(flame_tiled): Allowing tilesets with images in the same folder to load

erickzanardo:
# Description

Before, when trying to use tilesets where their images were stored in the same folder as their tsx were, the loading would break.

This PR addresses that.

## Checklist

## Breaking Change?

## Related Issues


erickzanardo:
@spydon I've updated one of the goldens. I was curious on why it changed, because in theory, what I changed, shouldn't produce different goldens. Here is a screenshoot comparing both goldens:

Screenshot 2023-10-11 at 14 05 17

After debugging, I reliazed that this probably changed because since I map some paths, the order in which images are written in the atlas is different, which generates a different packing, IMO, this isn't an issue, since all the images are still there, so I just updated the golden. But let me know if you agree with that

PR #2818, opened 2023-10-13
fix: Do not scale debug texts with zoom

l1553k:
# Description
Right now the debug texts (certain coordinates of components) use 12px size, but `zoom: 10` turns that into `120px`. This PR fixes that, debug informations are displayed using 12px size font, regardless of zoom.

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame/lib/src/components/core/component.dart
+++ b/packages/flame/lib/src/components/core/component.dart
@@ -974,9 +974,10 @@ class Component {
   /// Returns a [TextPaint] object with the [debugColor] set as color for the
   /// text.
   TextPaint get debugTextPaint {
+    final zoom = CameraComponent.currentCamera?.viewfinder.zoom ?? 1.0;
     if (!_debugTextPaintCache.isCacheValid([debugColor])) {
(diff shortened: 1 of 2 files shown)


packages/flame/lib/src/components/core/component.dart
```diff
@@ -974,9 +974,10 @@ class Component {
   /// Returns a [TextPaint] object with the [debugColor] set as color for the
   /// text.
   TextPaint get debugTextPaint {
+    final zoom = CameraComponent.currentCamera?.viewfinder.zoom ?? 1.0;
```
spydon, line 977: It would be good to take the scale of the viewport into account too

l1553k: Are you talking about `camera.viewport.size`?

spydon: Ah, apparently only the `FixedResolutionViewport` exposes the scale, so we can skip that for now.

spydon (approved):

PR #2835, opened 2023-11-02
feat: MoveAlongPathEffect should maintain initial angle of the component

### LINKED ISSUE #2833: MoveAlongPathEffect oriented should keep inital angle of the component
# What could be improved
MoveAlongPathEffect when oriented is true should keep the original orientation of the component

# Why should this be improved
The component is always oriented in the same direction, event if the need is to have it oriented differently

# Any risks?
Not really as it gets the initial component settings and if and angle is prvided it will keep that orientation

# More information
Test case
```
import 'dart:math';

void main() {
  final game = MainGame();
  runApp(
    MaterialApp(
      home: Scaffold(
        body: Stack(
          children: [
            GameWidget(
              game: game,
            ),
          ],
        ),
      ),
    ),
  );
}

class MainGame extends FlameGame {
  MainGame()
      : super(
          camera: CameraComponent.withFixedResolution(
            width: 400,
            height: 600,
          )..viewfinder.anchor = Anchor.topLeft,
          world: _MainGametWorld(),
        );
}

class _MainGametWorld extends World {
  @override
  Future<void> onLoad() async {
    super.onLoad();

    final path2 = Path()..addOval(const Rect.fromLTRB(80, 230, 320, 470));
    final List<Vector2> vertices = [
      Vector2(40, 30),
      Vector2(100, 30),
      Vector2(70, 200)
    ];
    for (var i = 0; i < 20; i++) {
      add(
        PolygonComponent(vertices)
          ..position = Vector2(i * 10, 0)
          ..paint = (Paint()
            ..color = i.remainder(2) == 0 ? Colors.white : Colors.blue)
          ..angle = pi // THIS Angle should be kept after the effect is applied
          ..add(
            MoveAlongPathEffect(
              path2,
              EffectController(
                duration: 6,
                startDelay: i * 0.3,
                infinite: true,
              ),
              absolute: true,
              oriented: true,
            ),
          ),
      );
    }
  }
}
```

ivanfemia:
# Description

See #2833 

## Checklist

## Breaking Change?

## Related Issues

Closes #2833 


packages/flame/lib/src/effects/move_along_path_effect.dart
```diff
@@ -85,7 +85,8 @@ class MoveAlongPathEffect extends MoveEffect {
         'An `oriented` MoveAlongPathEffect cannot be applied to a target that '
         'does not support rotation',
       );
-      (target as AngleProvider).angle = _lastAngle = -start.angle;
+      _lastAngle = -start.angle;
+      (target as AngleProvider).angle += -start.angle;
```
spydon, line 89: I don't think this is correct, since `followAngle` is true, it should be following the angle of the effect, and not be affected by what its current angle is?

ivanfemia: Well the problem is with the preexisting angle of the component, in the current release this is always initialized to the tangent. It should keep consider the angle of the component itself. Please have a look to the demo in the issue.

spydon: Then you should be using `nativeAngle` to set that, not `angle`.

ivanfemia: Sounds good, we can close this as intended behaviour

ivanfemia:
@spydon Coding style preference
```

      (target as AngleProvider).angle = (target is PositionComponent)
          ? -start.angle + (target as PositionComponent).nativeAngle
          : -start.angle;
```

instead of
```
(target as AngleProvider).angle = _lastAngle = -start.angle;
(target as AngleProvider).angle += (target is PositionComponent)
          ? (target as PositionComponent).nativeAngle
          : 0;
```

spydon:
> @spydon Coding style preference

I think it is even clearer to read it if you separate them and just add:
```dart
final _target = target;
(_target as AngleProvider).angle = -start.angle;
if(_target is PositionComponent) {
  _target.angle += _target.nativeAngle;
}
```

packages/flame/lib/src/effects/move_along_path_effect.dart
```diff
@@ -85,7 +85,12 @@ class MoveAlongPathEffect extends MoveEffect {
         'An `oriented` MoveAlongPathEffect cannot be applied to a target that '
         'does not support rotation',
       );
-      (target as AngleProvider).angle = _lastAngle = -start.angle;
+      _lastAngle = -start.angle;
+      final targetObj = target;
```
spydon, line 89: We don't allow abbreviations in the codebase, you could name it `angleTarget` maybe.

spydon (approved):
Lgtm, only a variable name change needed I think!

packages/flame/lib/src/effects/move_along_path_effect.dart
ivanfemia: I would suggest targetProvider as this is not actually an angle

PR #2846, opened 2023-11-10
feat: Add TimeTrackComponent and ChildCounterComponent

erickzanardo:
# Description

Adds `TimeTrackComponent` and a `ChildCounterComponent` to help developers track down perfomance issues in their game.

## Checklist

## Breaking Change?

## Related Issues

Replace or remove this text.


packages/flame/lib/src/components/component_track_component.dart
```diff
@@ -0,0 +1,38 @@
+import 'dart:async';
+import 'dart:ui';
+
+import 'package:flame/components.dart';
+import 'package:flame/text.dart';
+
+class ComponentTrackComponent<T> extends TextComponent {
```
spydon, line 7: This is a bit of a confusing name :sweat_smile:

packages/flame/lib/src/components/time_track_component.dart
```diff
@@ -0,0 +1,55 @@
...
+class TimeTrackComponent extends TextComponent {
+  TimeTrackComponent({super.position});
+
+  static final Map<String, int> _startTimes = {};
+  static final Map<String, int> _endTimes = {};
+
+  static void clear() {
+    _startTimes.clear();
+    _endTimes.clear();
+  }
```
spydon, line 16: Hmm, do we really want all these to be static?

spydon (commented):
Looks good, no harm in providing debug helpers!
Maybe they could reside in a subdirectory, like `components/debug` or something like that?

packages/flame/lib/src/components/component_track_component.dart
erickzanardo: Yes, this is so freaking confusing hahaha, but like I just submitted this as it was based on my perfomance debugging, and I was not really caring about names hahahahaha

Maybe `ChildTrackComponent`? Since this will track the number of a certain child of a parent?

alestiago:
I'm in favour of moving debug components to a `debug` subdirectory and maybe even prefixing all of them with `Debug` if that helps to avoid confusion.

packages/flame/lib/src/components/time_track_component.dart
erickzanardo: Yeah this is also part of the rough implementation. But at the same time, I thought that keeping then static would make the tracking be less impactful and make the tracking more accurate? But anyhow totally open to suggestion, at the time I just did it static because I needed something quick.

spydon:
> I'm in favour of moving debug components to a `debug` subdirectory and maybe even prefixing all of them with `Debug` if that helps to avoid confusion.

I don't think we have to prefix every component with debug, but we could skip exporting them in `components.dart` and just export them in a separate `debug.dart` barrel file.

packages/flame/lib/src/components/time_track_component.dart
spydon: Yeah, if it is an explicit debug component it can probably just be kept as static, since debugging is supposed to be "quick & dirty" anyways.

packages/flame/lib/src/components/component_track_component.dart
alestiago: Ngl, the first time I read this I thought it was about a race car game! Since I interpreted "track" as the "race track" 😂

spydon: > Maybe ChildTrackComponent?

I don't think that really conveys what it does either, maybe `ChildCounterComponent` or something like that?

doc/flame/other/debug.md
```diff
@@ -35,3 +35,45 @@ commonly want to show the current FPS somewhere when the `FpsComponent` is used.
 
 
 [TextComponent]: ../rendering/text_rendering.md#textcomponent
+
+
+### ChildCounterComponent
+
+`ChildCounterComponent` is a component that can be added to a game and will every second render
+the number of children of type `T` from a component. So for example:
```
spydon, line 43:
```suggestion
`ChildCounterComponent` is a component that renders the number of children of
type `T` from a component (`target`) every second. So for example:
```

doc/flame/other/debug.md
```diff
@@ -35,3 +35,45 @@ commonly want to show the current FPS somewhere when the `FpsComponent` is used.
...
+
+```dart
+add(
+  ChildCounterComponent<SpriteAnimationComponent>(
+    target: world,
+  ),
+);
+```
+
+Will render the number of `SpriteAnimationComponent` that are children of the game `world`.
```
spydon, line 53: It would be better to merge this with "So for example:" and put it above the code, since it sounds like it is continuing that sentence.
> So for example, the following will render the number of `SpriteAnimationComponent` that are children of the game `world`:

packages/flame/lib/src/components/child_counter_component.dart
```diff
@@ -0,0 +1,42 @@
...
+        fontSize: 12,
+      ),
+    );
+
+    add(
+      TimerComponent(
+        period: 1,
+        repeat: true,
+        onTick: () {
+          text = '$_label: ${target.children.whereType<T>().length}';
```
spydon, line 37:
```suggestion
          text = '$_label: ${target.children.query<T>().length}';
```
Much cheaper to use `query`.

packages/flame/lib/components.dart
```diff
@@ -48,6 +49,7 @@ export 'src/components/sprite_group_component.dart';
 export 'src/components/text_box_component.dart';
 export 'src/components/text_component.dart';
 export 'src/components/text_element_component.dart';
+export 'src/components/time_track_component.dart';
```
spydon, line 52: Move these to a `debug.dart` barrel file like we discussed :)

packages/flame/lib/src/components/time_track_component.dart
```diff
@@ -0,0 +1,65 @@
+import 'dart:async';
```
spydon, line 1: Can you create a subdirectory inside `component` called `debug` too where these can live? And eventually we can move over other ones too, like the ones tracking fps (probably add those to the debug barrel file too).

spydon (approved):

PR #2851, opened 2023-11-14
feat: Adding way to configure a layer paint in flame tiled

erickzanardo:
# Description

Creates a configuration in Flame Tiled that allows developers to use custom paint objects which can be useful when we need to render tilesets that require a different Paint configuration other than the default one. Like low rez pixel art tilesets which benefits from using filterQuality.none

## Checklist

## Breaking Change?

## Related Issues

Replace or remove this text.

### DIFF
--- a/packages/flame_tiled/lib/flame_tiled.dart
+++ b/packages/flame_tiled/lib/flame_tiled.dart
@@ -1,8 +1,21 @@
 library flame_tiled;
 
+import 'dart:ui';
+
 export 'package:tiled/tiled.dart';
 
 export 'src/flame_tsx_provider.dart';
 export 'src/renderable_tile_map.dart';
 export 'src/simple_flips.dart';
 export 'src/tiled_component.dart';
+
+Paint _defaultPaintFactory(double opacity) =>
+    Paint()..color = Color.fromRGBO(255, 255, 255, opacity);
+
+class FlameTiled {
+  FlameTiled._() : paintFactory = _defaultPaintFactory;
+
+  Paint Function(double opacity) paintFactory;
+
+  static final FlameTiled instance = FlameTiled._();
+}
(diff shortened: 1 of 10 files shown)


spydon (commented):
This does things a bit differently than what we do everywhere else where we pass along a `Paint` object. Meanwhile this looks quite handy, shouldn't we first and foremost go for consistency?

I also don't understand fully, why is only the `opacity` an input to the factory function?

erickzanardo:
> This does things a bit differently than what we do everywhere else where we pass along a `Paint` object. Meanwhile this looks quite handy, shouldn't we first and foremost go for consistency?
> 
> I also don't understand fully, why is only the `opacity` an input to the factory function?

Yeah, so, from what I understood from Flame Tiled code (I am still getting used to all of its structure haha). There is one paint per layer. That paint should contain the opacity of such layer, hence why the factory receives the opacity, so the paint can be built taking that into consideration. I am not sure how we could do if `TiledComponent` received the paint object directly instead, because that object would already arrive built to the layer and it would be built without the layer opacity.

Usually I would be more for the consistency, but in this case I don't see how we could accommodate by passing the objects.

spydon:
> Yeah, so, from what I understood from Flame Tiled code (I am still getting used to all of its structure haha). There is one paint per layer. That paint should contain the opacity of such layer, hence why the factory receives the opacity, so the paint can be built taking that into consideration. I am not sure how we could do if `TiledComponent` received the paint object directly instead, because that object would already arrive built to the layer and it would be built without the layer opacity.
> 
> Usually I would be more for the consistency, but in this case I don't see how we could accommodate by passing the objects.

It should be possible to pass down a factory function through the `FlameTileLayer`, exactly the same path as the `filterQuality` goes today I think? Going down the route of having these disconnected singletons is a bit dangerous I think.

erickzanardo:
Ah, alright, I thought you mentioned passing down the instantiated objects, not the factory itself. That we can do!!!!

spydon (approved):

PR #2876, opened 2023-11-27
refactor!: Replace `Offset` with `opacityFrom` and `opacityTo` in ColorEffect

ufrshubham:
# Description

`ColorEffect`'s API was a bit confusing because it used `Offset` for getting values of opacity start and end from user. This PR changes that to use optional `opacityFrom` and `opacityTo` double parameters. It also adds validate checks on these value to make sure that are between 0 and 1.

## Checklist

## Breaking Change?

### Migration instructions

To specify start and end opacities for `ColorEffect` use the optional named parameters `opacityFrom` and `opacityTo`. So `offset.dx` should be set as `opacityFrom` and `offset.dy` should be set as `opacityTo`.

## Related Issues

NA

### DIFF
--- a/doc/flame/effects.md
+++ b/doc/flame/effects.md
@@ -535,13 +535,14 @@ Usage example:
 ```dart
 final effect = ColorEffect(
   const Color(0xFF00FF00),
-  const Offset(0.0, 0.8),
   EffectController(duration: 1.5),
+  opacityfrom = 0.2,
+  opacityTo: 0.8,
 );
 ```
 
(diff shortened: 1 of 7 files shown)


spydon:
This is on the #1938 list, but maybe it is worth doing this directly, because it is indeed really confusing.

ufrshubham:
> This is on the #1938 list, but maybe it is worth doing this directly, because it is indeed really confusing.

Didn't know this was already in that list 😅

spydon (approved):
Lgtm!

PR #2926, opened 2023-12-16
fix: Allow setting `bounds` while `BoundedPositionBehavior`'s target is null

### LINKED ISSUE #2655: BoundedPositionBehavior fails with target == null
# Current bug behavior
When I call CameraComponent.setBounds it fails, because BoundedPositionBehavior's _target is null. I traced it down to line 45 in packages/flame/lib/src/camera/behaviors/bounded_position_behavior.dart where _target! is used. In this case it is null, and the exception occurs.

# Expected behavior
I would expect setBounds not to fail.

# Steps to reproduce
Here is the example that causes a null exception https://zapp.run/edit/flame-zl3i06lvl3j0?entry=lib/main.dart&file=lib/main.dart

# Flutter doctor output
I don't believe flutter doctor output is relevant here since there is an example from zapp.run attached.

* Flame version: 1.5.0

# Log information

# More information

ufrshubham:
# Description

The `bounds` setter of `BoundedPositionBehavior` tries to update the target's position when bounds are updated. But it wasn't checking if the target is null. This was causing null exceptions while updating bounds of an unmounted `BoundedPositionBehavior` with null target (as seen in [this failing test case](https://github.com/flame-engine/flame/actions/runs/7231182930/job/19704091006#step:5:1291)). This PR fixes that by checking if the target is null before updating the position.

## Checklist

## Breaking Change?

## Related Issues

Closes #2655 

### DIFF
--- a/packages/flame/lib/src/camera/behaviors/bounded_position_behavior.dart
+++ b/packages/flame/lib/src/camera/behaviors/bounded_position_behavior.dart
@@ -36,7 +36,9 @@ class BoundedPositionBehavior extends Component {
     _bounds = newBounds;
     if (!isValidPoint(_previousPosition)) {
       _previousPosition.setFrom(_bounds.center);
-      update(0);
+      if (_target != null) {
+        update(0);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


spydon (approved):
Good catch!

PR #2930, opened 2023-12-17
fix: Misalignment of the hittest area of PolygonHitbox

### LINKED ISSUE #2758: `HasGestureHitboxes` is slightly off
# Current bug behavior
If you try this example: https://examples.flame-engine.org/#/Input_Gesture%20Hitboxes
you can see that the hitboxes aren't perfect if you are tapping right outside some edges of polygons for example.
And also, the hover callbacks don't seem to trigger at all...
The issue was the same in v1.8.2.

# Expected behavior
That the gesture hitboxes are exact like they used to be.

# Steps to reproduce
Use this example: https://examples.flame-engine.org/#/Input_Gesture%20Hitboxes

# Flutter doctor output

```
[✓] Flutter (Channel stable, 3.13.4, on Arch Linux 6.5.3-arch1-1, locale en_DK.UTF-8)
    • Flutter version 3.13.4 on channel stable at /home/spydon/bin/flutter
    • Framework revision 367f9ea16b (9 days ago), 2023-09-12 23:27:53 -0500
    • Engine revision 9064459a8b
    • Dart version 3.1.2
    • DevTools version 2.25.0

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at /opt/android-sdk
    • Platform android-33, build-tools 30.0.3
    • ANDROID_HOME = /opt/android-sdk
    • ANDROID_SDK_ROOT = /opt/android-sdk
    • Java binary at: /home/spydon/bin/android-studio/jbr/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b829.9-10027231)

    • Chrome at google-chrome

[✓] Linux toolchain - develop for Linux desktop
    • clang version 16.0.6
    • cmake version 3.27.5
    • ninja version 1.11.1
    • pkg-config version 1.8.1

[✓] Android Studio (version 2022.3)
    • Android Studio at /home/spydon/bin/android-studio
    • Flutter plugin version 75.1.2
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b829.9-10027231)

    • Linux (desktop) • linux  • linux-x64      • Arch Linux 6.5.3-arch1-1
    • Chrome (web)    • chrome • web-javascript • Google Chrome 117.0.5938.88

[✓] Network resources

• No issues found!
```

* Flame version: 1.8.2 and 1.9.0 at least

# More information
I'll do a git bi-sect tomorrow to find when this was introduced.

organic-nailer:
# Description

When performing a pointer hit test using PolygonHitbox, there may be a discrepancy between the expected range and the range in which the hit test succeeds. If the polygon shape has an offset at the top left, the hit test will behave as if there is no offset.

Investigation revealed that the cause was in the calculation of containsLocalPoint() in the PolygonComponent. The unnecessary offset calculation was removed, and the hit test is now performed in the expected range.

## Checklist

## Breaking Change?

## Related Issues

Closes #2758


packages/flame/test/geometry/polygon_component_test.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:flame/components.dart';
+import 'package:flame/geometry.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('PolygonComponent', () {
```
spydon, line 6: Could you move this to shape_component_test where the other tests for this are?
There are some inconsistencies of which directories the tests live currently so it can be quite hard to find the correct place. :sweat_smile:

spydon (commented):
Thank you so much for looking into this! I've been meaning to look into this for months.

spydon:
If you run `melos format` locally it should fix the formatting issue.

spydon (approved):
Lgtm!

PR #2943, opened 2023-12-21
fix: Package flame_riverpod, setState() or markNeedsBuild() called during build.

markvideon:
# Description

The forceRebuild function now has an "isDirty" check to determine whether setState should be called. Previously, setState was uncritically invoked which *could* produce the Flutter framework error that occurs when setState is called while widgets are still being rebuilt. 

The setState call inside ref.watch has been replaced with a call to forceRebuild, and forceRebuild now has checks to prevent setState from being called while a build is in progress. A persistent frame callback has been added to reset supporting flags. 

## Checklist

## Breaking Change?

## Related Issues

N/A. Customer reported issue[ via Discord.](https://discord.com/channels/509714518008528896/1186421112217813062/1186908306179100723) 

### DIFF
--- a/packages/flame_riverpod/lib/src/widget.dart
+++ b/packages/flame_riverpod/lib/src/widget.dart
@@ -35,8 +35,11 @@ class RiverpodAwareGameWidget<T extends Game> extends GameWidget<T> {
 }
 
 class RiverpodAwareGameWidgetState<T extends Game> extends GameWidgetState<T>
-    implements WidgetRef {
+    with WidgetsBindingObserver implements WidgetRef {
   RiverpodGameMixin get game => widget.game! as RiverpodGameMixin;
+
+  bool _isForceBuilding = false;
+  bool _hasQueuedBuild = false;
 
   late ProviderContainer _container = ProviderScope.containerOf(context);
   var _dependencies =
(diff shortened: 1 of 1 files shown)


packages/flame_riverpod/lib/src/widget.dart
```diff
@@ -47,14 +51,31 @@ class RiverpodAwareGameWidgetState<T extends Game> extends GameWidgetState<T>
   List<_ListenManual<Object?>>? _manualListeners;
 
   /// Rebuilds the [RiverpodAwareGameWidget] by calling [setState].
+  // Undesirable to call [setState] while the widget may be building.
+  // Honour requests to rebuild by setting a flag.
```
spydon, line 55: I guess this is meant as a comment and not a dartdoc? Never seen them mixed after each other before, does it even work? :sweat_smile:

spydon (approved):
Lgtm, just a comment about a comment

packages/flame_riverpod/lib/src/widget.dart
spydon: Maybe the comments can be moved into the function?

packages/flame_riverpod/lib/src/widget.dart
```diff
@@ -51,8 +51,9 @@ class RiverpodAwareGameWidgetState<T extends Game> extends GameWidgetState<T>
   List<_ListenManual<Object?>>? _manualListeners;
 
   /// Rebuilds the [RiverpodAwareGameWidget] by calling [setState].
-  // Undesirable to call [setState] while the widget may be building.
-  // Honour requests to rebuild by setting a flag.
+  /// As it is undesirable to call [setState] while the widget may be building,
+  /// this function honours (potential) subsequent requests to rebuild by
+  /// setting a flag.
```
spydon, line 56: Even better

spydon (approved):

packages/flame_riverpod/lib/src/widget.dart
markvideon: I had considered that but have just modified the comments outside the function!

PR #2955, opened 2024-01-01
refactor!: Remove the Projector interface that is no longer used for coordinate transformations

luanpotter:
# Description

Remove the Projector interface that is no longer used for coordinate transformations.

While this is exposed to end-users there is not much point in doing so as it was always an internal detail of our implementation.
Since our last batch of removals of the old event system, this was left orphaned.

I don't think deprecation is necessary as it is very unlikely that anyone is using this interface (and if they are, they can easily declare their own as we do not care about it anywhere).

## Checklist

## Breaking Change?

### Migration instructions

This interface is no longer necessary to interop with Flame in any way shape or form.
If you wish to use it internally in your game you can just declare it yourself.


.github/.cspell/words_dictionary.txt
```diff
@@ -34,5 +34,3 @@ tappable
 tappables
 toolset
 underutilize
-unproject
-unscale
```
luanpotter, line 38: my advanced spell check linter with unused validation is working :relieved:

spydon (approved):
Lgtm, I also think it makes sense to remove without deprecation

PR #2956, opened 2024-01-04
feat: Notifier for changing current sprite/animation in group components

spydon:
# Description

With these new notifiers you can get notified when the current animation or sprite changes in the group components.
The value notifier is not initialized unless the user tries to use it, so it doesn't carry any extra weight on the components.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/components/sprite_animation_group_component.dart
+++ b/packages/flame/lib/src/components/sprite_animation_group_component.dart
@@ -3,6 +3,7 @@ import 'dart:ui';
 import 'package:flame/components.dart';
 import 'package:flame/src/effects/provider_interfaces.dart';
 import 'package:flame/src/sprite_animation_ticker.dart';
+import 'package:flutter/foundation.dart';
 import 'package:meta/meta.dart';
 
 export '../sprite_animation.dart';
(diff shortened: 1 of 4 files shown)


packages/flame/lib/src/components/sprite_animation_group_component.dart
```diff
@@ -3,6 +3,7 @@ import 'dart:ui';
 import 'package:flame/components.dart';
 import 'package:flame/src/effects/provider_interfaces.dart';
 import 'package:flame/src/sprite_animation_ticker.dart';
+import 'package:flutter/foundation.dart';
 import 'package:meta/meta.dart';
```
luanpotter, line 8: nice!

luanpotter (approved):
LGTM bar the nice github analyzer notes

Also wondering whether we should have a standard/philosophy on where to add these listeners or not. theoretically we could add them on every single field, but I suspect we want to strike some sort of balance.

spydon:
> LGTM bar the nice github analyzer notes
> 
> Also wondering whether we should have a standard/philosophy on where to add these listeners or not. theoretically we could add them on every single field, but I suspect we want to strike some sort of balance.

Yeah, I think we should apply YAGNI, and now someone asked for it on Discord and I estimated this to be a legit usecase.

PR #2958, opened 2024-01-04
fix: Set hitbox `debugColor` to yellow

spydon:
# Description

To easily be able to spot whether a it is a hitbox that is being drawn in `debugMode = true` it is now yellow instead of purple like the other debug boxes.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
+++ b/packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
@@ -1,5 +1,6 @@
 import 'package:flame/collisions.dart';
 import 'package:flame/components.dart';
+import 'package:flame/extensions.dart';
 import 'package:flame/game.dart';
 import 'package:flame/geometry.dart';
 import 'package:flame/src/geometry/shape_intersections.dart'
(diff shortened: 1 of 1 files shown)


packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
```diff
@@ -1,5 +1,6 @@
 import 'package:flame/collisions.dart';
 import 'package:flame/components.dart';
+import 'package:flame/extensions.dart';
```
erickzanardo, line 3: Why import extensions here? 😅

erickzanardo (approved):
A question but approved

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
spydon: It's instead of importing `dart:ui`, shouldn't really matter. :)

erickzanardo: Def not, was just curious haha

PR #2964, opened 2024-01-07
fix: Change return type of RiverpodComponentMixin.onLoad to FutureOr<void>

### LINKED ISSUE #2961: Change `RiverpodComponentMixin.onLoad` to return `FutureOr<void>`
# What could be improved

The function signature of `RiverpodComponentMixin.onLoad` could be changed from `void onLoad()` to `FutureOr<void> onLoad`.

# Why should this be improved

For example, when trying to mixin `RiverpodComponentMixin` to a `TextBoxComponent`, a function signature mismatch occurs, as `RiverpodComponentMixin.onLoad` has signature `void onLoad` and `TextBoxComponent.onLoad` has signature `Future<void> onLoad`.
I know i could just make a separate _riverpod aware_ component that extends `Component` (which in fact has the signature `FutureOr<void> onLoad`) (such as in the [documentation](https://docs.flame-engine.org/latest/bridge_packages/flame_riverpod/component.html#riverpodcomponentmixin)), but i would like to have the provider logic inside the actual component. 
Here is some code to see what i am talking about.
```dart
class MyTextBox extends TextBoxComponent with RiverpodComponentMixin { // ERROR: can't mixin  (signature mismatch)
  @override
  void onMount() {
    addToGameWidgetBuild(() {
      ref.listen(myProvider, (previous, next) {
        // do some things such as updating text contents
      });
    });
    super.onMount();
  }
}
```

# Any risks?

I don't know if having the `RiverpodComponentMixin` attached to an actual component other than the plain `Component` class poses any problems, but i don't think so.
Theoretically this is an API change, since the signature changes. However, this signature change will not break any code, because the way `FutureOr<T>` is defined.

# More information

i could do a PR if this is considered useful. But for such a small change i wouldn't be bothered if you would change it yourself real quick.

markvideon:
# Description

This PR changes the return type of RiverpodComponentMixin.onLoad to match that of Component.onLoad. 

This resolves #2961 which describes an error preventing use of RiverpodComponentMixin on Component subclasses that return a Future in their implementation of onLoad due to a signature mismatch. 

## Checklist

## Breaking Change?

## Related Issues

Closes #2961.

### DIFF
--- a/.github/.cspell/gamedev_dictionary.txt
+++ b/.github/.cspell/gamedev_dictionary.txt
@@ -6,7 +6,6 @@ AHSV # alpha hue saturation value
 antialiasing # set of techniques to combat the problems of aliasing
 ARGB # alpha red green blue
 arities # plural of arity
-arity # number of parameters a function takes
 autofocus # auto focus event
 backgrounded # moving the app to the background
 backgrounding # moving the app to the background
--- a/packages/flame_riverpod/lib/src/consumer.dart
+++ b/packages/flame_riverpod/lib/src/consumer.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'package:flame/components.dart';
 import 'package:flame/game.dart';
 import 'package:flame_riverpod/src/widget.dart';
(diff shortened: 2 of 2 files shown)


spydon (approved):

spydon:
@markvideon I have no idea why the spellchecker triggered for `arity` not being used now, could you remove it from the game dev dictionary?

PR #2968, opened 2024-01-09
feat: Add a hitboxFilter argument to raycast()

### LINKED ISSUE #2966: Let raycast dynamically ignore some hitboxes
# Problem to solve
Currently, when calling `CollisionDetection.raycast()`, you can only ignore hitboxes with a pre-existing list of hitboxes.

```dart
  /// ...
  ///
  /// [ignoreHitboxes] can be used if you want to ignore certain hitboxes, i.e.
  /// the rays will go straight through them. For example the hitbox of the
  /// component that you might be casting the rays from.
  RaycastResult<T>? raycast(
    Ray2 ray, {
    double? maxDistance,
    List<T>? ignoreHitboxes,
    RaycastResult<T>? out,
  });
```

This means that we have to have (or construct) this list in advance. This is prohibitively expensive for games with more complex systems.

Imagine, for example, that a game entity in an top-down 2D game is trying to check whether it sees an enemy. If both the enemy and the entity are on the ground, then the raycast cannot ignore buildings. But if one of them is in the air, it can. (At least the buildings closer to the flying entity.) Similarly, a line of fire algorithm can ignore enemy units (i.e. I don't care that an enemy unit is partly blocking my line of fire — worst case scenario, I hit it instead of my target) but it shouldn't ignore friendlies.

Constructing the lists anew for every raycast could be very expensive if the raycast is done often. And managing a big variety of pre-populated lists is a recipe for bugs and memory leaks.

# Proposal

Apart from `List<T>? ignoreHitboxes`, the `raycast()` method should also accept a callback, something like:

```dart
  RaycastResult<T>? raycast(
    Ray2 ray, {
    double? maxDistance,
    List<T>? ignoreHitboxes,
    bool Function(T other)? filter,  // <-- added this
    RaycastResult<T>? out,
  });
```

The new `filter` callback is then used alongside of `ignoreHitboxes` like so:

```dart
    // ...
    for (final item in items) {
      if (ignoreHitboxes?.contains(item) ?? false) {
        continue;
      }
      if (!filter?.call(item) ?? false) {
        // When filter(item) returns `false`, we skip this item.
        continue;
      }
      if (!item.aabb.intersectsWithAabb2(_temporaryRayAabb)) {
        continue;
      }
      // ...
```

The naming could probably be improved. Something like `whereHitbox` instead of `filter`? I'd like to try staying positive (as in, when the callback returns `true`, we process that hitbox, otherwise no.

**Sidenote:** We should probably test performance. Even now, I'm not sure whether going through a `List` of `ignoreHitboxes` for every item is conducive to good performance. A `Set` would be better? Or maybe we can phase out `ignoreHitboxes` altogether in lieu of the callback? Of course, we'd need to measure first. It's also possible that `aabb.intersectsWithAabb2` is generally faster than checking `ignoreHitboxes` or running a callback, so that it should go first.

If the performance is too bad, we could do what Unity does (see below). Every component has something like a [`BitField`](https://api.flutter.dev/flutter/foundation/BitField-class.html) attached that's similar in practice to Unity's `layers` bit field. All tight loops (such as `raycast` or collision detection) then accept a bit mask.

# More information

- FWIW, Unity uses the following approaches for raycasts:
  - When calling the raycast, you can specify a bitmask for the [layers](https://docs.unity3d.com/Manual/Layers.html) you want to ignore. That is really low level, but is clearly very fast. (Every GameObject has a layer bit set.)
    - `Physics.Raycast(transform.position, transform.forward, out hit, 3f, layerMask)`
  - Some layers can be ignored altogether (every GameObject in those layers is ignored in every raycast call.)

filiph:
# Description

Adds a new argument callback to all `raycast*` methods that lets the user ignore hitboxes dynamically. The callback is called with every prospective hitbox, and the hitbox is only considered when the callback returns `true`.

This is faster, in the general case, than the current `ignoreHitboxes` aproach. And it lets the developer have dynamic rules about what the rays collide with. (For example, a line-of-fire raycast should take friendlies into account, but could easily ignore other enemies that get in the way.)

## Checklist

## Breaking Change?

For people who just _use_ raycasts, this is not a breaking change.

People who extend `CollisionDetection` will need to support the new argument. I'd say this is pretty rare but not unheard of.

If you have your own custom collision detection logic (a class implementing or extending Flame's `CollisionDetection`), you'll have to update its three `ray*` methods. Add the new argument to the base `raycast()` method and make sure that the ray actually only collides with hitboxes that return `true` from `hitboxFilter()`. The other two methods (`raycastAll()` and `raytrace()`) will, in most cases, just call `raycast()`, so just make sure you pass the argument.

## Related Issues

Closes #2966

### DIFF
--- a/packages/flame/lib/src/collisions/collision_detection.dart
+++ b/packages/flame/lib/src/collisions/collision_detection.dart
@@ -101,15 +101,22 @@ abstract class CollisionDetection<T extends Hitbox<T>,
   /// [maxDistance] can be provided to limit the raycast to only return hits
   /// within this distance from the ray origin.
   ///
-  /// [ignoreHitboxes] can be used if you want to ignore certain hitboxes, i.e.
-  /// the rays will go straight through them. For example the hitbox of the
-  /// component that you might be casting the rays from.
+  /// You can provide a [hitboxFilter] callback to define which hitboxes
+  /// to consider and which to ignore. This callback will be called with
+  /// every prospective hitbox, and only if the callback returns `true`
+  /// will the hitbox be considered. Otherwise, the ray will go straight
+  /// through it. One common use case is ignoring the component that is
+  /// shooting the ray.
+  ///
+  /// If you have a list of hitboxes to ignore in advance,
+  /// you can provide them via the [ignoreHitboxes] argument.
   ///
   /// If [out] is provided that object will be modified and returned with the
   /// result.
   RaycastResult<T>? raycast(
     Ray2 ray, {
     double? maxDistance,
+    bool Function(T candidate)? hitboxFilter,
     List<T>? ignoreHitboxes,
     RaycastResult<T>? out,
   });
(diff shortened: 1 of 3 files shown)


spydon (approved):
Lgtm!

filiph:
Thanks! Just a reminder that I don't have commit rights here so you'll have to land this yourself when you're happy.

spydon:
> Thanks! Just a reminder that I don't have commit rights here so you'll have to land this yourself when you're happy.

Yep, only Blue Fire has merging rights here.

PR #2979, opened 2024-01-13
fix: Updated PolygonComponent.containsPoint to account for concave polygons

davidjan3:
# Description

Previously, the `PolygonComponent.containsPoint()` and `.containsLocalPoint()` functions consisted of duplicate code that checked whether a given point lies within a convex polygon. They didn't function properly with concave polygons.

I created a new `_containsPoint()` function that is called from both functions to reduce redundancies. This new function uses a different approach to figure out whether a point lies within a polygon, which should also work for concave polygons, or even polygons with holes. The algorithm is vaguely explained within code comments, and is visualized in this post: https://stackoverflow.com/a/218081/5008997

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/geometry/polygon_component.dart
+++ b/packages/flame/lib/src/geometry/polygon_component.dart
@@ -207,30 +207,44 @@ class PolygonComponent extends ShapeComponent {
       return false;
     }
 
-    // Count the amount of edges crossed by a straight horizontal line going left from the point
-    int count = 0;
+    // Count the amount of edges crossed by going left from the point
+    var count = 0;
     for (var i = 0; i < vertices.length; i++) {
       final edge = getEdge(i, vertices: vertices);
 
-      // If the edge is entirely to the right, above or below the point then it can't be crossed
+      // Skip if the edge is entirely to the right, above or below the point
       if (edge.from.x > point.x && edge.to.x > point.x ||
           min(edge.from.y, edge.to.y) > point.y ||
           max(edge.from.y, edge.to.y) < point.y) {
         continue;
       }
 
-      // If the edge is crossed by the line, increase the count
-      if (edge.from.y == edge.to.y ||
-          ((point.y - edge.from.y) * (edge.to.x - edge.from.x)) /
-                      (edge.to.y - edge.from.y) +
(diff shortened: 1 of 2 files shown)


spydon (commented):
Looks good after a quick look, how's the performance impact?

davidjan3:
Just stress-tested the performance within my current project.
Calling `containsPoint(vertex)` for 4 vertices on 10,000 randomly generated polygons with 10-24 vertices each:
Old method avg: ~370ms
New method avg: ~530ms

So it takes about 43% longer on average.

However, by replacing 
```dart
final edge = getEdge(i, vertices: vertices);
```
with
```dart
final from = vertices[i];
final to = vertices[(i + 1) % vertices.length];
``` 

It goes down to ~350ms and still survives every test. 
Though this also speeds up the old version to ~320ms.

It's important to note that in my test, the old function returns false almost all the time. And looking at the code, returning false can happen much faster than checking all edges before returning true.

spydon:
@davidjan3 can you try with a bit more vertices, like 10?

davidjan3:
All polygons had 10-24 vertices. My test just ran the `.containsPoint(vertex)` method 4 times per polygon, to see if all 4 corners of a rectangle lie within it. I can share my code sample after work.

spydon:
> Otherwise I'd just add a couple tests and hope for approval.

Sounds good, seems like a reasonable performance trade-off.

davidjan3:
@spydon So would you prefer keeping the `getEdge()` function for consistency, or should I push this?:


> However, by replacing
> 
> ```dart
> final edge = getEdge(i, vertices: vertices);
> ```
> 
> with
> 
> ```dart
> final from = vertices[i];
> final to = vertices[(i + 1) % vertices.length];
> ```
> 
> It goes down to ~350ms and still survives every test.

spydon:
@davidjan3 sounds like it's definitely worth changing it then! :)

davidjan3:
@spydon Added said optimization, and a little test :)

spydon (approved):
Lgtm! Sorry for the delay, I've been away skiing in the alps the last week :sweat_smile:

PR #2988, opened 2024-01-21
feat: Make `Component.key` public

### LINKED ISSUE #2985: getter for ComponentKey on Component itself
# Problem to solve
Currently im trying to implement collision and need an id for the components. Since there is already the componentkey it would be pretty nice if I could just reuse it. But there is no getter for the key.

# Proposal
Add a getter for the ``ComponentKey _key`` to ``Component``.

# More information
No.

spydon:
# Description

Since the user might want to use the `ComponentKey` after it has been assigned it should be made public.

It also updates `Consumer.key` to `Consumer.widgetKey` in flame_riverpod to avoid a name clash. (Slightly breaking)

## Checklist

## Breaking Change?

### Migration instructions

If you are using `consumer.key` from flame_riverpod you have to now use `consumer.widgetKey` instead.

## Related Issues

Closes #2985

### DIFF
--- a/packages/flame/lib/src/components/core/component.dart
+++ b/packages/flame/lib/src/components/core/component.dart
@@ -69,9 +69,8 @@ class Component {
   Component({
     Iterable<Component>? children,
     int? priority,
-    ComponentKey? key,
-  })  : _priority = priority ?? 0,
-        _key = key {
+    this.key,
+  }) : _priority = priority ?? 0 {
     if (children != null) {
       addAll(children);
     }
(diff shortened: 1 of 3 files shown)


packages/flame/lib/src/components/core/component.dart
```diff
@@ -989,7 +988,7 @@ class Component {
   /// A key that can be used to identify this component in the tree.
   ///
   /// It can be used to retrieve this component from anywhere in the tree.
-  final ComponentKey? _key;
+  final ComponentKey? key;
```
luanpotter, line 991: I wonder if we should call it componentKey for the same reason you have exposed; it is possible other people are using key for something else on their components, and this would clash. no strong opinion though

luanpotter (approved):
LGTM

packages/flame/lib/src/components/core/component.dart
spydon: I think it is rare enough, I've never seen it anywhere else except in the very new flame_riverpod instance

PR #2994, opened 2024-01-22
feat: Bumped integral_isolates package for flame_isolate

lohnn:
# Description

Updated package dependencies for flame_isolate.
Added support for TailoredIsolate and computeStream (called isolateComputeStream in flame_isolate).

Updated the example package with new sprites created by yours truly, and updated the license for the package, now that there is no assets with copy-left CC license remaining.

## Checklist

## Breaking Change?

### Migration instructions

Computation function is renamed from `isolate` to `isolateCompute`.
There are future breaking changes that are, as of now, mitigated with a `@Deprecated` annotation with information regarding what function to use instead.


packages/flame_isolate/lib/flame_isolate.dart
```diff
@@ -6,8 +6,9 @@ export 'package:integral_isolates/integral_isolates.dart';
 /// Mixin on [Component] that holds an instance of a long running isolate using
 /// the library integral_isolates.
 ///
-/// Using isolate is done by just running [isolate] function the same way you
-/// would run Flutter's compute function.
+/// Using isolate is done by just running [isolateCompute] function the same way
```
spydon, line 9:
```suggestion
/// Using the isolate is done by just running [isolateCompute] function the same way
```

packages/flame_isolate/lib/flame_isolate.dart
```diff
@@ -50,12 +51,41 @@ mixin FlameIsolate on Component implements IsolateGetter {
     _isolate = null;
   }
 
-  @override
+  /// Deprecated in favor of [isolateCompute].
+  @Deprecated('Use isolateCompute instead.')
```
spydon, line 55: Write which version it will be removed in, v0.6.0 I guess?

spydon (approved):
Lgtm! Should be a `feat` since it exposes new methods.

packages/flame_isolate/lib/flame_isolate.dart
lohnn: Well, it is already removed in the underlying package, so this is a soft deprecation that can be removed in the flame_isolate package whenever.

Hence a bit up to you and the best practices of Flame :)

spydon: v0.7.0 then, since this one will be 0.6.0 now due to the `feat` :)

spydon (approved):

PR #2997, opened 2024-01-23
fix: CameraComponent no longer throws Concurrent modification on stop

lohnn:
# Description

When camera is following a component and trying to follow another target I got an exception:
```
dart:core                                              Iterable.forEach
package:flame/src/camera/camera_component.dart 326:25  CameraComponent.stop
package:flame/src/camera/camera_component.dart 309:5   CameraComponent.follow
test/camera/camera_component_test.dart 69:14           main.<fn>.<fn>
package:flame_test/src/test_flame_game.dart 80:21      testWithGame.<fn>

Concurrent modification during iteration: _Set len:0.
```

Copying viewfinder children before iterating through it and removing child from parent does the trick, as we are not iterating through the same list as we are removing items from, but rather a copy of it.

I tried to use camera from `game` provided in the test but it is AFAIK not mounted and will hence not queue modifications (adds and removes). Hence I create a new camera and mount it.

## Checklist

## Breaking Change?

## Related Issues

Replace or remove this text.

### DIFF
--- a/packages/flame/test/camera/camera_component_test.dart
+++ b/packages/flame/test/camera/camera_component_test.dart
@@ -61,10 +61,11 @@ void main() {
     });
 
     testWithFlameGame('camera should be able to retarget follow', (game) async {
-      final world = World()..addToParent(game);
-      final camera = CameraComponent(world: world)..addToParent(game);
-      final player = PositionComponent()..addToParent(world);
-      final player2 = PositionComponent()..addToParent(world);
+      // Creating new camera as the one included with game is not mounted and
+      // will therefore not be queued.
+      final camera = CameraComponent(world: game.world)..addToParent(game);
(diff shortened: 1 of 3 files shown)


packages/flame/test/camera/camera_component_test.dart
```diff
@@ -60,6 +60,19 @@ void main() {
       }
     });
 
+    testWithFlameGame('camera should be able to retarget follow', (game) async {
+      final world = World()..addToParent(game);
+      final camera = CameraComponent(world: world)..addToParent(game);
```
spydon, line 65: No need to create a new world and camera here, you can use `game.camera` and `game.world`.

spydon (approved):
Lgtm, just a minor comment on the test

spydon (approved):

PR #3007, opened 2024-01-27
fix: Lifecycle completers to be called for FlameGame

### LINKED ISSUE #3003: Make `loaded`,`mounted` and `removed` Futures on FlameGame instance working
# Problem to solve

Knowing when the game has finished loading/mounting or was removed.
Although one can argue the `removed` Future isn't really useful, so we can omit this maybe.

# Proposal

As `Component` already has the `mounted`,`loaded`,`removed` Futures with their respective Completers, we just need to access the completers `_mountCompleter`,`_loadCompleter`,`_removeCompleter` and mark them as completed. Right now these member variables are private and cannot be accessed by subclasses, so maybe we can introduce a protected getter for them.

For the mount case we can complete `_mountCompleter` in `FlameGame.mount`, after the `super.mount()` call.
For the loading case we can complete `_loadCompleter` in `GameWidgetState.loaderFuture`, before the `game.mount()` call.
One note: This way both `_mountCompleter` and `_loadCompleter` would complete before the mounting/loading bit is set through `setMounted()`, which is probably undesirable.
I think it would be best to make `_finishLoading` protected and introduce a similar function `_finishMounting`, which does the same things (set state bit, call completer). Then we can use these methods inside `FlameGame` to propagate loading/mounting completion events, and it would handle the state bits as well. The `setMounted` then is a bit unnecessary, as loading/mounting bits were already set (by `_finishdLoading`/`_finishMounting`), but `setMounted` is only used by `FlameGame`, so we can maybe change the behaviour or API there a bit.
We maybe could use the new `_finishMounting` function inside `Component._mount` as well, altough the order of `_clearMountingBit` and `_setMountedBit` are reversed there. Is there any reason why `_mount` does this:
```dart
// inside Component._mount:
    onMount();
    _setMountedBit();
    _mountCompleter?.complete();
    _mountCompleter = null;
    _parent!.children.add(this);
    _reAddChildren();
    _parent!.onChildrenChanged(this, ChildrenChangeType.added);
    _clearMountingBit();
```
instead of this:
```dart
// inside Component._mount:
    onMount();
    // clear mounting bit here instead of at the end
    _clearMountingBit();
    _setMountedBit();
    _mountCompleter?.complete();
    _mountCompleter = null;
    _parent!.children.add(this);
    _reAddChildren();
    _parent!.onChildrenChanged(this, ChildrenChangeType.added);
    
```

For the remove case i have no proposal yet, but it should be similar.

# More information

The relevant methods are:

```dart
class Component {
  // [...]
  void _finishLoading() {
    _clearLoadingBit();
    _setLoadedBit();
    _loadCompleter?.complete();
    _loadCompleter = null;
  }
}
```
```dart
class GameWidgetState<T extends Game> extends State<GameWidget<T>> {
   // [...]
  Future<void> get loaderFuture => _loaderFuture ??= (() async {
        final game = currentGame;
        assert(game.hasLayout);
        final onLoad = game.onLoadFuture;
        if (onLoad != null) {
          await onLoad;
        }
        game.mount();
        if (!game.paused) {
          game.update(0);
        }
      })();
}
```
```dart
class FlameGame<W extends World> extends ComponentTreeRoot with Game implements ReadOnlySizeProvider {
  // [...]
  @override
  @internal
  void mount() {
    super.mount();
    setMounted();
  }
}
```

I can work on a PR for this.

This would unblock #2971.

spydon:
# Description

Previously the `mounted`, `loaded` and `removed` completers for `FlameGame` weren't called since `FlameGame` doesn't go through the normal component lifecycle flow, this PR adds so that the completers are completed properly.

## Checklist

## Breaking Change?

## Related Issues

Closes #3003 


packages/flame/lib/src/components/core/component.dart
```diff
@@ -927,13 +926,34 @@ class Component {
...
+  void setLoaded() {
     _setLoadedBit();
+    _loadCompleter?.complete();
+    _loadCompleter = null;
+  }
+
+  /// Used by the [FlameGame] to set the mounted state of the component, since
+  /// the game isn't going through the whole normal component life cycle.
+  @internal
+  void setMounted() {
```
erickzanardo, line 941: This method already existed, so I wonder if the docs here is correct? Other components that are not a flame game don't call it?

packages/flame/lib/src/game/game_widget/game_widget.dart
```diff
@@ -191,10 +191,7 @@ class GameWidgetState<T extends Game> extends State<GameWidget<T>> {
   Future<void> get loaderFuture => _loaderFuture ??= (() async {
         final game = currentGame;
         assert(game.hasLayout);
-        final onLoad = game.onLoadFuture;
```
erickzanardo, line 194: Is it safe to remove this call and rely only on super.load?

Sorry I am on mobile so would took me a while to check on the source of that method

erickzanardo (approved):
Couple of doubts. But LGTM already

packages/flame/lib/src/components/core/component.dart
spydon: It was previously only called by `FlameGame` too, but the completer wasn't completed here previously

packages/flame/lib/src/game/game_widget/game_widget.dart
spydon: It's basically just a name change

erickzanardo: Hum, but is the onLoad method gone????

spydon: Haha no, that would be a massive breaking change 😄
Only the internal `onLoadFeature` getter has been renamed to `load` and made into a method.

erickzanardo: Got ya! Phew, that massive breaking change was my concern haha

erickzanardo (approved):

PR #3008, opened 2024-01-28
feat: Add support for base64 encoded images to be manually added to Images cache.

erickzanardo:
# Description

Adds a method so base64 encoded images can be manually added to the Images cache.

## Checklist

## Breaking Change?

## Related Issues

Replace or remove this text.


packages/flame/lib/src/cache/images.dart
```diff
@@ -49,6 +49,14 @@ class Images {
     _assets[name] = _ImageAsset.fromImage(image);
   }
 
+  /// Transform the base64 encoded image into an [Image] and adds it into the
+  /// cache.
+  Future<void> addBase64EncodedImage(String name, String base64Data) async {
```
spydon, line 54:
```suggestion
  Future<void> addFromBase64Data(String name, String base64Data) async {
```
Maybe? That it's an image should already be implicit from the cache name

spydon (approved):

spydon:
@erickzanardo I just stumbled upon this method which is already in the `Images` class:
image
Doesn't that do the same thing, except with the missing dispose?

erickzanardo:
@spydon not really, that method will fetch from cache if the image is already there, the method that I introduced will always force the cache to be updated.

spydon:
> @spydon not really, that method will fetch from cache if the image is already there, the method that I introduced will always force the cache to be updated. 

Ah, alright, it seems like it should just have been a flag to this method instead of a new method then?

We should really standardize how the caches work for V2, it feels quite unstructured now.

spydon:
> Yeah, agree on the flag, that was what I thought at first, but I preferred to follow the other method `add` that already exists to keep the consistency.

Makes sense, I'll put it in the V2 list that we should look through the cache classes :)

erickzanardo:
Sounds good!!!!

PR #3010, opened 2024-01-29
feat: Allow ComponentRef access in RiverpodGameMixin

### LINKED ISSUE #2971: Allow ComponentRef access in RiverpodGameMixin
# What could be improved

I was wondering if it is possible to allow `RiverpodComponentMixin` and `RiverpodGameMixin` to be a mixin on the same class. Currently this is not possible due to the same variable name being used in both mixins (`_onBuildCallbacks`).
One example on how this could be used then:
```dart
class MyGame extends FlameGame with RiverpodGameMixin, RiverpodComponentMixin {
  @override
  void onMount() {
    addToGameWidgetBuild(() {
      ref.listen(pausedStateStreamProvider, (p0, p1) {
        if (p1.hasValue) {
         paused = p1.value!;
        }
      });
    });
    super.onMount();
  }
}
```

# Why should this be improved

It would make it easier to react upon provider changes on the Game instance level, for example to pause the game or do other things where access to the game instance is needed. Currently, a separate Component, which uses `findGame()` or `HasGameRef`, is needed for these use cases. Nevertheless, maybe a separate component is in fact the cleaner and recommended solution, so what do you think about this? 

# Any risks?

As far as i know, the mixins don't interfere with each other. However, based on the mixin names it might be a bit irritating to have both `RiverpodGameMixin` and `RiverpodComponentMixin` on one `FlameGame` class, but as `FlameGame` also extends `Component`, i think this is justified.

# More information

I can write the PR if this is considered useful.

tibotix:
# Description

This PR allows `ComponentRef` access through the `RiverpodGameMixin`.

## Checklist

## Breaking Change?

## Related Issues

Closes #2971 


packages/flame_riverpod/lib/src/consumer.dart
```diff
@@ -139,8 +139,30 @@ mixin RiverpodGameMixin<W extends World> on FlameGame<W> {
   /// Used to facilitate [Component] access to the [ProviderContainer].
   GlobalKey<RiverpodAwareGameWidgetState>? widgetKey;
 
+  final ComponentRef ref = ComponentRef(game: null);
   final List<void Function()> _onBuildCallbacks = [];
 
+  /// Adds a callback method to be invoked in the build method of
+  /// [RiverpodAwareGameWidgetState].
+  void addToGameWidgetBuild(Function() cb) {
```
spydon, line 147: Hmm, not too fond of this one, it feels like this should be possible to handle under the hood for the user?
What do you think @markvideon?

markvideon: @spydon there doesn't appear to be any mechanism for removing callbacks from _onBuildCallbacks in RiverpodGameMixin at the moment

packages/flame_riverpod/test/widget_test.dart
```diff
@@ -136,4 +146,46 @@ void main() {
...
+
+    // Replace the widget tree so that the GameWidget gets disposed
+    await widgetTester.pumpWidget(Container());
+    await widgetTester.pumpAndSettle();
+
+    // Expect that the component has been removed from the game.
+    expect(game.isAttached, false);
+
+    // Expect that the GameWidget is no longer listening to
+    // numberProvider as the watching component has been removed.
```
markvideon, line 188: I think the wording of this should be tweaked to reflect the comparison being made in the line.  

E.g. 

> Expect that the key no longer has access to a state.

packages/flame_riverpod/lib/src/consumer.dart
```diff
@@ -139,8 +139,30 @@ mixin RiverpodGameMixin<W extends World> on FlameGame<W> {
...
+    _onBuildCallbacks.add(cb);
+  }
+
+  @override
+  void onMount() {
+    super.onMount();
+    mounted.whenComplete(() {
+      widgetKey!.currentState!.forceBuild();
```
markvideon, line 155: Can we add callbacks to the game widget before we call forceBuild?

packages/flame_riverpod/lib/src/consumer.dart
```diff
@@ -139,8 +139,30 @@ mixin RiverpodGameMixin<W extends World> on FlameGame<W> {
...
+    });
+  }
+
```
markvideon, line 158: Can we define onRemove to remove callbacks from the game widget?

markvideon (commented):
See individual comments

spydon:
@tibotix did you see the comments on here? :)

tibotix:
Yes I saw, them. Thank you for the reminder :) I’m a little busy right now due to upcoming exams, I will come back to this after a week.

packages/flame_riverpod/lib/src/consumer.dart
tibotix: I think we don’t need this, as dependencies are already disposed in `RiverpodAwareGameWidget.dispose`. And afaik manual removals of callbacks are not supported in ComponentMixin either.

packages/flame_riverpod/lib/src/consumer.dart
tibotix: i think so. This is whats done in RiverpodComponentMixin as well. first the user can add callbacks to `_onBuildCallbacks` through `addToGameWidgetBuild`, and then we call forceBuild to have the callbacks be executed. Maybe I'm not understanding your question correct.

packages/flame_riverpod/lib/src/consumer.dart
tibotix: The problem here is that some ComponentRef methods, like read or watch, return a value. So we need to wrap these calls (like read or watch) in a callback, as they are by design executed later whenever the game builds. So something like
```dart
void onMount() {
    final myValue = ref.watch(myProvider);
}
```
will not work as the actual watching of myProvider happens later when the game builds. As to the exact reason why we do this behaviour - i don't really know. Maybe @markvideon can explain it better.

markvideon: > The problem here is that some ComponentRef methods, like read or watch, return a value. So we need to wrap these calls (like read or watch) in a callback, as they are by design executed later whenever the game builds. So something like
> 
> ```dart
> void onMount() {
>     final myValue = ref.watch(myProvider);
> }
> ```
> 
> will not work as the actual watching of myProvider happens later when the game builds. As to the exact reason why we do this behaviour - i don't really know. Maybe @markvideon can explain it better.

If you take a look at the implementation of watch, doing this _will_ work in the sense that it will create a subscription and obtain a value automatically, but it _won't_ work in the sense that the subscription can not be expected to persist in the way that a user might expect if they're coming from flutter_riverpod. In flutter_riverpod, all current subscriptions are terminated before an application's ConsumerWidget/ConsumerStatefulWidget build code is executed - which, if it contains calls to watch or listen operations, re-creates the subscriptions.

packages/flame_riverpod/lib/src/consumer.dart
markvideon: > I think we don’t need this, as dependencies are already disposed in `RiverpodAwareGameWidget.dispose`. And afaik manual removals of callbacks are not supported in ComponentMixin either.

The use case that I have in the back of my mind is a swapping out the game associated with a RiverpodGameWidget. In this case I think we would want to remove the various subscriptions associated with the former game, and add the various subscriptions associated with the new game. E.g. imagine that our Flutter app is an implementation of the SNES Classic, which features a small library of different games for the user to select from, with only one being active at any given time. 

In any case, I don't think that we need to ensure that this functionality is supported as of this PR, more of a longer-term goal.

spydon (approved):

PR #3016, opened 2024-02-04
docs: Fix positioning of git commit message exclusion blocks in the PR template

luanpotter:
# Description

Fix positioning of git commit message exclusion blocks in the PR template. I found it a bit confusing as I think what we want to encourage is that the migration instructions go into the git message, but rn there is just a "Replace or remove this text." lost after the related issues. I don't think we want to include related issues? So it seems just out of order.

## Checklist

## Breaking Change?


spydon (approved):
Maybe we should have and explanation of what should go inbetween those tags instead, before the start of the tag? So many people forget to remove the "Replace or remove" text.

luanpotter:
This is the explanation, no?

> If the PR is breaking, uncomment this header and add instructions for how to migrate from the

Or you mean replace `Replace or remove` with that message?

spydon:
@luanpotter I mean something like this:
```md
Add the migration instruction in-between the two following tags:
<!-- End of exclude from commit message -->
<!-- Exclude from commit message -->
```

luanpotter:
Fixed

spydon (approved):

PR #3043, opened 2024-02-17
feat: Add `HasPerformanceTracker` mixin on `Game`

ufrshubham:
# Description

This PR adds a `HasPerformanceTracker` mixin on `Game`. This mixin will allow for tracking the update and render time of the game.

## Checklist

## Breaking Change?

## Related Issues

NA


packages/flame/lib/src/game/game_render_box.dart
```diff
@@ -118,15 +128,31 @@ class GameRenderBox extends RenderBox with WidgetsBindingObserver {
     if (!attached) {
       return;
     }
+
+    _stopwatch.reset();
+    _stopwatch.start();
```
spydon, line 133: Hmm, I wonder if this could be implemented so that it doesn't run by default, it's probably not very heavy, but still?

ufrshubham: Sure, I can add a boolean called `trackPerformance` to control this.

spydon: The `if` will probably be more expensive than actually running it, so an if is probably not an option :sweat_smile:
I'm not sure if there is any solution, since it can't be made in to a mixin or such...

ufrshubham: One way I can think of is to add `ValueChange` callbacks for update and render. Then based on `trackPerformance` we can swap the callbacks with the normal or tracked versions.

```
```

spydon: Then we also don't add code in the core which would be preferable

ufrshubham: Okay, let me try that 👍🏼

ufrshubham: That works exactly the same! Thanks for suggesting the mixin approach 🙂

spydon: Super, looks good! 🙂

spydon (approved):
Lgtm!

PR #3050, opened 2024-02-25
fix: Check for removing state while adding a child

ufrshubham:
# Description

Adding a child component to a parent which is in `removing` state, caused the lifecycle processing to go into a cyclic dependency when the parent is re-added. It happens because, while processing the lifecycle events, child's add event causes itself and the parent to get added to the blocked list. As a result of this, when the parent's add event is processed next, it gets skipped due to being in the blocked list.

This PR makes sure that the child does not get enqueued when parent is about to be removed.

## Checklist

## Breaking Change?

## Related Issues

NA

### DIFF
--- a/packages/flame/lib/src/components/core/component.dart
+++ b/packages/flame/lib/src/components/core/component.dart
@@ -594,7 +594,7 @@ class Component {
         _clearRemovingBit();
       }
       game.enqueueMove(child, this);
-    } else if (isMounted && !child.isMounted) {
+    } else if (isMounted && !isRemoving && !child.isMounted) {
       child._parent = this;
       game.enqueueAdd(child, this);
     } else {
--- a/packages/flame/test/components/component_test.dart
+++ b/packages/flame/test/components/component_test.dart
@@ -556,6 +556,27 @@ void main() {
           expect(wrapper.contains(child), true);
         },
       );
+
+      testWithFlameGame('when parent is in removing state', (game) async {
+        final parent = Component();
+        final child = Component();
+
+        await game.add(parent);
+        await game.ready();
+
+        // Remove the parent and add the child in the same tick.
+        parent.removeFromParent();
+        await parent.add(child);
+
+        // Timeout is added because processLifecycleEvents of ComponentTreeRoot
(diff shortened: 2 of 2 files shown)


denisgl7:
Thanks!

spydon (commented):
Can you describe in the PR description what the PR solves? 

What happens when enqueeing an add when the parent is removing? Shouldn't it be reconciled at mounting still? 🤔

ufrshubham:
> What happens when enqueeing an add when the parent is removing? Shouldn't it be reconciled at mounting still? 🤔

Due to the sequence of the queue, the child's `handleLifecycleEventAdd` caused itself and its parent, both to get added to the `blocked` set here
https://github.com/flame-engine/flame/blob/892052b99a21a8e371c4163e1e1918fd187c6e11/packages/flame/lib/src/components/core/component_tree_root.dart#L109-L110

So in the next iteration when the parent's re-add event was encountered, it was getting skipped here:
https://github.com/flame-engine/flame/blob/892052b99a21a8e371c4163e1e1918fd187c6e11/packages/flame/lib/src/components/core/component_tree_root.dart#L85-L88

spydon (approved):
Lgtm!

ufrshubham:
Posting the MRE just in case anyone is interested:

```dart
class RemoveAddGame extends FlameGame with HasKeyboardHandlerComponents {
  final RectangleComponent rectangle = RectangleComponent(
    size: Vector2.all(100),
    position: Vector2.all(100),
  );

  @override
  Future<void> onLoad() async {
    await add(rectangle);
  }

  @override
  KeyEventResult onKeyEvent(
    KeyEvent event,
    Set<LogicalKeyboardKey> keysPressed,
  ) {
    if (event is! KeyDownEvent) {
      return KeyEventResult.handled;
    }

    if (event.logicalKey == LogicalKeyboardKey.keyA) {
      add(rectangle);
    } else if (event.logicalKey == LogicalKeyboardKey.keyR) {
      rectangle.removeFromParent();
    }

    rectangle.add(Component());

    return super.onKeyEvent(event, keysPressed);
  }
}
```

PR #3089, opened 2024-03-21
refactor: Change the ClipComponent factory Constructor to redirect Constructor

### LINKED ISSUE #3083: Convert ClipComponent factory Constructor to redirecting Constructor.
### What could be improved

The ClipComponent Class has 3 factory Constructor "circle", "rectangle", "polygon". 
I suggest that "circle" and "rectangle" factory constructor can be converted into redirecting constructor.

As of my understanding, redirecting constructor are more performant than factory constructor as they initialize an object in one step and not taking one step per parameter.

But beside that, if they are converted to redirecting constructor, they allow subclasses to call them during initialization.
```dart
// Example:
Subclass.rectangle() : super.rectangle(); 
```

### Risks

As the notation do not change, this won't change the API for current users. 
Risks could be, that assertions in the constructor itself are no longer possible for further parameter (Thats why, ClipComponent.polygon should stay the way it is).

It will also no longer be possible, to change this behavior back without adding the risk to break implementations of user.

### More information

_No response_

### Other

- [X] Are you interested in working on a PR for this?

YukiAttano:
# Description

The ClipComponent factory Constructor "circle", "rectangle" and "polygon" are now redirect Constructor while they persist their previous behaviour and syntax.
This allows subclasses of ClipComponent to directly address them.
```dart
/// Examples: 

SubClass.fancyShape() : super.polygon();
Smiley() : super.circle();
```

## Checklist

## Breaking Change?

## Related Issues

Closes #3083

### DIFF
--- a/packages/flame/lib/src/components/clip_component.dart
+++ b/packages/flame/lib/src/components/clip_component.dart
@@ -88,7 +88,7 @@ class ClipComponent extends PositionComponent implements SizeProvider {
     int? priority,
     ComponentKey? key,
   }) : this(
-          builder: _polygonBuilderCallback(points),
+          builder: _polygonShapeBuilder(points),
           position: position,
           size: size,
           scale: scale,
@@ -131,7 +131,7 @@ class ClipComponent extends PositionComponent implements SizeProvider {
   ///
   /// this allows us to use an assertion during Constructor initialization
   /// rather than at the execution of the builder function.
-  static ShapeBuilder _polygonBuilderCallback(List<Vector2> points) {
+  static ShapeBuilder _polygonShapeBuilder(List<Vector2> points) {
     assert(
       points.length >= 3,
       'PolygonClipComponent requires at least 3 points.',
--- a/packages/flame/lib/src/components/clip_component.dart
+++ b/packages/flame/lib/src/components/clip_component.dart
@@ -29,7 +29,7 @@ class ClipComponent extends PositionComponent implements SizeProvider {
   /// {@macro circle_clip_component}
   ///
   /// Clips the canvas in the form of a circle based on its size.
-  factory ClipComponent.circle({
+  ClipComponent.circle({
     Vector2? position,
     Vector2? size,
     Vector2? scale,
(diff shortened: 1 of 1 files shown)


packages/flame/lib/src/components/clip_component.dart
```diff
@@ -144,4 +126,26 @@ class ClipComponent extends PositionComponent implements SizeProvider {
...
+  }
+
+  static Shape _polygonBuilder(List<Vector2> points, Vector2 size) {
+    final translatedPoints = points
+        .map(
+          (p) => p.clone()..multiply(size),
+        )
+        .toList();
+    return Polygon(translatedPoints);
+  }
```
spydon, line 150: I would merge these two methods and call the resulting method `_polygonBuilder`, since it isn't really a callback.

YukiAttano: The idea behind them is, that the first gets executed during initialization and therefor the assertion is run imediatly while it then returns the builder function that is used later.

If the assertion is in the builder itself, it gets first executed in the onLoad() method if i am right.

spydon: Right, that makes sense, the second function could still be inlined, but now when looking at it that will be a bit more expensive so probably worth having it like it is now, but coming up with a better name for  `_polygonBuilderCallback`. As always, hard with naming though. :sweat_smile:

spydon: Maybe just `_polygonShapeBuilder`?

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -91,31 +87,17 @@ class ClipComponent extends PositionComponent implements SizeProvider {
...
-      scale: scale,
-      angle: angle,
-      anchor: anchor,
-      children: children,
-      priority: priority,
-      key: key,
-    );
-  }
+  }) : this(
+          builder: _polygonBuilderCallback(points),
```
spydon, line 91:
```suggestion
          builder: _polygonShapeBuilder(points),
```

packages/flame/lib/src/components/clip_component.dart
```diff
@@ -144,4 +126,26 @@ class ClipComponent extends PositionComponent implements SizeProvider {
   bool containsLocalPoint(Vector2 point) {
     return _shape.containsPoint(point);
   }
+
+  /// Returns the [ShapeBuilder] function that builds a polygon
+  ///
+  /// this allows us to use an assertion during Constructor initialization
+  /// rather than at the execution of the builder function.
+  static ShapeBuilder _polygonBuilderCallback(List<Vector2> points) {
```
spydon, line 134:
```suggestion
  static ShapeBuilder _polygonShapeBuilder(List<Vector2> points) {
```

packages/flame/lib/src/components/clip_component.dart
spydon: I'll fix it up then, since you're in a hurry :)

spydon (approved):

PR #3101, opened 2024-03-27
feat: Add custom long tap delay

lucasnlm:
# Description

Currently `longTapDelay` is hardcoded to 300ms.
This is good for most of the games. But same games required custom long press values.
This PR allows it.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/events/component_mixins/tap_callbacks.dart
+++ b/packages/flame/lib/src/events/component_mixins/tap_callbacks.dart
@@ -16,13 +16,18 @@ mixin TapCallbacks on Component {
   void onTapUp(TapUpEvent event) {}
   void onTapCancel(TapCancelEvent event) {}
 
+  /// The delay (in milliseconds) after which a tap is considered a long tap.
+  int get longTapDelay => 300;
+
   @override
   @mustCallSuper
   void onMount() {
     super.onMount();
     final game = findRootGame()!;
     if (game.findByKey(const MultiTapDispatcherKey()) == null) {
-      final dispatcher = MultiTapDispatcher();
+      final dispatcher = MultiTapDispatcher(
+          longTapDelay: longTapDelay,
+      );
       game.registerKey(const MultiTapDispatcherKey(), dispatcher);
       game.add(dispatcher);
     }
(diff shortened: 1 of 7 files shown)


packages/flame/lib/src/events/flame_game_mixins/multi_tap_dispatcher.dart
```diff
@@ -118,9 +125,9 @@ class MultiTapDispatcher extends Component implements MultiTapListener {
 
   //#region MultiTapListener API
 
-  /// The delay (in seconds) after which a tap is considered a long tap.
+  /// The delay (in milliseconds) after which a tap is considered a long tap.
   @override
-  double get longTapDelay => 0.300;
+  int get longTapDelay => _longTapDelay;
```
spydon, line 130: In Flame we use seconds everywhere, so this should not be changed to milliseconds

packages/flame/lib/src/events/component_mixins/tap_callbacks.dart
```diff
@@ -16,13 +16,18 @@ mixin TapCallbacks on Component {
   void onTapUp(TapUpEvent event) {}
   void onTapCancel(TapCancelEvent event) {}
 
+  /// The delay (in milliseconds) after which a tap is considered a long tap.
+  int get longTapDelay => 300;
```
spydon, line 20: It's a bit strange that it is coming from this mixin, since it'll only have effect the first time a component is added. I think for this case it might be worth adding a `static` variable to the dispatcher.

spydon (commented):
It's a good idea, we just need to discuss the implementation details

packages/flame/lib/src/events/component_mixins/tap_callbacks.dart
lucasnlm: You mean a mutable static? I don't like the ideia. Is there any alternative?

spydon: There are few times where static variables are useful, but since passing this value in from the mixin just makes it more confusing for the user since they'll believe that they can set it per tappable component, I think it is a better option.
However, one downside of it is that it does affect the possibility of having multiple different games in the same app with different long tap delays (a very rare use-case). I don't have any other ideas for how to solve this, if you do, feel free to suggest them.

lucasnlm: @spydon Added an alternative implementation here: https://github.com/flame-engine/flame/pull/3110

lucasnlm:
Canceled in favor of https://github.com/flame-engine/flame/pull/3110

PR #3103, opened 2024-03-27
feat!: Use `Flame.images` in flame_texturepacker

### LINKED ISSUE #3096: texture packer can't override images
### Problem to solve

When using an Animation from TexturePacker where the assets are in a separate package, it is required to override the prefix of both the `AssetsCache` as well as as the `Images` to include `packages/my_package`

The `AssetsCache` can be overridden easily as texture_packer uses `Flame.assets` https://github.com/flame-engine/flame/blob/b1f01986b440ac18bb35b0d76963b2c66f49ea33/packages/flame_texturepacker/lib/src/texture_packer_atlas.dart#L104

However when loading the texture it uses a private custom image storage.

https://github.com/flame-engine/flame/blob/b1f01986b440ac18bb35b0d76963b2c66f49ea33/packages/flame_texturepacker/lib/src/texture_packer_atlas.dart#L147

This can't be overriden.

### Proposal

Use `Flame.images` to load the texture.
When loading the texture from a Storage this is already hapenning
https://github.com/flame-engine/flame/blob/b1f01986b440ac18bb35b0d76963b2c66f49ea33/packages/flame_texturepacker/lib/src/texture_packer_atlas.dart#L142

Making https://github.com/flame-engine/flame/blob/b1f01986b440ac18bb35b0d76963b2c66f49ea33/packages/flame_texturepacker/lib/src/texture_packer_atlas.dart#L147 the only usage of the private `_images` field

### More information

_No response_

### Other

ABausG:
# Description

Using `Flame.images` instead of a custom `ImageCache` in `TexturePackerAtlas` allowing users to override the `ImageCache`'s prefix to support loading Atlas Animations from different packages.
Due to this change the `prefix` of the used `ImageCache` changed from `assets/` to `assets/images/`

## Checklist

## Breaking Change?

### Migration instructions

Either move all sprites and textures from `assets/` to `assets/images/` or provide a custom cache using `images: Images(prefix: 'assets/')`or use `Flame.images.prefix = 'assets/'` to point to the same Location as before. Be aware that the latter approach might interfere with loading other Image Assets

## Related Issues
Closes #3096 


spydon (approved):
I was going to suggest that we let the user pass in an optional custom cache too, but since that argument will have to go through so many methods I think we can leave that as a follow-up for when someone needs it.

ABausG:
> [...] that argument will have to go through so many methods

If I counted correctly it would be ~6 methods with a maximum call chain of 3. This would then still be kind of reasonable I guess.

If this route is taken would you suggest we keep a `Images(prefix: 'assets/')` as the default? Then this wouldn't even be a breaking change!

Happy to adjust that @spydon

spydon:
> If I counted correctly it would be ~6 methods with a maximum call chain of 3. This would then still be kind of reasonable I guess.

Alright, if you want to adjust it you can follow how it's done in the parallax:
https://github.com/flame-engine/flame/blob/main/packages/flame/lib/src/parallax.dart#L147

> If this route is taken would you suggest we keep a `Images(prefix: 'assets/')` as the default? Then this wouldn't even be a breaking change!

I think the breaking change is still worth it, it's good to default to `Flame.images`.

ABausG:
@spydon adjusted to pass the custom cache through. Also applying the custom cache for the Assets Loading.

Also adjusted the PR description to include the 3rd way of migrating the version

PR #3113, opened 2024-04-03
feat: Loading builder for Route

PistonShot1:
# Description

This change is mainly to make it easier for specifying loading screen along when routing between component using the ``RouterComponent``. Currently the change is only made on ``Route``, that now have optional ``loadingBuilder`` to Route constructor which would expect a ``Component``, which would be added as the loading page/screen while  ``onLoad`` of the ``builder`` (the main routing component) to be completed.

## Checklist

## Breaking Change?


packages/flame/lib/src/components/route.dart
```diff
@@ -118,17 +125,31 @@ class Route extends PositionComponent
   /// also be added as a child component.
   Component? _page;
 
+  /// The loadingpage that was built and is now owned by this route. This
+  /// loadingpage will also be added as a child component.
+  Component? _loadingpage;
```
spydon, line 130:
```suggestion
  Component? _loadingPage;
```

spydon (approved):
Lgtm, very clean implementation.
Could you add a simple test that just tests that both the loadingPage and then the page is shown?
It would be in here:
https://github.com/flame-engine/flame/blob/main/packages/flame/test/components/route_test.dart

packages/flame/lib/src/components/route.dart
spydon: The spell checker step in the pipeline will also be happy once this is updated.

PistonShot1:
wait ya , dont merge yet, adding test for this

PistonShot1:
Don't merge yet.

hi @spydon , sorry I was away for awhile. Just pushed new one with test. Anyways, I just have problem it making a full test to check if the loading component was mounted before it was removed. I was only able to check for loading component was removed and the builder component ( page component) was mounted right after it was removed.

If you look at the file route_test.dart , I added test and also did add few comments , to explain the situation for the test coverage. Sorry I am not that good at coming with test, but I did manually check it by linking my version of flame repo to one of my old flame game, I mean my flame in pubspec refer to my git repo instead of pub.dev and test this feature I added, according to my observation it did work to what I expected.

PistonShot1:
I mean does checking the loading component was removed technically means it also verifies that it was mounted before right . So if that is the case then the test I added already suffice ?

packages/flame/test/components/route_test.dart
```diff
@@ -474,6 +475,51 @@ void main() {
...
+      // mounted before it was removed.
+
+      // expect(
+      //   loadingComponent.isMounted,
+      //   isTrue,
+      // );
+      // expect(
+      //   pageComponent.isMounted,
+      //   isFalse,
+      // );
```
spydon, line 510: I don't think I follow why this doesn't work, what are you getting here currently?

PistonShot1: I'm sorry , False is being expected for 503, as for 507 it does pass

spydon: Why is false expected for 503? Since the route is added it should be showing the `loadingComponent` right?

PistonShot1: I just found, I made an error in the checking at line 514 is not accurate in this commit, it should be expecting True, but I checked there for False and it passed

That means my loading component was never mounted even before and after the main component was mounted? 
Maybe, that mean my implementation may have some error ?

spydon: I'm guessing that due to how `Future.delayed` works in tests the `loadingComponent` was added and removed in the same tick. You could pass in a future to the `_HeavyComponent` that you manually complete after you have checked that the `loadingComponent` is mounted.

renancaraujo (approved):

packages/flame/test/components/route_test.dart
spydon: Did you have any more time to look at this @PistonShot1?

PistonShot1: Sorry for the delays got caught up with another project, but I am trying to implement the approach you mentioned , to pass a future that I can manually complete, I am not exactly sure how to implement this . I am assuming with `Completer` or sort ?

spydon: Yeah, `Completer` should work

PistonShot1: the test for checking if the `loadingComponent` was mounted and remove, both doesn't pass (at line 499 and at line 509), as I have commented, it is some how implying that the loadingPage was never mounted. I am not sure why, but the changes I added at `packages/flame/lib/src/components/route.dart` (line 146-151), supposed have mounted and removed after awaiting `_page!.loaded;`

spydon:
@PistonShot1 have you made any progress on this? :)

PistonShot1:
@spydon . I have fixed it, sorry took so long for such a small feature addition, I gave up after the test keep on fail, but as you suspected it was the way I wrote the test, now it works fine and should be working and passed the expected tests

packages/flame/lib/src/components/route.dart
```diff
@@ -53,6 +55,11 @@ class Route extends PositionComponent
   /// in which case the user must override the [build] method.
   final Component Function()? _builder;
 
+  /// The function that will be invoked that will build the loading page
+  /// component when this route first becomes active. Can be left `null`,
+  /// in which case it is not required.
```
spydon, line 60:
```suggestion
  /// The function that will build the loading page component, which is shown
  /// when this route first becomes active, but hasn't fully loaded yet.
```

packages/flame/lib/src/components/route.dart
```diff
@@ -118,17 +125,32 @@ class Route extends PositionComponent
   /// also be added as a child component.
   Component? _page;
 
+  /// The loadingPage that was built and is now owned by this route. This
+  /// loadingPage will also be added as a child component.
```
spydon, line 129:
```suggestion
  /// The loadingPage that was built and is now owned by this route. The
  /// [_loadingPage] will also be added as a child component.
```

packages/flame/lib/src/components/route.dart
```diff
@@ -118,17 +125,32 @@ class Route extends PositionComponent
...
+    (_loadingBuilder != null) ? _addLoadingPage() : _page!.addToParent(this);
     onPush(previousRoute);
   }
 
+  /// Adds the [_loadingPage] to the parent and invoked by [didPush] , when
+  /// [_loadingBuilder] is specified
+  Future<void> _addLoadingPage() async {
+    _loadingPage ??= _loadingBuilder!()..addToParent(this);
+    await _loadingPage!.loaded;
+    await _page!.addToParent(this);
```
spydon, line 149:
```suggestion
    await add(_page!);
```

spydon (approved):
Lgtm!

PR #3123, opened 2024-04-14
feat: Add completed future for effects

ufrshubham:
# Description

This PR adds a `completed` future for effects which completes when the effect finishes.

## Checklist

## Breaking Change?

## Related Issues

NA

### DIFF
--- a/doc/flame/effects.md
+++ b/doc/flame/effects.md
@@ -91,6 +91,8 @@ functionality inherited by all other effects. This includes:
 - Optional user-provided `onComplete`, which will be invoked when the effect has just
   completed its execution but before it is removed from the game.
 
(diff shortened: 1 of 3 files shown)


spydon (commented):
Missing docs, otherwise it looks good!
Maybe we should remove the callback in V2 and just have completers everywhere?

ufrshubham:
> Missing docs, otherwise it looks good!

Added a line about this in the docs now. 

> Maybe we should remove the callback in V2 and just have completers everywhere?

Maybe it is fine to keep both because there is slight difference between the two. Callbacks can execute multiple times if the event repeats. On the other hand, we set the completers to null once it is triggered.

spydon (approved):

PR #3129, opened 2024-04-18
fix: Fix cascading and fallback propagation of text styles

luanpotter:
# Description

Fix cascading and fallback propagation of text styles.

While using text nodes and flame_markdown, I noticed that basic propagating of styles was not working.

For example, if you create a standard document, it will have some margins by default. Even when setting it to zero, it was still overwritten by the default style.

I noticed that there was no consistency in the methods `copyWith` and `merge` of which parameter/receiver should be the master and which should be the fallback. Both the implementations and the callers would use it either way indiscriminately.

This consolidates the definition and adds comments to the methods enforcing the priority order.

In order to fully test this, I had to expose a few internal details of text elements; I think this is fine since Dart supports the `@visibleForTesting` annotation, and it makes the tests cleaner.

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame/lib/src/text/styles/flame_text_style.dart
+++ b/packages/flame/lib/src/text/styles/flame_text_style.dart
@@ -20,7 +20,7 @@ abstract class FlameTextStyle {
   FlameTextStyle copyWith(covariant FlameTextStyle other);
 
   /// Merges two [FlameTextStyle]s together, preferring the properties of
-  /// [style1] if present, falling back to the properties of [style2].
+  /// [style2] if present, falling back to the properties of [style1].
   static T? merge<T extends FlameTextStyle>(T? style1, T? style2) {
     if (style1 == null) {
       return style2;
@@ -28,7 +28,7 @@ abstract class FlameTextStyle {
       return style1;
     } else {
       assert(style1.runtimeType == style2.runtimeType);
-      return style2.copyWith(style1) as T;
+      return style1.copyWith(style2) as T;
     }
   }
 }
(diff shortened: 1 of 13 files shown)


packages/flame/lib/src/text/styles/flame_text_style.dart
```diff
@@ -15,16 +15,20 @@ import 'package:flame/text.dart';
...
+  /// [style1] if present, falling back to the properties of [style2].
   static T? merge<T extends FlameTextStyle>(T? style1, T? style2) {
     if (style1 == null) {
       return style2;
     } else if (style2 == null) {
       return style1;
     } else {
       assert(style1.runtimeType == style2.runtimeType);
-      return style1.copyWith(style2) as T;
+      return style2.copyWith(style1) as T;
```
spydon, line 31: Doesn't it make more sense in the order it was in before? I think that is what I would expect from a merge function, that the first argument is the base and you overwrite its properties with the second one, especially if this would be a merge function taking in more than two arguments.

spydon: Similarly to this for example: https://api.flutter.dev/flutter/package-collection_collection/mergeMaps.html

luanpotter: Good point; I had chosen this particular order because it makes the copyWith neater:
image
But I think your argument is stronger, so I will swap

spydon (approved):
Lgtm!

PR #3139, opened 2024-04-27
feat: Expand flame_lint to respect required pub.dev checks

luanpotter:
# Description

Expand flame_lint to respect required pub.dev checks.

We are currently losing pub points due to lint violations:

image

Turns out `flame_lint` does not respect lints/core which is being enforced now by pub.

This adds that as a dependency on `flame_lint`, updates `flutter_lint` for the packages that use that, and fix all existing violations (luckily very few).

This change will ensure that us and everyone else who uses `flame_lint` get all the points they deserve on pub. We can consider expanding to `lints/recommended` in the future, but that is definitely not a requirement at this stage (and a much bigger change).

## Checklist

## Breaking Change?

### DIFF
--- a/doc/tutorials/platformer/app/lib/ember_quest.dart
+++ b/doc/tutorials/platformer/app/lib/ember_quest.dart
@@ -57,19 +57,19 @@ class EmberQuestGame extends FlameGame
   void loadGameSegments(int segmentIndex, double xPositionOffset) {
     for (final block in segments[segmentIndex]) {
       final component = switch (block.blockType) {
-        GroundBlock => GroundBlock(
+        GroundBlock _ => GroundBlock(
             gridPosition: block.gridPosition,
             xOffset: xPositionOffset,
           ),
-        PlatformBlock => PlatformBlock(
+        PlatformBlock _ => PlatformBlock(
             gridPosition: block.gridPosition,
             xOffset: xPositionOffset,
           ),
-        Star => Star(
+        Star _ => Star(
             gridPosition: block.gridPosition,
             xOffset: xPositionOffset,
           ),
-        WaterEnemy => WaterEnemy(
+        WaterEnemy _ => WaterEnemy(
             gridPosition: block.gridPosition,
             xOffset: xPositionOffset,
           ),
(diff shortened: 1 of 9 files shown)


packages/flame_bloc/pubspec.yaml
```diff
@@ -23,7 +23,7 @@ dev_dependencies:
   dartdoc: ^6.3.0
   flame_lint: ^1.1.2
   flame_test: ^1.16.1
-  flutter_lints: ^2.0.1
+  flutter_lints: ^3.0.2
```
spydon, line 26: Why is this using both flutter_lints and flame _lints?

doc/tutorials/platformer/app/lib/ember_quest.dart
```diff
@@ -57,19 +57,19 @@ class EmberQuestGame extends FlameGame
   void loadGameSegments(int segmentIndex, double xPositionOffset) {
     for (final block in segments[segmentIndex]) {
       final component = switch (block.blockType) {
-        GroundBlock => GroundBlock(
+        GroundBlock _ => GroundBlock(
```
spydon, line 60: I think there might be a better way of doing type checks in switches

packages/flame_bloc/pubspec.yaml
luanpotter: I imagine it is because this has flutter code? flame_lints doesn't include any flutter-specific rules.
(not sure though, I am just bumping the version)

doc/tutorials/platformer/app/lib/ember_quest.dart
luanpotter: the other option suggest is this:
```
        const (GroundBlock) => GroundBlock(
            gridPosition: block.gridPosition,
            xOffset: xPositionOffset,
          ),
```
I thought the underscore looked cleaner as it implies in my mind a pattern matching variable that is just not captured

packages/flame_bloc/pubspec.yaml
luanpotter: it is actually never used, so let me remove

spydon (approved):

PR #3142, opened 2024-04-28
fix!: Update IsometricTileMapComponent to have better defined position and size

luanpotter:
# Description

Update IsometricTileMapComponent to have better defined position and size.

Before, the isometric component "zero" would be the center of the 0,0 block.
However, that does not play nicely with our component system if you want to know the size (i.e. bounding box) of a component.

This changes so that the 0,0 of the component is the 0,0 of the AABB around the isometric tiles. Then, it also computes the size of the component accordingly.

This also changes the example to allow toggling between half and full size more easily.

In our example, this is what it looks like:

image

The example still shows how to compute the previous origin (the purple dot) if you want to.

With full size blocks:

image

I also [updated isomash](https://github.com/luanpotter/isomash/pull/1) to check if more complex functionality plays well with the new system:
image

## Checklist

## Breaking Change?

### Migration instructions

This is a minor breaking change as you might need to "reposition" your tile components, essentially remove the "compensation" for its location that you probably did yourself. If you were centering the tile component based on the available methods such as `map.getBlockCenterPosition`, then just make sure you are adding the `map.position` to that and it should work as before.

### DIFF
--- a/packages/flame/lib/src/components/isometric_tile_map_component.dart
+++ b/packages/flame/lib/src/components/isometric_tile_map_component.dart
@@ -60,7 +60,7 @@ class IsometricTileMapComponent extends PositionComponent {
 
   /// Displacement applied so that the origin of the component
   /// matches the origin of the AABB.
-  final Vector2 _ds = Vector2.zero();
+  final Vector2 _offset = Vector2.zero();
 
   IsometricTileMapComponent(
     this.tileset,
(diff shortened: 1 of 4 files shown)


packages/flame/lib/src/components/isometric_tile_map_component.dart
```diff
@@ -101,6 +104,11 @@ class IsometricTileMapComponent extends PositionComponent {
     }
   }
 
+  @override
+  void update(double dt) {
+    _recomputeSize();
```
luanpotter, line 109: since the block matrix is not final, theoretically it could be changed at any time. luckily recomputing the size is quite cheap so I think this should be fine.

spydon (approved):
Nice! Lgtm

PR #3147, opened 2024-05-05
feat: Support text align on new text rendering pipeline

luanpotter:
# Description

Support text align on new text rendering pipeline.

### Notes

I am re-using Flutter's `TextAlign` enum for simplicity; however, it has extra constants we don't quite need. It has `start` and `end` that for us are synonyms of `left` and `right` because the text rendering pipeline does not support rtl text directions. It also has `justify` which this PR does not support as it is much more involved (though we could support later).

The other options were to either make a new enum, use a double from 0 to 1, or use the anchor class (but only take the horizontal component - vertically it doesn't make sense as it is always exactly line height). Of those options, I believe re-using the enum is cleaner.

That being said, this does touch on a crucial point that I am noticing: it seems we are rebuilding a lot of what Flutter already gives us for free. There is a whole `TextPainter` based rendering pipeline available in Flutter is decomposed from flutter components and could be leverage more directly leverage by Flame. It includes layouting, line breaking, text metrics, handles rtl text, etc etc. I have some doubts about our current structure, even though it was vastly simplified [with this series of PRs](https://github.com/flame-engine/flame/pull/2663), we still have two minorly overlapping systems that both also have varying degrees of overlap with inherited Flutter abstractions, with often less powerful features.

### Results

image

image

## Checklist

## Breaking Change?

### DIFF
--- a/examples/lib/stories/rendering/rendering.dart
+++ b/examples/lib/stories/rendering/rendering.dart
@@ -67,7 +67,15 @@ void addRenderingStories(Dashbook dashbook) {
     )
     ..add(
       'Rich Text',
-      (_) => GameWidget(game: RichTextExample()),
+      (context) => GameWidget(
+        game: RichTextExample(
+          textAlign: context.listProperty(
+            'Text align',
+            TextAlign.left,
+            TextAlign.values,
+          ),
+        ),
+      ),
       codeLink: baseLink('rendering/rich_text_example.dart'),
       info: RichTextExample.description,
     );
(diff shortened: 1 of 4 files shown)


spydon (approved):
Lgtm! If we can use flutters text rendering system I think that would definitely be worth a breaking change.

luanpotter:
I need to do some digging before coming up with a concrete proposal, but I do see it as a necessary improvement.

PR #3152, opened 2024-05-08
feat: Add `onReleased` action to `AdvancedButtonComponent` which will be called within `onTapUp`

### LINKED ISSUE #3150: `AdvancedButtonComponent` doesn't show `downSkin` when changing the route
### What could be improved

`AdvancedButtonComponent` doesn't show the `downSkin` if the action of `onPressed` changes the router to another page. That is because the router is changed before the `downSkin` is showen as the `onPressed` action is called in `onTapDown` instead of `onTapUp`.

My suggestion is to move calling `onPressed` action to `onTapUp` method instead of `onTapDown`.

That makes the `AdvancedButtonComponent` more natural and similar to many other platforms. That also allows users to skip the action if they move the mouse pointer or their finger outside the button box.

### Risks

The current behavior of the `AdvancedButtonComponent` will slightly get changed, as the action will be called on tap up. With normal click/press, that should not be noticable.

### More information

_No response_

### Other

- [X] Are you interested in working on a PR for this?

mabdelaal86:
# Description

Add `onReleased` action which will be called within `onTapUp`. That fixed some scenarios which require the pressed action to be called after the user taps up the button.

For example, `AdvancedButtonComponent` doesn't show the `downSkin` if the action of `onPressed` changes the router to another page. That is because the router is changed before the `downSkin` is shown as the `onPressed` action is called in `onTapDown`.

## Checklist

## Breaking Change?

## Related Issues

Fixes #3150

### DIFF
--- a/packages/flame/lib/src/components/input/advanced_button_component.dart
+++ b/packages/flame/lib/src/components/input/advanced_button_component.dart
@@ -86,13 +86,16 @@ class AdvancedButtonComponent extends PositionComponent
     if (_isDisabled) {
       return;
     }
-    onPressed?.call();
     isPressed = true;
     updateState();
   }
 
   @override
   void onTapUp(TapUpEvent event) {
+    if (_isDisabled) {
+      return;
+    }
+    onPressed?.call();
     isPressed = false;
     updateState();
   }
--- a/packages/flame/lib/src/components/input/advanced_button_component.dart
+++ b/packages/flame/lib/src/components/input/advanced_button_component.dart
@@ -19,6 +19,7 @@ class AdvancedButtonComponent extends PositionComponent
     with HoverCallbacks, TapCallbacks {
   AdvancedButtonComponent({
     this.onPressed,
+    this.onReleased,
     this.onChangeState,
     PositionComponent? defaultSkin,
     PositionComponent? downSkin,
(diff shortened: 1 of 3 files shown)


mabdelaal86:
The failed test case seems unrelated to my changes and all tests are passed locally with `melos run test`!
Is there a way to rerun the checks again?

spydon:
I don't think this is worth a breaking change, it's better to expose a callback for onTapUp. In UI's buttons often trigger in onTapUp, but in games, when mashing a fire button for example, you usually want onTapDown.

mabdelaal86:
Yeah, that makes sense. I didn't think about it, since I'm developing mainly board games, which are kind of UI apps.

mabdelaal86:
Does it worth adding a callback to the `AdvancedBottonComponent` for onTapUp? or just implement a custom version in my game?

spydon:
> Does it worth adding a callback to the `AdvancedBottonComponent` for onTapUp? or just implement a custom version in my game?

I think it would be worth adding! I guess we should add `onTapDown` then too and deprecate `onPressed` in the long run.

mabdelaal86:
I have added `onReleased` event which will be called from `onTapUp`.

spydon:
> I have added `onReleased` event which will be called from `onTapUp`.

Sounds good! Then we can refactor it to `onTapDown` and `onTapUp` for V2 instead (in a distant future)

spydon (approved):
Lgtm, thanks for your contribution!

mabdelaal86:
Thanks a lot for the fast review

PR #3168, opened 2024-05-24
fix: Invoke `setToStart` on child effect controller of wrapping effect controllers

ufrshubham:
# Description

`DelayedEffectController` was not calling `setToStart` on its child, causing it to not behave as expected when used with `InfiniteEffectController`. This was reported by a discord member [in this message](https://discord.com/channels/509714518008528896/516639688581316629/1242258377766080666).

This PR introduces a `HasSingleChildEffectController` mixin which can be added to any effect controller that wraps a single child effect controller. This mixin makes sure that the `setToStart`, `setToEnd` and `onMount` methods always get invoked on the child controllers.

## Checklist

## Breaking Change?

## Related Issues

NA

### DIFF
--- a/packages/flame/lib/src/effects/controllers/mixins/has_single_child_effect_controller.dart
+++ b/packages/flame/lib/src/effects/controllers/mixins/has_single_child_effect_controller.dart
@@ -1,0 +1,28 @@
+import 'package:flame/effects.dart';
+import 'package:meta/meta.dart';
+
+/// This mixin must be used with [EffectController]s that wrap a single child
+/// effect controller of type [T].
+mixin HasSingleChildEffectController<T extends EffectController>
+    on EffectController {
+  /// Returns the wrapped child effect controller.
+  T get child;
+
+  @mustCallSuper
+  @override
+  void setToStart() {
+    child.setToStart();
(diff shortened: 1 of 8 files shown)


packages/flame/lib/src/effects/controllers/mixins/has_single_child_effect_controller.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:flame/effects.dart';
+import 'package:meta/meta.dart';
+
+/// This mixin must be used with [EffectController]s that wrap a single child
+/// effect controller of type [T].
+mixin HasSingleChildEffectController<T extends EffectController>
```
spydon, line 6: Doesn't all effect controllers wrap a single child, so that this could be directly on the `EffectController` instead of a mixin? 🤔

ufrshubham: `DurationEffectController`s do not need a child and `SequenceEffectContoller` has a list of children. So these two controllers won't satisfy the single child condition.

spydon (approved):

PR #3182, opened 2024-06-04
feat: Implement basic light abstraction [flame_3d]

luanpotter:
# Description

Implement a basic light abstraction.

Still only supports one light for now, and no parameters other than its position.
Note: I will fix the 0-lights case when I add support for multiple lights.

A few complications arose that led me to revamp our component system a bit. Basically splitting the `Component3D` class into a more fundamental version of itself that only includes the 3D positioning stuff, and none of the rendering stuff, and what used to be `Component3D` becomes `Object3D` (I am not sold on the names, feel free to suggest replacements).

I also added a `Group3D` class to allow for people to create groups of components without carrying the rendering code. Of course you can still add children to Object3D as well. I am open to removing it and just making `Component3D` concrete instead.

The image below explains the new structure:

image

---

The difference between the 3 "light" classes:

* `LightComponent` → interacts with FCS. is Component3D, thus carries the position
* `LightSource` → describes the light in abstract terms (minus position). used both by `LightComponent` and `Light`
* `Light` → base description of a Light as a Resource. FCS agnostic. includes properties (source) + position

Again, I am open to any renames.

---

#### why separate component3d and object 3d?

not really needed but I didn't want light component to "carry" the rendering code that is not used. that also makes the distinction more clear as objects are completely different than lights. that way we don't have to have a definition of "non-light objects/components"

#### ~why do we need group3d?~

i removed group3d!

~again not really needed, you can add children to objects, but if you want a pure container (no rendering), this would make more semantic sense. other options would be to not have anything (tell people to group things using object3d) or make component3d concrete (tell people to use that). having group3d consolidates the separation of light as its own thing.~

#### why some classes end in 3D and some classes end in Component

[answer](https://martinfowler.com/bliki/TwoHardThings.html) (please help)

## Checklist

## Breaking Change?


packages/flame_3d/lib/src/resources/light/light.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:flame_3d/components.dart';
+import 'package:flame_3d/graphics.dart';
+
+abstract class Light extends Component3D {
```
wolfenrain, line 4: A light should be a resource while this is a component. The `graphics` and `resources` directory should always stay flame agnostic as we can in the long term pull it out as a 3D rendering framework for others to use.

We should have light describe all the "shader values" just like how materials and such do that. And on bind it should set the uniforms on the current active shader/material somehow. I am not sure how we can best do that but the rendering pipeline should imho go:
```
- Mesh
   - Surface
      - Material
         - Texture
         - Shaders
         - Uniforms
- Light
   - Uniforms on Shaders
```

I am of course open to ideas, maybe potentially light should not even be a graphics resource (like you did here) but that we have to make sure that you still can control resource uniforms so that the light can change aspects of materials

luanpotter: I see. let me think about that.

we still need a way in the end to refer to the list of lights in the `bind` method of the mesh right?

so either we store it on the graphics device, or create another parameter to bind, like a `context`. what do you think?

I will try to take a look tonight on how other engines do this too, see if it can... shed any light on us.

luanpotter: @wolfenrain made an attempt based on what I learned from playing with Godot.

It is hard to be 1:1 because their architecture is very different, and they have no separation between their "components" (called Nodes) and the features (like Light).

packages/flame_3d/example/lib/rotating_light.dart
```diff
@@ -0,0 +1,23 @@
...
+
+  @override
+  void update(double dt) {
+    const radius = 15;
+    final angle = DateTime.now().millisecondsSinceEpoch / 4000;
+    final x = cos(angle) * radius;
+    final z = sin(angle) * radius;
+    light.position.setValues(x, 10, z);
+  }
+
```
luanpotter, line 22:
```suggestion
```

packages/flame_3d/lib/src/components/light_component.dart
```diff
@@ -0,0 +1,16 @@
...
+class LightComponent extends Component3D {
+  final Light light;
+
+  LightComponent({required this.light});
+
+  @override
+  void bind(GraphicsDevice device) {
+    world.device
+      ..model.setFrom(transformMatrix)
+      ..bindLight(light);
```
wolfenrain, line 14: This will grow the `lights` list in the graphics device on each frame right? We should clear the lights list in the `begin` method to ensure we dont have duplicate lights.

packages/flame_3d/lib/src/resources/light/light.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:flame_3d/game.dart';
+import 'package:flame_3d/graphics.dart';
+import 'package:flame_3d/resources.dart';
+
+/// {@template light}
+/// A [Resource] that represents a light source that changes how the scene is
+/// rendered.
+///
+/// This class isn't a true resource, it does not upload it self to the GPU.
+/// Instead, it is used to modify how meshes are uploaded.
```
wolfenrain, line 10: 🔥

packages/flame_3d/lib/src/resources/light/light.dart
```diff
@@ -0,0 +1,24 @@
...
+/// {@endtemplate}
+class Light extends Resource<void> {
+  // TODO(luan): Add more attributes and light types
```
wolfenrain, line 13: In the long run we can make `Resource` a macro and allowing us to have more dynamic structs for shaders.

packages/flame_3d/lib/src/resources/light/light.dart
```diff
@@ -0,0 +1,24 @@
...
+  final Vector3 position;
+
+  /// {@macro light}
+  Light({
+    required this.position,
+  }): super(null);
```
wolfenrain, line 19:
```suggestion
  /// {@macro light}
  Light({
    required this.position,
  }): super(null);
  
  final Vector3 position;
```

I know this isn't how we do it in Flame but that is how I did it in `flame_3d` :P so we should either keep it or refactor the rest.

packages/flame_3d/lib/src/resources/material/spatial_material.dart
```diff
@@ -61,18 +60,18 @@ class SpatialMaterial extends Material {
 
   @override
   void bind(GraphicsDevice device) {
+    final lightPosition = device.lightPosition;
+    if (lightPosition == null) {
+      return;
+    }
```
wolfenrain, line 66: Instead of returning we should maybe do `?? Vector3.zero()` because now a mesh wont have a material if someone didn't add a light.

Or better (and a follow-up PR) have a default light like godot so that will be used **unless** there is a light in the scene.

packages/flame_3d/lib/src/components/light_component.dart
luanpotter: oh good catch, I didn't realize that in my testing

packages/flame_3d/lib/src/resources/light/light.dart
luanpotter: we can change that later to match Flame

packages/flame_3d/lib/src/resources/material/spatial_material.dart
luanpotter: that would cause us to actually add a light at (0, 0, 0), which would be kinda weird?

I think we can add a default light later when we support an omni light. these are spot lights so far

luanpotter: (right now if you don't add any lights nothing is renderer, which kinda makes sense)

packages/flame_3d/lib/src/graphics/graphics_device.dart
```diff
@@ -93,6 +95,17 @@ class GraphicsDevice {
 
   void clearBindings() {
     _renderPass.clearBindings();
+    _lights.clear();
+  }
```
wolfenrain, line 99: I don't think `flame_3d` is currently using the `clearBindings` in the render loop is it? We should probably call it somewhere 😅  maybe by default in the `end()`?

packages/flame_3d/lib/src/components/group_3d.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:flame_3d/components.dart';
+
+/// {@template group_3d}
+/// [Group3D] is a [Component3D] that acts as a container for other
+/// [Component3D]s.
+///
+/// Changing the position, rotating or scaling the [Component3D] will affect
+/// the whole group as if it was a single entity.
```
luanpotter, line 8: I'm not convinced this will "just work", neither before nor now. But that is a test for a followup.

luanpotter:
@wolfenrain I think I came up with a nifty solution to solve the problem of LightComponent vs Light.
however I got excited and also attempted to solve other problems :P

packages/flame_3d/lib/src/components/component_3d.dart
```diff
@@ -1,48 +1,44 @@
...
+  final Transform3D transform;
+
   /// {@macro component_3d}
   Component3D({
     Vector3? position,
     Vector3? scale,
     Quaternion? rotation,
-  }) : transform = Transform3D()
+    List<Component3D> children = const [],
+  })  : transform = Transform3D()
```
wolfenrain, line 37: Does this mean children can be positioned in 3D space relative to the parent? I don't think that will work right now without changes to t he `Transform3D` right?

packages/flame_3d/lib/src/components/group_3d.dart
```diff
@@ -0,0 +1,19 @@
...
+///
```
wolfenrain, line 9:
```suggestion
```

packages/flame_3d/lib/src/components/group_3d.dart
wolfenrain: But the class itself would not be needed anyway because when it works the `Component3D` does the same thing :P

packages/flame_3d/lib/src/components/light_component.dart
```diff
@@ -0,0 +1,38 @@
...
+    Vector3? position,
+  }) : this(
+          source: SpotLight(),
+          position: position,
+        );
+
+  final LightSource source;
+
+  late final Light _light = Light(
+    transform: Transform3D(),
```
wolfenrain, line 23: Shouldn't this be the component's transform?

packages/flame_3d/lib/src/components/light_component.dart
```diff
@@ -0,0 +1,38 @@
...
+  Light get light {
+    return _light..transform.setFrom(transform);
+  }
+
+  @override
+  void onMount() {
+    assert(
+      parent is World3D,
+      'Lights must be added to the root of the World3D',
+    );
```
wolfenrain, line 36: Instead of an assert we should maybe just throw an exception? Asserts dissapear in release mode and children can be dynamically added/removed so you might never know why it fails in release mode

packages/flame_3d/lib/src/components/object_3d.dart
```diff
@@ -0,0 +1,62 @@
...
+/// [Object3D]s are the basic building blocks for a 3D [FlameGame].
+///
+/// It is an object that is positioned in 3D space and can be bind to be
+/// rendered by a [GraphicsDevice].
+///
+/// However, it has no visual representation of its own (except in
+/// debug mode). It is common, therefore, to derive from this class
+/// and implement a specific rendering logic.
+///
+/// See the [MeshComponent] for a [Component3D] that has a visual representation
```
wolfenrain, line 19:
```suggestion
/// See the [MeshComponent] for a [Object3D] that has a visual representation
```

packages/flame_3d/example/lib/rotating_light.dart
```diff
@@ -0,0 +1,20 @@
...
+          position: Vector3.zero(),
+        );
+
+  @override
+  void update(double dt) {
+    const radius = 15;
+    final angle = DateTime.now().millisecondsSinceEpoch / 4000;
+    final x = cos(angle) * radius;
+    final z = sin(angle) * radius;
+    position.setValues(x, 10, z);
```
wolfenrain, line 18: A little TODO to say that this is temporary would be nice :P the CI fails already anyway

packages/flame_3d/lib/src/components/component_3d.dart
luanpotter: it should, but I don't think it would right now. this is exactly how it was before

packages/flame_3d/lib/src/components/light_component.dart
luanpotter: it's being set to that value the line below:
```
  Light get light {
    return _light..transform.setFrom(transform);
  }
```
makes sure they are in sync

packages/flame_3d/lib/src/components/light_component.dart
luanpotter: this is the idiomatic way we do elsewhere, so I'd rather keep it consistent and change all at once later

packages/flame_3d/example/lib/rotating_light.dart
luanpotter: wouldn't say it is temporary, it is just an example

is the CI failing for this? I can only see the "expected" failures

packages/flame_3d/lib/src/components/group_3d.dart
luanpotter: they actually all seem to have this extra line. I can remove en mass if you don't want it

packages/flame_3d/lib/src/components/object_3d.dart
luanpotter, line 19:
```suggestion
/// See the [MeshComponent] for an [Object3D] that has a visual representation
```

packages/flame_3d/lib/src/components/object_3d.dart
```diff
@@ -0,0 +1,62 @@
...
+    // We set the priority to the distance between the camera and the object.
+    // This ensures that our rendering is done in a specific order allowing for
+    // alpha blending.
+    //
+    // Note(wolfenrain): we should optimize this in the long run it currently
+    // sucks.
+    priority = -(CameraComponent3D.currentCamera!.position - position)
+        .length
+        .abs()
+        .toInt();
```
wolfenrain, line 52: Shouldn't this live in `Component3D`?

packages/flame_3d/lib/src/game/notifying_quaternion.dart
```diff
@@ -170,5 +170,5 @@ class NotifyingQuaternion extends Quaternion with ChangeNotifier {
   }
 
   @override
-  Float32List get storage => UnmodifiableFloat32ListView(super.storage);
+  Float32List get storage => super.storage.asUnmodifiableView();
```
wolfenrain, line 173: Ah this is why we had it at `3.3.0` :P

packages/flame_3d/lib/src/graphics/graphics_device.dart
```diff
@@ -48,6 +48,10 @@ class GraphicsDevice {
 
   Size _previousSize = Size.zero;
 
+  /// Must be set by the rendering pipeline before elements are bound.
+  /// Can be accessed by elements in their bind method.
+  Iterable<Light> lights = [];
```
wolfenrain, line 53: Do we want to make this final? And/or private with a `Iterable` getter?

packages/flame_3d/lib/src/resources/light/light.dart
```diff
@@ -0,0 +1,29 @@
...
+    required this.source,
+  }) : super(null);
+
+  void apply(Shader shader) {
+    shader.setVector3('Light.position', transform.position);
+    // apply additional parameters
+    source.apply(shader);
+  }
+
+  static UniformSlot shaderSlot = UniformSlot.value('Light', {'position'});
```
wolfenrain, line 28: The additional parameters would only work if they are part of the shader slot right? Shouldn't this be determined by the resource of a light then?

packages/flame_3d/lib/src/resources/material/spatial_material.dart
```diff
@@ -61,29 +59,40 @@ class SpatialMaterial extends Material {
...
+    final cameraPosition = invertedView.transform3(Vector3.zero());
+    fragmentShader.setVector3('Camera.position', cameraPosition);
+  }
+
+  void _bindLights(GraphicsDevice device) {
+    final light = device.lights.firstOrNull;
+    if (light == null) {
+      return;
+    }
+    light.apply(fragmentShader);
```
wolfenrain, line 95: So are we applying or binding, the method name indicates something else. 

I vote for naming it `_applyLights` and calling it in the `_bindMaterials` instead. Because in the end it gets applied to the shader of the material.

packages/flame_3d/lib/src/components/group_3d.dart
wolfenrain: I still don't think we need the class tho, grouping is a core feature of any component.

packages/flame_3d/lib/src/components/group_3d.dart
wolfenrain: Yes please!

packages/flame_3d/lib/src/components/light_component.dart
wolfenrain: Wouldn't passing it be better? `Light` does not copy it and that way we don't have to use more memory.

packages/flame_3d/lib/src/components/component_3d.dart
wolfenrain: I agree it should be :P so one day we will be glad you already did this part of the code haha.

packages/flame_3d/lib/src/components/light_component.dart
luanpotter: I see. I was concerned with someone modifying the reference or something. but I guess it is final on the Component so should work

packages/flame_3d/lib/src/components/object_3d.dart
luanpotter: no, this is only relevant to components that are rendered, right?
that is the different between comp3d and obj3d

packages/flame_3d/lib/src/graphics/graphics_device.dart
luanpotter: cannot be final because it is modified by the world3d.
I opted to make it not final rather than to clear/addAll for performance reasons, but I can change

packages/flame_3d/lib/src/resources/light/light.dart
luanpotter: I am not sure I understand this comment, I did extract it to the Light resource exactly because this class is responsible for defining and binding these values (right now only position but more in the future)

packages/flame_3d/lib/src/resources/material/spatial_material.dart
luanpotter: I can rename to _applyLights, or I can rename light.apply to light.bind - no preference whatsoever
however I don't know if this should be inside the _bindMaterials, feels like light is a separate concept that the shader can use for several things... not super strong opinion though

wolfenrain: The spatial material is light aware and will always apply lights on it's material.

packages/flame_3d/lib/src/components/light_component.dart
wolfenrain: They still can :P but that is fine as that is the same as modifying the transform of the component.

packages/flame_3d/lib/src/resources/material/spatial_material.dart
luanpotter: I know, I was just thinking that other parts of the shader might want to know about light information, seemed more like a global concept than something that belongs in the material. but I can move it inside if you prefer, doesn't make any difference

luanpotter: in fact it is not even inside the `Material.` namespace, it has it's own `Light.` namespace... that is how it is currently organized

packages/flame_3d/lib/src/resources/light/light.dart
luanpotter: discussed offline - we can address this when we add custom lights

luanpotter:
@wolfenrain removed the group component, please lmk if this makes more sense to you

still keeping the separation of obj3d/comp3d; lmk if you want me to drop that as well.

packages/flame_3d/lib/src/components/light_component.dart
wolfenrain: I do think we should still change this to use the transform by reference right?

wolfenrain:
> @wolfenrain removed the group component, please lmk if this makes more sense to you
> 
> still keeping the separation of obj3d/comp3d; lmk if you want me to drop that as well.

Yea let's park that for now as is, we can always talk about it with the rest of the team later!

Left one comment on an open thread but apart from that LGTM!

packages/flame_3d/lib/src/components/light_component.dart
luanpotter: yes, I can do that!

packages/flame_3d/lib/src/components/group_3d.dart
luanpotter: removed

luanpotter:
@wolfenrain addressed final comments!

wolfenrain (approved):

PR #3192, opened 2024-06-08
fix: Fix text rendering issue where spaces are missing

luanpotter:
# Description

Fix text rendering issue where spaces end up missing.

What happens is that if a single paragraph is split into multiple spans, due to having different formatting, for example, the layouting engine will run into an edge case when trying to place the last span. It will:

* split it and try to put each word at a time with the preceding space
* it will eventually find a word that cannot fit, and return to a partial state
* however, the layouting engine then keeps going and try the same element again. that is because there could be more elements, and the engine conflates bits and elements
* now the element "starts" at the next word, and suddenly it fits, because a new space is not added. this only happens if the final word is such that it fits by itself but doesn't with the leading space.

That is not normally an issue because if it doesn't fit, it will go to the next line and the space becomes unnecessary in that case. But in case we are still trying to populate the same line, we need to make sure we don't "eat up" the space.

There could be other similar edge cases where a space could go missing, so I opted to make it explicit whether the initial space should be added or not by adding a parameter to determine if it is the start of a new line or not.

### Examples

Before:

image

image

After:

image

image

## Checklist

## Breaking Change?


spydon (approved):
Lgtm!

PR #3224, opened 2024-07-17
feat: Enhance TexturePackerSprite

ani3llyon:
# Description

- Add feature to enable to use original size instead always to use packed size
- Support trimmed/stripped whitespace with `useOriginalSize = true`
- Fix issue when rendering rotated image (wrong size and position)

## Checklist

## Breaking Change?

## Related Issues


packages/flame_texturepacker/lib/src/texture_packer_sprite.dart
```diff
@@ -97,42 +150,67 @@ class TexturePackerSprite extends Sprite {
...
+      _tmpRenderPosition.setFrom(_tmpRenderOffset);
+      return super.render(
+        canvas,
+        position: _tmpRenderPosition,
+        size: _tmpRenderSize,
+        overridePaint: overridePaint,
+      );
+    }
 
+    // Calculate and render for non-rotated image, must call function render
```
luanpotter, line 199:
```suggestion
    // Calculate and render for rotated image, must call function render
```

packages/flame_texturepacker/lib/src/texture_packer_sprite.dart
```diff
@@ -97,42 +150,67 @@ class TexturePackerSprite extends Sprite {
...
 
-    _tmpRenderSize.setValues(tempWidth, tempHeight);
+    if (!rotate) {
+      // Calculate and render for non-rotated image, must call function render
+      // from super class with anchor = Anchor.topLeft, because we already
+      // calculated size and position based on Anchor.topLeft to rendered by
+      // super class function.
+      _tmpRenderSize.setFrom(_tmpRenderImageSize);
+      _tmpRenderPosition.setFrom(_tmpRenderOffset);
+      return super.render(
```
luanpotter, line 191: I'm curious why this branch doesn't need the decorator

packages/flame_texturepacker/test/new_format_test.dart
```diff
@@ -44,8 +44,8 @@ void main() {
       expect(jumpSprite.rotate, true);
       expect(jumpSprite.name, 'robot_jump');
       expect(jumpSprite.srcPosition, Vector2(196, 260));
-      expect(jumpSprite.packedHeight, 192);
-      expect(jumpSprite.packedWidth, 256);
+      expect(jumpSprite.packedHeight, 256);
+      expect(jumpSprite.packedWidth, 192);
```
luanpotter, line 48: can we add a test using the new useOriginalSize option and showcasing the difference? just need to assert the size variables

packages/flame_texturepacker/lib/src/texture_packer_sprite.dart
ani3llyon: Sorry my bad..
With update it at next commit.

packages/flame_texturepacker/lib/src/texture_packer_sprite.dart
ani3llyon: Because this branch use non-rotated src image.
Which is at this point the position of the canvas should be like this one.
image

due the src image should be like this one below..
So we do not need decorator for this branch to rotate the src image..
image

<hr>

But if we use rotated src image..
First we calculate for this position..
image

but our src image will be like this
image

So we need to calculate and render to be this position
image

at last we use the decorator to rotate the previous rendered image..
and the result should be like this below
image

packages/flame_texturepacker/test/new_format_test.dart
ani3llyon: Sure, will add it at next commit

packages/flame_texturepacker/lib/src/texture_packer_sprite.dart
ani3llyon: done

spydon (approved):
Lgtm, thanks for your contribution!

packages/flame_texturepacker/lib/src/texture_packer_sprite.dart
luanpotter: makes perfect sense, thanks for the explanation :)

PR #3230, opened 2024-07-20
refactor: Use a temp vector for delta calculations of `FollowBehavior`

ufrshubham:
# Description

This PR add a temp vector in `FollowBehavior` to avoid creating a new vector every frame for the delta calculations.

## Checklist

## Breaking Change?

## Related Issues

NA

### DIFF
--- a/packages/flame/lib/src/camera/behaviors/follow_behavior.dart
+++ b/packages/flame/lib/src/camera/behaviors/follow_behavior.dart
@@ -68,8 +68,9 @@ class FollowBehavior extends Component {
     );
 
     final distance = _tempDelta.length;
-    if (distance > _speed * dt) {
-      _tempDelta.scale(_speed * dt / distance);
+    final deltaOffset = _speed * dt;
+    if (distance > deltaOffset) {
+      _tempDelta.scale(deltaOffset / distance);
     }
     if (_tempDelta.x != 0 || _tempDelta.y != 0) {
       owner.position = _tempDelta..add(owner.position);
--- a/packages/flame/lib/src/camera/behaviors/follow_behavior.dart
+++ b/packages/flame/lib/src/camera/behaviors/follow_behavior.dart
@@ -3,6 +3,7 @@ import 'package:flame/src/camera/viewport.dart';
 import 'package:flame/src/components/core/component.dart';
 import 'package:flame/src/components/position_component.dart';
 import 'package:flame/src/effects/provider_interfaces.dart';
+import 'package:vector_math/vector_math_64.dart';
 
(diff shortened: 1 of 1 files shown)


packages/flame/lib/src/camera/behaviors/follow_behavior.dart
```diff
@@ -59,19 +62,17 @@ class FollowBehavior extends Component {
...
-    final distance = delta.length;
+    _tempDelta.setValues(
+      verticalOnly ? 0 : target.position.x - owner.position.x,
+      horizontalOnly ? 0 : target.position.y - owner.position.y,
+    );
+
+    final distance = _tempDelta.length;
     if (distance > _speed * dt) {
-      delta.scale(_speed * dt / distance);
+      _tempDelta.scale(_speed * dt / distance);
```
luanpotter, line 72: nit: if you want you can extract `_speed * dt` to avoid computing it twice

packages/flame/lib/src/camera/behaviors/follow_behavior.dart
```diff
@@ -59,19 +62,17 @@ class FollowBehavior extends Component {
...
     }
-    if (delta.x != 0 || delta.y != 0) {
-      owner.position = delta..add(owner.position);
+    if (_tempDelta.x != 0 || _tempDelta.y != 0) {
+      owner.position = _tempDelta..add(owner.position);
```
luanpotter, line 75: to go all the way avoiding object creation entirely, this could be owner.position.setValues(owner.position.x + delta.x, ...)

sadly we lose all the easiness of using vectorial math at all when doing things like this

luanpotter (approved):
LGTM

packages/flame/lib/src/camera/behaviors/follow_behavior.dart
ufrshubham: > this could be owner.position.setValues(owner.position.x + delta.x, ...)

I thought the same. But unfortunately doing so causes some of the tests to fail, because the `.setValues` bypasses the `position` setter of `PositionProvider`. Some components do extra stuff in that setter.

packages/flame/lib/src/camera/behaviors/follow_behavior.dart
ufrshubham: Done.

spydon (approved):
Lgtm

packages/flame/lib/src/camera/behaviors/follow_behavior.dart
luanpotter: oof, seems like a footgun waiting to happen :(

PR #3249, opened 2024-07-28
feat: Create ellipse component

arthasyou:
# Description

Replace this text.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/geometry/ellipse_component.dart
+++ b/packages/flame/lib/src/geometry/ellipse_component.dart
@@ -1,0 +1,194 @@
+import 'package:flame/components.dart';
+import 'package:flame/extensions.dart';
+import 'package:flame/geometry.dart';
+import 'package:flame/src/effects/provider_interfaces.dart';
+import 'package:flame/src/math/solve_quadratic.dart';
+import 'package:meta/meta.dart';
+
+class EllipseComponent extends ShapeComponent implements SizeProvider {
+  /// Constructor to create an [EllipseComponent] from width and height.
+  EllipseComponent({
+    double? width,
+    double? height,
+    super.position,
+    super.scale,
+    super.angle,
+    super.anchor,
+    super.children,
+    super.priority,
+    super.paint,
+    super.paintLayers,
+    super.key,
+  }) : super(size: Vector2(width ?? 0, height ?? 0));
+
+  /// Constructor to create an [EllipseComponent] relative to the parent size.
+  /// The width and height are determined by [widthRelation] and [heightRelation]
+  /// which are fractions of the parent's width and height.
+  EllipseComponent.relative(
+    double widthRelation,
+    double heightRelation, {
+    required Vector2 parentSize,
+    super.position,
+    super.scale,
+    super.angle,
+    super.anchor,
+    super.paint,
+    super.paintLayers,
+    super.children,
+    super.key,
+  }) : super(
+          size: Vector2(
+            widthRelation * parentSize.x,
+            heightRelation * parentSize.y,
+          ),
+        );
+
+  @override
+  @mustCallSuper
+  Future<void> onLoad() async {
+    void updateCenterOffset() => _centerOffset = Offset(size.x / 2, size.y / 2);
+    size.addListener(updateCenterOffset);
+    updateCenterOffset();
+  }
+
+  late Offset _centerOffset;
+
+  /// Get the width of the ellipse before scaling.
+  double get width {
+    return size.x;
+  }
+
+  /// Set the width of the ellipse (and therefore the [size]).
+  set width(double value) {
+    size.setValues(value, size.y);
+  }
+
+  /// Get the height of the ellipse before scaling.
+  double get height {
+    return size.y;
+  }
+
+  /// Set the height of the ellipse (and therefore the [size]).
+  set height(double value) {
+    size.setValues(size.x, value);
+  }
+
+  // Used to not create new Vector2 objects every time scaledWidth or scaledHeight is called.
+  final Vector2 _scaledSize = Vector2.zero();
+
+  /// Get the width of the ellipse after it has been sized and scaled.
+  double get scaledWidth {
+    _scaledSize
+      ..setFrom(size)
+      ..multiply(absoluteScale);
+    return _scaledSize.x;
+  }
+
+  /// Get the height of the ellipse after it has been sized and scaled.
+  double get scaledHeight {
+    _scaledSize
+      ..setFrom(size)
+      ..multiply(absoluteScale);
+    return _scaledSize.y;
+  }
+
+  @override
+  void render(Canvas canvas) {
+    if (renderShape) {
+      if (hasPaintLayers) {
+        for (final paint in paintLayers) {
+          canvas.drawOval(
+            Rect.fromCenter(
+              center: _centerOffset,
+              width: scaledWidth,
+              height: scaledHeight,
+            ),
+            paint,
+          );
+        }
+      } else {
+        canvas.drawOval(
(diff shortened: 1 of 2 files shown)


packages/flame/lib/src/geometry/ellipse_component.dart
```diff
@@ -0,0 +1,200 @@
...
+            paint,
+          );
+        }
+      } else {
+        canvas.drawOval(
+          Rect.fromCenter(
+            center: _centerOffset,
+            width: scaledWidth,
+            height: scaledHeight,
+          ),
```
luanpotter, line 121: nit: you can extract the repeated rect definition to avoid duplication

luanpotter (commented):
Can we add some basic tests?
You can look at other shapes tests like RectangleComponentTest
In particular the intersection method

packages/flame/lib/src/geometry/ellipse_component.dart
arthasyou: fixed

arthasyou:
> Can we add some basic tests? You can look at other shapes tests like RectangleComponentTest In particular the intersection method

I can't find the related demo. Could you please provide the specific test demo file location?

In particular, I need to add some basic tests for the intersection method and would like to reference tests like RectangleComponentTest.

Thank you!

luanpotter:
Not a demo, just unit tests: `packages/flame/test/components/rectangle_component_test.dart`

spydon:
@arthasyou do you need some guidance on how to get started with the tests? The rest of this PR looks great :)

arthasyou:
> @arthasyou do you need some guidance on how to get started with the tests? The rest of this PR looks great :)

Yes just need a example

spydon:
@arthasyou here is the simplest one to start with:
https://github.com/flame-engine/flame/blob/refs%2Fheads%2Fmain/packages%2Fflame%2Ftest%2Fcomponents%2Frectangle_component_test.dart

And then intersections etc have to be tested here:
https://github.com/flame-engine/flame/blob/refs/heads/main/packages/flame/test/collisions/collision_detection_test.dart#L259

I think you can just search for `RectangleComponent` in the test directory and follow the same thing as is done for that one.

A follow up that would be great is if we also got a `EllipseHitbox`. 😃

arthasyou:
> @arthasyou here is the simplest one to start with: https://github.com/flame-engine/flame/blob/refs%2Fheads%2Fmain/packages%2Fflame%2Ftest%2Fcomponents%2Frectangle_component_test.dart
> 
> And then intersections etc have to be tested here: https://github.com/flame-engine/flame/blob/refs/heads/main/packages/flame/test/collisions/collision_detection_test.dart#L259
> 
> I think you can just search for `RectangleComponent` in the test directory and follow the same thing as is done for that one.
> 
> A follow up that would be great is if we also got a `EllipseHitbox`. 😃

ty I'll do it

arthasyou:
adding the test file, pls check if these code can work well. ty

packages/flame/test/components/ellipse_component_test.dart
```diff
@@ -0,0 +1,62 @@
...
+        ..anchor = Anchor.center;
+
+      final lineSegment = LineSegment(
+        Vector2(50, 50),
+        Vector2(150, 50),
+      );
+
+      final intersections = component.lineSegmentIntersections(lineSegment);
+      expect(intersections.length, 2);
+      expect(intersections[0].x < intersections[1].x, isTrue);
```
spydon, line 59: Can you check the approximate intersection points here too? 
You can use the `closeToVector` matcher, like here: https://github.com/flame-engine/flame/blob/267d6801cb7e6cbbaa450e24e38aaa7d8fcfc03f/packages/flame/test/collisions/collision_detection_test.dart#L1037

spydon (commented):
A small comment, otherwise the `ellipse_component_test.dart` file looks good, tests in `collision_detection_test.dart` is missing though.

arthasyou:
updated test file

packages/flame/test/components/ellipse_component_test.dart
arthasyou: It's very difficult to write a class similar to CircleCircleIntersections for ellipses. I tried but couldn't figure it out. It seems like the EllipseComponent I wrote isn't very useful.
the CircleCircleIntersections code as bellow
···dart
class CircleCircleIntersections
    extends Intersections<CircleComponent, CircleComponent> {
  @override
  Set<Vector2> intersect(CircleComponent shapeA, CircleComponent shapeB) {
    final centerA = shapeA.absoluteCenter;
    final centerB = shapeB.absoluteCenter;
    final distance = centerA.distanceTo(centerB);
    final radiusA = shapeA.radius;
    final radiusB = shapeB.radius;
    if (distance > radiusA + radiusB) {
      // Since the circles are too far away from each other to intersect we
      // return the empty set.
      return {};
    } else if (distance < (radiusA - radiusB).abs()) {
      // When one circle is contained within the other there is only a collision
      // if the outer circle isn't hollow.
      final outerShape = radiusA > radiusB ? shapeA : shapeB;
      if (outerShape.isSolid) {
        final center = outerShape == shapeA ? centerB : centerA;
        return {center};
      } else {
        return {};
      }
    } else if (distance == 0 && radiusA == radiusB) {
      // The circles are identical and on top of each other, so there are an
      // infinite number of solutions. Since it is problematic to return a
      // set of infinite size, we'll return 4 distinct points here.
      return {
        shapeA.absoluteCenter + Vector2(radiusA, 0),
        shapeA.absoluteCenter + Vector2(0, -radiusA),
        shapeA.absoluteCenter + Vector2(-radiusA, 0),
        shapeA.absoluteCenter + Vector2(0, radiusA),
      };
    } else {
      // There are definitely collision points if we end up in here.
      // To calculate these we use the fact that we can form two triangles going
      // between the center of shapeA, the point in between the shapes which the
      // intersecting line goes through, and then two different triangles are
      // formed with the two intersection points as the last corners.
      // The length to the point in between the circles is first calculated,
      // this is [lengthA], then we calculate the length of the other cathetus
      // [lengthB]. Then the [centerPoint] is calculated, which is the point
      // which the intersecting line goes through in between the shapes.
      // At this point we know the two first points of the triangles, the center
      // of [shapeA] and the [centerPoint], the two third points of the
      // different triangles are the intersection points that we are looking for
      // and we get those points by calculating the [delta] from the
      // [centerPoint] to the intersection points.
      // The result is then [centerPoint] +- [delta].
      final lengthA = (pow(radiusA, 2) - pow(radiusB, 2) + pow(distance, 2)) /
          (2 * distance);
      final lengthB = sqrt((pow(radiusA, 2) - pow(lengthA, 2)).abs());
      final centerPoint = shapeA.absoluteCenter +
          (shapeB.absoluteCenter - shapeA.absoluteCenter) * lengthA / distance;
      final delta = Vector2(
        lengthB *
            (shapeB.absoluteCenter.y - shapeA.absoluteCenter.y).abs() /
            distance,
        -lengthB *
            (shapeB.absoluteCenter.x - shapeA.absoluteCenter.x).abs() /
            distance,
      );
      return {
        centerPoint + delta,
        centerPoint - delta,
      };
    }
  }
}
···

spydon: Yeah, it's a quite hard problem I guess.
Should we close this PR then, or do you want to keep trying?

arthasyou: yeah, you can close it sry about that

spydon: No problem, next person that wants to try can build on your PR :)

PR #3263, opened 2024-08-10
feat: Adding x,y,width and height inputs to position components on Dev Tools

erickzanardo:
# Description

Adds fields to allow developers to change the x, y, width and height of position components from the dev tools.

https://github.com/user-attachments/assets/70f6fe15-488c-4eb2-83bf-6cdc050526d8

## Checklist

- [x] I have followed the [Contributor Guide] when preparing my PR.
- [x] I have updated/added tests for ALL new/updated/fixed functionality.
- [x] I have updated/added relevant documentation in `docs` and added dartdoc comments with `///`.
- [x] I have updated/added relevant examples in `examples` or `docs`.

## Breaking Change?

- [ ] Yes, this PR is a breaking change.
- [x] No, this PR is not a breaking change.

## Related Issues

[Contributor Guide]: https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md
[Conventional Commit]: https://conventionalcommits.org
[CHANGELOG]: https://github.com/flame-engine/flame/blob/main/CHANGELOG.md

PR #3267, opened 2024-08-12
feat: Adding scale and angle to devtools attributes

erickzanardo:
# Description

Adds scaling and angle to the dev tools form. Also fix a caching issue that was presenting outdated values to the user.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame_devtools/lib/widgets/position_component_attributes_form.dart
+++ b/packages/flame_devtools/lib/widgets/position_component_attributes_form.dart
@@ -20,6 +20,10 @@ class PositionComponentAttributesForm extends ConsumerWidget {
       ),
     );
 
+    // We don't want to keep the attributes in cache since the user will most
+    // likely change them often.
+    ref.invalidate(positionComponentAttributesProvider(componentId));
+
     return attributesData.when(
       error: (e, s) => const Text(
         'Error loading component attributes',
@@ -36,6 +40,7 @@ class PositionComponentAttributesForm extends ConsumerWidget {
             Row(
               children: [
                 IncrementalNumberFormField<double>(
+                  key: Key('x_field_$componentId'),
                   label: 'X',
                   initialValue: attributes.x,
                   onChanged: (v) {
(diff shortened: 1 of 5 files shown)


luanpotter (approved):
LGTM

packages/flame_devtools/lib/widgets/position_component_attributes_form.dart
```diff
@@ -20,6 +20,10 @@ class PositionComponentAttributesForm extends ConsumerWidget {
       ),
     );
 
+    // We don't want to keep the attributes in cache since the user will most
+    // likely change them often.
+    ref.invalidate(positionComponentAttributesProvider(componentId));
```
spydon, line 25: Are there more places listening to this provider? Otherwise it will dispose itself once it's not listened to in the tree anymore. I think adding invalidate in the build method is probably not really how it's meant to be used.

erickzanardo: It is used only here. Do you have any suggestion on how to make this provider never cache? I couldn't find any alternative 🙃

spydon: When the widget leaves the tree the state is disposed if it is not listened to, so it shouldn't be cached if it is only used here? :thinking:

erickzanardo: I thought that too. But for some reason, the provider function is being executed only once per component, so if the user changed any values, selected a different component, then returned to the one they changed before, they would see the initial values before their changes, because somehow the provider were cached

spydon: Aaah, autodispose is missing from here:
https://github.com/flame-engine/flame/pull/3263/files#diff-d87cdf5db8e0f1fb3c312e7b2cd3425ca724554c5e01791cdd23033cc299dd42R5

erickzanardo: Right, changed to autoDispose and it is better now!

spydon (approved):
Lgtm!

PR #3276, opened 2024-08-21
fix: ViewportAwareBounds component and lifecycle issues

TheMaverickProgrammer:
# Description
The behavior for `CameraComponent.setBounds` was not behaving correctly. The viewport-aware behavior was not triggering until a window resize event. All supported bound variants were not respecting the fixed resolution viewport, or in some cases, behaving unpredictably.

The issue was with the fact the viewport-aware behavior component depends on the bounded position component to be in its parent, but was returning `null` during `onMount`. The viewport math was using the logical size instead of the virtual size. I made some math changes to be accurate.

This PR adds new a getter `CameraComponent.considerViewport`. `ViewportAwareBoundsBehavior` is now added as a side effect of mounting `BoundedPositionBehavior` by waiting for the `mounted` future. This guarantees that 
Flames life cycle is respected and the components behave as expected. 

Tests pass on my own local project. 

Videos are linked on the discord thread [here](https://discord.com/channels/509714518008528896/1275814019235709003/1275888732481785925).

Melos dry run completed successfully.

## Checklist

## Breaking Change?

### Migration instructions
No work to be done.

## Related Issues
https://github.com/flame-engine/flame/pull/2769

## QUESTIONS FOR THE DEVS
So I'm not satisfied with the fixes for `Circle` bounds. The bounds uses the maxima of the viewport, and yes, creates a circle that stays within the size of the bounds circle, but I'd expect that `Circle` bounds should keep my viewport _inside_ the circle as it is documented to stay _inside_ the desired bounds shape. Therefore, I think `Circle` case should have 4 incident points to fit the viewport. That is to say, the viewport is not the maxima of the newly calculated `Circle`, but will be fully contained by this new `Circle`. This way, you'll never see outside of the bounds as it, seems to me anyway, implies. However I do not know if this is intuitive for others and expected behavior. Thoughts?

### DIFF
--- a/packages/flame/lib/src/camera/camera_component.dart
+++ b/packages/flame/lib/src/camera/camera_component.dart
@@ -371,9 +371,13 @@ class CameraComponent extends Component {
     _considerViewport = considerViewport;
 
     final boundedBehavior = viewfinder.firstChild<BoundedPositionBehavior>();
+    final viewPortAwareBoundsBehavior =
+        viewfinder.firstChild<ViewportAwareBoundsBehavior>();
 
     if (bounds == null) {
+      // When bounds is null, all bounds-related components need to be dropped.
       boundedBehavior?.removeFromParent();
+      viewPortAwareBoundsBehavior?.removeFromParent();
       return;
     }
 
@@ -383,6 +387,26 @@ class CameraComponent extends Component {
       );
     } else {
       boundedBehavior.bounds = bounds;
+    }
+
+    if (!considerViewport) {
+      // Edge case: remove pre-existing viewport aware components.
+      viewPortAwareBoundsBehavior?.removeFromParent();
+      return;
+    }
+
+    // Param `considerViewPort` was true and we have a bounds.
+    // Add a ViewportAwareBoundsBehavior component with
+    // our desired bounds shape or update the boundsShape if the
+    // component already exists.
+    if (viewPortAwareBoundsBehavior == null) {
+      viewfinder.add(
+        ViewportAwareBoundsBehavior(
+          boundsShape: bounds,
+        ),
+      );
+    } else {
+      viewPortAwareBoundsBehavior.boundsShape = bounds;
     }
   }
 
--- a/packages/flame/lib/src/camera/camera_component.dart
+++ b/packages/flame/lib/src/camera/camera_component.dart
@@ -122,17 +122,6 @@ class CameraComponent extends Component {
 
   Viewfinder _viewfinder;
 
-  /// The [considerViewport] flag is read-only and cannot be set except through
-  /// [setBounds] as an optional parameter. If true, this represents that the
-  /// camera's [viewfinder] will recalculate your desired bounds with respect
-  /// to your viewport dimensions by ensuring that [ViewportAwareBoundsBehavior]
-  /// component is mounted correctly when the tree updates.
-  ///
-  /// The default value is false.
-  bool get considerViewport => _considerViewport;
-
-  bool _considerViewport = false;
-
   /// Special component that is designed to be the root of a game world.
   ///
   /// Multiple cameras can observe the same [world] simultaneously, and the
@@ -370,8 +359,6 @@ class CameraComponent extends Component {
   /// Note that this option only works with [Rectangle], [RoundedRectangle] and
   /// [Circle] shapes.
   void setBounds(Shape? bounds, {bool considerViewport = false}) {
-    _considerViewport = considerViewport;
-
     final boundedBehavior = viewfinder.firstChild<BoundedPositionBehavior>();
     final viewPortAwareBoundsBehavior =
         viewfinder.firstChild<ViewportAwareBoundsBehavior>();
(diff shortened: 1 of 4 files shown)


packages/flame/lib/src/camera/viewfinder.dart
```diff
@@ -239,6 +241,37 @@ class Viewfinder extends Component
...
+            ViewportAwareBoundsBehavior(
+              boundsShape: child.bounds,
+            ),
+          );
+        }
+      } else {
+        viewPortAwareBoundsBehavior?.removeFromParent();
+      }
+    }
+  }
```
spydon, line 273: I would do this within the `setBounts` by listening to the `mounted` completer of the `BoundedPositionBehavior` instead.

spydon (commented):
I think the current behavior of the circle is fine, the behavior's main function is to restrict the movement of the viewfinder (the clipping can be done elsewhere).

Some tests would be good, so that we don't end up with regressions here later.
Here is the previous (very lacking) test:
https://github.com/flame-engine/flame/blob/main/packages/flame/test/camera/behaviors/viewport_aware_bounds_behavior_test.dart

packages/flame/lib/src/camera/viewfinder.dart
TheMaverickProgrammer: I did at first but the code looked messy since the bound behavior can also be mounted already and that needed its own code-path to handle. The code was much simpler to read and understand here. Also, `ViewportAwareBoundsBehavior` strictly depends on the existence of `BoundedPositionBehavior` so trying to treat them as two separate classes seems incorrect to me. Putting the setup code here illustrates their relationship better.

If this is a final decision, I can revisit it tomorrow and move it back to the camera component `setBounds`. Lemme know. 👍

spydon: I still think that it is better to keep the code clean by not having dependencies in both directions here.
This way is also more expensive, even though it doesn't matter that much in the viewfinder, since it's quite rare to add many components it.

TheMaverickProgrammer: What do you think about merging `ViewportAwareBoundsBehavior` into `BoundedPositionBehavior` and tossing the first?

spydon: The `BoundedPositionBehavior` can be used without the `ViewportAwareBoundsBehavior`, it can target anything that is a `PositionProvider`.

TheMaverickProgrammer: Understood.

spydon:
@TheMaverickProgrammer any update on this? :)

TheMaverickProgrammer:
Need to test something I forgot before pushing. I will re-open after verifying the mount operation happens correctly.

packages/flame/lib/src/camera/camera_component.dart
```diff
@@ -120,6 +122,17 @@ class CameraComponent extends Component {
...
   Viewfinder _viewfinder;
 
+  /// The [considerViewport] flag is read-only and cannot be set except through
+  /// [setBounds] as an optional parameter. If true, this represents that the
+  /// camera's [viewfinder] will recalculate your desired bounds with respect
+  /// to your viewport dimensions by ensuring that [ViewportAwareBoundsBehavior]
+  /// component is mounted correctly when the tree updates.
+  ///
+  /// The default value is false.
+  bool get considerViewport => _considerViewport;
```
spydon, line 132: Maybe this should be clarified in the name that it is related to the viewfinder, since this is in the `CameraComponent`. Or maybe the variable should live in the viewfinder instead?

spydon (commented):
Just one small comment, other than that it looks good

packages/flame/lib/src/camera/camera_component.dart
TheMaverickProgrammer: Since it relates to a pre-existing convenience for the end-user (b/c of the effect of `setBounds(considerViewport: true)`), adding a comment that this getter is related to the camera's view finder seems reasonable.

spydon: This getter didn't exist before though, in the `setBounds` method it is clear that `considerViewport` is related to the bounds, meanwhile it isn't in the general `CameraComponent`.

TheMaverickProgrammer: What do you want to name it? The only thing that comes to mind is some variant of `hasBoundsAwareViewport`.

spydon: Yeah, I can only come up with long names too, so I think it would be more natural to move it to the viewfinder, it's not something that the user should have to access often anyways (if ever).

TheMaverickProgrammer: Done

spydon: Super, thanks!

spydon (approved):
Lgtm! Thanks for your contribution

Issue #3293, opened 2024-09-01
SpriteButtonComponent throws error with button setter. (Unhandled Exception: Null check operator used on a null value) 


puneet-739:
### What happened?

I am using `SpriteButtonComponent`, with flame (v1.18.0), everything was fine, but after updating to the latest flame version (v1.19.0), I got this error :-

E/flutter ( 8758): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Null check operator used on a null value
E/flutter ( 8758): #0      SpriteGroupComponent.updateSprite (package:flame/src/components/sprite_group_component.dart:103:13)
E/flutter ( 8758): #1      SpriteButtonComponent.button= (package:flame/src/components/input/sprite_button_component.dart:50:5)
E/flutter ( 8758): #2      DotButtonComponent.onLoad (package:pitu/pitu.dart:31:5)
E/flutter ( 8758): #3      Component._startLoading (package:flame/src/components/core/component.dart:856:26)
E/flutter ( 8758): #4      Component._addChild (package:flame/src/components/core/component.dart:606:20)
E/flutter ( 8758): #5      Component.add (package:flame/src/components/core/component.dart:568:46)
E/flutter ( 8758): #6      Pitu.onLoad (package:pitu/pitu.dart:22:5)
E/flutter ( 8758): <asynchronous suspension>
E/flutter ( 8758): #7      FlameGame.load (package:flame/src/game/flame_game.dart:105:5)
E/flutter ( 8758): <asynchronous suspension>
E/flutter ( 8758): #8      GameWidgetState.loaderFuture.<anonymous closure> (package:flame/src/game/game_widget/game_widget.dart:194:9)
E/flutter ( 8758): <asynchronous suspension>
E/flutter ( 8758): #9      _FutureBuilderState._subscribe.<anonymous closure> (package:flutter/src/widgets/async.dart:638:31)
E/flutter ( 8758): <asynchronous suspension>
E/flutter ( 8758): 

### What do you expect?

Just expecting my code to run error free after updating flame engine `: )`
`''''`
`(oo)`
`/  \`
` ||`

### How can we reproduce this?

step 1: create a sample project in flutter. 
`flutter create <project_name>`
step 2: add/update flame latest version (v1.19.0)
`flutter pub add flame`
step 3: create a class with extends `SpriteButtonComponent` class.
step 4: in the onLoad() method, set the button setter with a `Sprite` object.
[...]

PR #3302, opened 2024-09-13
fix: SpriteButtonComponent to initialize sprites in `onLoad`

puneet-739:
# Description

Just replaced the null check operator ```!``` with null aware ```?```. 
```void updateSprite(T key, Sprite sprite) {```
```_sprites?[key] = sprite;```
``` _resizeToSprite();```
```}```
the value of ```_sprites``` set on the ```mount()``` method of the ```SpriteButtonComponent``` class

## Checklist

- [x] I have followed the [Contributor Guide] when preparing my PR.
- [ ] I have updated/added tests for ALL new/updated/fixed functionality.
- [x] I have updated/added relevant documentation in `docs` and added dartdoc comments with `///`.
- [ ] I have updated/added relevant examples in `examples` or `docs`.

## Breaking Change?

- [ ] Yes, this PR is a breaking change.
- [x] No, this PR is not a breaking change.

## Related Issues

#3293 

[Contributor Guide]: https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md
[Conventional Commit]: https://conventionalcommits.org
[CHANGELOG]: https://github.com/flame-engine/flame/blob/main/CHANGELOG.md

PR #3318, opened 2024-09-21
test: Test case for extended `SpriteButtonComponent`

puneet-739:
# Description

Added an extra test case for testing `SpriteButtonComponent` class. 
which includes a `CustomSpriteButtonComponet` class to test the working with `onLoad` instead of directly using Constructor.

## Checklist

## Breaking Change?

## Related Issues

It may help working on the issues like  #3293 


spydon (commented):
Tests for the `SpriteButtonComponent` should live in the `sprite_button_component_test.dart` file.

spydon (commented):
These tests are just copies of already existing tests, but going through an extended version that isn't exposed to the user. I don't think we want this type of testing pattern in the code-base.
If you disagree with a good argument I can open it again, but I'll close it for now.

puneet-739:
> These tests are just copies of already existing tests, but going through an extended version that isn't exposed to the user. I don't think we want this type of testing pattern in the code-base. If you disagree with a good argument I can open it again, but I'll close it for now.

Oh, ok.
Actually, I've added the test cases due to the issue #3293. (I think there is still an issue) Please refer to this comment https://github.com/flame-engine/flame/pull/3302#discussion_r1769626058
The reason of adding the test cases is to test the working of `SpriteButtonComponent` with extended version. because the issue #3293 only occurs when the user use the extended version but not on the Constructor.
So, we need to test if the issue is resolved for the extended version as well, as the present test cases only test for Constructor version, it will always give positive result (it gives error if we use the extended version).

spydon:
@puneet-739 you're right, can you try https://github.com/flame-engine/flame/pull/3327? It should solve the issue so that the user doesn't have to call `super.onLoad` last.

spydon:
@puneet-739 I changed my mind, due to the intricacies of the `SpriteButtonComponent` it's probably good to have these tests so that we don't get any regressions later. Can you fix the DCM warning? (Make the `CustomSpriteButtonComponent` private)

packages/flame/test/components/input/sprite_button_component_test.dart
```diff
@@ -228,6 +404,69 @@ Future<void> main() async {
...
+      },
+    );
+  });
+}
+
+/// test class [_CustomSpriteButtonComponent] extending
+/// the [SpriteButtonComponent] class,
+/// To test [SpriteButtonComponent] without using Constructor
+/// but setting properties in [onLoad]
+///
```
spydon, line 447:
```suggestion
/// This is used to test [SpriteButtonComponent] without using the constructor
/// and setting properties in [onLoad] of an extending class instead.
```

spydon (approved):

PR #3322, opened 2024-09-24
feat!: New `AlternatePattern` argument for `SequenceEffect`

TheMaverickProgrammer:
* This PR fixes math in `MoveToEffect` so that negative progress did not reflect the direction of the delta offset vec. 
* Corrected progress getter overload in `SequenceEffect` to correctly map to the progress when `_index < 0`. 
* Added `onStart()` in `recede()` initial condition so that effects like `MoveToEffect` can recalculate their initial data.
* Removed unnecessary math in `DurationEffectController` to avoid potential negative zeros.
* Introduced a new type `AlternatePattern` which allows users to include or exclude the last effect in the sequence when the pattern reverses. **NOTE:** When playing forward, the last effect is `List.last`, but in reverse it is `List.first`.
* Default value is now `AlternatePattern.repeatLast`.
* Made the TiledGame example project more aesthetically interesting.
    * The camera now pans to all 4 corners WRT the bounds of the loaded map and uses the new `AlternatePattern` feature along with all the new fixes.

# TO DISCUSS
1. I have skipped the test checkbox until I had a review and we could discuss these changes.
2. We may need to consider an `onEnd()` for `Effect`s. Or provide `onStart()` with an initial `progress` value to calculate from.

## Breaking Change?

## Related Issues
I looked but didn't immediately see anyone reporting these issues. I discovered them myself.


examples/lib/stories/effects/sequence_effect_example.dart
```diff
@@ -28,7 +28,7 @@ class SequenceEffectExample extends FlameGame {
               ScaleEffect.by(Vector2.all(1.5), duration(0.7)),
               MoveEffect.to(Vector2(400, 500), duration(0.7)),
             ],
-            alternate: true,
+            alternatePattern: AlternatePattern.excludeLast,
```
spydon, line 31: This should be done additionally to `alternate`, so that it's not a breaking change, and also follows the same pattern as the other effects.

packages/flame/lib/src/effects/move_to_effect.dart
```diff
@@ -40,7 +40,7 @@ class MoveToEffect extends MoveEffect {
   @override
   void apply(double progress) {
     final dProgress = progress - previousProgress;
-    target.position += _offset * dProgress;
+    target.position += _offset * dProgress.abs();
```
spydon, line 43: Hmm, is this really correct?

packages/flame/lib/src/effects/sequence_effect.dart
```diff
@@ -22,6 +22,32 @@ EffectController _createController({
...
+/// of the alternate pattern, effectively playing it twice in a row.
+///
+/// [AlternatePattern.excludeLast] will not replay the last [Effect] and will
+/// jump to the second-to-last [Effect], if available, at the start of the
+/// alternate pattern instead, effectively playing the last [Effect] only once
+/// throughout the original and alternating pattern.
+enum AlternatePattern {
+  none(0),
+  includeLast(1),
+  excludeLast(2);
```
spydon, line 45: If `inifinite` (or N runs) is used together with `alternate`, shouldn't the first also be excluded on runs that are after the first and not the last?

spydon (commented):
Good idea, it needs tests and docs though.

packages/flame/lib/src/effects/move_to_effect.dart
TheMaverickProgrammer: After thinking about it some more yesterday I think the issue wasn't the reflected vector itself, but rather that the direction vector was incorrect when the camera has a viewport-aware bounds behavior. This causes it to never reach its destination and so the reflective vector is incorrect. That's why taking the absolute value was ok after I added the step to recalculate the vector.

The camera won't be the only component that may never complete its effect with regards to `MoveToEffect` so this matter should be resolved in the PR imo.

spydon: But this will potentially overshoot the destination right? And also it won't be matching the time anymore?
If you are at 0.99 progress and the delta progress is -0.02 when the effect is going forward it should have moved 0.03 "steps", since it'll start going back again, but with this it will simply move to 1.01.

TheMaverickProgrammer: I'm not sure I'm understanding your question precisely, but it seems that you've noticed the possibility of under-fitting the expected step amounts, time-wise. Yes this does happen with the viewport-aware behavior as described (the camera _cannot_ exceed the bounds, and therefore never arrives to its destination). ATM with the changes I've added, I have not seen any noticeable problems.

packages/flame/lib/src/effects/sequence_effect.dart
TheMaverickProgrammer: I've corrected the default pattern behavior. The alternating pattern now plays both the first and last `Effect` elements in the list, instead of just the last element as it did before this PR. This default behavior is `AlternatePattern.repeatLast`. The term "last" used here is relative to the playback direction.

To omit placing first and last twice-in-a-row when the sequence reverses, `AlternatePattern.doNotRepeatLast` is appropriate.

TheMaverickProgrammer:
There's still the topic to discuss: I'm in favor of providing `onStart()` with an initial progress value to calculate the starting conditions from since `MoveEffect` can be affected by bounds checks.

packages/flame/lib/src/effects/move_to_effect.dart
spydon: Simply put; I don't think it will always be at the expected place at each time step when `abs` is used, that was what I tried to describe in the last comment, the delta doesn't become correct. If it overshoots its destination it should start going back, in the cases where it is alternating.

packages/flame_tiled/example/.gitignore
```diff
@@ -0,0 +1,43 @@
+# Miscellaneous
```
spydon, line 1: This gitignore shouldn't be needed, there is already one in the root of the repo that should be covering everything.

packages/flame/lib/src/effects/move_to_effect.dart
TheMaverickProgrammer: We make explicit use of `onStart` so that we are recalculating the vector from the equation `v = to - from` even when the pattern has reversed. This means our progress values shouldn't be negative otherwise it will reverse the vector `v`, which already has correct directionality in its scalar's sign from the explicit call to `onStart`.

spydon: It does sound like it makes sense but I'm on my phone, but if that is what is happening, then how does it then make sense to make it `abs`?

TheMaverickProgrammer: `progress` will be decrementing `dt` units from `1.0` during the reverse phase of the alternating pattern, while `previousProgress` retains the `progress+dt` value form before. Therefore it would be negative, which we don't want to use. It's better to recalculate the starting position and use positive scalars instead.

spydon: But it'll surely overshoot the end value then?

TheMaverickProgrammer: Hi. No it never does for this effect or any as far as I can tell.

packages/flame/lib/src/effects/sequence_effect.dart
```diff
@@ -124,9 +173,21 @@ class _SequenceEffectEffectController extends EffectController {
...
+    // Abort early
+    if (totalDuration == 0.0) {
+      return totalDuration;
+    }
+
     if (alternate) {
       totalDuration *= 2;
+
+      if (alternatePattern == AlternatePattern.doNotRepeatLast) {
+        totalDuration -= effects.first.controller.duration ?? 0;
```
spydon, line 186: How come the first one is removed here too? :thinking:

packages/flame/lib/src/effects/sequence_effect.dart
```diff
@@ -136,7 +197,7 @@ class _SequenceEffectEffectController extends EffectController {
   }
 
   @override
-  double get progress => (_index + 1) / n;
+  double get progress => (_index < 0 ? -_index : _index + 1) / n;
```
spydon, line 200: Can you explain this a bit? In what cases is the `_index` negative? This doesn't seem like it would follow how `progress` works for the rest of the effects.

packages/flame/lib/src/effects/move_to_effect.dart
spydon: In the case that `dProgress` becomes negative it is supposed to go backwards. Say that the previous progress is 0.9 and it it supposed to move backwards to 0.8, then `dProgress` should be `-0.1`, not `0.1`.
Try running only the the `MoveToEffect` with `alternate: true`.

packages/flame/lib/src/effects/sequence_effect.dart
TheMaverickProgrammer: Hi. This has been explained in the PR, code comments, and here:
https://github.com/flame-engine/flame/pull/3322#discussion_r1828604667

packages/flame/lib/src/effects/sequence_effect.dart
TheMaverickProgrammer: The way the `SequenceEffect` has been written for some time now is that it uses a negative `_index` to determine if the pattern is in its reverse phase.

TheMaverickProgrammer:
```
In the case that dProgress becomes negative it is supposed to go backwards. Say that the previous progress is 0.9 and it it supposed to move backwards to 0.8, then dProgress should be -0.1, not 0.1.
Try running only the the MoveToEffect with alternate: true.
```

Please try running the `tiled_example/lib/main.dart` example with from my PR and compare it with the same example in upstream after adding the `ViewportAwareBounds` to the camera in the upstream version. From our conversation, this issue seems to be missed and it's crucial to understanding the problem that is being addressed with the addition to `onStart()` in the `recede()` function and using a non-negative duration value.

spydon:
I'll give it a proper try when I'm back home again next week!

spydon:
@TheMaverickProgrammer I haven't had the time to test it properly yet unfortunately, but there is a test failing, could you have a look at that first?
```
 test/effects/sequence_effect_test.dart 271:9
 ```

TheMaverickProgrammer:
> @TheMaverickProgrammer I haven't had the time to test it properly yet unfortunately (forgot that it was on my table 😅), but there is a test failing, could you have a look at that first?
> 
> ```
>  test/effects/sequence_effect_test.dart 271:9
> ```

The test failing is related to the issue at hand. I have thought a lot about it and can propose a quick solution but it's not ideal: There can be a `MoveTo` effect and a `MoveToWithCorrection` effect. The latter can be used for components that may not arrive at their destination and need to be adjusted when the effect begins. This new effect would update its deltas every time `onStart()` is invoked (see PR for the additional callsite for `onStart()`) and the former effect would simply cache its delta the first time `onStart()` is ran. This way the test can pass and then this `MoveToWithCorrection` component can be used on components like the `Camera` when `ViewportAwareBounds` behavior is used.

spydon:
> There can be a MoveTo effect and a MoveToWithCorrection effect.

Hmm, I'm not too fond of that idea, it feels like a less clean API.
How come that it currently passes the test? I didn't think `MoveToEffect` had any correction built in to it, since it wasn't designed to stack with multiple effects of the same type (like `MoveByEffect` is)?

TheMaverickProgrammer:
> > There can be a MoveTo effect and a MoveToWithCorrection effect.
> 
> Hmm, I'm not too fond of that idea, it feels like a less clean API. How come that it currently passes the test? I didn't think `MoveToEffect` had any correction built in to it, since it wasn't designed to stack with multiple effects of the same type (like `MoveByEffect` is)?

Because in the PR I said I added a new callsite to `onStart()` for effects in order to allow `MoveTo` effect to recalculate its starting position whenever it was unable to ensure its component arrived to its destination. This edge case happens in the camera with viewport aware bounds behavior but can potentially happen for other effects/behaviors that dictate a component's position.

spydon:
@TheMaverickProgrammer will you continue on this PR at some point, or what is the state of it? :)

spydon:
> It seems to be the case that the proposed solutions would be less than ideal. I provided context as to why these would be needed but I did not hear back.

Ah, sorry for that! Thanks for the quick reply. :)

TheMaverickProgrammer:
> > It seems to be the case that the proposed solutions would be less than ideal. I provided context as to why these would be needed but I did not hear back.
> 
> Ah, sorry for that! Thanks for the quick reply. :)

I can implement whatever changes seem best but I would need your explicit instructions in order to quickly resolve the matter.

packages/flame/lib/src/effects/move_to_effect.dart
spydon: @TheMaverickProgrammer maybe it would be best to split this PR into two? First one that fixes the `alternate: true` for the `SequenceEffect` and then one that introduces the alternation modes?

TheMaverickProgrammer: I can do that. I'll move the proposed fixes to a new PR and leave the `alternate` change set in this PR.

spydon:
> I can implement whatever changes seem best but I would need your explicit instructions in order to quickly resolve the matter.

So basically:
* One PR that fixes `alternate: true`. This PR shouldn't change how the delta progress behaves compared to other effects, which I think should be possible, since `alternate` works for other effects. `alternate` is obviously a bit more complicated here though, when it involves propagating the progress changes to the effects in the chain.
* A follow-up PR that introduces the `AlternatePattern`, which  I think shouldn't be too hard if the first PR keeps that functionality in mind.

TheMaverickProgrammer:
> > I can implement whatever changes seem best but I would need your explicit instructions in order to quickly resolve the matter.
> 
> So basically:
> 
>     * One PR that fixes `alternate: true`. This PR shouldn't change how the delta progress behaves compared to other effects, which I think should be possible, since `alternate` works for other effects. `alternate` is obviously a bit more complicated here though, when it involves propagating the progress changes to the effects in the chain.
> 
>     * A follow-up PR that introduces the `AlternatePattern`, which  I think shouldn't be too hard if the first PR keeps that functionality in mind.

Roger. I'm on vacation now but when I'm home I'll start. Should I go ahead and close this branch and just reference it in the next one?

spydon:
> Roger. I'm on vacation now but when I'm home I'll start. Should I go ahead and close this branch and just reference it in the next one?

Sounds good! 🚀

TheMaverickProgrammer:
Closing PR to start two other, modular PRs. This PR will be closed but used as a reference in the next two.

PR #3334, opened 2024-10-10
feat: Adding spawnWhenLoaded flag on SpawnComponent

erickzanardo:
# Description

Adds a new flag on `SpawnComponent` to spawn a component when it is loaded, similar to how we have on `TimerComponent`. 

Its default is set to false to avoid breaking the current behaviour.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/components/spawn_component.dart
+++ b/packages/flame/lib/src/components/spawn_component.dart
@@ -26,6 +26,7 @@ class SpawnComponent extends Component {
     this.within = true,
     this.selfPositioning = false,
     this.autoStart = true,
+    this.spawnWhenLoaded = false,
     Random? random,
     super.key,
   })  : assert(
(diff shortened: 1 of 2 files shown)


packages/flame/lib/src/components/spawn_component.dart
```diff
@@ -107,6 +109,9 @@ class SpawnComponent extends Component {
   /// Whether the timer automatically starts or not.
   final bool autoStart;
 
+  /// Whether the timer should start when the [SpawnComponent] is loaded.
+  final bool spawnWhenLoaded;
```
spydon, line 113: `spawnWhenMounted` I would say is more accurate?

erickzanardo: But this happens on the onLoad phase doesn't it? That is when the `TimerComponent` is being created. Also this name would be consistent with `tickWhenLoaded` from the TimerComponent.

spydon: Hmm, true, it should definitely be consistent. I think in reality it will happen when it is mounted though, because `onLoad` of `SpawnComponent` won't wait until the `TimerComponent` is mounted. But keep it like it is, it matters very little to the user.

spydon (approved):

Issue #3333, opened 2024-10-10
raytrace outside of circle?


dailei2018:
### What happened?

https://examples.flame-engine.org/      Raytracing page，why outside of circle.
![ray](https://github.com/user-attachments/assets/17f6910e-6fad-4278-b3c2-35805e545a7e)

### What do you expect?

...

### Execute in a terminal and put output into the code block below

Output of: flutter doctor -v

### Affected platforms

All

### Are you interested in working on a PR for this?

- [ ] I want to work on this

spydon:
Interesting, I wonder why this has started to happen all of a sudden, we haven't made any changes to that code for a long while. Maybe some subtle changes with float numbers in Dart maybe... I'll have a closer look later.

spydon:
This is fixed and released in Flame v1.21.0 :)

PR #3341, opened 2024-10-13
fix: Ray should not be able to escape `CircleHitbox`

spydon:
# Description

A raycasted ray should not be able to escape a `CircleHitbox` if it originates from inside of it (or on the edge and has a direction inwards).
Fixes: #3333

## Checklist

## Breaking Change?

## Related Issues


erickzanardo (approved):

packages/flame/lib/src/collisions/hitboxes/circle_hitbox.dart
```diff
@@ -56,6 +57,13 @@ class CircleHitbox extends CircleComponent with ShapeHitbox {
   }) {
     var isInsideHitbox = false;
     _temporaryLineSegment.from.setFrom(ray.origin);
+    // Adding a small value to the origin to avoid the ray to be on the edge
+    // of the circle and then directly intersecting and causing the reflecting
+    // ray to go in the wrong direction.
+    _temporaryOrigin.setValues(
+      ray.origin.x + ray.direction.x * 0.00001,
```
luanpotter, line 64: should we put that into an `epsilon`?

luanpotter (approved):
Curious though what changed that caused this to break? And couldn't it be that the point started in a "good" place but when you add the small epsilon it ends in a "bad" place?

wolfenrain (approved):

spydon:
> Curious though what changed that caused this to break? And couldn't it be that the point started in a "good" place but when you add the small epsilon it ends in a "bad" place?

It can happen, but it's very unlikely since the epsilon is so small. So the problem was that it properly calculates the first ray, but then it's calculating what the reflection ray will intersect and since its origin is straight on the edge of the circle sometimes the origin point was counted as the intersection, so we add a small vector to origin with the same direction as the ray to "get out" of the origin.

PR #3358, opened 2024-10-28
fix: Remove extra `implements SizeProvider`s

HosseinYousefi:
I accidentally came across this. `PositionComponent` already implements `SizeProvider` but as it was added later it wasn't cleaned up everywhere.


spydon (approved):
Super, thanks for your contribution!

PR #3407, opened 2024-12-14
fix: Unmultiply before brightening

jtmcdole:
The image data read back is in a pre-multiplied state. Convert it back to straight alpha before brightening and then premultiply.

This just makes the colors go to white though:
image

You could alternatively use the HSV color space and increase the Value (brightness); this would be more subtle and keep the colors correct.

image

### DIFF
--- a/packages/flame/lib/src/extensions/image.dart
+++ b/packages/flame/lib/src/extensions/image.dart
@@ -4,6 +4,7 @@ import 'dart:ui';
 
 import 'package:flame/extensions.dart';
 import 'package:flame/palette.dart';
+import 'package:flutter/painting.dart';
 
 export 'dart:ui' show Image;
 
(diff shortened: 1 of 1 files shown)


packages/flame/lib/src/extensions/image.dart
```diff
@@ -78,17 +78,37 @@ extension ImageExtension on Image {
...
-        pixelData[i + 0],
-        pixelData[i + 1],
-        pixelData[i + 2],
-      ).brighten(amount);
-
-      newPixelData[i] = (color.r * 255).round();
-      newPixelData[i + 1] = (color.g * 255).round();
-      newPixelData[i + 2] = (color.b * 255).round();
-      newPixelData[i + 3] = (color.a * 255).round();
+      final a = pixelData[i + 3] / 255;
```
luanpotter, line 81: do we need to do this on darken as well? [even though the issue is not visible with the example image on darken I imagine there could be similar thing going on]
or even put it on `pixelsInUint8()` so everyone gets it

jtmcdole: Yes, I believe you want to operate on unmultiplied alpha to be accurate. I'll peek at the CI checks later.

jtmcdole: I did want to ask: why not use HSL and increase L by the amount? I know the stack overflow comment makes it sound incorrect, but its more correct because it shouldn't be changing the hue or saturation. That being said, these functions are "brighten" and "darken" which I'm not clear on what the desired results are (white-out of colors?)

jtmcdole: Also; what is "ammount" saying here? Are you wanting each pixel's individual's brightness to increase by some percentage amount (0 to 1) or are you trying to level all the brightness to a set level?  If its individual; then 0 to 1 needs to be over the range of <pixel brightness>-<1.0> with 0 to 1 covering that.

spydon: > I did want to ask: why not use HSL and increase L by the amount? I know the stack overflow comment makes it sound incorrect, but its more correct because it shouldn't be changing the hue or saturation. That being said, these functions are "brighten" and "darken" which I'm not clear on what the desired results are (white-out of colors?)

Sounds good to be, I think this was just our naive attempt at the implementation without doing any research. 😅
But maybe we should just be redirecting to those `painting` methods you mentioned?

jtmcdole: I switched to HSL color from Painting and updated the `darken` call.  The test errors and everything else has nothing to do with this code 😅

spydon: > Also; what is "ammount" saying here? Are you wanting each pixel's individual's brightness to increase by some percentage amount (0 to 1) or are you trying to level all the brightness to a set level?  If its individual; then 0 to 1 needs to be over the range of <pixel brightness>-<1.0> with 0 to 1 covering that.

Makes sense, I think we could also do a breaking change here since it hasn't really worked before. Or deprecate this method and create a new one if we have some name that could fit.

jtmcdole: I think I captured what was previously done here lighten = (color + (1 - color) * amount; darken = color * amount.

luanpotter:
Thank you so much @jtmcdole !
I will merge this as is to Spydon's branch so I can work on it more easily, and then fix the analyzer stuff.
I also want to extract the shared code between brighten and darken.

PR #3426, opened 2024-12-29
feat: Add support for strike-through text for flame_markdown

luanpotter:
# Description

image

Add support for strike-through text for flame_markdown, if enabled by the user.

Basically parses the `del` HTMl tag and maps it to a new inline text renderer in Flame's text rendering pipeline.

The style can be controlled with a new property if desired.

In order to parse `~~`-wrapped text into the new node element, the underlying markdown parser must have the strikethrough option enabled (or an equivalent custom option), which can be controlled by the user if providing their own document.

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame_markdown/example/pubspec.yaml
+++ b/packages/flame_markdown/example/pubspec.yaml
@@ -11,6 +11,7 @@ dependencies:
   flame_markdown: ^0.2.2+3
   flutter:
     sdk: flutter
+  markdown: ^7.1.1
 
 dev_dependencies:
   flame_lint: ^1.2.1
--- a/.github/.cspell/words_dictionary.txt
+++ b/.github/.cspell/words_dictionary.txt
@@ -27,6 +27,7 @@ rebalance
 refreshable
 renderable
 rescan
+strikethrough # of a text with a horizontal line across
 tappable
 tappables
 toolset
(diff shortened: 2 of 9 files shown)


luanpotter:
Note: according to [Wikipedia](https://en.wikipedia.org/wiki/Strikethrough), "Strikethrough" is correct as a single word so I added to the dictionary but happy to rename everything to "Strike Through" if that is preferred.

packages/flame_markdown/example/pubspec.yaml
```diff
@@ -11,6 +11,7 @@ dependencies:
   flame_markdown: ^0.2.2+3
   flutter:
     sdk: flutter
+  markdown: ^7.1.1
```
luanpotter, line 14: note: in order to provide custom documents, a user would need to add this as a direct dependency. another option is that we export all the relevant classes on flame_markdown; not sure what would be the best practice here

spydon (approved):

packages/flame_markdown/example/pubspec.yaml
spydon: If there aren't that many I think we should export them

luanpotter: let me make a followup as a proposal and we can assess together - sadly I think there is quite a few

PR #3461, opened 2025-01-27
chore!: Rename `PositionComponent` to `PositionedComponent`

alpozcan:
# Description

Position**ed** is the correct usage of the word. Rename `PositionComponent` to `PositionedComponent`.
Given the nature of this change, I'm not _really_ expecting this to be merged, but in case it can be considered for v2 perhaps, here it is.

## Checklist

## Breaking Change?

### Migration instructions

Rename all occurrences of  `PositionComponent` to `PositionedComponent`.

## Related Issues

### DIFF
--- a/doc/bridge_packages/flame_bloc/bloc.md
+++ b/doc/bridge_packages/flame_bloc/bloc.md
@@ -69,7 +69,7 @@ Listening to states changes at the component level can be done with two approach
 By using `FlameBlocListener` component:
 
 ```dart
-class Player extends PositionComponent {
+class Player extends PositionedComponent {
   @override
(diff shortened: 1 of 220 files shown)


alpozcan:
nevermind, tests are failing and I don't want to sink more time into this.

pcornelissen:
You could add a transitional PositionComponent that inherits from PositionedComponent

spydon:
In v2 it will most likely be called `TransformComponent`. 

`PositionComponent` isn't wrong grammatically either though, it refers to that it has a position, not that it is positioned. But it has a lot more transform properties than position, which is why it'll most likely change name in v2.

PR #3471, opened 2025-02-03
feat: Add flip components around a given anchor

Treorai:
# Description

This methods flip a component around a Anchor passed to the method. When you have a component with a offset hitbox, changing the component's anchor and flipping it don't automatically flip the hitbox. With this function, both component and hitbox are flipped together.

## Checklist

## Breaking Change?

## Related Issues


packages/flame/lib/src/components/position_component.dart
```diff
@@ -416,6 +416,26 @@ class PositionComponent extends Component
     transform.flipVertically();
   }
 
+  /// Flip the component horizontally around a given anchor point.
+  void flipHorizontallyAroundAnchor(Anchor anchorPoint) {
+    if (anchor.x != 0.5) {
```
spydon, line 421: Shouldn't this just check if `anchor` and `anchorPoint` have the same X values?

packages/flame/lib/src/components/position_component.dart
```diff
@@ -416,6 +416,26 @@ class PositionComponent extends Component
...
+      final delta = (1 - 2 * anchorPoint.x) * width * transform.scale.x;
+      transform.x += delta * math.cos(transform.angle);
+      transform.y += delta * math.sin(transform.angle);
+    }
+    transform.flipHorizontally();
+  }
+
+  /// Flip the component vertically around a given anchor point.
+  void flipVerticallyAroundAnchor(Anchor anchorPoint) {
+    if (anchor.y != 0.5) {
```
spydon, line 431: Same here

spydon (commented):
Looks like a good addition, it needs some tests though.

spydon:
> When you have a component with a offset hitbox, changing the component's anchor and flipping it don't automatically flip the hitbox. 

The flip should affect all children too, including hitboxes, it should be visible with `debugMode = true`. Possibly the hitbox calculations don't take the scaling into account properly though.

packages/flame/lib/src/components/position_component.dart
Treorai: Not at all. We expect the user to call the methods I propose IF the wants to use a anchor point different that it's propriety. But I did do a mistake here: it should be checking for anchorPoint.x (and anchorPoint.y below).
The check skips the function because an anchor 0.5 is just a simple flip which don't need to do the calculations.

spydon: > The check skips the function because an anchor 0.5 is just a simple flip which don't need to do the calculations.

That only applies if the `anchorPoint` is the same as `anchor`, otherwise it will flip around the component's `anchor` and not the `anchorPoint`.

Treorai:
> > When you have a component with a offset hitbox, changing the component's anchor and flipping it don't automatically flip the hitbox.
> 
> The flip should affect all children too, including hitboxes, it should be visible with `debugMode = true`. Possibly the hitbox calculations don't take the scaling into account properly though.

I was expecting that too. (and keep in mind I just started learning flame, so I might be doing something wrong myself) but in my project although in the debugMode the hitbox seem to be flipping, the collision start happening at weird positions, outside the hitbox, but inside the sprite size. So I figured the hitbox was not flipping correctly with the sprite. My conclusion might be mistaken, but I made the methods locally to use in my project and the collisions worked great, didn't had to do any kind of adjustments.

packages/flame/lib/src/components/position_component.dart
Treorai: Oh I understand now what you mean. Leave this on hold, I'll do more tests myself and will get back to you.

spydon:
> I have a video demonstrating the usage of these features [here](https://youtu.be/fGIbyqLbJs8).

After watching your video I understand why you need this, but it is a very rare use-case, and I think there is a better solution.
The usual set-up is to just have the anchor of the component aligned with the center of the feet* of the sprite, and then all flipping would work as intended with just `flipHorizontally`.

* The feet so that it is easier to calculate the position of the component in relation to the ground, but it could just be horizontally lined up with the visual part of the sprite too.

Treorai:
Oh, I see. I thought this would be a recurrent situation. This usual set-up is indeed interesting, although I would hate to have any anchor point different than 0,0 but I guess this is a me thing LOL.
Anyway, thank you for your time, I learned a lot this week. Have a great one!

spydon:
> Oh, I see. I thought this would be a recurrent situation. This usual set-up is indeed interesting, although I would hate to have any anchor point different than 0,0 but I guess this is a me thing LOL.
> Anyway, thank you for your time, I learned a lot this week. Have a great one!

Most of the time the Anchor point would be 0 at X, but your component isn't equally wide on both sides around the center, maybe it would be  if the size it would be tighter around the visual representation?

Treorai:
In my case the sprite is not equally wide around the center because of the attacking sprite that is the centralized one. The weapon bend forward and then the sprite is finally centered (although the char's hitbox is still offset).
I don't know if this is the common setup, but I'm not the artist that drew the art, so I assume this is the way it usually is done?

spydon:
@Treorai usually the hitbox would be static in relation to where the anchor is and then you can just flip the component normally, but maybe that is not the case for you? If you set `debugMode = true` you should be able to see the hotbox visualized.

Treorai:
When I flip the component the flipped hitbox does not match the flipped sprite. Probably because I like to keep the anchor at 0,0.

spydon:
> When I flip the component the flipped hitbox does not match the flipped sprite. Probably because I like to keep the anchor at 0,0.

Can you send an image with debug mode activated to show how it looks before and after a flip?

PR #3472, opened 2025-02-05
perf: Switch from forEach to regular for-loops for about 30% improvement in raw update performance

filiph:
# Description

Replaces uses of `.forEach()` with regular `for` loops. This has significant impact on performance in hot paths such as `Component.updateTree()` and `Component.renderTree()`.

Updating Components

In the graph above, you see 50 runs of `benchmark/update_components_benchmark.dart`. The forEach results are blue, the for-loop results are green. I could see this effect after just replacing the `forEach` calls in `component.dart`. Data [here](https://docs.google.com/spreadsheets/d/e/2PACX-1vRk_yGmLN6o0oqSUWDBh7ODx7B8EIToeahZcZBS3VKHX8AbEnmrgmEqDt98cZLoBjIKQX3MlOc0XwsP/pubhtml).

> Aside for posterity: `for i in {1..50}; do flutter test benchmark/main.dart --no-pub -r silent 2>/dev/null >> benchmarks_for_loop.txt; done`, then get the data from the text file. 

I went ahead and replaced additional `forEach` calls elsewhere in the engine codebase, but there was no additional effect on the benchmark. Still, I kept those changes in. I only replaced `forEach` in places that seemed relatively hot (e.g. `ComponentSet.reorder()`). There are more `forEach` calls in the codebase but those seem fine to me as they aren't likely to be called too often.

It should be noted that I needed to update the benchmark to add children to the components. Every `_BenchmarkComponent` now has 10 children. This feels a bit more realistic use of the framework than having a flat array of components with no children. By changing the benchmark code in this way, I made it a bit slower, so I'm not sure if the effect will be seen in the CI/CD.

I also tried whether the change will have effect on my game's benchmark (which is a lot more involved and uses `flutter driver` to test the whole game in AOT mode). For the game, the effect is negligible but that was kind of expected since my game spends a significant amount of its CPU time on AI, raycasting, smoke simulation and drawVertices, none of which really depend on the speed of the engine `update()` mechanism.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
+++ b/packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
@@ -89,10 +89,11 @@ mixin ShapeHitbox on ShapeComponent implements Hitbox<ShapeHitbox> {
       _validAabb = false;
       onAabbChanged?.call();
     };
-    ancestors(includeSelf: true).whereType<PositionComponent>().forEach((c) {
-      _transformAncestors.add(c.transform);
-      c.transform.addListener(_transformListener);
-    });
+    for (final ancestor
+        in ancestors(includeSelf: true).whereType<PositionComponent>()) {
+      _transformAncestors.add(ancestor.transform);
+      ancestor.transform.addListener(_transformListener);
+    }
 
     if (shouldFillParent) {
       _parentSizeListener = () {
--- a/packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
+++ b/packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
@@ -89,7 +89,8 @@ mixin ShapeHitbox on ShapeComponent implements Hitbox<ShapeHitbox> {
       _validAabb = false;
       onAabbChanged?.call();
     };
-    fina positionComponents = ancestors(includeSelf: true).whereType<PositionComponent>();
+    final positionComponents =
+        ancestors(includeSelf: true).whereType<PositionComponent>();
     for (final ancestor in positionComponents) {
       _transformAncestors.add(ancestor.transform);
       ancestor.transform.addListener(_transformListener);
--- a/packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
+++ b/packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
@@ -89,8 +89,8 @@ mixin ShapeHitbox on ShapeComponent implements Hitbox<ShapeHitbox> {
       _validAabb = false;
       onAabbChanged?.call();
     };
-    for (final ancestor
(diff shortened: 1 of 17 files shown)


spydon:
> Ok, it looks like the benchmark GitHub action failed with `Error: + Failed to create or update comment on GitHub.: Resource not accessible by integration - https://docs.github.com/rest/issues/comments#create-an-issue-comment`.

It can only run for members of the organization unfortunately.

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
```diff
@@ -89,10 +89,11 @@ mixin ShapeHitbox on ShapeComponent implements Hitbox<ShapeHitbox> {
       _validAabb = false;
       onAabbChanged?.call();
     };
-    ancestors(includeSelf: true).whereType<PositionComponent>().forEach((c) {
-      _transformAncestors.add(c.transform);
-      c.transform.addListener(_transformListener);
-    });
+    for (final ancestor
+        in ancestors(includeSelf: true).whereType<PositionComponent>()) {
```
erickzanardo, line 93: Nit
```suggestion
    fina positionComponents = ancestors(includeSelf: true).whereType<PositionComponent>();
    for (final ancestor in positionComponents) {
```

erickzanardo (approved):
One small nit, but LGTM already, Thanks for this!

spydon (approved):
Lgtm! I thought we had already done this... I think there is a DCM rule that we can enable that checks for these.

I hope one day the compiler will be able to make `forEach` et. al as efficient as normal loops.

erickzanardo:
> Lgtm! I thought we had already done this... I think there is a DCM rule that we can enable that checks for these.
> 
> I hope one day the compiler will be able to make `forEach` et. al as efficient as normal loops.

I don't think we need DCM for this, I am pretty sure I saw a lint rule for that already somewhere...I will take a look for it

packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
spydon: Hehe, you missed an `l` in final :smile:

erickzanardo: Tried my best without a compiler analyzer at my disposal :P

filiph:
Thanks for the quick turnaround!

FWIW, I don't think enabling a lint and then converting each and every `forEach` call into a for loop is something you'd want. The effect is only felt in very hot paths. In this PR, the only real performance upside came from the two lines in `Component`. Elsewhere, the effect is negligible. (Unless you're doing something like resizing your game every frame.) And it's kind of nice to be able to express things in a one-liner.

PR #3481, opened 2025-02-09
fix: Remove forced location of the atlas file.

exoad:
# Description

This PR removes the a line where `texture_packer` forces devs to put their atlases under a folder named `/images/`. In Flame, devs can change prefixes of both `AssetCache` and `Images`, but `Future<_TextureAtlasData> _parse()` does not respect this change. This change would make it respect that change.

## Checklist

## Breaking Change?

Simply remove the "/images/" line from any atlas Strings.

## Related Issues

### DIFF
--- a/packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
+++ b/packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
@@ -26,9 +26,8 @@ class TexturePackerAtlas {
   /// instead of the packed size. For animation sprites, load with origin size
   /// is recommended for smooth result.
   ///
-  /// If [prefix] is not specified, the atlas file will be expected to under
-  /// a folder named "images" under [Flame.assets]. This is only applicable when
-  /// reading from the Flutter assets folder.
+  /// If [prefix] is not specified, the atlas file [path] will be resolved
+  /// relative to the `assets/images/` directory.
   static Future<TexturePackerAtlas> load(
     String path, {
     bool fromStorage = false,
--- a/packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
+++ b/packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
@@ -25,18 +25,22 @@ class TexturePackerAtlas {
   /// If [useOriginalSize] is true, the sprites loaded will be use original size
   /// instead of the packed size. For animation sprites, load with origin size
   /// is recommended for smooth result.
+  ///
+  /// If [prefix] is not specified, the atlas file will be expected to under
+  /// a folder named "images" under [Flame.assets]
   static Future<TexturePackerAtlas> load(
     String path, {
     bool fromStorage = false,
     Images? images,
     bool useOriginalSize = true,
+    String? prefix,
   }) async {
     final _TextureAtlasData atlasData;
 
     if (fromStorage) {
-      atlasData = await _fromStorage(path, images: images);
+      atlasData = await _fromStorage(path, images: images, prefix: prefix);
     } else {
-      atlasData = await _fromAssets(path, images: images);
+      atlasData = await _fromAssets(path, images: images, prefix: prefix);
     }
 
     return TexturePackerAtlas(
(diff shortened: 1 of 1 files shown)


packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
```diff
@@ -114,7 +114,7 @@ Future<_TextureAtlasData> _parse(
   if (fromStorage) {
     fileAsString = await File(path).readAsString();
   } else {
-    fileAsString = await Flame.assets.readFile('images/$path');
+    fileAsString = await Flame.assets.readFile(path);
```
spydon, line 117: The change makes sense, but we need to avoid making it a breaking change. Maybe take the prefix from the `Images` object that is passed in and if that is null use the `Flame.images` prefix. 
Or another possibility would be to add an optional prefix argument that defaults to `images/`.

So that the dev at least gets some configuration possibilities, but without making it a breaking change.

spydon: Hmm, on the other hand the dev already has a configurable prefix on the assets cache like you say, so it's very weird to have it here too. So maybe we should treat this like a bug and make a breaking change.

exoad: It's used here because it's expected that the atlas is right next to the image under "assets/images". I am thinking that we can just have an optional parameter for this prefix, it could solve the breaking change.

spydon: Yeah, let's go with an optional parameter then.

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
```diff
@@ -25,18 +25,22 @@ class TexturePackerAtlas {
   /// If [useOriginalSize] is true, the sprites loaded will be use original size
   /// instead of the packed size. For animation sprites, load with origin size
   /// is recommended for smooth result.
+  ///
+  /// If [prefix] is not specified, the atlas file will be expected to under
+  /// a folder named "images" under [Flame.assets]
```
spydon, line 30:
```suggestion
  /// If [prefix] is not specified, the atlas file [path] will be resolved
  /// relative to the `assets/images/` directory.
```
Since `Flame.assets` is an object I think the updated version will read a bit clearer.

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
```diff
@@ -25,18 +25,22 @@ class TexturePackerAtlas {
...
   static Future<TexturePackerAtlas> load(
     String path, {
     bool fromStorage = false,
     Images? images,
     bool useOriginalSize = true,
+    String? prefix,
```
spydon, line 36:
```suggestion
    String prefix = 'images',
```

Set the default here (so that the user can see it in their IDE), and set the internal methods to have this argument as required and non-nullable.

spydon (commented):
Two nits, but otherwise it looks good

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
spydon: Maybe name it `assetsPrefix` instead, so that it isn't mixed up with the image cache prefix.

exoad: Ok got it, I have made the `_fromStorage` also not accept the parameter just for simplicity sakes. 

It should be good now

spydon: Don't worry about the commits, they are all squashed on merge :)

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
```diff
@@ -114,7 +141,9 @@ Future<_TextureAtlasData> _parse(
   if (fromStorage) {
     fileAsString = await File(path).readAsString();
   } else {
-    fileAsString = await Flame.assets.readFile('images/$path');
+    fileAsString = await Flame.assets.readFile(
+      '${assetsPrefix ?? 'images/'}$path',
+    );
```
spydon, line 146:
```suggestion
    assert(
      assetsPrefix != null,
      'When reading from storage, the assetsPrefix needs to be provided.',
    );
    fileAsString = await Flame.assets.readFile('$assetsPrefix!/$path');
```
To avoid having defaults on multiple places and also it's a faulty state to end up in this `else` case without an `assetsPrefix`.

spydon (approved):
Thanks for your contribution!

PR #3483, opened 2025-02-09
feat(audio): Set audio context AudioContextConfigFocus.mixWithOthers by default

Gustl22:
# Description

AudioPlayers used to mix their sounds with other audio on the device, which is not the proposed experience on Mobile devices.  So this behavior was changed in AudioPlayers, but it wasn't enforced due to a bug. Obviously this behavior is not wanted for Games as multiple sounds should play simultaneously, so this recovers the old functionality again.

## Checklist

Not sure, if I can test this properly, as it's on a system level. I could ensure the values are set though (?)

## Breaking Change?
I'm actually not sure, if this is breaking, I think it's not, as `audioplayers` used to work like `mixWithOthers` due to a bug, but is not anymore.

## Related Issues

No related issues found

### DIFF
--- a/packages/flame_audio/example/lib/main.dart
+++ b/packages/flame_audio/example/lib/main.dart
@@ -39,9 +39,9 @@ class AudioGame extends FlameGame with TapDetector {
 
   Rect get button => Rect.fromLTWH(20, size.y - 300, size.x - 40, 200);
 
-  void startBgmMusic() async {
-    FlameAudio.bgm.initialize();
-    FlameAudio.bgm.play('music/bg_music.ogg');
+  Future<void> startBgmMusic() async {
+    await FlameAudio.bgm.initialize();
+    await FlameAudio.bgm.play('music/bg_music.ogg');
   }
 
   void fireOne() {
--- a/packages/flame_audio/example/lib/main.dart
+++ b/packages/flame_audio/example/lib/main.dart
@@ -39,7 +39,7 @@ class AudioGame extends FlameGame with TapDetector {
 
   Rect get button => Rect.fromLTWH(20, size.y - 300, size.x - 40, 200);
 
-  void startBgmMusic() {
+  void startBgmMusic() async {
     FlameAudio.bgm.initialize();
     FlameAudio.bgm.play('music/bg_music.ogg');
   }
--- a/packages/flame_audio/lib/flame_audio.dart
+++ b/packages/flame_audio/lib/flame_audio.dart
@@ -1,5 +1,4 @@
 import 'package:audioplayers/audioplayers.dart';
-
 import 'package:flame_audio/bgm.dart';
 
 export 'package:audioplayers/audioplayers.dart';
@@ -49,9 +48,12 @@ class FlameAudio {
     String file,
     double volume,
     ReleaseMode releaseMode,
-    PlayerMode playerMode,
-  ) async {
+    PlayerMode playerMode, {
+    required AudioContext? audioContext,
+  }) async {
     final player = AudioPlayer()..audioCache = audioCache;
+    audioContext ??= _defaultAudioContext;
+    await player.setAudioContext(audioContext);
     await player.setReleaseMode(releaseMode);
     await player.play(
       AssetSource(file),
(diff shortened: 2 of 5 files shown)


Gustl22:
How do you handle TODO comments in the code. I saw you don't have any (?).

packages/flame_audio/example/lib/main.dart
```diff
@@ -39,9 +39,9 @@ class AudioGame extends FlameGame with TapDetector {
 
   Rect get button => Rect.fromLTWH(20, size.y - 300, size.x - 40, 200);
 
-  void startBgmMusic() {
-    FlameAudio.bgm.initialize();
-    FlameAudio.bgm.play('music/bg_music.ogg');
+  Future<void> startBgmMusic() async {
+    await FlameAudio.bgm.initialize();
+    await FlameAudio.bgm.play('music/bg_music.ogg');
```
Gustl22, line 44: It also works withoud a Future, but it's cleaner and more future proof to await the initialize method.

spydon:
> How do you handle TODO comments in the code. I saw you don't have any (?).

We have `TODO(username): comment` as the format of our todos, we do have them, but we try to fix them soon so that they are not forgotten.

Gustl22:
> We have TODO(username): comment as the format of our todos, we do have them, but we try to fix them soon so that they are not forgotten.

Thank you for the hint. So yeah so the pipeline fails, as the spell checker won't accept my username^^

spydon:
> > We have TODO(username): comment as the format of our todos, we do have them, but we try to fix them soon so that they are not forgotten.
> 
> Thank you for the hint. So yeah so the pipeline fails, as the spell checker won't accept my username^^

Add it to the list! It's in the .github directory :)

packages/flame_audio/lib/flame_audio.dart
```diff
@@ -49,9 +48,12 @@ class FlameAudio {
     String file,
     double volume,
     ReleaseMode releaseMode,
-    PlayerMode playerMode,
-  ) async {
+    PlayerMode playerMode, {
+    required AudioContext? audioContext,
```
luanpotter, line 52: should we make the new parameter not-required?

packages/flame_audio/lib/flame_audio.dart
```diff
@@ -112,12 +134,21 @@ class FlameAudio {
     String sound, {
     required int maxPlayers,
     int minPlayers = 1,
-  }) {
+    AudioContext? audioContext,
+  }) async {
+    audioContext ??= _defaultAudioContext;
+    // TODO(gustl22): Probably set context for each player individually,
+    //  as soon as functionality is supported.
```
luanpotter, line 141: IIRC that was how I wanted to impl but iOS didn't support it so I had to make it global, but that was half a decade ago so my memory fails me

luanpotter (approved):
LGTM with a comment about required AudioContext param

packages/flame_audio/lib/flame_audio.dart
Gustl22: It's only used internally, wanted to make sure every method calling it, actually hands it over, so the dev is able to alter it, if needed.

packages/flame_audio/lib/flame_audio.dart
Gustl22: In fact, for iOS it's still global. We changed that behavior to align it along with Android, so on iOS, if setting it for one player, it is actually applied to all players. So the dev experience is much worse, needing to differentiate the cases themselves, than actually experiencing no change at all for that given method, which we got a lot of tickets for. It's all documented though.

packages/flame_audio/lib/flame_audio.dart
luanpotter: ah yes agree makes sense if it is internal!

packages/flame_audio/lib/flame_audio.dart
luanpotter: yeah I recall there was some sort of nuisance between platforms, as often is the case...

spydon (approved):

PR #3513, opened 2025-02-27
feat: Layout shrinkwrap

### LINKED ISSUE #3509: LayoutComponent shrink wrapping
### Problem to solve

In situations where you need a linear list of elements (which you would use LayoutComponent for) but would prefer this list to expand or shrink as defined by its children, the current experimental LayoutComponent is insufficient.

For example, a long scrollable list (expanding) or a short list that is centered (shrinking)

The calculation of inherent size is not a problem: it can be determined by iterating over the children, and taking into consideration the direction of the layout, the various alignment settings, and the gaps. The tricky part is in deciding when to perform this calculation, and when not to.

To wit: setting size automatically triggers a layout. If the layout itself calculates the inherent size and resizes, it'll trigger layout again, and we end up with a stack overflow.

### Proposal

Add an encapsulated boolean property: shrinkWrap, defaulting to false.

Add a function `syncSizeListener`, which adds `layoutChildren` as a size listener, if shrinkWrap is false, and removes it otherwise.

In both the constructor body and the setter for shrinkWrap, call `syncSizeListener`.

### More information

Related to #1944 

### Other

livtanong:
# Description

Implements shrink wrapping of the LayoutComponents, such that instead of a top-down influence of size, the LayoutComponent's size is dictated by the sizes of the children.

## Checklist

## Breaking Change?

## Related Issues

Closes #3509

### DIFF
--- a/examples/lib/stories/experimental/layout_component_example.dart
+++ b/examples/lib/stories/experimental/layout_component_example.dart
@@ -1,12 +1,13 @@
 import 'dart:async';
 
 import 'package:flame/components.dart';
+import 'package:flame/events.dart';
 import 'package:flame/experimental.dart';
 import 'package:flame/game.dart';
 import 'package:flame/input.dart';
 import 'package:flutter/material.dart';
 
-class LayoutComponentExample extends FlameGame {
+class LayoutComponentExample extends FlameGame with DragCallbacks {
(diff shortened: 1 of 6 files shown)


spydon (approved):
Lgtm! Just one nit

PR #3519, opened 2025-03-12
feat: Support custom attributes syntax to allow for multiple styles in the text rendering pipeline

luanpotter:
# Description

This is a proposal to extend both our text rendering pipeline and our markdown-parsing capabilities to support more fine-grained styling by leveraging a subset of the extended markdown definition of "custom attributes" (for example, see [markdown-it-attrs](https://www.npmjs.com/package/markdown-it-attrs) as an exemplary implementation).

On the markdown side, the syntax will look like this:

```markdown
[This is a custom class]{.my-custom-class}
This word will be [red]{.red} and this one will be [blue]{.blue}.
```

Note that the current implementation only supports specifying the class name and will skip other possible permutations.

On the Flame side, we add a new `customStyles` map to the `DocumentStyle` class, allowing users to provide as many styles as desired.

Example:

image
image

Currently, users have to hijack the existing syntaxes for the desired styles; with the addition of code and strike-through recently I have doubled the possibilities for custom styles, but the solution still remained a hack. Also I now need more styles, and the only way forward is to support arbitrary styling.

While we could consider supporting other attributes, such as id, css styling, etc, within the attribute syntax, the class name is the most flexible and versatile option, and it already solves all use cases, as class names can be used as ids (if desired) and can be used to apply the styling indirectly. I don't see any immediate advantage to supporting more complex constructs within the attributes block (though of course we can iterate as needed).

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame_markdown/lib/custom_attribute_syntax.dart
+++ b/packages/flame_markdown/lib/custom_attribute_syntax.dart
@@ -1,0 +1,111 @@
+import 'package:markdown/markdown.dart';
+
+// cSpell:ignore charcode.dart (file name from another package)
+// NOTE: values obtained from file `charcode.dart` from the markdown package
+class _Chars {
+  /// Character `[`.
+  static const int leftBracket = 0x5B;
+
+  /// Character `{`.
+  static const int leftBrace = 0x7B;
+
+  /// Character `}`.
+  static const int rightBrace = 0x7D;
+}
+
+/// Allows for a toned-down version of custom attributes extension for markdown,
+/// inspired by the markdown-it-attrs package.
+///
+/// This allows users to specify a custom class name to a span of text:
+///
+/// ```markdown
+/// [This is a custom class]{.my-custom-class}
+/// This word will be [red]{.red} and this one will be [blue]{.blue}.
+/// ```
+///
+/// This is based on the standard Link markdown parser (which matches the
+/// `[text](url)` and `[text][ref]` syntaxes).
+class CustomAttributeSyntax extends LinkSyntax {
+  /// Creates a new custom attribute syntax.
+  CustomAttributeSyntax()
+      : super(
+          pattern: r'\[',
+          startCharacter: _Chars.leftBracket,
+        );
+
+  @override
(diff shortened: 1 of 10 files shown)


packages/flame_markdown/lib/custom_attribute_syntax.dart
```diff
@@ -0,0 +1,110 @@
+import 'package:markdown/markdown.dart';
+
+// NOTE: values obtained from file `charcode.dart` from the markdown package
```
luanpotter, line 3: sadly we cannot access the file as it is not exported. tbh I do prefer my naming schema though :P

packages/flame_markdown/lib/flame_markdown.dart
```diff
@@ -51,7 +51,20 @@ class FlameMarkdown {
         .map(_castCheck<InlineTextNode>)
         .toList();
     final child = _groupInlineChildren(children);
+
+    final customClassName = element.attributes['class'];
+    if (customClassName != null) {
+      if (element.tag != 'span') {
+        throw Exception(
```
luanpotter, line 58: we could consider support for other inline elements, for example `**foo**{.super-bold}`, which is part of markdown-it-attrs.
but I think it is unnecessary because you can just make your custom style bold or whatever you want.

packages/flame_markdown/lib/custom_attribute_syntax.dart
```diff
@@ -0,0 +1,111 @@
+import 'package:markdown/markdown.dart';
+
+// cSpell:ignore charcode.dart (file name from another package)
+// NOTE: values obtained from file `charcode.dart` from the markdown package
+class _Chars {
```
spydon, line 5: Does this even need a class?

spydon (approved):

packages/flame_markdown/lib/custom_attribute_syntax.dart
luanpotter: No but I thought it was more organized to group these together than let them laying around

Issue #3363, opened 2024-11-01
Flame 1.21.0 removes Components from World by itself


dvmatyun:
### What happened?

1. Updated from Flame 1.18.0 to 1.21.0 (so issue was not present on 1.18.0)
2. During game session, flame randomly can remove my components (I have no clue why that happens on 1.21 and not on 1.18)

### What do you expect?

I expect to know the reason why the component is removed. Both `removeFromParent` or `onRemove` are NOT called on removed components. So they just dissapear from World `children ` set. Maybe Flame introduced some sort of garbage collector?

### How can we reproduce this?

To be honest, I have no idea, maybe you can give me a clue why that can be happening.

### What steps should take to fix this?

Make `World`'s children set immutable and all add or remove operations can be logged if `debug` is true?

### Do have an example of where the bug occurs?

no

### Execute in a terminal and put output into the code block below

fvm flutter doctor -v
[√] Flutter (Channel stable, 3.24.3, on Microsoft Windows [Version 10.0.22631.4317], locale ru-RU)
    • Flutter version 3.24.3 on channel stable at C:\Users\dima\fvm\versions\3.24.3
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 2663184aa7 (7 weeks ago), 2024-09-11 16:27:48 -0500
    • Engine revision 36335019a8
    • Dart version 3.5.3
    • DevTools version 2.37.3

[√] Windows Version (Installed version of Windows is version 10 or higher)

[√] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at C:\sdk\Android\android-sdk
    • Platform android-34, build-tools 34.0.0
    • ANDROID_HOME = C:\sdk\Android\android-sdk
    • Java binary at: C:\sdk\AndroidStudio\jbr\bin\java
    • Java version OpenJDK Runtime Environment (build 17.0.7+0-b2043.56-10550314)
    • All Android licenses accepted.

[√] Chrome - develop for the web
    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[√] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.7.4)
    • Visual Studio at C:\Program Files\Microsoft Visual Studio\2022\Community
    • Visual Studio Community 2022 version 17.7.34031.279
    • Windows 10 SDK version 10.0.19041.0

[!] Android Studio (version 2022.3)
    • Android Studio at D:\Programs\Android
[...]

PR #3530, opened 2025-03-18
fix: Fix priority rebalancing causing concurrent mutation of component ordered_set

luanpotter:
# Description

Fix priority rebalancing causing concurrent mutation of component ordered_set.

This fixes https://github.com/flame-engine/flame/issues/3363

Basically priority changes and rebalancing was not using the component lifecycle event queue, which caused concurrent mutations and iterations of the component set, and due to race conditions, loss of components as they are removed and re-added for rebalancing.

The issue is not with ordered_set itself, but rather how it is being used. It was never designed to support concurrent mutations.

A few notes for the future:

* I think we should consider changing the whole event (input) system so that every event is queued and only processed predictably within the game loop, avoiding a whole class of concurrent modification issues
* we should consider more efficient ways of rebalancing than literally removing every single component + re-adding (which is also error prone if there is concurrent access).

## Checklist

## Breaking Change?


spydon (approved):
Super, thanks for the quick fix! ❤️

packages/flame/test/components/component_test.dart
```diff
@@ -1,3 +1,6 @@
+import 'dart:async';
```
spydon, line 1:
```suggestion
```

luanpotter: how come analyzer is not catching that???

packages/flame/test/components/priority_test.dart
```diff
@@ -163,10 +164,13 @@ void main() {
         c1.priority = 10;
         a2.priority = 0;
         game.update(0);
+        await game.ready();
 
         a.assertCalled(1);
         b.assertCalled(0);
-        c.assertCalled(1);
+        // TODO(luan): consider optimizing so that multiple priority changes
+        // on the same parent on the same tick only trigger one reorder call
```
spydon, line 172: Didn't we use to have this? 🤔

luanpotter: yes the previous implementation did this naturally due to how it was setup

spydon: It will be quite expensive to not have that if there are a lot of priority changes, any ideas of how this could be implemented with the new way?

luanpotter: that is only if you have multiple priority changes on the same parent on the same tick, so I don't think it is super relevant. we can eventually revamp the whole lifecycle queue to re-group and optimize events in general between the 4 types

spydon: I remember that there were complaints about it before we implemented it. Since we don't have any other way to do z-index, priority is heavily used like that in for example isometric games, it would go from O(n) to O(n²).

packages/flame/lib/src/components/core/component.dart
```diff
@@ -881,11 +881,10 @@ class Component {
 
   @internal
   LifecycleEventStatus handleLifecycleEventRebalance(int newPriority) {
-    if (_priority == newPriority) {
-      return LifecycleEventStatus.skip;
```
luanpotter: skip isn't really supported upstream

packages/flame/test/components/priority_test.dart
luanpotter: That would be `O(n^2)` only if you are changing the order of every single component within a parent on every single tick, which I think is unrealistic (and already would not perform well regardless). But I changed the implementation to batch reorders, to the cost of some minor overhead for the common case.

spydon: When you measure big O you do it for the asymptotic upper bound, hence `O(n²)`.
The point is, just reordering many (N) per tick would be N times as expensive, which was causing the problem I mentioned previously.
The updated solution looks good!

spydon (approved):
Lgtm!

PR #3552, opened 2025-04-08
feat: Adding tickCount to TimerComponent

erickzanardo:
# Description

Adds a `tickCount` to `TimerComponent`.

## Checklist

## Breaking Change?

## Related Issues


ufrshubham (commented):
@erickzanardo it would be good if we can have `tickCount` in `Timer` as well. What do you think?

packages/flame/lib/src/components/timer_component.dart
```diff
@@ -21,15 +23,23 @@ class TimerComponent extends Component {
...
+  /// This is is only used when [repeat] is true. If null,
+  /// the timer will run indefinitely.
   TimerComponent({
     required double period,
     bool repeat = false,
     bool autoStart = true,
     this.removeOnFinish = false,
     VoidCallback? onTick,
     this.tickWhenLoaded = false,
+    this.tickCount,
```
spydon, line 36: I would probably name it `tickLimit` or something like that which indicates that it is controlling when it should stop.

spydon (commented):
I like @ufrshubham's idea of moving it one step down into the timer

doc/flame/other/util.md
```diff
@@ -153,6 +153,12 @@ class MyFlameGame extends FlameGame {
 }
 ```
 
+```{note}
+A `Timer` or `TimerComponent` can repeat indefinitely by providing a
```
spydon, line 157:
```suggestion
A `Timer` or `TimerComponent` can repeat indefinitely by providing the
```

doc/flame/other/util.md
```diff
@@ -153,6 +153,12 @@ class MyFlameGame extends FlameGame {
...
+`repeat: true` property to it or it can repeat a certain number of
```
spydon, line 158:
```suggestion
`repeat: true` argument or it can be repeated a certain number of
```

doc/flame/other/util.md
```diff
@@ -153,6 +153,12 @@ class MyFlameGame extends FlameGame {
...
+times by using the `tickCount` property alongside the `repeat`.
```
spydon, line 159:
```suggestion
times by using the `tickCount` argument together with `repeat: true`.
```

packages/flame/lib/src/timer.dart
```diff
@@ -41,20 +46,29 @@ class Timer {
...
           _current -= limit;
-          onTick?.call();
+          _callTicker();
         }
       }
     }
   }
 
+  void _callTicker() {
+    if (tickCount != null && _currentTick >= tickCount!) {
```
spydon, line 64: Shouldn't we stop directly after the last tick has been made, instead of in the beginning of the next tick?

erickzanardo: Good point, updated!

spydon: If tick count is 0 for whatever reason we'd still tick now :)

erickzanardo: hum, yeah, but how should we handle this cause that really doesn't makes much sense right?

In the `TimerComponent` I  added the following assertion

```dart
assert(
  tickCount == null || tickCount > 0,
  'tickCount must be null or bigger than 0',
);
```

But I forgot to move it here when I moved the logic to the `Timer` class instead. Do you think moving the assertion here would be enough?

spydon: That that sounds good!

erickzanardo: Moved the assertion 🥳

spydon (approved):
Lgtm!

packages/flame/lib/src/timer.dart
```diff
@@ -15,19 +15,28 @@ class Timer {
...
     this.onTick,
     this.repeat = false,
     bool autoStart = true,
-  }) : _running = autoStart;
+    this.tickCount,
+  })  : assert(
+          tickCount == null || tickCount > 0,
+          'tickCount must be null or bigger than 0',
+        ),
+        _running = autoStart;
```
luanpotter, line 31: should we throw if tickCount!=null and repeat==false?

luanpotter (approved):
LGTM

gnarhard:
Thank you guys!

PR #3564, opened 2025-04-18
fix: Priority change should be reflected directly

spydon:
# Description

After #3554 we can reflect the priority change directly on the component.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/components/core/component.dart
+++ b/packages/flame/lib/src/components/core/component.dart
@@ -845,8 +845,8 @@ class Component {
   /// The smaller the priority, the sooner your component will be
   /// updated/rendered.
   /// It can be any integer (negative, zero, or positive).
-  /// If two components share the same priority, they will probably be drawn in
-  /// the order they were added.
+  /// If two components share the same priority, they will be updated and
+  /// rendered in the order they were added.
   ///
   /// Note that setting the priority is relatively expensive if the component is
   /// already added to a component tree since all siblings have to be re-added
@@ -855,12 +855,11 @@ class Component {
   int _priority;
   set priority(int newPriority) {
     if (_priority != newPriority) {
+      _priority = newPriority;
       final parent = _parent;
       final game = findGame();
       if (game != null && parent != null) {
         game.enqueuePriorityChange(parent, this, newPriority);
-      } else {
-        _priority = newPriority;
       }
     }
   }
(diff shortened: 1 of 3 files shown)


packages/flame/test/components/component_test.dart
```diff
@@ -1022,16 +1023,26 @@ void main() {
...
+          expect(c.priority, 10);
+          expect(
+            game.queue.any(
+              (e) =>
+                  e.child == c &&
+                  e.parent == game.world &&
+                  e.kind == LifecycleEventKind.rebalance,
+            ),
+            isTrue,
+          );
```
luanpotter, line 1041: can we also check the order on the set is untouched? (can be followup!)

luanpotter: actually I'd rather only check that than to make everything public

spydon: Don't we already do that in other tests? :thinking:

packages/flame/lib/src/components/core/component.dart
```diff
@@ -855,12 +855,11 @@ class Component {
   int _priority;
   set priority(int newPriority) {
     if (_priority != newPriority) {
+      _priority = newPriority;
       final parent = _parent;
       final game = findGame();
       if (game != null && parent != null) {
         game.enqueuePriorityChange(parent, this, newPriority);
```
luanpotter, line 862: we don't need to pipe in the newPriority anymore right?

spydon: What do you mean?

packages/flame/test/components/component_test.dart
spydon: Maybe we don't, I'll create a test for that!

packages/flame/lib/src/components/core/component.dart
spydon: Aaah now I get what you mean, it showed the comment on the wrong line on the phone. 😅

spydon: Fixed in [9233b40](https://github.com/flame-engine/flame/pull/3564/commits/9233b408306c5b447311f2fb3248140f000e51f6)

luanpotter (approved):
LGTM

PR #3579, opened 2025-04-22
fix: Export necessary classes to build custom commands, update docs [flame_console]

luanpotter:
# Description

Export necessary classes to build custom commands; currently you don't have access to the `ArgResults` class so you cannot override it.
This also fixes the docs which had an unsupported override. Also tidy up some stuff on the docs (unnecessary constructor, etc).

## Checklist

## Breaking Change?


packages/flame_console/example/.metadata
```diff
@@ -4,7 +4,7 @@
 # This file should be version controlled and should not be manually edited.
```
luanpotter, line 4: I honestly don't know why we must commit these generated files 🤷

spydon (approved):

PR #3600, opened 2025-05-08
feat(Jenny): Add onCommandFinish lifecycle method to DialogueView

### LINKED ISSUE #3596: Lifecycle for `onCommandExecuted` in Jenny
### Problem to solve

Sometimes a user-defined command sets a globally-scoped variable. Currently, there's no way to access the updated value of a variable just after it has been set by the user-defined command. `onCommand` is the only lifecycle method that is tied to commands, and it is triggered before the user-defined command finishes executing.

### Proposal

- Add a new method for DialogueView: `onCommandExecuted` (or a more appropriate name)
- (less sure about this) in `DialogueRunner.deliverCommand`, modify the function body such that `onCommandExecuted` is called after the future returned by `command.execute` resolves.

### More information

_No response_

### Other

livtanong:
# Description

Add onCommandFinish lifecycle method to DialogueView

## Checklist

## Breaking Change?

## Related Issues

Closes #3596

### DIFF
--- a/examples/lib/stories/bridge_libraries/flame_jenny/jenny_command_lifecycle_example.dart
+++ b/examples/lib/stories/bridge_libraries/flame_jenny/jenny_command_lifecycle_example.dart
@@ -8,7 +8,7 @@ import 'package:jenny/jenny.dart';
 
 class JennyCommandLifecycleExample extends FlameGame {
   static const String description = '''
-    This is an example of how the lifecycles relating to user-defined
+    This is an example of how the lifecycle methods relating to user-defined
     commands work.
   ''';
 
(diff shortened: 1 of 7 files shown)


examples/lib/stories/bridge_libraries/flame_jenny/jenny_command_lifecycle_example.dart
```diff
@@ -0,0 +1,71 @@
...
+import 'package:flame/components.dart';
+import 'package:flame/experimental.dart';
+import 'package:flame/game.dart';
+import 'package:flutter/services.dart';
+import 'package:jenny/jenny.dart';
+
+class JennyCommandLifecycleExample extends FlameGame {
+  static const String description = '''
+    This is an example of how the lifecycle methods relating to user-defined
+    commands work.
```
spydon, line 12:
```suggestion
This is an example of how the lifecycles relating to user-defined
commands work.
```
I think we have this in a lot of examples doing this, so not surprising you also added the spacing, but with the multiline strings like this the spacing is included in the string.

spydon (commented):
Maybe `onCommandFinished` would be a better name? Because possibly "executed" could be thought of as that the command has just been "executed", but the result not awaited for.

Otherwise it looks good!

spydon (approved):
Super, thanks!

PR #3606, opened 2025-05-21
fix: Only expose `ReadOnlyOrderedset` from `component.children`

spydon:
# Description

This PR updates to ordered_set 8.0.0 and makes sure that `children` are of the `ReadOnlyOrderedSet` type, so that `children.add/remove` isn't directly exposed to the user, since they skip the component lifecycle.

Since we previously had the `ComponentSet` to hide `add/remove` we now had to introduce a new internal getter in the component to make sure that the set only is initialized when it is needed. This was a previous optimization to keep components without children more lightweight.

It also introduces the static `Component.strictQueryMode` variable so that the user can set the `ComponentSetFactory` to only create components sets with strict mode turned on (this functionality also previously lived in the `ComponentSet` class).

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/components/core/component.dart
+++ b/packages/flame/lib/src/components/core/component.dart
@@ -287,7 +287,15 @@ class Component {
     }
   }
 
+  /// This field should be used internally for functionality when you don't need
+  /// to create a component set for the children if one doesn't already exist.
+  ///
+  /// This makes it possible to have lighter components that don't have any
+  /// children.
   OrderedSet<Component>? _children;
+
+  /// This field should be used internally for functionality when you need to
+  /// make sure that the component set is created if it doesn't already exist.
   OrderedSet<Component> get _internalChildren =>
       _children ??= createComponentSet();
 
--- a/packages/flame/lib/src/components/core/component.dart
+++ b/packages/flame/lib/src/components/core/component.dart
@@ -12,7 +12,7 @@ import 'package:flame/src/effects/provider_interfaces.dart';
 import 'package:flutter/painting.dart';
 import 'package:meta/meta.dart';
 import 'package:ordered_set/ordered_set.dart';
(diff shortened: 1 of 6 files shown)


packages/flame/lib/src/components/core/component.dart
```diff
@@ -287,14 +287,30 @@ class Component {
...
+  /// This field should be used internally for functionality when you don't need
+  /// to create a component set for the children if one doesn't already exist.
+  ///
+  /// This makes it possible to have lighter components that don't have any
+  /// children.
+  OrderedSet<Component>? _children;
+
+  /// This field should be used internally for functionality when you need to
+  /// make sure that the component set is created if it doesn't already exist.
+  OrderedSet<Component> get _internalChildren =>
```
luanpotter, line 299: Another option could be mutableChildren

packages/flame/test/components/component_test.dart
```diff
@@ -1474,18 +1473,19 @@ void main() {
         final component0 = Component();
         expect(component0.children.strictMode, false);
 
-        Component.childrenFactory = () => QueryableOrderedSet(
-              OrderedSet.comparing(Comparing.on((e) => e.priority)),
-              strictMode: false,
+        Component.childrenFactory = () => OrderedSet.mapping<num, Component>(
+              (e) => e.priority,
+              // ignore: avoid_redundant_argument_values
+              strictMode: true,
```
luanpotter, line 1479: Can we just remove?

luanpotter (approved):
LGTM

packages/flame/test/components/component_test.dart
spydon: Then it's not as easy to read and see what is happening in the test, so I choose to have it there explicitly.

PR #3616, opened 2025-06-03
fix: Calculate zoom ratio before rendering svg

### LINKED ISSUE #3615: SvgComponent blurred when camera.viewFinder.zoom is larger than 1
### What happened?

When camera.viewFinder.zoom is larger than 1, the svg appears to be rendered in the unzoomed resolution and then scaled up:

Image

### What do you expect?

The svg should be rendered with native resolution regardless of zoom. Here is a quick fix I applied to the flame_svg package: 0be4940f0c748a7fcb2d2f722d67a2915623b469. I'm however unsure if it would play well with the `pixelRatio` parameter. I'd be glad to open a pr if this is considered the "proper" way.

Image

Set `game.camera.viewFinder.zoom` to anything higher than 1, and add an `SvgComponent` to the world.

The example is setup to have scroll zooming. Scroll up to zoom in, the heart should then appear pixellated without the patch.

zapp seem to have outdated dart version and can't resolve the flame_svg package. Below is the singular `main.dart` that can be used to reproduce this behavior. Only the `flame` and `flame_svg` package is needed.

```
import 'dart:math' as math;

import 'package:flame/events.dart';
import 'package:flame_svg/flame_svg.dart';

void main() {
  runApp(const MainApp());
}

final svgStr = '''
<svg xmlns="http://www.w3.org/2000/svg" width="800" height="800" viewBox="0 0 32 32"><path d="M26.996 12.898c-.064-2.207-1.084-4.021-2.527-5.13-1.856-1.428-4.415-1.69-6.542-.132-.702.516-1.359 1.23-1.927 2.168-.568-.938-1.224-1.652-1.927-2.167-2.127-1.559-4.685-1.297-6.542.132-1.444 1.109-2.463 2.923-2.527 5.13-.035 1.172.145 2.48.788 3.803 1.01 2.077 5.755 6.695 10.171 10.683l.035.038.002-.002.002.002.036-.038c4.415-3.987 9.159-8.605 10.17-10.683.644-1.323.822-2.632.788-3.804"/></svg>
''';

class SvgWorld extends World {
  final double svgScale;

  SvgWorld({this.svgScale = 1});

  @override
  Future<void> onLoad() async {
    super.onLoad();

    add(
      SvgComponent(
        anchor: Anchor.center,
        svg: await Svg.loadFromString(svgStr),
        position: Vector2(0, 0),
        size: Vector2(100 * svgScale, 100 * svgScale),
      ),
    );
  }

  @override
  void render(Canvas canvas) {
    canvas.drawRect(canvas.getLocalClipBounds(), Paint()..color = Colors.white);

    super.render(canvas);
  }
}

class MreGame extends FlameGame with ScrollDetector {
  MreGame({super.world});

  @override
  void onScroll(PointerScrollInfo info) {
    super.onScroll(info);
    camera.viewfinder.zoom *= math.max(
      1 - (info.scrollDelta.global.y.sign * 0.1),
      0.1,
    );
  }
}

class MainApp extends StatelessWidget {
  const MainApp({super.key});

  @override
    return GameWidget(game: MreGame(world: SvgWorld()));
  }
}
```

### Relevant log output

```shell

```

[✓] Flutter (Channel stable, 3.29.3, on Arch Linux 6.14.7-zen2-1-zen, locale en_US.UTF-8) [231ms]
    • Flutter version 3.29.3 on channel stable at /home/vun/softwares/flutter/flutter
    • Framework revision ea121f8859 (7 weeks ago), 2025-04-11 19:10:07 +0000
    • Engine revision cf56914b32
    • Dart version 3.7.2
    • DevTools version 2.42.3

[!] Android toolchain - develop for Android devices (Android SDK version 35.0.1) [340ms]
    • Android SDK at /home/vun/Android/Sdk
      See https://flutter.dev/to/linux-android-setup for more details.

[✓] Chrome - develop for the web [183ms]
    • CHROME_EXECUTABLE = /usr/bin/chromium

[✓] Linux toolchain - develop for Linux desktop [494ms]
    • clang version 19.1.7
    • cmake version 4.0.2-dirty
    • ninja version 1.12.1
    • pkg-config version 2.4.3

[✓] Android Studio (version 2024.3) [180ms]
    • Android Studio at /home/vun/softwares/android-studio
    • Flutter plugin version 85.2.3
    • Java version OpenJDK Runtime Environment (build 21.0.6+-13355223-b631.42)

[✓] IntelliJ IDEA Community Edition (version 2025.1) [17ms]
    • IntelliJ at /usr/share/idea

[✓] Connected device (2 available) [292ms]
    • Linux (desktop) • linux  • linux-x64      • Arch Linux 6.14.7-zen2-1-zen
    • Chrome (web)    • chrome • web-javascript • Chromium 137.0.7151.40 Arch Linux

[✓] Network resources [852ms]

! Doctor found issues in 1 category

### Affected platforms

All

### Other information

_No response_

- [x] I want to work on this

vunyunt:
# Description

SvgComponent currently renders in lower-than-native resolution when `game.camera.viewfinder.zoom` is larger than 1. This is addressed by taking a ratio of `canvas.getDestinationClipBounds().size` and `canvas.getLocalClipBounds().size`, and applying it to the existing canvas scaling.

I'm still not quite sure how well this would work with `pixelRatio`, but I've verified that when running the "render sharply" test which has the device pixel ratio set to 3, the canvas size ratio mentioned above is 1, and not rendering in excessive resolution.

The test uses a `Transform.scale` to simulate viewfinder zoom. I've also verified that when running without the patch, this test generates a blurry image:
|before|after|
|-|-|
|Screenshot_20250604_073801|image|

## Checklist

## Breaking Change?

## Related Issues

Fixes #3615 


packages/flame_svg/test/svg_test.dart
```diff
@@ -93,5 +93,39 @@ void main() {
...
+              ),
+            ),
+          ),
+        );
+
+        await tester.pumpAndSettle();
+        await expectLater(
+          find.byType(MaterialApp),
+          matchesGoldenFile(
+              './_goldens/render_sharply_with_viewfinder_zoom.png'),
```
spydon:
```suggestion
              './_goldens/render_sharply_with_viewfinder_zoom.png'),
```
Add a trailing comma here to make the analyzer happy.

packages/flame_svg/test/svg_test.dart
```diff
@@ -93,5 +93,39 @@ void main() {
         );
       },
     );
+
+    testWidgets(
+      'render sharply with viewfinder zoom',
```
spydon, line 98: It would be great if we could use the "new" way for doing the golden test:

https://github.com/flame-engine/flame/blob/main/packages%2Fflame%2Ftest%2Fcamera%2Fcamera_component_test.dart#L16

packages/flame_svg/lib/svg.dart
```diff
@@ -55,11 +55,24 @@ class Svg {
...
+        canvas.getLocalClipBounds().size.width;
+    final heightRatio = canvas.getDestinationClipBounds().size.height /
+        canvas.getLocalClipBounds().size.height;
+
+    final localSize = Size(size.x * widthRatio, size.y * heightRatio);
     final image = _getImage(localSize);
 
     canvas.save();
-    canvas.scale(1 / pixelRatio);
+    canvas.scaleVector(
```
spydon: Since we don't have a `Vector2` already we should avoid creating a new one here, just use the normal `canvas.scale(x, y)`.

packages/flame_svg/lib/svg.dart
```diff
@@ -55,11 +55,24 @@ class Svg {
...
-    final localSize = size.toSize();
+    // Scale the canvas to the size of the destination clip bounds
+    // This is necessary to avoid blurriness when having a
+    // camera.viewfinder.zoom larger than 1.0
+    final widthRatio = canvas.getDestinationClipBounds().size.width /
+        canvas.getLocalClipBounds().size.width;
+    final heightRatio = canvas.getDestinationClipBounds().size.height /
+        canvas.getLocalClipBounds().size.height;
+
+    final localSize = Size(size.x * widthRatio, size.y * heightRatio);
```
spydon: What happens if the width and height ratio aren't the same?

spydon (commented):
Looks good, just a few small comments

packages/flame_svg/test/svg_test.dart
vunyunt: Applied in 5a5a302188c928ba54027f846977f0850b4358c2

packages/flame_svg/test/svg_test.dart
vunyunt: The `testGolden` method seems to be defined in the `flame_test` package. With a search locally, I've found that only files under `packages/flame/test/` or `packages/flame_test/` references this method. Should this test be moved into `packages/flame/test`? or should the `flame_test` package be added as a dev dependency?

packages/flame_svg/lib/svg.dart
vunyunt: Applied in 7d0dcb46673f37dda87fc80ae27ea1ba837ef57f

vunyunt:
While trying to test different width/height scale, a bug in 0be4940f0c748a7fcb2d2f722d67a2915623b469 was discovered. Looking at the original code for `_getImage`:

```dart
  Image _getImage(Size size) {
    final image = _imageCache.getValue(size);

    if (image == null) {
      final recorder = PictureRecorder();
      final canvas = Canvas(recorder);
      canvas.scale(pixelRatio);
      _render(canvas, size);
      final picture = recorder.endRecording();
      final image = picture.toImageSync(
        (size.width * pixelRatio).ceil(),
        (size.height * pixelRatio).ceil(),
      );
      picture.dispose();
      _imageCache.setValue(size, image);
      return image;
    }
   
    return image;
  }
  ```
  
  We can see that the unscaled size is passed into `_render(canvas, size);`. Hence instead of passing the scaled size into `_getImage`, the width and height ratio should be passed in instead, and `_getImage` should apply the scaled size only where pixelRatio is used. This is applied in 5a5a302188c928ba54027f846977f0850b4358c2.
  
  Here's a comparison of a scaling of scaleX=3, scaleY=1.5 with and without this pr now:
|before|after|
|-|-|
|Screenshot_20250606_082444|Screenshot_20250606_082514|

Should the test be updated to have different X and Y scaling, or should an additional test be added?

File: packages/flame_svg/lib/svg.dart (hunk 4 above)
Comment (vunyunt) [reply packages/flame_svg/lib/svg.dart]: A bug was discovered when trying this out, addressed [here](https://github.com/flame-engine/flame/pull/3616#issuecomment-2947264181)

**Comment by vunyunt** [comment]
Some comments about the tests:

Golden files were mistakenly updated in 5a5a302188c928ba54027f846977f0850b4358c2 due to the `_render` function being incorrectly changed. They are restored in f97e3287fb3864514fc9896eaa861eac0cb067dd. I've verified that the golden file generated with `render sharply` after f97e3287fb3864514fc9896eaa861eac0cb067dd has the same hash with the original version.

Upon further inspection, the `pixelRatio` in `svg.dart` is always set to 3 regardless of `tester.view.devicePixelRatio`. When I tried to set `tester.view.devicePixelRatio = 1`, the output is at 100x100 but `pixelRatio` is still set to 3 for some reason. This causes the image to come out sharp even without the patch with a 2x scale, as the image is always rendered with a 3x pixel ratio. To avoid this false pass, I've updated the test to set `tester.view.devicePixelRatio = 3` explicity for now.

File: packages/flame_svg/test/svg_test.dart (hunk 2 above)
Comment (spydon) [reply packages/flame_svg/test/svg_test.dart:98]: You can add flame_test as a dev dependency here

**Comment by spydon** [comment]
> Upon further inspection, the `pixelRatio` in `svg.dart` is always set to 3 regardless of `tester.view.devicePixelRatio`. When I tried to set `tester.view.devicePixelRatio = 1`, the output is at 100x100 but `pixelRatio` is still set to 3 for some reason. This causes the image to come out sharp even without the patch with a 2x scale, as the image is always rendered with a 3x pixel ratio. To avoid this false pass, I've updated the test to set `tester.view.devicePixelRatio = 3` explicity for now.

In the `svg.dart` constructor, change to use `WidgetsBinding` instead, otherwise it won't make any difference if the DPR is changed in the test, so like this:

```dart
...
import 'package:flutter/widgets.dart' hide Image; // Remember to hide the other Image type

/// A [Svg] to be rendered on a Flame [Game].
class Svg {
  /// Creates an [Svg] with the received [pictureInfo].
  /// Default [pixelRatio] is the device pixel ratio.
  Svg(this.pictureInfo, {double? pixelRatio})
      : pixelRatio = pixelRatio ??
            WidgetsBinding
                .instance.platformDispatcher.views.first.devicePixelRatio;
```

packages/flame_svg/test/svg_test.dart
vunyunt: Applied in 85e885b2c81f809cd9aa64c25d4654289c5bd656. The test now uses the provided game instance with `camera.viewfinder.zoom` instead of a `Transform` component

vunyunt:
> In the svg.dart constructor, change to use WidgetsBinding instead

Seems to work 👍, with the new testing method however, I'm not sure how to set the device pixel ratio. Currently a teardown step is added to the 'render sharply' test that sets the device pixel ratio to 1. Using `resetDevicePixelRatio()` doesn't seem to work (stays at 3).

spydon:
> Seems to work 👍, with the new testing method however, I'm not sure how to set the device pixel ratio. Currently a teardown step is added to the 'render sharply' test that sets the device pixel ratio to 1. Using `resetDevicePixelRatio()` doesn't seem to work (stays at 3).

Just set the device pixel ratio inside of the test, or in the setup function. The tear down is called after each test has run.

vunyunt:
> > Seems to work 👍, with the new testing method however, I'm not sure how to set the device pixel ratio. Currently a teardown step is added to the 'render sharply' test that sets the device pixel ratio to 1. Using `resetDevicePixelRatio()` doesn't seem to work (stays at 3).
> 
> Just set the device pixel ratio inside of the test, or in the setup function. The tear down is called after each test has run.

Ah, I'm not sure how should this be done under either `setUp()` or `testGolden()`? The `render sharply` test seems to do this by setting the `view.devicePixelRatio` field under the `WidgetTester` instance, but `testGolden` seems to provide a game instance instead of the `WidgetTester` instance. To clarify, the cleanup step is added to the original `render sharply` test and not the newly added `Svg.withCameraZoom` test.

My assumption is that the reason it's set to 3 with the new test is because of `tester.view.devicePixelRatio = 3` in the `render sharply` test. I got the impression that it's affecting the new test from this function description of `WidgetTester.binding.setSurfaceSize`: `To avoid affecting other tests by leaking state, a test that uses this method should always reset the surface size to the default. For example, using addTearDown:`

spydon:
@vunyunt ah right, I forgot that you can't access the tester from within `testGoldens` (we should probably fix that), but for now, let's just use the old method for this test then.

vunyunt:
Alright, the test has been restored to just before 85e885b2c81f809cd9aa64c25d4654289c5bd656

spydon (approved):
Thanks for your contribution!

PR #3629, opened 2025-06-22
fix: Fix angleTo and lookAt considering parental transformations [WIP]

luanpotter:
# Description

FOO

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame/lib/src/components/position_component.dart
+++ b/packages/flame/lib/src/components/position_component.dart
@@ -398,23 +398,60 @@ class PositionComponent extends Component
   /// Note: If target coincides with the current component, then it is treated
   /// as being north.
   double angleTo(Vector2 target) {
-    final absoluteScale = this.absoluteScale;
-    var atanA = target.x - absolutePosition.x;
-    var atanB = absolutePosition.y - target.y;
-    if (absoluteScale.x.isNegative) {
-      atanA = absolutePosition.x - target.x;
-    }
-    if (absoluteScale.y.isNegative) {
-      atanB = target.y - absolutePosition.y;
-    }
-    var angleDifference =
-        math.atan2(atanA, atanB) - (nativeAngle + absoluteAngle);
-    if (absoluteScale.y.isNegative) {
-      angleDifference -= math.pi;
-    }
-    final flippedRotation =
-        absoluteScale.x.isNegative ^ absoluteScale.y.isNegative;
-    return (flippedRotation ? -angleDifference : angleDifference);
+    //final absoluteScale = this.absoluteScale;
+    //var atanA = target.x - absolutePosition.x;
+    //var atanB = absolutePosition.y - target.y;
+    //if (absoluteScale.x.isNegative) {
+    //  atanA = absolutePosition.x - target.x;
+    //}
+    //if (absoluteScale.y.isNegative) {
+    //  atanB = target.y - absolutePosition.y;
+    //}
+    //var angleDifference =
+    //    math.atan2(atanA, atanB) - (nativeAngle + absoluteAngle);
+    //if (absoluteScale.y.isNegative) {
+    //  angleDifference -= math.pi;
+    //}
+
+    final parentAbsoluteScale =
+        ancestors().whereType<ReadOnlyScaleProvider>().fold<Vector2>(
+              Vector2.all(1.0),
+              (totalScale, c) => totalScale..multiply(c.scale),
+            );
+
+    final direction = target - absolutePosition;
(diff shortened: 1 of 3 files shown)


packages/flame/lib/src/components/position_component.dart
```diff
@@ -411,8 +417,9 @@ class PositionComponent extends Component
     final hasSelfYFlip =
         !parentAbsoluteScale.y.isNegative && scale.y.isNegative;
 
-    return (hasOddFlips ? -1 : 1) * angleDifference +
+    final result = (hasOddFlips ? -1 : 1) * angleDifference +
         (hasSelfYFlip ? 1 : 0) * math.pi;
+    return result % tau;
```
spydon: I'd say we should normalize this between [-pi, pi], since then the user knows which way to rotate the shortest way without having to do anything more.

luanpotter: ah yes good call. I originally had that but it was failing tests, so I changed to 0,tau. but that also failed the tests. I am not sure what the test authors were expecting exactly. eventually I gave up and just add he same normalize to the test
I think there is a bigger q of do we normalize angles on flame, where, and do we have a standard. we could export a common util for now, but I think we should start doing it on angle setters for example

spydon: >  I think there is a bigger q of do we normalize angles on flame, where, and do we have a standard. we could export a common util for now, but I think we should start doing it on angle setters for example

I think this place is the most important one, so we could start here in this PR and then we could make a list of other places that we'd like to normalize and do that in a follow-up?

spydon: Done in c8a30b2, will add an extra note about it in the release notes.

luanpotter: I was a bit skittish of doing it and left a TODO instead, but let's do it! should've been done since the beginning
I suspect it could potentially break some super edge case handling but nothing that couldn't be very easily fixed by the user (also I can't come up with a specific example rn)

packages/flame/lib/src/components/position_component.dart
```diff
@@ -229,21 +227,46 @@ class PositionComponent extends Component
...
+      }
+
+      if (ancestor is ReadOnlyAngleProvider) {
+        final reflected = totalScaleX.isNegative ^ totalScaleY.isNegative;
+        final localAngle = (ancestor as ReadOnlyAngleProvider).angle;
+        angle += reflected ? -localAngle : localAngle;
+      }
+    }
+
+    return angle % tau;
```
luanpotter: use toNormalizedAngle here?

luanpotter (commented):
@spydon LGTM but I left a few comments [at the commit level](https://github.com/flame-engine/flame/commit/c8a30b23158f91fc6cb300c017c4399d125d934b)

gnarhard:
Hey guys, I just recently went down a trig rabbit hole watching a video about the Tau Manifesto and it brought up Euler's Identity. This equation could save the engine from calling atan2 every frame and solve angle wrapping. It's also a beautiful equation when written with Tau and might visually simplify your calculations. I know there's A LOT more going on than just making one vector look at another, but I wanted to bring it to your attention just in case you found it to be a more elegant solution.

spydon:
@gnarhard I'm not sure it would help, but if you have an idea of how to simplify it further, please post an example code snippet for it. :)

spydon (approved):
Lgtm! We'll continue to look at whether it makes sense to take anchors into consideration in a follow-up.

itsdax:
This may be categorized as a breaking change(?) since there may be existing code calling lerpDouble to a certain angle.

```
set angle(double a) => transform.angle = a.toNormalizedAngle();
```

With this change, this will cause the lerp to never reach the target (since it's normalized) and cause an infinite lerp.

~~A simple fix would be to call lerpDouble on a normalized target angle first, but it took me a while to figure out what was breaking my rotation animations.~~ Ah, this causes problems when I'm trying to rotate full rotations, since if I want to rotate a full 360, a normalized angle is a no-op.

spydon:
@itsdax ah true, it should indeed have been marked as a breaking change. I wonder if all `RotateEffect`s that are larger than pi are broken now... Maybe we'll have to revert normalization on the setter.

spydon:
@itsdax the `RotateEffect`s still work, since they calculate the diff when they start and then just increment by deltas, maybe you could use an effect instead of manually lerping?

itsdax:
Effects are a valid alternative :) Can confirm it works

PR #3645, opened 2025-07-16
feat!: AnimatedSpriteBatchComponent

gnarhard:
# Description

1. Add AnimatedSpriteBatchComponent for optimizing the display of animated sprites.
2. Add priority param to SpriteBatchComponent for more flexible rendering priority.
3. Add children param to SpriteBatchComponent to allow RemoveEffect.
4. Make SpriteBatchComponent properties final to gain minor performance improvements.

AnimatedSpriteBatchComponent could probably be optimized using the replace() method versus clearing and re-adding every frame. Maybe one of you guys knows how to do that. I got intimidated by the RSTransform object.

## Checklist

## Breaking Change?

Make sure to set spriteBatch in the constructor.

## Related Issues

### DIFF
--- a/packages/flame/lib/src/components/animated_sprite_batch_component.dart
+++ b/packages/flame/lib/src/components/animated_sprite_batch_component.dart
@@ -19,66 +19,69 @@ class AnimatedBatchComponent extends SpriteBatchComponent {
 
   final SpriteAnimation _animation;
   final List<Rect> _sources;
-  final List<PositionComponent> _components = [];
-  final List<SpriteAnimationTicker> _tickers = [];
-  final List<int> _animationIndexes = [];
+
+  final List<_BatchItem> _items = [];
 
   // Reuse these objects to avoid allocations
   final _offset = Vector2.zero();
-  final _anchor = Vector2.zero();
-  final _componentScale = Vector2.zero();
 
   void addToBatch(PositionComponent component) {
-    _components.add(component);
-    _tickers.add(SpriteAnimationTicker(_animation));
-    _animationIndexes.add(0);
-    _anchor.setValues(
+    final anchor = Vector2(
       component.anchor.x * component.size.x,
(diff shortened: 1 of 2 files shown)


packages/flame/lib/src/components/animated_sprite_batch_component.dart
```diff
@@ -0,0 +1,64 @@
+import 'package:flame/components.dart';
+import 'package:flame/extensions.dart';
+
+class AnimatedBatchComponent extends SpriteBatchComponent {
```
erickzanardo, line 5: This is interesting, but I am a big confused about the use case of this component?

It has the prefix `Animated`, but it feels more like a SpriteBacth that renders at target position components?

Can you elaborate a bit to help me understand?

gnarhard: You pass in your sources (sprite.srcPosition, sprite.srcSize) and your desired stepTime of your Sprites in an animation, then it automatically handles animating between your sprites for you versus managing stepTimes and animation indexes outside of the AnimatedSpriteBatchComponent. It's more performant because of that and also far more convenient.

spydon (commented):
Is there a lot of difference in doing this compared to using a `SpriteAnimationTicker`, because that would be more consistent with the other classes.

gnarhard:
An array of tickers would allow each animation to have unique animation start and end times which would be nice. Im primarily using this class to improve performance so as long as the tickers don't have much of an impact that could be a worthwhile enhancement.

gnarhard:
@erickzanardo You were right to question this. My class adds a dependency on component which makes it so that you have to add a component every time you want to a display an animation. I think using .onFrame() and .onComplete() from SpriteAnimationTicker to manage batch item state is actually more flexible, performant, and removes the need for a separate class to maintain. Regardless, I appreciate you guys humoring me on this.

PR #3647, opened 2025-07-17
perf!: TexturePacker optimizations

### LINKED ISSUE #3644: [flame_texturepacker] make Region immutable
### What could be improved

The Region class should be immutable and all redundant fields in TexturePackerSprite should be removed. WIP: `perf/immutable_region_data`

https://github.com/gnarhard/flame/tree/perf/immutable_region_data

This _parse method is also not the friendliest to read. I think there could be some improvement there too.

I've already confirmed that this will reduce TexturePackerSprite memory consumption by ~50%. These fields that are in the TexturePackerSprite class that I've removed are redundant and already stored in the region data. My branch removes all of that code that shouldn't be mutated anyways, and makes Region immutable. However, I'm hung up on the tests and need help here.

_parse method could be improved to be easier to read and maintain.

### Risks

Currently, I don't have access to the texture pack files in the testing environment. Each field needs to be set every time, but since I don't have access to the texture pack save files, I can't check and see which algorithm is being used to update the atlas files accordingly.

### More information

_No response_

### Other

gnarhard:
# Description

1. Make Region class immutable and bit pack properties.
2. Refactor _parse method to be more readable and to support the new immutable Region object instantiation.
3. Breaking Change: Remove redundant fields in TexturePackerSprite, but make region public.
4. Breaking Change: Limit sprite texture pack to 4096x4096px for massive performance gains.
5. Add Region data test.

## Checklist

## Breaking Change?

Region data contained within TexturePackerSprite is now encapsulated in the region property.
Texture packs are now limited to 4096x4096px.

## Related Issues
Closes #3644

### DIFF
--- a/packages/flame_texturepacker/lib/src/model/region.dart
+++ b/packages/flame_texturepacker/lib/src/model/region.dart
@@ -71,6 +71,6 @@ final class Region {
     this.degrees = 0,
     this.rotate = false,
     this.index = -1,
-  })  : originalWidth = originalWidth ?? width,
-        originalHeight = originalHeight ?? height;
+  }) : originalWidth = originalWidth ?? width,
+       originalHeight = originalHeight ?? height;
 }
--- a/packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
+++ b/packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
@@ -233,8 +233,9 @@ Future<TextureAtlasData> _parse(
       ? await XFile(path).readAsString()
       : await (assets ?? Flame.assets).readFile('${assetsPrefix!}/$path');
 
-  final lines =
-      LineSplitter.split(fileContent).where((line) => line.trim().isNotEmpty);
+  final lines = LineSplitter.split(
+    fileContent,
+  ).where((line) => line.trim().isNotEmpty);
 
   final lineQueue = ListQueue<String>.from(lines);
   images ??= Flame.images;
@@ -313,8 +314,9 @@ Future<Page> _parsePage(
   page.textureFile = lineQueue.removeFirst();
 
   final parentPath = (path.split('/')..removeLast()).join('/');
-  final texturePath =
-      parentPath.isEmpty ? page.textureFile : '$parentPath/${page.textureFile}';
+  final texturePath = parentPath.isEmpty
+      ? page.textureFile
+      : '$parentPath/${page.textureFile}';
 
   if (fromStorage) {
     final bytes = await XFile(texturePath).readAsBytes();
--- a/packages/flame_texturepacker/lib/src/texture_packer_sprite.dart
+++ b/packages/flame_texturepacker/lib/src/texture_packer_sprite.dart
@@ -13,14 +13,14 @@ import 'package:flutter/material.dart';
 class TexturePackerSprite extends Sprite {
   /// {@macro _texture_packer_sprite}
   TexturePackerSprite(this.region, {this.useOriginalSize = true})
-      : super(
-          region.page.texture,
-          srcPosition: Vector2(region.left, region.top),
-          srcSize: Vector2(
-            useOriginalSize ? region.originalWidth : region.width,
-            useOriginalSize ? region.originalHeight : region.height,
-          ),
-        ) {
+    : super(
+        region.page.texture,
+        srcPosition: Vector2(region.left, region.top),
+        srcSize: Vector2(
+          useOriginalSize ? region.originalWidth : region.width,
+          useOriginalSize ? region.originalHeight : region.height,
+        ),
+      ) {
     if (region.rotate) {
       final transform = Transform2D()..angle = math.pi / 2;
       _decorator = Transform2DDecorator(transform);
@@ -54,9 +54,9 @@ class TexturePackerSprite extends Sprite {
   Vector2 get _packedSize => Vector2(region.width, region.height);
 
   Vector2 get _originalSize => Vector2(
-        region.originalWidth,
-        region.originalHeight,
-      );
+    region.originalWidth,
+    region.originalHeight,
+  );
 
   Vector2 get offset => useOriginalSize ? _offset : Vector2.zero();
 


gnarhard:
Just wanna mention that the breaking change impact of this is extremely low. I don't think most people will be creating textures with sizes over 4096x4096. I'm currently still using 2048x2048 for max compatibility on phones. The properties that were in TexturePackerSprite and moved within region should be an extremely simple and easy fix.

erickzanardo (approved):
Changes looks good to me, but I would like a second review from @spydon before landing it!

packages/flame_texturepacker/lib/src/model/region.dart
```diff
@@ -1,18 +1,122 @@
+import 'dart:typed_data';
 import 'package:flame_texturepacker/src/model/page.dart';
+import 'package:flutter/foundation.dart';
 
 /// Represents a region within the texture packer atlas.
+@immutable
 class Region {
```
spydon: This could be final, because it is never extended right?

packages/flame_texturepacker/lib/src/model/region.dart
```diff
@@ -1,18 +1,122 @@
...
+  /// Uses 12 bits (supports 0-4095) with some values packed together.
+  /// Format:
+  /// [left|top, width|height, offsetX|offsetY, originalWidth, originalHeight]
+  final Uint16List _packedData;
+
+  /// Packed rotation data:
+  /// bits 0-8 for degrees, bit 9 for rotate flag, bits 10-31 for index
+  final int _rotationAndIndex;
+
+  static const int _maxValue = 4095; // 12 bits max
```
spydon: Is this really needed? It should be possible to create an array of the size that is needed without restricting the max value and without impacting memory or performance.

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
```diff
@@ -184,172 +226,238 @@ Future<TextureAtlasData> _parse(
...
   final regions = <Region>[];
-  final String fileAsString;
+  var hasIndexes = false;
+
+  final fileContent = fromStorage
+      ? await XFile(path).readAsString()
+      : await (assets ?? Flame.assets).readFile('${assetsPrefix!}/$path');
+
+  final lines = LineSplitter.split(fileContent).toList();
+  final it = _LookaheadIterator(lines.iterator);
```
spydon: Since lines already is a materialized list, does it really make sense to use the lookahead iterator? This could be a Stack or a ListQueue, then we don't need a custom class for the iterator.

gnarhard: Good call! I wasn't aware of ListQueue before you mentioned it. I updated the code accordingly.

packages/flame_texturepacker/lib/src/model/region.dart
gnarhard: No, it wasn't. Whoops! Since Uint16List allows values from 0-65K, we should have plenty of headroom here. I looked into Uint8List, but that's too limiting at 0-255. I removed the hard limit in the code and in the docs. We should be good to go.

packages/flame_texturepacker/lib/src/texture_packer_sprite.dart
```diff
@@ -34,77 +24,39 @@ class TexturePackerSprite extends Sprite {
...
   }
 
   /// Region object for [clone] function, don't modify this object properties.
-  final Region _region;
+  final Region region;
 
-  /// If true, use [originalWidth] and [originalHeight] as size; otherwise use
-  /// [packedWidth] and [packedHeight] as size.
+  // /// If true, use [originalWidth] and [originalHeight] as size; otherwise use
+  // /// [width] and [height] as size.
```
spydon:
```suggestion
  /// If true, use [originalWidth] and [originalHeight] as size; otherwise use
  /// [width] and [height] as size.
```

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
```diff
@@ -184,172 +227,251 @@ Future<TextureAtlasData> _parse(
...
-  final String fileAsString;
+  var hasIndexes = false;
+
+  final fileContent = fromStorage
+      ? await XFile(path).readAsString()
+      : await (assets ?? Flame.assets).readFile('${assetsPrefix!}/$path');
+
+  final lines = LineSplitter.split(fileContent)
+      .where((line) => line.trim().isNotEmpty)
+      .toList();
```
spydon: I don't think you need to do `toList` here, it should be possible to materialize it in the `ListQueue` creation instead.

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
```diff
@@ -184,172 +227,251 @@ Future<TextureAtlasData> _parse(
...
+      }
+
+      regions.add(region);
+    }
+  }
+
+  if (hasIndexes) {
+    regions.sort(
+      (a, b) {
+        final i1 = a.index == -1 ? double.maxFinite.toInt() : a.index;
```
spydon: You could declare a `0x7fffffff` int const here and then you don't have to do the conversion from double.

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
```diff
@@ -184,172 +227,251 @@ Future<TextureAtlasData> _parse(
...
+  final offsets = values['offsets'];
+  final rotate = values['rotate'];
+  final index = values['index'];
+
+  final offsetOrNull = offsets ?? offset;
+
+  final offsetX = offsetOrNull != null ? double.parse(offsetOrNull[0]) : 0.0;
+  final offsetY = offsetOrNull != null ? double.parse(offsetOrNull[1]) : 0.0;
+
+  final origWidth = offsets != null
```
spydon: No shortened words

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
```diff
@@ -184,172 +227,251 @@ Future<TextureAtlasData> _parse(
...
+      ? double.parse(offsets[2])
+      : (orig != null ? double.parse(orig[0]) : 0.0);
+
+  final originalWidth = origWidth == 0.0 ? null : origWidth;
+
+  final origHeight = offsets != null
```
spydon: No shortened words

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
gnarhard: Good catch. Removed.

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
gnarhard: Hey, well that's a pretty neat little trick!

gnarhard:
Before 1,443.5KB:
Screenshot 2025-07-22 at 2 56 07 PM
After 631.6KB:
Screenshot 2025-07-22 at 2 54 43 PM

That's a 56.3% reduction in memory usage! That's what's up!

spydon:
@gnarhard there are some analyze issues on the latest Dart. Just some imports missing for dartdocs (in a future version of Dart there will be special imports for these types of imports).

gnarhard:
@spydon not sure what you need

spydon:
> @spydon not sure what you need

```
Dart Analyzer - Infos
  /home/runner/work/flame/flame/packages/flame_texturepacker/lib/src/texture_packer_sprite.dart:35:21 The referenced name isn't visible in scope.
  Notice: Try adding an import for the referenced name.
  See https://dart.dev/lints/comment_references to learn more about this problem.
  
  /home/runner/work/flame/flame/packages/flame_texturepacker/lib/src/texture_packer_sprite.dart:35:41 The referenced name isn't visible in scope.
  Notice: Try adding an import for the referenced name.
  See https://dart.dev/lints/comment_references to learn more about this problem.
  
  /home/runner/work/flame/flame/packages/flame_texturepacker/lib/src/texture_packer_sprite.dart:36:8 The referenced name isn't visible in scope.
  Notice: Try adding an import for the referenced name.
  See https://dart.dev/lints/comment_references to learn more about this problem.
  
  /home/runner/work/flame/flame/packages/flame_texturepacker/lib/src/texture_packer_sprite.dart:36:20 The referenced name isn't visible in scope.
  Notice: Try adding an import for the referenced name.
  See https://dart.dev/lints/comment_references to learn more about this problem.
  ```

File: packages/flame_texturepacker/lib/src/model/region.dart [hunk 9]
```diff
@@ -1,18 +1,108 @@
...
+    int degrees = 0,
+    bool rotate = false,
+    int index = -1,
+  })  : assert(left >= 0 && left <= 65535, 'left must be 0-65535'),
+        assert(top >= 0 && top <= 65535, 'top must be 0-65535'),
+        assert(width >= 0 && width <= 65535, 'width must be 0-65535'),
+        assert(height >= 0 && height <= 65535, 'height must be 0-65535'),
+        assert(offsetX >= 0 && offsetX <= 65535, 'offsetX must be 0-65535'),
+        assert(offsetY >= 0 && offsetY <= 65535, 'offsetY must be 0-65535'),
+        _packedData = Uint16List.fromList([
```
spydon: I don't see the point of having the values represented within a list here, it might save a few bytes in memory, but it will make each access to the values much more expensive than if they would just be declared as final variables instead, since the getters will have to handle conversions and list bounds checks for each call.

gnarhard: The whole point is saving bytes in memory. As you could see from my profile above, TexturePackerSprite and Region can bloat your memory pretty quickly (especially if you're not whitelisting the loading of your textures, which I was in my profile). To your point, protecting the CPU is typically more valuable than RAM. It seems like lower-end devices seem to have capable RAM and lack capable CPUs. Do you know of some sort of test we can do to get a better gauge of how this will impact the CPU so we can make an informed decision? Technically, it's only accessing left/top, and original or adjusted width/height in the constructor of TexturePackerSprite. Nothing else is being accessed every frame. For a high-quality game, you _should_ be loading your sprites into memory before the game begins anyways. . .

spydon: So from your example it saves like half a megabyte for 5774 Regions, it seems to me like this isn't the place to be chasing memory optimizations. Even if you'd have a looot more of these, which won't be very common, the memory footprint wouldn't be a problem for anything that can run Flutter, which is why I'd say that this is a premature optimization and thus we should prioritize simpler and more CPU efficient code instead.

gnarhard: I agree with you. The CPU's resources are more precious here. I'll remove the bit packing.

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
```diff
@@ -184,172 +227,247 @@ Future<TextureAtlasData> _parse(
...
+/// [line] - The line to check
+///
+/// Returns true if the line looks like a texture file path.
+bool _isTextureFile(String line) {
+  final trimmed = line.trim();
+
+  // Check for common image file extensions
+  const imageExtensions = ['.png', '.jpg', '.jpeg', '.bmp', '.tga', '.webp'];
+
+  for (final ext in imageExtensions) {
```
spydon: No abbreviations

luanpotter (approved):
LGTM

spydon (approved):
Apparently I had one last comment from a long time ago that hadn't been submitted, but after the tests pass and the branch is up to date I think we can merge this. :)

packages/flame_texturepacker/lib/src/texture_packer_atlas.dart
luanpotter: spydon's relentless crusade against abbreviations :P

gnarhard:
@spydon you also wanted me to refactor the way we parse the file. Still want that done before merge too?

luanpotter:
@gnarhard I'd leave it for a followup, since this is already a good chunk of work on its own :)

are you looking into the merge conflicts / test failures?

gnarhard:
@luanpotter @spydon I am not sure what to do on that failing test. I don't have enough context nor do I have time to gain the context right now :/ I addressed everything else.

luanpotter:
No worries at all! As far as I can see on the logs the broken test appears to potentially be unrelated. Let me take a look and see if I can figure something out.
```
❌ /home/runner/work/flame/flame/packages/flame/test/game/transform2d_test.dart: Transform2D random [seed=2021996283] (failed)
  Expected: a Vector2 object within 0.00001721491496360162 of (2.0070884227752686, 2.1730268001556396)
    Actual: Vector2:<[2.007079839706421,2.173011302947998]>
     Which: is at distance 0.00001771531866861336
  
  package:matcher                                expect
  test/game/transform2d_test.dart 255:9          main.<fn>.<fn>
  package:flame_test/src/random_test.dart 50:17  testRandom.<fn>
  ```

luanpotter:
ok it was just a fluke, does not seem related at all to the PR. in fact you seemed to have gotten EXTREMELY lucky - from my local testing the seed `2021996283` fails consistently, but no other random seed that I've tried.

I will open an issue but this is definitely unrelated and probably just a precision issue with this particular seed.

luanpotter:
Added an issue for the unrelated issue https://github.com/flame-engine/flame/issues/3692
Will merge this for the next release!

PR #3648, opened 2025-07-18
fix: Attach layout listeners to new children

livtanong:
# Description

Prior to this fix, when adding a child to the `LayoutComponent`, resizing the child wouldn't cause the `LayoutComponent` to resize. This was because the children size listeners were only attached on instantiation of `LayoutComponent` *and* during setting of `shrinkWrap`.

This fix attaches the size listeners to a single child when it is added to `LayoutComponent` via `onChildrenChanged`.
`setupSizeListeners` is not used because it iterates through all the children, and is unnecessary.

## Checklist

## Breaking Change?

## Related Issues


packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -175,6 +175,15 @@ abstract class LayoutComponent extends PositionComponent {
 
   @override
   void onChildrenChanged(Component child, ChildrenChangeType type) {
+    if (child is! PositionComponent) {
```
spydon, line 178: This could be `ReadOnlySizeProvider` instead

spydon (approved):
Except for the small nit it looks good!

packages/flame/lib/src/experimental/layout_component.dart
spydon: On the other hand, we're not doing that for the initial setup either, so I guess this will be enough for now. 

Another thing I thought about is that we probably should also listen to the position of the children, since that could move children outside of the shrinkwrap.

livtanong: @spydon but the LayoutComponents are supposed to fully control the positions of its children, no? Like, they align the children in rows or columns, and align them in the center or whatnot. So for any of the children to be positioned in an unexpected way could be considered a misuse of the LayoutComponent.

spydon: Ah, that's true! Then it is all correct :)

PR #3650, opened 2025-07-21
feat: Add optional ID to BatchItem and methods for managing items by ID in SpriteBatch

gnarhard:
# Description

This PR adds functionality to make it easier to manage BatchItems within a SpriteBatch. You can pass in an ID which is a unique string to various methods that allow you to add, find, replace, and remove a specific BatchItem. I don't think this needs examples as it's pretty self-explanatory so I marked that as done. I'm happy to add tests for this if you guys want me to move forward with that. Let me know!

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/packages/flame/lib/src/sprite_batch.dart
+++ b/packages/flame/lib/src/sprite_batch.dart
@@ -37,10 +37,14 @@ class BatchItem {
   BatchItem({
     required this.source,
     required this.transform,
+    this.id,
     Color? color,
     this.flip = false,
   })  : paint = Paint()..color = color ?? const Color(0x00000000),
         destination = Offset.zero & source.size;
+
+  /// Optional identifier for the batch item.
+  final String? id;
 
   /// The source rectangle on the [SpriteBatch.atlas].
   final Rect source;
@@ -143,6 +147,13 @@ class SpriteBatch {
   }
 
   FlippedAtlasStatus _flippedAtlasStatus = FlippedAtlasStatus.none;
+
+  /// A map to keep track of the index of each batch item by its id.
+  final Map<String, int> _idToIndex = {};
+
+  /// Returns all ids currently in the batch (excluding nulls).
+  Iterable<String> get ids =>
+      _batchItems.where((item) => item.id != null).map((item) => item.id!);
 
   /// List of all the existing batch items.
   final _batchItems = <BatchItem>[];
@@ -300,8 +311,10 @@ class SpriteBatch {
     RSTransform? transform,
     bool flip = false,
     Color? color,
+    String? id,
   }) {
     final batchItem = BatchItem(
+      id: id,
       source: source,
       transform: transform ??= defaultTransform ?? RSTransform(1, 0, 0, 0),
       flip: flip,
@@ -385,12 +398,101 @@ class SpriteBatch {
     );
   }
 
+  /// Finds the index of the batch item with the given [id].
+  int? findIndexById(String id) {
+    if (_idToIndex.containsKey(id)) {
+      return _idToIndex[id];
+    }
+    for (var i = 0; i < _batchItems.length; i++) {
+      if (_batchItems[i].id == id) {
+        _idToIndex[id] = i; // repair mapping
+        return i;
+      }
+    }
+    return null;
+  }
+
+  /// Removes a batch item by its [id].
+  void removeById(String id) {
+    final index = _idToIndex[id];
+    if (index == null) {
+      return;
+    }
+
+    removeAt(index);
+    _idToIndex.remove(id);
+
+    // adjust indices > removed index
+    _idToIndex.updateAll((key, idx) => idx > index ? idx - 1 : idx);
+  }
+
+  /// Removes a batch item at the given [index].
+  void removeAt(int index) {
+    if (index < 0 || index >= length) {
+      throw ArgumentError('Index out of bounds: $index');
+    }
+
+    _batchItems.removeAt(index);
+    _sources.removeAt(index);
+    _transforms.removeAt(index);
+    _colors.removeAt(index);
+  }
+
+  /// Adds a new batch item with the given [id].
+  int addWithId(
+    String id, {
+    required Rect source,
+    RSTransform? transform,
+    bool flip = false,
+    Color? color,
+  }) {
+    final idx = findIndexById(id);
+    if (idx != null) {
+      replace(idx, source: source, transform: transform, color: color);
+      return idx;
+    }
+
+    final item = BatchItem(
+      id: id,
+      source: source,
+      transform: transform ?? defaultTransform ?? RSTransform(1, 0, 0, 0),
+      flip: flip,
+      color: color ?? defaultColor,
+    );
+
+    _batchItems.add(item);
+    _sources.add(item.source);
+    _transforms.add(item.transform);
+    _colors.add(color ?? _defaultColor);
+
+    final newIdx = _batchItems.length - 1;
+    _idToIndex[id] = newIdx;
+
+    return newIdx;
+  }
+
+  /// Replaces the batch item identified by [id] with new data.
+
+  void replaceById(
+    String id, {
+    Rect? source,
+    Color? color,
+    RSTransform? transform,
+  }) {
+    final index = _idToIndex[id];
+    if (index == null) {
+      throw ArgumentError('No BatchItem found with id: $id');
+    }
+    replace(index, source: source, color: color, transform: transform);
+  }
+
   /// Clear the SpriteBatch so it can be reused.
   void clear() {
     _sources.clear();
     _transforms.clear();
     _colors.clear();
     _batchItems.clear();
+    _idToIndex.clear();
   }
 
   // Used to not create new Paint objects in [render] and


packages/flame/lib/src/sprite_batch.dart
```diff
@@ -385,12 +398,101 @@ class SpriteBatch {
...
+    }
+
+    _batchItems.removeAt(index);
+    _sources.removeAt(index);
+    _transforms.removeAt(index);
+    _colors.removeAt(index);
+  }
+
+  /// Adds a new batch item with the given [id].
+  int addWithId(
```
erickzanardo: Can't we just include an optional id argument in the `add` method instead?

Even if internally on the `add` method we did something like:

```dart
if (id != null) {
  _addWithId(.....)'// <- private method
}
```

That way we don't need to add a somehow redundant method in the SpriteBatch API, and it would work the same.

Additionally, I feel it is important to mention on the docs of the method that when adding with an id already existent, it will replace the old one instead.

erickzanardo (commented):
Great idea! I added one comment and I think we need tests for the added functionality!

packages/flame/lib/src/sprite_batch.dart
gnarhard: Thanks! I just updated the branch to include:
1. Simplified logic based on your suggestion.
2. SpriteBatch tests that target this new functionality.
3. Docs for the new ID management feature.

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -144,6 +148,13 @@ class SpriteBatch {
 
   FlippedAtlasStatus _flippedAtlasStatus = FlippedAtlasStatus.none;
 
+  /// A map to keep track of the index of each batch item by its id.
+  final Map<String, int> _idToIndex = {};
+
+  /// Returns all ids currently in the batch (excluding nulls).
+  Iterable<String> get ids =>
+      _batchItems.where((item) => item.id != null).map((item) => item.id!);
```
spydon: Maybe `_batckItems` could be a map instead, so that you'd get constant lookup.

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -144,6 +148,13 @@ class SpriteBatch {
 
   FlippedAtlasStatus _flippedAtlasStatus = FlippedAtlasStatus.none;
 
+  /// A map to keep track of the index of each batch item by its id.
+  final Map<String, int> _idToIndex = {};
```
spydon: What is this used for?

packages/flame/lib/src/sprite_batch.dart
spydon: Ah right, that wouldn't work since they can be null. But you at least don't have to do two iterations here, you can remove the where not null one.

packages/flame/lib/src/sprite_batch.dart
gnarhard: I use this in all of the new methods for finding the index of the BatchItem from the id.

packages/flame/lib/src/sprite_batch.dart
gnarhard: Done.

spydon:
Can you explain your use-case a bit further? I'm not sure I understand why this is needed to do animations based on sprite batches, and whether that is even anything that the sprite batch should be aware of. It seems to me like one would want all the sprites related to an animation within the atlas, not replace the entries.

gnarhard:
In my use case, I often spawn hundreds of orbs that contain health. Sprite batching significantly improves performance here. Each orb has four states: materialize, idle, dissipate, and collect.
	•	Materialize plays once and transitions to idle, which loops 4 times.
	•	Then it switches to dissipate, unless a collision occurs.
	•	On collision, the current animation is interrupted and replaced with collect.

All animations are texture-packed into a single image. I’m generating BatchItems dynamically for the materialize animation, then replacing and eventually removing them.

Currently, SpriteBatch management is index-based. While I could store and track indices in my orb model, it’s unreliable—especially when animations change rapidly. More importantly, there’s no built-in way to remove a BatchItem by index, which is what inspired this PR.

You wouldn’t manage database indexes outside the database—likewise, managing BatchItems by a unique ID makes CRUD operations more reliable and intuitive. This approach ensures I always target the correct item, even as orb states change.

Regarding your comment about animation sprites needing to be within the atlas: to use drawAtlas() efficiently for hundreds of animated orbs, I need to create and replace BatchItems on every frame. Unless there’s a better methodology I’ve missed, that’s currently the only way to animate with sprite batching.

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -144,6 +148,12 @@ class SpriteBatch {
...
+
+  /// Returns all ids currently in the batch (excluding nulls).
+  Iterable<String> get ids => _batchItems.map((item) => item.id!);
```
spydon: This would currently crash if the ids are null (like you did check for before).

Since you already have the ids in `_idToIndex` you can use the keys from that.
```suggestion
  Iterable<String> get ids => _idToIndex.keys;
```

Another way would be the following (but not necessary here):
```suggestion
  Iterable<String> get ids => [for (final item in _batchItems) ?item.id];
```

spydon (commented):
* It would be good to have a simple example of how to use it with animations.
* `removeAt`, `removeById` and `ids` are lacking tests.

Other than that and the comment on `ids` I think it looks good!

gnarhard:
@spydon A problem I'm running into currently with this implementation is race conditions if you're mutating the index from separate components. I've implemented a Free List Strategy to account for this and it's working perfectly in my game. Here's an overview of my changes:

### Error safety improvements
- Fixes race conditions caused by mutating indexes from multiple different sources
  - By not shifting indices or compacting arrays during mutation, render/update logic stays consistent.
  - Prevents bugs caused by index mismatches or invalid lookups during concurrent access within a single frame.
- Enables Safe Deferred Mutation
  - Works perfectly with a “schedule and flush” mutation model (e.g. deferred removeById()), which avoids unsafe access during the render loop.

### BONUS! Performance improvements
- No shifting of indices
  - removeAt() is O(1), no map updates
- Reuse of index slots
  - Reduces memory churn, avoids growth
- Sparse batch safety
  - Works even when many items are removed
- Fast add and remove
  - Constant time for typical operations

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -144,37 +148,83 @@ class SpriteBatch {
...
+  /// The transforms that should be applied on the [_sources], stored sparsely.
+  final Map<int, RSTransform> _transforms = {};
 
-  /// The transforms list shouldn't be modified directly, that is why an
-  /// [UnmodifiableListView] is used. If you want to add transforms use the
-  /// [add] or [addTransform] method.
-  UnmodifiableListView<RSTransform> get transforms {
-    return UnmodifiableListView<RSTransform>(_transforms);
+  /// Returns a compact list of transforms for rendering.
+  List<RSTransform> get transforms {
```
spydon: Getting the transforms looks like it will be considerably more expensive after this change.

gnarhard: Ah, totally. I'll implement some caching.

spydon: Is it really worth doing this second change? Have you measured the performance differences?

gnarhard: Ya, that was really bad. Let me know if the new code looks better to you. Sorry about that slop!

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -144,37 +148,99 @@ class SpriteBatch {
...
+  bool _dirty = true;
+
+  /// The lists used for rendering the batch items.
+  final List<Rect> _sourcesList = [];
+
+  /// The transforms used for rendering the batch items.
+  final List<RSTransform> _transformsList = [];
+
+  /// The colors used for rendering the batch items.
+  final List<Color> _colorsList = [];
```
spydon: Why are the `BatchItem`s not responsible for their own properties? :thinking: 
It feels a quite messy having them here, and I don't see the performance gain?

spydon (commented):
Maybe this PR should be split up in two, one that is introducing the `id`, which should be quite simple to land. And then one with the performance changes.

spydon:
@gnarhard can you have a look at why the tests are failing?
I updated the PR to align with Flutter 3.35, so remember to pull first.

gnarhard:
@spydon I need your help on fixing these tests. These tests were passing before the move to Flutter 3.35 and other changes to main. I'm wondering if the precision issues are related to Matrix4 returning a Float32List instead of a Float64List. I combed through the file and compared it to main, but I can't figure out why the goldens are different and I don't see a reason for that to happen from my updates.

spydon:
> @spydon I need your help on fixing these tests. These tests were passing before the move to Flutter 3.35 and other changes to main. I'm wondering if the precision issues are related to Matrix4 returning a Float32List instead of a Float64List. I combed through the file and compared it to main, but I can't figure out why the goldens are different and I don't see a reason for that to happen from my updates.

I squashed all the commits in this PR so that it can easily cherry-picked into an older version of Flame and try the tests there. I don't think this is related to any precision issues, because as you can see in the image diffs that I linked, some tiles are just completely gone.

spydon:
@gnarhard I cherry-picked this PR onto Flame v1.29.0 and used Flutter 3.27.1 to run the tests, and the same goldens still fail, so I'm pretty certain the bug is within this PR and not external.

gnarhard:
Okay, thanks for doing that. I'll look into it again.

gnarhard:
@spydon I’ve taken several 1-2 hour bite-sized chunks of time trying to figure out why those flame_tiled tests are failing and I’m still puzzled. Now that I’m home from tour I have a bit more resources and time to look into this. Just wanted to give you an update.

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -307,26 +317,15 @@ class SpriteBatch {
...
-    );
-
-    if (flip && useAtlas && _flippedAtlasStatus.isNone) {
-      _makeFlippedAtlas();
-    }
-
-    _batchItems.add(batchItem);
-    _sources.add(
-      flip
+      source: flip
```
ufrshubham: Hey @gnarhard, I took a quick look at the failing tests and I suspect it has something to do with the `flip` and `source` rects. Before your change, `addTransform` was always creating `BatchItem` with the original `source`. `flip` was considered only while adding to `_sources` list.

I modified the `addTransform` method to create `BatchItem` with default `source` and some of the tests started passing.

```dart
int addTransform({
  required Rect source,
  RSTransform? transform,
  bool flip = false,
  Color? color,
}) {
  final index = _allocateIndex();
  final batchItem = BatchItem(
    source: source,
    transform: transform ??= defaultTransform ?? RSTransform(1, 0, 0, 0),
    flip: flip,
    color: color ?? defaultColor,
  );

  if (flip && useAtlas && _flippedAtlasStatus.isNone) {
    _makeFlippedAtlas();
  }

  _batchItems[index] = batchItem;

  return index;
}
```

I didn't get enough time to look further into this, but I hope this helps you to narrow down the root cause 🙂

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -419,27 +432,38 @@ class SpriteBatch {
...
 
     if (useAtlas && !_flippedAtlasStatus.isGenerating) {
+      final transforms = _batchItems.values
+          .map((e) => e.transform)
+          .toList(growable: false);
+      final sources = _batchItems.values
+          .map((e) => e.source)
+          .toList(growable: false);
+      final colors = _batchItems.values
+          .map((e) => e.color)
```
ufrshubham: Along with my previous [comment](https://github.com/flame-engine/flame/pull/3650/files#r2417214005), if you apply the flip transformation on the sources here, it should work. But I don't think it is a good idea to do this in the render method.

```suggestion
          .map((e) => e.flip
              ? Rect.fromLTWH(
                  // The atlas is twice as wide when the flipped atlas is generated.
                  (atlas.width * (_flippedAtlasStatus.isGenerated ? 1 : 2)) -
                      e.source.right,
                  e.source.top,
                  e.source.width,
                  e.source.height,
                )
              : e.source)
```

packages/flame_tiled/test/tiled_test.dart
```diff
@@ -338,49 +337,6 @@ void main() {
       }
       expect(allGreen, true);
     });
-
-    test('[useAtlas = false] Green tile pixels are in correct spots', () {
```
spydon, line 342: How come these tests are removed?

gnarhard: I thought it didn't do anything, but I guess I just don't understand it. I now recall it providing different test results than the one before so I restored it in the latest commit.

spydon (approved):
Lgtm!

PR #3653, opened 2025-07-23
`feat:` [flame_tiled] Layers respect draw-order of added Components

### LINKED ISSUE #3594: Design Limitation /Feature Request : Insert Object Between Tiled Layers using (Leap + Flame)
### What could be improved

i have map animated.tmx in my asset that i created with Tiled software, 

🧭 Context and Goal
I'm building a game using Flame (v1.11.0) with the Leap platformer library, leveraging .tmx maps and object layers for level content.

My goal is to insert animated objects (custom Flame components) — such as cow_zone areas — between specific Tiled map layers during gameplay. These animated tiles are parsed from ObjectGroup layers and drawn using custom PositionComponents (e.g., UniversalAnimationLayer), each with an assigned priority.

Flutter: 3.29.3
Dart : 3.7.2
Flame: 1.19.0

Below is GIF From TIled(that view want)
https://github.com/user-attachments/assets/5558bd9e-d1ce-410e-b2f7-f5ad3daa1e1e

this is the problems (that i got)
Image

  Leap: https://github.com/VeryGoodOpenSource/leap.git
  flame_tiled: https://github.com/flame-engine/flame.git
  flame_behaviors: ^1.0.0

tiled map details:-----
tiled Software V:--1.10.2

Tiled Map Properties:--

tileset atlas :-- 4048*4048

344*40 Map size Tiled
22016*2560 Dimension Figma Size
Orthogonal
Right Down

64*64 Tile && Grid Size

Flutter Anchor camera.viewfinder.zoom = 0.31;

Player:--128*128
Iteam:--75*75
enemy:--128*128

Layer Format:--

respawn   -Object
spawn   -Object
enemies   -Object
items   -Object
windMill_zone   -Object
fishhouse_zone   -Object
treehouse_back   -Layer
Ground   -Layer
water/spikes   -Layer
treehouse_front   -Layer
house_zone  -Object
cow zone  -Object
wateffall_zone  -Object
tree_zone  -Object
fabulous_zone  -Object
decorations_terrain   -Layer
decorations   -Layer
coconut zone   -Object

❗️The Core Problem
Leap loads and draws all Tiled map layers as a single internal structure, making it impossible to inject custom Components (like animations) between Tiled layers.

Even though I:

Gave negative or mid-range priority values (priority: -1000, priority: 250, etc.)

Used UniversalAnimationLayer as a PositionComponent

Parsed ObjectGroups from .tmx files and spawned Flame animation components correctly

…the result is:

Custom objects always render either completely behind or completely in front of all tile layers.

There’s no access to control per-layer draw order or inject content between tile layers.

⚠️⚠️Recap:-
 Object render between 2 layer (priority not work) @erickzanardo @spydon @kurtome 

To add animated object between layer

### Risks

_No response_

### More information

_No response_

### Other

TheMaverickProgrammer:
## Description
This PR allows a user of Flame to add components to a Tiled Component's `RenderableLayer`. Since these layers now draw their child component's after they draw their tile map data to the canvas, the outcome is that paintable components (such as `SpriteComponent`) are drawn in the order of the layers which enables foregrounds and other detail layers to visually behave as expected.
 
✏️ Note that this is not a breaking change!

### PR Demonstration
Video of the `flame_tiled` example application with changes.
https://github.com/user-attachments/assets/7a1b3bc7-7365-4a2d-81cc-7f23b6e12fc4

Video of components drawing as expected in a custom Flame game.
https://github.com/user-attachments/assets/bfa09575-be7b-4ae3-99af-96336ad44f67

Here's an example of adding the player to a layer beneath the "foreground" layer.
Note that the water layer is also on top of the foreground layer and both layers are on top of the player as expected.
image

## Checklist

❓ Does there need to be a test for this and how would that be constructed?

## Breaking Change?

## Related Issues

[Closes #3594](https://github.com/flame-engine/flame/issues/3594)


spydon:
@TheMaverickProgrammer This PR needs a little conflict resolving with main, but it should be quite trivial, it's mostly formatting.
Ping us when this is ready for review, we're planning to make a release in a few days if you think it'll be ready until then. :)

TheMaverickProgrammer:
> @TheMaverickProgrammer This PR needs a little conflict resolving with main, but it should be quite trivial, it's mostly formatting. Ping us when this is ready for review, we're planning to make a release in a few days if you think it'll be ready until then. :)

Will do, and I'll be done with this soon.

TheMaverickProgrammer:
@spydon There's only two tests failing by a margin: the orthographic map with parallax effect and the shifted isometric map test.

The ortho map has a questionable golden test. If I open up the map in Tiled and center the viewport on `(0,0)` it looks different from both of our results. I compared my changes on another Flame project and it looks stable, so I think we may just need to upgrade the golden test.

The shifted isomorphic map has me reviewing the code many times over. I'm unsure what I missed. It's just the group is shifting too much. So some extra offset is happening somewhere. What's odd is that all other tests pass without issue. Maybe if you do a quick review you'll catch it before I do. I'll try to be done with this tomorrow.

spydon:
> The ortho map has a questionable golden test

Can you see what is wrong in the test? Golden tests on Flutter differ slightly on different OS:es, but I guess the diff is bigger than just a pixel off?

I'll see if I have time to review it today, and I'll add some other reviewers that are more familiar with this code.

spydon:
@kurtome @ufrshubham @jtmcdole could you maybe review this PR if you have time? Since you are much more familiar with the flame_tiled code than I am. 😇

packages/flame_tiled/lib/src/renderable_layers/tile_layers/unsupported_layer.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:flame_tiled/flame_tiled.dart';
+import 'package:meta/meta.dart';
+
+// Represents a [RenderableLayer] that cannot be parsed by this package.
+@internal
+class UnsupportedLayer extends RenderableLayer {
```
kurtome, line 12: Can you add a comment explaining why we want to include these in the component tree? I assume it's for the position and offset etc

kurtome (commented):
If I remember correctly the original implementation combined the layers into a single image, which was a performance optimization. So, I assume this CL gets rid of that optimization? 

Personally I'm fine with that, I always try to keep my Tiled maps small to avoid performance issues.

The code changes look reasonable to me on a quick review. I could try to test this on my game later this week if that's helpful.

TheMaverickProgrammer:
> > The ortho map has a questionable golden test
> 
> Can you see what is wrong in the test? Golden tests on Flutter differ slightly on different OS:es, but I guess the diff is bigger than just a pixel off?
> 
> I'll see if I have time to review it today, and I'll add some other reviewers that are more familiar with this code.

Yes for the `test_shifted.tmx` particularly, the group offset seems to be skipped which is odd considering the components add their offset from their parent's. All others are a tiny percentile off due to pixel artifacts which can be patched by upgrading the golden test since layers now render differently.

packages/flame_tiled/lib/src/renderable_layers/tile_layers/unsupported_layer.dart
TheMaverickProgrammer: This is correct - will do! 🫡

TheMaverickProgrammer: Done.

TheMaverickProgrammer:
I found and patched the issue for the `test_shifted.tmx`. However it revealed I need to re-crunch the numbers for the parallax behavior optimization. Because I'm accumulating the total offset for the optimization, I do need to track that, but also each layer is already potentially translating the canvas, we don't want to apply the accumulate offset again during that step. However the canvas translation does require that we take into consideration the parallax offset which does accumulate.

So it's a bit of a mess but solvable now that I see the mistake. My next push ought to be the final.

*EDIT:* The math revealed we can translate the canvas for each layer with local parallax values as long as our local parallax values takes into account their parent's as well. Writing up a fix.

jtmcdole:
> If I remember correctly the original implementation combined the layers into a single image, which was a performance optimization. So, I assume this CL gets rid of that optimization?
> 
> Personally I'm fine with that, I always try to keep my Tiled maps small to avoid performance issues.

I don't think that's correct; TiledAtlas was added to batch rendering, but I'm pretty sure that was done on a per-layer. I'm not sure if I was clipping the batch draw to the camera (which would have been a good idea), and I wanted to add some y-order optimizations so external components could be rendered between the layers.  It sounds like this PR is aiming to fix the later.

> The code changes look reasonable to me on a quick review. I could try to test this on my game later this week if that's helpful.

Its a bit large of a PR for me to review while sipping my morning coffee, for sure.  I don't see any tests showing "Foreground tiles obscure sprites." - or I am I missing some golden images?

TheMaverickProgrammer:
> > If I remember correctly the original implementation combined the layers into a single image, which was a performance optimization. So, I assume this CL gets rid of that optimization?
> > Personally I'm fine with that, I always try to keep my Tiled maps small to avoid performance issues.
> 
> I don't think that's correct; TiledAtlas was added to batch rendering, but I'm pretty sure that was done on a per-layer. I'm not sure if I was clipping the batch draw to the camera (which would have been a good idea), and I wanted to add some y-order optimizations so external components could be rendered between the layers. It sounds like this PR is aiming to fix the later.
> 
> > The code changes look reasonable to me on a quick review. I could try to test this on my game later this week if that's helpful.
> 
> Its a bit large of a PR for me to review while sipping my morning coffee, for sure. I don't see any tests showing "Foreground tiles obscure sprites." - or I am I missing some golden images?

Good point. I can add a test for this feature. Do I also have permission to update the orthogonal golden?

spydon:
> Do I also have permission to update the orthogonal golden?

Go for it. :)
When possible, create new golden tests instead though.

TheMaverickProgrammer:
Moving this back into `DRAFT` status until I find a reasonable solution for parallax tests which expect visible **camera** offset.

PR #3660, opened 2025-07-28
refactor!: Remove shrinkwrap

### LINKED ISSUE #3657: Get rid of LayoutComponent.shrinkWrap
### What could be improved

I think we can actually get rid of shrinkWrap entirely. I've been using RowComponent and ColumnComponent a lot, and I realized that I set this to true more than I set this to false.

What I'm thinking is, by default the behavior should be shrinkwrapping UNLESS an explicit size is set.

To simplify the API

### Risks

The API will change, so this will be a breaking change. But also this is in the experimental side of flame anyway, so people using it should have had their expectations set anyway.

### More information

_No response_

### Other

livtanong:
# Description

Removes the `shrinkWrap` property from `LayoutComponent` and its subclasses, and make the `size` setter nullable. Setting the `size` to null causes the `LayoutComponent` to behave as if it were `shrinkWrap`ed.

## Checklist

## Breaking Change?

In all cases with `shrinkWrap=true`, simply remove it and ensure `size` is either set to null, or not set at all.

## Related Issues

Closes #3657

### DIFF
--- a/examples/lib/stories/bridge_libraries/flame_jenny/jenny_command_lifecycle_example.dart
+++ b/examples/lib/stories/bridge_libraries/flame_jenny/jenny_command_lifecycle_example.dart
@@ -38,7 +38,6 @@ commands work.
     add(dialogueControllerComponent);
     add(
       ColumnComponent(
-        shrinkWrap: true,
         children: [
           onCommandLabel,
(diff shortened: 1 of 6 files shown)


spydon (approved):
Great, thanks for your contribution!

PR #3662, opened 2025-07-30
feat!: Implements ExpandedComponent

### LINKED ISSUE #3656: ExpandedComponent
### Problem to solve

RowLayout and ColumnLayout have no way to selectively make some of its children expand to fill up remaining space.

### Proposal

1. Under `_layoutMainAxisImpl`, look for direct children of type `ExpandedComponent`.
2. Solve for available expandable space by summing the mainAxis lengths of non-ExpandedComponent children
3. Divide the expandable space among all `ExpandedComponent`s, and size them to their corresponding spaces.
4. `ExpandedComponent` subclasses `PaddingComponent` (or some common superclass) with `shrinkWrap` set to false, so that setting its size also sets its child size automatically.

### More information

_No response_

### Other

livtanong:
# Description

Implements an `ExpandedComponent`, which takes up as much space as possible along the main axis within a `LinearLayoutComponent`. With more than one `ExpandedComponent`, this expansion is divided equally.

## Checklist

## Breaking Change?

### Migration instructions

currently released version in-between the two following tags:

`PaddingComponent` now inherits from `SingleLayoutComponent`.

## Related Issues

Closes #3656

### DIFF
--- a/packages/flame/lib/src/experimental/linear_layout_component.dart
+++ b/packages/flame/lib/src/experimental/linear_layout_component.dart
@@ -92,6 +92,7 @@ abstract class LinearLayoutComponent extends LayoutComponent {
     Vector2? position,
     Vector2? size,
     Iterable<Component> children = const [],
+    ComponentKey? key,
   }) {
     switch (direction) {
       case Direction.horizontal:
@@ -102,6 +103,7 @@ abstract class LinearLayoutComponent extends LayoutComponent {
           size: size,
           position: position,
           children: children,
+          key: key,
         );
       case Direction.vertical:
         return ColumnComponent(
@@ -111,6 +113,7 @@ abstract class LinearLayoutComponent extends LayoutComponent {
           size: size,
           position: position,
           children: children,
+          key: key,
         );
     }
   }


livtanong:
Closing to clean up commit history

livtanong:
Actually, I'll just reopen this. I think you guys squash the commits anyway before merging

spydon:
> Actually, I'll just reopen this. I think you guys squash the commits anyway before merging

We do indeed!

packages/flame/lib/src/experimental/expanded_component.dart
```diff
@@ -0,0 +1,69 @@
...
+    if (inflateChild && child != null) {
+      // We want to set the child's size.
+      // BUT it'll trigger the child size listener and trigger [layoutChildren]
+      // which will trigger [parent.layoutChildren()], which will set [size],
+      // resulting in an infinite loop.
+      // So, we have to first remove the listener, then reattach it afterwards.
+      child.size.removeListener(layoutChildren);
+      child.size = size;
+      child.size.addListener(layoutChildren);
+    }
```
spydon: I think this needs to be done in a listener on `this.size` instead since the size doesn't need to be changed through the setter, but can for example be changed by doing `size.setValues(x, y);`, which is what effects etc are doing.

livtanong: Fixed this along with a bunch of other issues I found :D

livtanong:
Discovered a infinite loop new issue involving nested LinearLayoutComponents (a row inside a column, and the column using CrossAxisAlignment.stretch)

These listener interactions are messing with my head T_T

livtanong:
@spydon i think I may have to do another refactor:

One of the bigger challenges I have right now is the fact that a LayoutComponent may want to shrink wrap in one axis, while being explicit in another.

For example:
```
ColumnComponent(
  key: ComponentKey.named("alice"),
  size: Vector2(600, 400),
  crossAxisAlignment: CrossAxisAlignment.stretch,
  children: [
    ExpandedComponent(
      key: ComponentKey.named("bob"),
      // child omitted for brevity
    ),
    RowComponent(
      key: ComponentKey.named("charlie"),
      mainAxisAlignment: MainAxisAlignment.end,
      children: [
        TextComponent(text: "text")
      ],
    ),
  ],
);
```

- `alice` has crossAxisAlignment = stretch
- `charlie` is a child of alice and therefore its width is set to 600 (explicit setting)
- `charlie` was originally intended to shrinkwrap down to its child's size (say, 24).

The intent is for `charlie` to shrinkwrap along the y axis, but have explicit sizing along the x axis.

Because a lot of the logic around adding and removing listeners centers around the value of `shrinkWrapMode`, this particular case is troublesome.

spydon:
Sounds good @livtanong 
Since these components luckily are marked as experimental, don't be afraid to do breaking changes if it helps to clean them up.

livtanong:
@spydon this is a significant departure, and I'm almost sure you won't be very happy with this.

I introduce `layoutWidth` and `layoutHeight`. These are both nullable doubles, and this is the key to solving the layout infinite loop problem. This allows us to both set height/width and track shrinkwrapping behavior in an axis-independent way.

Because these are scalars, and aren't notifiers, a lot of logic is moved to the setters of these values.
We still have listeners on size in general, but these layout components no longer listen to their own size.

We make a distinction between the size that a layout component *is*, and the size that a layout component *intends to be*. The latter is captured by the nullable doubles: layoutWidth and layoutHeight. And a component never needs to listen to the former. It only needs to *set* the former in reaction to the latter.

I have also toyed with the idea of making a `NullableVector2`, where its components can be nullable doubles, but I realize that I should make it work with scalars first. I don't even know if making an entirely new data structure just for this is even sane.

So with all this ugliness out and explained--I think this is the only way we can get to properly implementing `ExpandedComponent`.

spydon:
> I introduce layoutWidth and layoutHeight. These are both nullable doubles, and this is the key to solving the layout infinite loop problem. This allows us to both set height/width and track shrinkwrapping behavior in an axis-independent way.

Hmm, it feels a bit dirty... I'd hope there is another way we can solve this without creating new arguments. Would it help if the `NotifyingVector2` had a way of setting the values without calling the registered listeners?

livtanong:
I think the bigger issue right now is a nullable vector 2. Anticipating your discomfort, I've already started working on a NullableVector2. Then after i'll see if I run into any trouble extending it into a NotifyingNullableVector2 with ChangeNotifier.

livtanong:
@spydon so I modified it so that the instantiation API is cleaner. Rather than a separate `layoutWidth` and `layoutHeight`, it's just `size`, which also uses `Vector2`. Internally, it is converted to `NullableVector2`.

Caveats
- This still doesn't use listeners for reasons I'm detailing below.
- Modifying the 'layout' size after instantiation requires either setting `layoutSize` directly, or using the `setLayoutAxisLength`. BUT, these are part of layouting logic and are not really meant for use by end-users.

> Would it help if the NotifyingVector2 had a way of setting the values without calling the registered listeners?

Yeah this would definitely help now.
I'm running into problems because when I set a size in the mainAxis (invocation A), then set a size in the crossAxis (invocation B), the listener that gets triggered by invocation A is working with incomplete sizing information, and as a listener it gets triggered "sometime later".

If it becomes possible to set `layoutSize` without triggering any of its listeners, and then triggering them at a later date, that would make working with listeners possible... possibly.

packages/flame/lib/src/experimental/notifying_nullable_vector_2.dart
```diff
@@ -0,0 +1,23 @@
...
+  @override
+  void operator []=(int i, double? v) {
+    super[i] = v;
+    notifyListeners();
+  }
+
+  @override
+  void setFrom(NullableVector2? other) {
+    super.setFrom(other);
+    notifyListeners();
```
spydon: There are a few more cases where you can set the values which currently aren't calling the listeners here, you can compare with the original class.

spydon (commented):
So you can add a method to `NotifyingVector2` that is something like `setValuesSilently` (or if you come up with a better name) that isn't calling the listeners, and then make sure that it is possible to notify the listeners from outside of the class.

livtanong:
> So you can add a method to `NotifyingVector2` that is something like `setValuesSilently` (or if you come up with a better name) that isn't calling the listeners, and then make sure that it is possible to notify the listeners from outside of the class.

Perhaps it should be a mixin because it also needs to apply to `NotifyingNullableVector2`. I certainly wouldn't want to make a change to ChangeNotifier because that's upstream in Flutter.

livtanong:
Is it possible to evaluate this feature as it is, without the listeners? Because I'm not actually 100% sure implementing this with listeners will be viable. Like, I *think* it might work, but I don't know for sure. Besides, the reason listeners are appealing is because a lot of things interact with `NotifyingVector2` and its dimensions, but `NullableVector2` is sort of an implementation detail. An end user wouldn't ever have to touch `layoutSize`, except in some edge cases.

In other words, we don't know if switching to listeners will preserve functionality, but we know it works now. So it might be worth evaluating and merging this in, to have a checkpoint. *And then* attempt the listener refactor afterwards.

spydon:
> An end user wouldn't ever have to touch layoutSize, except in some edge cases.

I'm thinking people will want to use this to build responsive layouts, and in those cases the listeners would be quite useful?

But yes, we can evaluate it as is if you don't think it is feasible.

livtanong:
@spydon I gave it my best shot here: https://github.com/livtanong/flame/blob/layout-listener/packages/flame/lib/src/experimental/linear_layout_component.dart#L310

It's still not behaving. T_T Perhaps someone else can figure out how to make this work with listeners, but I think my brain is fried at this point.

As for the case of responsive layouts, the developer can still set `layoutSize`, so there's no real loss of functionality. It may be slightly inefficient because you're passing a new object rather than modifying it, but it's not like this would be set 60 times in a second.

Given the persistent difficulties with listeners and my argument re: responsive layouts, I think this is ready to be evaluated.

examples/lib/stories/experimental/layout_component_example.dart
```diff
@@ -36,8 +48,9 @@ layouts to reflect the chosen values.
           return ButtonComponent(
```
spydon: I think all these buttons should be kept in the settings section now, since the root component is now so large that it is hard to get an overview, and with the current button implementation you also can't see which settings that are active, you can take inspiration from [here](https://github.com/flame-engine/flame/blob/main/examples/lib/stories/rendering/rendering.dart#L56-L67) (it should be fairly easy to change this since most of it is built-in to dashbook).

examples/lib/stories/experimental/layout_component_example.dart
```diff
@@ -100,63 +118,86 @@ layouts to reflect the chosen values.
...
         TextBoxComponent(
           boxConfig: TextBoxConfig(maxWidth: defaultSize.x),
-          text:
-              // ignore: lines_longer_than_80_chars
-              'The root of this example is a shrinkWrapped ColumnComponent. Notice how, when you update the size of the examples, the root component also resizes. This is a demonstration of the fact that shrinkWrapped LayoutComponents listen to changes in the sizes of their children.',
+          text: 'The root of this example is a shrinkWrapped ColumnComponent. '
+              'Notice how, when you update the size of the examples, the root '
+              'component also resizes. This is a demonstration of the fact '
+              'that shrinkWrapped LayoutComponents listen to changes in the '
+              'sizes of their children.',
```
spydon: I updated these so that they follow our linting standards.

examples/lib/stories/experimental/experimental.dart
```diff
@@ -15,7 +15,9 @@ void addExperimentalStories(Dashbook dashbook) {
       )
       .add(
         'Layout Components',
-        (_) => GameWidget(game: LayoutComponentExample()),
+        (_) => const GameWidget.controlled(
```
spydon: I swapped this to controlled so that it's less janky when resizing the window

packages/flame/lib/src/experimental/expanded_component.dart
```diff
@@ -0,0 +1,53 @@
+import 'package:flame/components.dart';
+import 'package:flame/experimental.dart';
+
+/// Works similarly to flutter's Expanded widget.
+/// This component must be a direct child of a [LinearLayoutComponent].
```
spydon, line 5: Add an assertion about this in `onMount`

spydon (commented):
Alright, let's go ahead without listeners!
But then I don't think the size notifying vectors are needed for those components, maybe it is better to just have doubles for the active axis like you had before, because then the vector isn't really giving much? And then the nullable vectors aren't needed either right?

packages/flame/lib/src/experimental/expanded_component.dart
livtanong: Won't an exception caused by the `ParentIsA` mixin happen in roughly the same time as an assertion in `onMount`?

livtanong:
@spydon 
> But then I don't think the size notifying vectors are needed for those components, maybe it is better to just have doubles for the active axis like you had before, because then the vector isn't really giving much? And then the nullable vectors aren't needed either right?

I'd say the NullableVector2 is still important because it sets up a future that *can* handle listeners.

Relatedly, I realize that a middle ground is an option.

`layoutSize` can still return a NotifyingNullableVector2, that one can attach listeners to, but the *internal implementation* can still avoid using listeners. The reason we can do this now and not before, is that `LayoutComponent` owns the private property, `_layoutSize` (in contrast, any subclass of `PositionComponent` cannot access `_size`)

examples/lib/stories/experimental/layout_component_example.dart
livtanong: Oh wow, I didn't even know this existed. I played around with it, and I'm happy to implement it, but I have 2 concerns:

1. It seems to simply not work with `GameWidget.controlled`. Perhaps I'm using it wrong? But so far I do not see any invocation of `GameWidget.controlled` that also uses dashbook's property controllers.
2. It seems to follow Flutter's paradigm, such that when you change a property, it reconstructs the widget. This is well and good, but with this model, we lose out on the ability to see what happens when you change the property of a component directly, vs constructing it anew. And I can't see an escape hatch in dashbook that lets me use a property in the "flamey" way.

livtanong: Also, do you have any advice for the case of nullable properties like size? I looked into the source of dashbook, and if the value of a property is found to be null, it uses the default value.

Right now, the best thing I can think of is make an enum for this to act as a mediating data type.

packages/flame/lib/src/experimental/expanded_component.dart
spydon: True, missed that mixin!

examples/lib/stories/experimental/layout_component_example.dart
spydon: @erickzanardo would it be possible to make so that we can change mutable variables directly, and not change the flutter state in dashbook?

erickzanardo: @spydon I think you could create a custom property that would not call the `onPropertyChanged`: https://github.com/bluefireteam/dashbook/blob/2c47e0012980b0db43d1fe55467b238c3997d1a4/lib/src/story.dart#L15

Will be a little bit of a hack though hahaha

It shouldn't be a problem to use the "conventional" properties, the rebuilds should maintain the state of the game widget shouldn't it?

spydon: > It shouldn't be a problem to use the "conventional" properties, the rebuilds should maintain the state of the game widget shouldn't it?

The problem is that when using `controlled` it won't rebuild the `GameWidget`, and when not using `controlled` then the resizing of the window is janky, which is something that one might want to try out on this example.

livtanong: @spydon To move things along, I followed the dashbook idiom for the examples. I can just modify them once the controls are updated to support the flutter layout loop bypass.

spydon:
@livtanong `shrinkWrap` seems to behave a bit strange in the examples?

livtanong:
> shrinkWrap seems to behave a bit strange in the examples?

@spydon in what way? It would help if you gave me a list of "transformations" to execute on the editor panel thingy.

spydon:
> @spydon in what way? It would help if you gave me a list of "transformations" to execute on the editor panel thingy.

I'm on my phone now, but I believe that I just had the default settings and pressed shrink wrap and that the last element in the column was outside of the parent's bounding box.
I can send a video when I get home if you can't reproduce it.

luanpotter:
This is what it looks like for me with and w/o the option:

image

image

Without much context, at first looked weird that didn't go to the end but I see that is because there is a padding component after the blue rect.

Then testing with shrink wrap:

image

image

It appears that if you have shrink wrap (regardless of ExpandedComponent or not), the bounding box of the wrapper is not computed correctly to accomodate the blue rect.

Is that the issue we are talking about @spydon @livtanong ?

examples/lib/stories/experimental/layout_component_example_1.dart
```diff
@@ -0,0 +1,161 @@
...
+
+  @override
+  FutureOr<void> onLoad() {
+    super.onLoad();
+    addAll(createComponentList(expandedMode: expandedMode, padding: padding));
+  }
+
+  PaddingComponent? get paddingComponent {
+    return descendants().whereType<PaddingComponent>().firstOrNull;
+    // return children.query<PaddingComponent>().firstOrNull;
```
luanpotter: nit: commented out code

examples/lib/stories/experimental/layout_component_example_1.dart
```diff
@@ -0,0 +1,161 @@
...
+          child: RectangleComponent(
+            size: Vector2(100, 70),
+            paint: Paint()..color = Colors.amber,
+          ),
+        )
+      else
+        RectangleComponent(
+          size: Vector2(100, 70),
+          paint: Paint()..color = Colors.amber,
+        ),
```
luanpotter, line 141: nit: I'd add a maybeWrap local function to avoid repeating the actual component

examples/lib/stories/experimental/layout_component_example_2.dart
```diff
@@ -0,0 +1,121 @@
...
+    addAll(
+      createComponentList(
+        direction: direction,
+      ),
+    );
+  }
+
+  PaddingComponent? get paddingComponent {
+    return descendants().whereType<PaddingComponent>().firstOrNull;
+    // return children.query<PaddingComponent>().firstOrNull;
```
luanpotter: nit: commented out code

examples/lib/stories/experimental/layout_component_example_size.dart
```diff
@@ -0,0 +1,24 @@
...
+  shrinkWrap(null, null),
+  small(640, 480),
+  large(1080, 720);
+
+  const LayoutComponentExampleSize(
+    this.x,
+    this.y,
+  );
+  final double? x;
+  final double? y;
```
luanpotter, line 13: can this use shining new NullableVector2 ?

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -1,108 +1,102 @@
...
-  }) {
-    // use the size setter rather than invoke [layoutChildren] because the
-    // latter needs the intent to shrinkwrap pre-set by [_shrinkWrapMode].
-    // At the time of construction, [_shrinkWrapMode] only has its default
-    // value.
-    this.size = size;
+  }) : _layoutSize = size == null
+            ? NullableVector2.blank()
+            : NullableVector2.fromVector2(size) {
+    // layoutSize.addListener(resetSize);
```
luanpotter: nit: commented out code

spydon:
@luanpotter yes exactly, thanks for the good explanation.

livtanong:
@luanpotter @spydon thanks for catching this! I apologize for my reticence, I've been really busy lately. I'll fix this as soon as  I can.

livtanong:
@luanpotter I think either you or @spydon fixed the issue with shrinkWrap's behavior on a PaddingComponent by setting `child` on construction.

shrinkWrap is, however, expected to break with ExpandedComponent. A shrink-wrapping parent tries to shrink to size, while expandedComponent tries to grow.

packages/flame/lib/src/experimental/nullable_vector_2.dart
```diff
@@ -0,0 +1,52 @@
+import 'package:flame/components.dart';
+
+class NullableVector2 {
```
erickzanardo: I am a bit conflicted by this. A vector is a pair of two coordinates, if one is null, it is not a vector anymore?

spydon: Do you have a better idea for how to implement this?

livtanong: Or perhaps, an alternate name to describe this data structure?

examples/lib/stories/experimental/layout_component_example_size.dart
livtanong: One might actually say this should be a Vector2. But neither Vector2 nor NullableVector2 can be used here, because neither of them have const constructors. They need to be const, because they are used with the `LayoutComponentExampleSize` enum.

examples/lib/stories/experimental/layout_component_example_1.dart
livtanong: implemented as `componentWrap` with a builder for flexibility

examples/lib/stories/experimental/layout_component_example_1.dart
```diff
@@ -127,36 +127,40 @@ class LayoutDemo1 extends LinearLayoutComponent {
...
           child: RectangleComponent(
             size: Vector2(96, 96),
             paint: Paint()..color = Colors.blue,
           ),
         ),
+      ),
     ];
   }
+
+  static Component componentWrapper({
```
spydon: If possible, it would be good to be very explicit in these examples and not use wrappers, so that they are easier to read for beginners.

livtanong: Ah, this was in response to @luanpotter 's recommendation: https://github.com/flame-engine/flame/pull/3662#pullrequestreview-3118083217

I can roll this back.

examples/lib/stories/experimental/layout_component_example_1.dart
spydon: @luanpotter that should not be done in examples, it's much better to be explicit so that the new users can read it more easily.

examples/lib/stories/experimental/layout_component_example_1.dart
spydon: Haha, sorry for giving you conflicting comments. :sweat_smile:

livtanong: No worries! Done!

livtanong: Hello! I hate to pester you... but any chance I can get a review? :3

spydon: Hey hey! I was away on FlutterCon last week, but now I can have a look :)

devtools_options.yaml
```diff
@@ -0,0 +1,3 @@
+description: This file stores settings for Dart & Flutter DevTools.
```
spydon: I don't think we have to check-in this file.

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -1,108 +1,101 @@
...
-    // we use [super.size] to benefit from the superclass's notifier mechanisms.
-    if (newSize == null) {
-      super.size = inherentSize;
-    } else {
-      super.size = newSize;
+  /// We also cannot use listeners at the moment, because listeners are
+  /// triggered when either height/width are set, when we need things to happen
+  /// *only* when height or width are set. Current functionality results in
+  /// a race condition.
+  void setLayoutAxisLength(int axisIndex, double? value) {
```
spydon: The axis shouldn't be an index, since that will be quite unintuitive for the user. It would be better to use an enum.

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -1,108 +1,101 @@
...
-    shrinkWrapMode = newShrinkWrapMode;
-    // we use [super.size] to benefit from the superclass's notifier mechanisms.
-    if (newSize == null) {
-      super.size = inherentSize;
-    } else {
-      super.size = newSize;
+  /// We also cannot use listeners at the moment, because listeners are
+  /// triggered when either height/width are set, when we need things to happen
+  /// *only* when height or width are set. Current functionality results in
+  /// a race condition.
```
spydon: This is an implementation detail, not a dartdoc documentation for the user.

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -1,108 +1,101 @@
...
+  /// [LayoutComponent] level, to selective set width or height without setting
+  /// both.
+  /// e.g. if [axisIndex] is 1, then that's the y axis.
   ///
-  /// Internally, this [size] setter should only ever be invoked upon
-  /// construction, and inside [layoutChildren] to make it easier to track and
-  /// reason about.
+  /// We cannot, for example, extend the accessor assignment of NullableVector2
+  /// to trigger some extra functionality (i.e. setting height/width) when the
+  /// accessor is set.
```
spydon: This is an implementation detail, not a dartdoc documentation for the user.

packages/flame/lib/src/experimental/layout_component.dart
```diff
@@ -1,108 +1,101 @@
...
-    if (shrinkWrapMode) {
-      size = null;
-    }
-  }
+  /// The method to refresh the layout, triggered by various events.
+  /// (e.g. [onChildrenChanged], size changes on both this component and its
+  /// children)
+  ///
+  /// By default, simply sets [size] to null, which sets [size] to
+  /// [intrinsicSize] under the hood.
```
spydon: It doesn't seem like  it does anything by default?

packages/flame/lib/src/experimental/linear_layout_component.dart
```diff
@@ -77,6 +87,8 @@ abstract class LinearLayoutComponent extends LayoutComponent {
     MainAxisAlignment mainAxisAlignment = MainAxisAlignment.start,
     double gap = 0.0,
     Vector2? position,
+    // double? layoutWidth,
+    // double? layoutHeight,
```
spydon: Remove?

packages/flame/lib/src/experimental/linear_layout_component.dart
```diff
@@ -107,73 +119,145 @@ abstract class LinearLayoutComponent extends LayoutComponent {
...
+  ///
+  /// Otherwise, returns the value of gap based on the expected behavior of
+  /// [mainAxisAlignment].
+  double get gap {
+    if ( // mainAxisAlignment is not an alignment that can override gaps
+    !gapOverridingAlignments.contains(mainAxisAlignment) ||
+        // There are [ExpandedComponent]s among the children
+        children.query<ExpandedComponent>().isNotEmpty) {
+      return _gap;
+    }
```
spydon: Put the comments above the `if` so that the code can be read separately from the code, the formatting is quite weird now.

packages/flame/lib/src/experimental/linear_layout_component.dart
```diff
@@ -107,73 +119,145 @@ abstract class LinearLayoutComponent extends LayoutComponent {
...
 
-  /// Sets the size of this [LinearLayoutComponent], then lays out the children
-  /// along both main and cross axes.
+  /// Lays out the children along both main and cross axes.
   @override
   void layoutChildren() {
-    if (shrinkWrapMode) {
-      size = null;
-    }
+    // resetSize();
```
spydon: ?

packages/flame/lib/src/experimental/linear_layout_component.dart
```diff
@@ -107,73 +119,145 @@ abstract class LinearLayoutComponent extends LayoutComponent {
...
     final mainAxisVectorIndex = direction.mainAxisVectorIndex;
     final availableSpace = size[mainAxisVectorIndex];
-    final unoccupiedSpace = availableSpace - _mainAxisOccupiedSpace;
-    final freeSpace = unoccupiedSpace - _gapSpace;
-    // Execute gap override. Internally, only does so if applicable.
-    _gapOverride(unoccupiedSpace);
+    final nonExpandingOccupiedSpace = positionChildren
+        .where((child) => child is! ExpandedComponent)
+        .map((child) => child.size[mainAxisVectorIndex])
+        .sum;
```
spydon:
```suggestion
    final nonExpandingOccupiedSpace = positionChildren.fold<double>(
      0.0,
      (sum, child) =>
          child is ExpandedComponent
              ? sum
              : sum + child.size[mainAxisVectorIndex],
    );
```
To make sure that the if the compiler doesn't manage to optimize this that it runs in `O(n)`.

packages/flame/lib/src/experimental/linear_layout_component.dart
```diff
@@ -182,54 +266,67 @@ abstract class LinearLayoutComponent extends LayoutComponent {
...
+    final expandedComponents = components.whereType<ExpandedComponent>();
+    if (
+    // Meaningless if this is shrinkWrapped.
+    shrinkWrappedIn(direction.mainAxisVectorIndex) ||
+        // There isn't actual any free space to expand or whatnot, sizing is
+        // standard.
+        freeSpace <= 0 ||
+        // There's actually no reason to perform any main axis sizing.
+        expandedComponents.isEmpty) {
       return;
```
spydon: Same here, put the comment above the `if`.

packages/flame/lib/src/experimental/linear_layout_component.dart
```diff
@@ -277,14 +373,40 @@ abstract class LinearLayoutComponent extends LayoutComponent {
...
+  }) {
+    final crossAxisVectorIndex = direction.crossAxisVectorIndex;
+    for (final component in components) {
+      if (crossAxisAlignment != CrossAxisAlignment.stretch) {
+        continue;
+      }
+      if (component is LayoutComponent) {
+        // Don't set value if the value is already correct.
+        if (component.layoutSize[crossAxisVectorIndex] != crossAxisLength) {
+          // component.layoutSize[crossAxisVectorIndex] = crossAxisLength;
```
spydon: ?

packages/flame/lib/src/experimental/linear_layout_component.dart
```diff
@@ -293,47 +415,59 @@ abstract class LinearLayoutComponent extends LayoutComponent {
...
-  double get _gapSpace {
-    return gap * (positionChildren.length - 1);
+    return positionChildren.map((child) {
+      if (child is ExpandedComponent) {
+        // Because ExpandedComponent size can be their expanded state
+        // and thus the occupied space will be inflated.
+        return child.intrinsicSize[direction.mainAxisVectorIndex];
+      }
+      return child.size[direction.mainAxisVectorIndex];
+    }).sum;
```
spydon: Rewrite to use fold, so that we can ensure `O(n)`

packages/flame/lib/src/experimental/linear_layout_component.dart
```diff
@@ -293,47 +415,59 @@ abstract class LinearLayoutComponent extends LayoutComponent {
...
-    final largestCrossAxisLength = positionChildren
-        .map((component) => component.size[crossAxisVectorIndex])
-        .max;
+    final largestCrossAxisLength = positionChildren.map((component) {
+      if (component is ExpandedComponent) {
+        return component.intrinsicSize[crossAxisVectorIndex];
+      } else {
+        return component.size[crossAxisVectorIndex];
+      }
+    }).max;
```
spydon: Rewrite to use fold, so that we can ensure `O(n)`

packages/flame/lib/src/experimental/linear_layout_component.dart
```diff
@@ -293,47 +415,59 @@ abstract class LinearLayoutComponent extends LayoutComponent {
...
-        positionChildren
-            .map((component) => component.size[mainAxisVectorIndex])
-            .sum;
+        positionChildren.map((component) {
+          if (component is ExpandedComponent) {
+            return component.intrinsicSize[mainAxisVectorIndex];
+          } else {
+            return component.size[mainAxisVectorIndex];
+          }
+        }).sum;
```
spydon: Rewrite to use fold, so that we can ensure `O(n)`

packages/flame/lib/src/experimental/nullable_vector_2.dart
spydon: Yeah, that would probably be a good idea, a name the reflects that it is specifically for the use-case that you're using it for in this PR.

packages/flame/lib/src/experimental/nullable_vector_2.dart
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  /// Set the values by copying them from [other].
+  void setFrom(NullableVector2? other) {
+    x = other?.x;
+    y = other?.y;
+  }
+
+  Vector2 toVector(Vector2 fallback) {
```
spydon: This needs dartdocs to explain how the `fallback` is used.

packages/flame/lib/src/experimental/single_layout_component.dart
```diff
@@ -0,0 +1,35 @@
+import 'package:flame/components.dart';
+import 'package:flame/experimental.dart';
+
+abstract class SingleLayoutComponent extends LayoutComponent {
```
spydon, line 11: dartdocs

spydon (commented):
Getting closer!
Missing docs, but I guess you're waiting with that until code is settled?

devtools_options.yaml
livtanong: For some reason this file keeps appearing. Is it okay if I add this to the project gitignore?

spydon: Sure

packages/flame/lib/src/experimental/layout_component.dart
livtanong: I don't mind implementing this, but this model sort of runs deep: It's eventually used to interact with NullableVector2, which is based off of Vector2. (Especially the bit with index access) Are you sure I should go ahead with the enum?

I'll go ahead implementing this while waiting for your answer. If you are sure about this, then i'll just commit it. Otherwise, I can just easily remove the enum code.

livtanong: Actually nvm. I figured out how to make the enum thing elegant enough that I don't need to change NullableVector2.

packages/flame/lib/src/experimental/layout_component.dart
livtanong: Oops, this is outdated behavior. Thanks for catching!

packages/flame/lib/src/experimental/linear_layout_component.dart
livtanong: Thanks for the commit suggestion!

livtanong: Oops, it was for a different thing! lmao

packages/flame/lib/src/experimental/nullable_vector_2.dart
livtanong: But... NullableVector2 is already my best effort at the name 😅 But if we want to make it specific to the use-case, would `LayoutVector2` work? In my opinion though, it's best to get close to its actual properties rather than what it was made for, because we don't know how useful it would be in the future for something else.

packages/flame/lib/src/experimental/nullable_vector_2.dart
livtanong: It's weird how sometimes github correctly detects that a comment is outdated, but sometimes (as in this case) it doesn't seem to realize that the file has been edited already.

packages/flame/lib/src/experimental/nullable_vector_2.dart
livtanong: How about `NullablePair`? I understand that @erickzanardo 's issue is that the spirit of a Vector is that it has two or more values. A "pair" avoids this baggage. However, with this name, the fields `x` and `y` feel out of place.

spydon: I'd say that a pair also expects two values, I'm fine with the `NullableVector2`.
`PartialVector2` could also be an option I guess.
What do you say @erickzanardo ?

erickzanardo: Hey folks, really sorry for the delay here, the notification got burried :/

I like `NullablePair`, I see your point Lukas that pair would expect two values, but a null value is still a value in this case because pair doesn't have the same mathematical meaning as a vector do?

I also don't mind the pair having a x and y? While x and y usually is used to represent the horizontal and vertical axis , it can also be two arbitrary variables.

If we can't come to agreement to a name, I can live with `NullableVector2` don't want to block this PR because a small thing like this.

wolfenrain: What about ConstrainedAxis? To clearly make it show what it is, what is used for.

wolfenrain: Another option would be to use a sealed class, and have a none, xAxis, yAxis and all implementation 

The sealed super class can provide factory constructors for each to make it available on location

livtanong: @wolfenrain that's an interesting idea, but a sealed class would require type checking every time it's used, and that would cause a significant refactor. If at all possible, I'd like to avoid this additional work.

spydon: I fiddled around a bit with the code to get a better understanding for how the nullable vector was used, and imho we don't need it at all, we can just use nullable doubles. I committed this one from my experiment: https://github.com/flame-engine/flame/pull/3662/commits/619b6f171856bd1b1aa630f3e28423080e1819c3
If you don't agree with it, feel free to revert it and we can continue the discussion. :)

livtanong: @spydon That actually works pretty well! I think let's go with this, and then if we find that there's an issue, we can just work on another PR. This is, after all in the experimental module, so we have a lot of leeway to, well... experiment.

examples/lib/stories/experimental/layout_component_example.dart
livtanong: Oops @spydon I may have spoken too soon. Somehow I'm getting a stack overflow on my existing project. Migrating to the new API you designed doesn't seem to be as straightforward as we thought.

livtanong: Wait this is the wrong thread lmao! sorry!

packages/flame/lib/src/experimental/nullable_vector_2.dart
livtanong: I responded to the wrong thread, but it seems @spydon already rolled back this change. For everyone else's context and convenience:

Oops @spydon I may have spoken too soon. Somehow I'm getting a stack overflow on my existing project. Migrating to the new API you designed doesn't seem to be as straightforward as we thought.

spydon: > I responded to the wrong thread, but it seems @spydon already rolled back this change.

Nope, I haven't rolled it back, it's still there.

> Oops @spydon I may have spoken too soon. Somehow I'm getting a stack overflow on my existing project. Migrating to the new API you designed doesn't seem to be as straightforward as we thought.

Where are you getting stack overflow? And are you sure that didn't happen before with the NullableVector?

livtanong: Oh you just merged main in there. Again, oops! 

I'll only be able to investigate further next week where exactly the stack overflow is happening. And it definitely didn't happen before with NullableVector2. When I switch back to commit 2e77b6a1f611894f68974f26e7004171dfd7a25c, it's working fine. I did this twice to verify.

spydon: Very curious... I can't see how the commit would change anything, must be something subtle.
Let's touch base next week then. :)

livtanong: Hallo! So I just remembered one reason why there might be value in storing layoutSize as a NullableVector2.
https://github.com/flame-engine/flame/pull/3662#issuecomment-3152901357

Because it can be listened to. (They're not listenable *yet*, but they *can* be, whereas two scalars cannot)

spydon: It doesn't seem like it was listened to internally though? And iirc, it didn't call the listeners in all cases.

livtanong: Yeah it didn't. It was more of a potentiality.

The implementation of this feature actually started out as two scalars.
https://github.com/flame-engine/flame/pull/3662#issuecomment-3146226783

And when I thought of merging them into a NullableVector2, we had a discussion about making them listenable for when we want to support responsive layouts
https://github.com/flame-engine/flame/pull/3662#issuecomment-3148429923

spydon: I think we can just add a `ChangeNotifier` to the class when that is needed.
You said that the current implementation doesn't work, did you figure out why?

livtanong: I have yet to figure out why (it's still quite a busy week), but it's reproducible in `layout_component_example_2`, at least. So it's good that it's not just in my project. If you switch over to that dashbook example, the whole thing freezes. If it goes on for long enough, we end up blowing the stack and the dashbook crashes.

livtanong: Figured out the issue! https://github.com/flame-engine/flame/pull/3662/commits/82e77956e1bee3738618df1ef8497bf363b5341f

It's now working perfectly fine for me :D

spydon: That's awesome to hear! :D

packages/flame/lib/src/experimental/linear_layout_component.dart
```diff
@@ -8,27 +9,30 @@ enum Direction {
...
-    return vector[mainAxisVectorIndex];
+  double mainAxisValue(Vector2 vector) {
+    return vector[mainAxis.axisIndex];
   }
 
   /// A helper for returning the cross axis for [vector], given this direction.
-  double crossAxis(Vector2 vector) {
-    return vector[crossAxisVectorIndex];
+  double crossAxisValue(Vector2 vector) {
+    return vector[crossAxis.axisIndex];
```
spydon, line 28: These could be a local extension on Vector2, but we can fix that up some other time.

spydon:
@livtanong I forgot that it doesn't have any docs yet, could you add that in a separate PR? :innocent:

livtanong:
> @livtanong I forgot that it doesn't have any docs yet, could you add that in a separate PR? 😇

Will do!

livtanong:
@spydon should I be writing it in here, and then say it's experimental? https://docs.flame-engine.org/latest/flame/layout/layout.html

spydon:
> @spydon should I be writing it in here, and then say it's experimental? https://docs.flame-engine.org/latest/flame/layout/layout.html

Sure, sounds good!

livtanong:
@spydon so I'm roughly aware that I should be updating this part of the repo: https://github.com/flame-engine/flame/blob/main/doc/index.md

But also there are bits about RowComponent and ColumnComponent and LayoutComponent, but they aren't on the docs website. I checked the docs website repo, and it sort of *should* be there, so I guess I'm trying to understand how this all works. 

The CONTRIBUTING.md guide does mention a how-to for adding to the documentation site, but I can't seem to find the actual instructions. They all seem to be for other things. https://github.com/flame-engine/flame/blob/main/CONTRIBUTING.md#documentation--miscellaneous

spydon:
@livtanong so if you look at this one it should be clearer I guess:
https://github.com/flame-engine/flame/blob/main/doc/flame/layout/layout.md

So in there the `RowComponent`, `ColumnComponent` and `ExpandedComponent` needs to be added too.
And a new file for the `ExpandedComponent` in the layout directory:
https://github.com/flame-engine/flame/tree/main/doc/flame/layout

Then it will look like this:
Menu: https://docs.flame-engine.org/main/flame/layout/layout.html
Components:
https://docs.flame-engine.org/main/flame/layout/align_component.html
https://docs.flame-engine.org/main/flame/layout/row_component.html (not yet added to index)

PR #3671, opened 2025-08-07
feat: Add setup command to flame_3d example

luanpotter:
# Description

This creates a support structure on flame_3d example to allow for different testing scenarios, using commands to easily switch between different setups.

I will followup with other useful commands, such as `spawn` (similar to minecraft) and more. The example can be basically a sandbox to explore several 3d concepts and ideas.

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame_3d/example/lib/scenarios/boxes_setup.dart
+++ b/packages/flame_3d/example/lib/scenarios/boxes_setup.dart
@@ -25,8 +25,6 @@ class BoxesScenario implements GameScenario {
         color: const Color(0xFFFF2255),
       ),
 
-      // Add the player
-
       // Floating crate
       Crate(
         position: Vector3(0, 5, 0),
--- a/packages/flame_3d/example/lib/scenarios/boxes_setup.dart
+++ b/packages/flame_3d/example/lib/scenarios/boxes_setup.dart
@@ -11,10 +11,6 @@ import 'package:flame_3d/components.dart';
 import 'package:flame_3d/resources.dart';
 
 class BoxesScenario implements GameScenario {
-  static final BoxesScenario instance = BoxesScenario._();
-
-  BoxesScenario._();
-
   @override
   void setup(ExampleGame3D game) {
     game.world.addAll([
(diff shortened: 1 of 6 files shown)


packages/flame_3d/example/lib/scenarios/boxes_setup.dart
```diff
@@ -0,0 +1,79 @@
...
+import 'package:example/components/rotating_light.dart';
+import 'package:example/example_game_3d.dart';
+import 'package:example/scenarios/game_setup.dart';
+import 'package:flame/extensions.dart';
+import 'package:flame/palette.dart';
+import 'package:flame_3d/components.dart';
+import 'package:flame_3d/resources.dart';
+
+class BoxesScenario implements GameScenario {
+  static final BoxesScenario instance = BoxesScenario._();
```
spydon: How come it needs to be a singleton?

luanpotter: it doesn't really, but it doesn't hold state, and we will only ever need one instance, so I didn't want to keep create new instances unnecessarily. but happy to follow other patterns if preferred

packages/flame_3d/example/lib/scenarios/boxes_setup.dart
```diff
@@ -0,0 +1,79 @@
...
+              ),
+            ),
+          ),
+        ),
+      );
+    }
+  }
+}
+
+extension on Random {
```
spydon: Not needed anymore

packages/flame_3d/example/lib/scenarios/boxes_setup.dart
spydon: Since it is only instantiated once I would keep it as a normal class

packages/flame_3d/example/lib/scenarios/boxes_setup.dart
luanpotter: will rebase in a bit!

packages/flame_3d/example/lib/scenarios/boxes_setup.dart
luanpotter: it is twice right now, I think it is more of an access pattern than an actual performance optimization, but I can remove, no problem

packages/flame_3d/example/lib/scenarios/boxes_setup.dart
```diff
@@ -0,0 +1,78 @@
...
+      RenderedPointLight(
+        position: Vector3(0, 0.1, 0),
+        color: const Color(0xFFFF00FF),
+      ),
+      RenderedPointLight(
+        position: Vector3(-2, 3, 2),
+        color: const Color(0xFFFF2255),
+      ),
+
+      // Add the player
```
spydon: For later I guess?

spydon (approved):
Lgtm!

packages/flame_3d/example/lib/scenarios/boxes_setup.dart
luanpotter: no that is a mistake, the player is added on the main file and not part of setup, good catch

PR #3680, opened 2025-08-10
refactor: Add collections library to flame_3d

luanpotter:
# Description

Add collections library to flame_3d to use the `sortBy` method instead of using `Comparing` from `ordered_set`.

Note: I noticed that while collections is used in a few other packages, versions are not consistent. If this proposal holds, I can followup with version unification.

## Checklist

## Breaking Change?


luanpotter:
yeah I agree, some of these methods are really basic stuff. I wonder if they would be open to some PRs

PR #3741, opened 2025-10-05
feat!: Support secondary taps (right click) on new callbacks system

luanpotter:
# Description

Support secondary taps (right click) on new callbacks system.

In order to follow through with our [event system migration](https://docs.google.com/document/d/1nBUup9QCPioVwWL1zs79z1hWF882tAYfNywFMdye2Qc), we need to make sure the new system is equipped to support all use cases; also changes the existing TapCallbacks to be primary-only.

I noticed that we don't support "secondary taps" (i.e. right clicks), so I am adding this. I honestly really dislike the fact that this is considered a completely different event from the left click, instead of just a property on the click event. But I kept this structure to replicate what Flutter does, so this is more familiar for users. I think that is worth the slight verbosity of having yet another detector. Also, it plays well this way with Flutter because that underlying events are a bit different (for example, the secondary ones don't support `pointId`).

Note: this is a slight breaking change because the existing detector works for BOTH left and right click, but there is NO WAY of distinguishing them because the `buttons` property is not propagated in the Flutter end (massive oversight I believe - might put a PR later). Since this provides the secondary as a solution, it also removes secondary clicks from triggering the primary. I think this is more versatile than having tap detector=`(primary OR secondary)` and secondary=`(secondary only)`.

I don't think this should affect basically any users because (1) desktop only and (2) this acceptance of right clicks was probably a bug anyway (for example, on the example it would rotate the square and also open the context menu, which is jarring).

However I am happy to add an option or pursue a different approach, I believe this is the best path forward, IMO.

## Checklist

## Breaking Change?


examples/lib/stories/input/secondary_tap_callbacks_example.dart
```diff
@@ -0,0 +1,63 @@
...
+  ''';
+
+  @override
+  Future<void> onLoad() async {
+    world.add(TappableSquare()..anchor = Anchor.center);
+    world.add(TappableSquare()..y = 350);
+  }
+}
+
+class TappableSquare extends PositionComponent
```
spydon, line 23: Why not extend `RectangleComponent`?

spydon (approved):
Lgtm, just one minor comment on the example

examples/lib/stories/input/secondary_tap_callbacks_example.dart
luanpotter: this is in line with how the other tap examples do it; I can followup and update all so they are consistent

luanpotter:
@felangel I was wondering if it was possible to add a link to the github check so we can see more details?

felangel:
> @felangel I was wondering if it was possible to add a link to the github check so we can see more details?

I’m investigating right now

felangel:
Can you try closing and reopening the pr in the meantime?

filiph:
FWIW, I'm also using middle click in my game (once you're on desktop and using keyboard+mouse input, you want to give players as many options as possible). I'm sharing it here just as a piece of context for later. I have my own workaround (I'm getting my clicks from `Listener.onPointerDown`).

The good news that there are no more buttons, basically. It's (from `flutter/lib/src/gestures/events.dart`): `kPrimaryButton`, `kSecondaryButton`, `kTertiaryButton`.

PR #3742, opened 2025-10-05
refactor: Re-organize internal event imports

luanpotter:
# Description

Re-organize internal event files using the `events.dart` export.

This changes the imports on all files to use the exposed `events.dart` export - this will vastly help us to catch issues of forgetting to expose things. If anything is imported from `src/`, we know it is not exposed by Flame.

This should be absolutely no-op whatsoever - just organizational.

## Checklist

## Breaking Change?


luanpotter:
Yeah I thought it was less bad. But I really wanted to change the first folder, and reclaim the name `input` once we delete the old system. Though that will take forever, and be user facing. This folder is purely internal so we can play around with it.

luanpotter:
I just do the import fix for now - we can discuss the directory structre in the doc.

PR #3770, opened 2025-11-15
feat: Add scaling gesture for components

### LINKED ISSUE #3768: Add component level scale gesture detection
### Problem to solve

my use case is to simulate a screen in game where pictures on it can be zoomed in. 
Since touch and drag work at component level it feels natural to make it work for scale gestures too.

### Proposal

No solutions yet will think about it.

### More information

_No response_

### Other

stilnat:
# Description

Currently, scaling is only working on device scale. This PR allows scaling gesture to be applied on components, very much like dragging works with drag callbacks. I added a ScaleDispatcher component which works very similarly to MultiDragDispatcher, basically sending the callbacks of ScaleGestureRecognizer to the components.

There was some tricky parts though. The goal was to preserve multi drag functionality while adding scale functionality but I run into this issue :
When running concurrently ScaleGestureRecognizer and ImmediateMultiDragGestureRecognizer, since drag in a simpler gesture it is usually recognized before scale, and therefore, drag wins the gesture arena and one pointer (finger) gets reserved for dragging. Scale gets recognized a bit later, but with a single pointer, and data such as scale or rotation are meaningless with a single pointer. 

More generally, multi dragging and scaling are two very similar gestures and it will inevitably lead to some ambiguities.

There was multiple solutions to this, all with their own drawbacks : 
1.  Not allowing multi dragging and scaling at the same time, with a mechanic to switch between the two. Simple to implement
but quite limiting.
2. Introducing a delay with DelayedMultiDragGestureRecognizer, the simplest solution, works well since scaling then gets recognized before dragging, removing the ambiguity.  The default delay is unacceptably long though but I did not experiment with shorter delay.
3. Using the data from the dragging pointer, to reconstruct rotation and scaling data, as if the dragging pointer was part of the scaling gesture.

We decided to go with the third solution as it preserves the ability to do multi drag instantaneously while allowing scaling at the same time. One issue inherent to this solution is also linked to the ambiguity between some multi drag gestures and scaling, preventing some multi drag gestures to be performed properly (basically all multi drag gestures similar to scaling).

One possibility would be to use the data from the single pointer scaling and send that to MultiDragDispatcher, basically what solution 3 does but the other way around.  this would allow for drag and scaling to work at the same time.

Another issue exists that might be fixable though : it doesn't behave great when three fingers are involved, basically when I lift one of the two first fingers, the third one is used in the scale and rotation computations leading to unwanted "jumps".
The issue is that pointers are not really tracked, so eventually the third finger position can be used as if it was the second one.

## Future things to do after this PR

- fix the 3 finger issues described above
- add a simple switch mechanism to activate/deactivate scale gesture recognizing at run time to allow for full real multi drag to work OR investigate "use the data from the single pointer scaling" solution.

## Checklist

## Breaking Change?

Closes #3768

### DIFF
--- a/doc/flame/inputs/scale_events.md
+++ b/doc/flame/inputs/scale_events.md
@@ -3,7 +3,6 @@
 **Scale events** occur when the user moves two fingers in a pinch in, or in a pinch out move.
 Only one single scale gesture can occur at the same time.
 
-
 For those components that you want to respond to scale events, add the `ScaleCallbacks` mixin.
 
 - This mixin adds three overridable methods to your component: `onScaleStart`, `onScaleUpdate`,
@@ -175,7 +174,6 @@ class ScaleOnlyRectangle extends RectangleComponent with ScaleCallbacks {
 
 
 ## Scale and drag gestures interactions
-
 
 A multi drag gesture can sometimes look exactly like a scale gesture.
 This is the case for instance, if you try to move two components toward each other at the same time.


spydon:
@stilnat did you see this from DCM?
```
packages/flame/test/events/component_mixins/scale_callbacks_test.dart (2 issues):
WARNING   Avoid top-level declarations in tests. Try making this declaration private.
          at packages/flame/test/events/component_mixins/scale_callbacks_test.dart:380:14
          avoid-top-level-members-in-tests : https://dcm.dev/docs/rules/common/avoid-top-level-members-in-tests

WARNING   Avoid top-level declarations in tests. Try making this declaration private.
          at packages/flame/test/events/component_mixins/scale_callbacks_test.dart:449:11
          avoid-top-level-members-in-tests : https://dcm.dev/docs/rules/common/avoid-top-level-members-in-tests
```

stilnat:
yeah working on fixing the few linting error left

packages/flame/lib/src/events/component_mixins/scale_callbacks.dart
```diff
@@ -0,0 +1,35 @@
+import 'package:flame/components.dart';
+import 'package:flame/events.dart';
+import 'package:flame/src/events/flame_game_mixins/scale_dispatcher.dart';
+import 'package:flutter/foundation.dart';
+
+mixin ScaleCallbacks on Component {
+  bool _isScaled = false;
+
+  /// Returns true while the component is being scaled.
+  bool get isScaled => _isScaled;
```
spydon: We should probably come up with a better name for this, since this sounds like it would just be when the `scale` isn't `Vector2(1.0, 1.0)`.
`isScaling` isn't consistent with `isDragged` from `DragCallbacks` but I can't think of anything better. I'm open to suggestions.

packages/flame/lib/src/events/flame_game_mixins/scale_dispatcher.dart
```diff
@@ -0,0 +1,356 @@
...
+      return 1.0;
+    }
+
+    final currentLineDistance = _currentLine!.pointerStartLocation
+        .toVector2()
+        .distanceTo(_currentLine!.pointerEndLocation.toVector2());
+
+    final firstLineDistance = _lineAtFirstUpdate!.pointerStartLocation
+        .toVector2()
+        .distanceTo(_lineAtFirstUpdate!.pointerEndLocation.toVector2());
```
spydon: This method creates a lot of new vector2 objects, it should be possible to calculate this without creating any new objects.

stilnat: fixed with last commit, simply used offset distance instead of going through vector2

doc/flame/inputs/scale_events.md
```diff
@@ -0,0 +1,186 @@
...
+    debugPrint('Scale ended with velocity ${event.velocity}');
+  }
+}
+
+```
+
+
+## Scale and drag gestures interactions
+
+
```
spydon:
```suggestion
```

doc/flame/inputs/scale_events.md
```diff
@@ -0,0 +1,186 @@
+# Scale Events
+
+**Scale events** occur when the user moves two fingers in a pinch in, or in a pinch out move.
+Only one single scale gesture can occur at the same time.
+
+
```
spydon:
```suggestion
```

examples/lib/stories/input/scale_example.dart
```diff
@@ -0,0 +1,294 @@
...
+
+    camera.viewport.add(debugText);
+  }
+
+  @override
+  void update(double dt) {
+    super.update(dt);
+
+    if (addCameraRotation) {
+      camera.viewfinder.angle += 0.001;
```
spydon: These two should be dependent on `dt`

stilnat: done

PR #3782, opened 2025-11-24
feat!: Update scale event with custom recognizer

### LINKED ISSUE #2635: ScaleDetector doesn't work when a Component with DragCallbacks is added
# Current bug behavior
`ScaleDetector` doesn't work when a Component with `DragCallbacks` is added.

# Expected behavior
Scale gesture (pinch-to-zoom) should work regardless if a Component with `DragCallbacks` is registered. I'd expect to distinguish between two-finger scale gestures from one-finger tap and drag gestures.

In the example below you can drag the green rectangle but you cannot scale the board. When you remove `DragCallbacks` mixin, scale gesture starts working. 

# Steps to reproduce
1. Use `ScaleDetector` with FlutterGame
2. Add a child with `DragCallbacks`
3. Notice `onScaleUpdate` is not being called after scale gesture
4. Remove `DragCallbacks` mixin and notice `onScaleUpdate` is now being called

```
import 'package:flame/events.dart';
import 'package:flame/palette.dart';

class ScaleDragGame extends StatelessWidget {
  const ScaleDragGame({super.key});

  @override
    final game = ScaleDragFlameGame();

    return GameWidget(game: game);
  }
}

/// `onScale` methods are not called when a Component with DragCallbacks is added as a child.
/// Green rectangle can be dragged around but scale gestures don't work. But if I remove DragCallbacks from
/// [DraggableRectangle] scale gestures start to work.
class ScaleDragFlameGame extends FlameGame with ScaleDetector {
  late final CameraComponent cameraComponent;
  late double startZoom;

  ScaleDragFlameGame();

  @override
  Color backgroundColor() => Colors.yellow;

  @override
  Future<void> onLoad() async {
    final world = World();
    cameraComponent = CameraComponent(world: world)
      ..viewfinder.anchor = Anchor.center
      ..viewport.anchor = Anchor.center;
    addAll([world, cameraComponent]);

    world.add(DraggableRectangle(
      size: size / 4,
      position: size / 6,
      paint: BasicPalette.darkGreen.paint(),
    ));
  }

  @override
  void onScaleStart(ScaleStartInfo info) {
    startZoom = cameraComponent.viewfinder.zoom;
  }

  @override
  void onScaleUpdate(ScaleUpdateInfo info) {
    /// not called when Component with DragCallbacks was added

    final currentScale = info.scale.global;
    if (!currentScale.isIdentity()) {
      final newZoom = (startZoom * currentScale.y).clamp(1.0, 6.0);
      cameraComponent.viewfinder.zoom = newZoom;
    } else {
      final delta = info.delta.game;
      cameraComponent.viewfinder.position.translate(-delta.x, -delta.y);
    }
  }
}

/// Remove [DragCallbacks] and see that pinch-to-zoom gestures start working
class DraggableRectangle extends RectangleComponent with DragCallbacks {
  DraggableRectangle({
    super.position,
    super.size,
    super.paint,
  });

  @override
  void onDragUpdate(DragUpdateEvent event) {
    super.onDragUpdate(event);
    position += event.delta;
  }
}
```

# Flutter doctor output

```
[✓] Flutter (Channel stable, 3.10.5, on macOS 13.2.1 22D68 darwin-arm64, locale en-US)
    • Flutter version 3.10.5 on channel stable at /Users/wojciechplesiak/Development/flutter
    • Framework revision 796c8ef792 (7 weeks ago), 2023-06-13 15:51:02 -0700
    • Engine revision 45f6e00911
    • Dart version 3.0.5
    • DevTools version 2.23.1
```

* Flame version: 1.8.1
* Platform affected: android
* Platform version affected: android 13

# More information
I want to have draggable components placed on the zoomable board. Imagine you have a puzzle board and you want to zoom in to be able to drag the smaller pieces with precision. `DragCallbacks` is taking over all touch inputs even when they happen outside of the Component. I see under the hood that whenever a Component with DragCallbacks is added `ImmediateMultiDragGestureRecognizer` is registered as one of the [gestureDetectors](https://github.com/flame-engine/flame/blob/dac2ebbf506ff48ca8f34d872bbc47cba3ad6c7b/packages/flame/lib/src/game/game.dart#L35) ([here](https://github.com/flame-engine/flame/blob/dac2ebbf506ff48ca8f34d872bbc47cba3ad6c7b/packages/flame/lib/src/events/flame_game_mixins/has_draggable_components.dart#L169)).

stilnat:
# Description

Currently flame handle scale input events in a very hacky way, by getting the drag gesture recognizer data and recomputing
the data for the scale gesture. This has multiple issues : 
- Recompute already computed stuff like rotation and scale factor -> inefficient
- Make the code difficult to understand, had to introduce hard dependencies between scale and drag dispatcher
- Limits the allowed multi drag gestures : those looking like scale gestures simply would not register as such but only as scale

This PR aims to fix all those issues by introducing a new gesture recognizer, which is basically just a mix of ScaleGestureRecognizer and immediateMultiDragGestureRecognizer, allowing pointers to be used for both gestures
without competing. I used the existing flutter code to write it.

I modified a bit ScaleCallbacks and DragCallbacks, so they use their original dispatcher if there is only one type
of them (it's a bit more efficient), and so they upgrade to using MultiDragScaleDispatcher if both mixins are mounted.
Transition between the two is  smooth as the old dispatcher wait for all gestures it started to finish before removing itself.

## Checklist

I wonder if gesture_input.md should be updated
## Breaking Change?

### Migration instructions

`MultiDragDispatcher` and `MultiDragDispatcherKey` have been renamed to `MultiDragScaleDispatcher`
and `MultiDragScaleDispatcherKey`. Deprecated aliases are provided so existing code continues to
compile, but usages should be updated:

```dart
// Before
MultiDragDispatcher
MultiDragDispatcherKey

// After
MultiDragScaleDispatcher
MultiDragScaleDispatcherKey
```

The separate `ScaleDispatcher` is no longer registered independently. Both drag and scale events
are now handled by a single `MultiDragScaleDispatcher`, backed by a combined gesture recognizer
instead of deriving scale data from drag input.

Behavior change for components using both `DragCallbacks` and `ScaleCallbacks`: single-finger
gestures now produce only drag events, and two-finger gestures produce both drag and scale events
simultaneously via the new `MultiDragScaleGestureRecognizer`.

## Related Issues

I believe it Closes #2635


doc/flame/inputs/scale_events.md
```diff
@@ -3,6 +3,7 @@
 **Scale events** occur when the user moves two fingers in a pinch in, or in a pinch out move.
 Only one single scale gesture can occur at the same time.
 
+
```
spydon:
```suggestion
```

examples/lib/stories/input/scale_drag_example.dart
```diff
@@ -1,8 +1,8 @@
 import 'dart:math';
-import 'package:flame/components.dart';
-import 'package:flame/events.dart';
-import 'package:flame/game.dart';
-import 'package:flutter/material.dart';
+import 'package:flame/components.dart' hide Matrix4;
+import 'package:flame/events.dart' hide PointerMoveEvent;
+import 'package:flame/game.dart' hide Matrix4;
+import 'package:flutter/material.dart' hide PointerMoveEvent, Matrix4;
```
spydon: Are these `hide`s really necessary?

packages/flame/lib/src/events/flame_game_mixins/multi_drag_dispatcher.dart
```diff
@@ -134,39 +111,59 @@ class MultiDragDispatcher extends Component implements MultiDragListener {
...
+
+  void markForRemoval() {
+    _shouldBeRemoved = true;
+    _tryRemoving();
+  }
+
+  bool _tryRemoving() {
+    // there's no more fingers
+    // that started dragging before _shouldBeRemoved flag was set to true.
+    if (_records.isEmpty && _shouldBeRemoved && isMounted) {
```
spydon: Shouldn't it remove it directly even if there are fingers still left on the screen?

packages/flame/lib/src/events/flame_game_mixins/scale_dispatcher.dart
```diff
@@ -127,175 +94,38 @@ class ScaleDispatcher extends Component implements ScaleListener {
...
 
+  void markForRemoval() {
+    _shouldBeRemoved = true;
+    _tryRemoving();
+  }
+
+  bool _tryRemoving() {
+    // there's no more fingers
+    // that started dragging before _shouldBeRemoved flag was set to true.
+    if (_records.isEmpty && _shouldBeRemoved && isMounted) {
```
spydon: Same here, I don't think `markForRemoval` should exist, we should just call `removeFromParent` and then do clean-up in `onRemove`.

spydon (commented):
Looks good, just a few comments

examples/lib/stories/input/scale_drag_example.dart
stilnat: might be leftover I'll check

packages/flame/lib/src/events/flame_game_mixins/multi_drag_dispatcher.dart
stilnat: so the thought behind this is that removing immediately the recognizer would abruptly end currently performed gestures which sounds like unwanted behavior to me (but let me know if that's acceptable). 

I didn't check if removing immediately would do that though but I wrote a test for that so I can try and run the test if I just call removeFromParent immediately instead.

packages/flame/lib/src/events/flame_game_mixins/scale_dispatcher.dart
stilnat: same comment as above

packages/flame/lib/src/events/flame_game_mixins/multi_drag_dispatcher.dart
spydon: I think it should send out a cancel event for the gesture when this happens.
Imagine a game where you are steering the player using some drag inputs and then you want to add a cut scene or something, so you remove the input capabilities, then if the gestures aren't ended directly the user can still move the player until they release the gesture.

examples/lib/stories/input/scale_drag_example.dart
```diff
@@ -52,7 +52,7 @@ class ScaleExample extends FlameGame {
 
     if (addScaleDragRectangle) {
       interactiveRectangle = InteractiveRectangle(
-        position: Vector2(200, 200),
+        position: Vector2(100, 100),
```
luanpotter: nit(existing): Vector2.all(100),

packages/flame/lib/src/events/flame_game_mixins/multi_drag_dispatcher.dart
```diff
@@ -30,6 +30,8 @@ class MultiDragDispatcher extends Dispatcher<FlameGame>
   /// The record of all components currently being touched.
   final Set<TaggedComponent<DragCallbacks>> _records = {};
 
+  bool _shouldBeRemoved = false;
```
luanpotter: on components this is called isRemoving - should we unify? (though there it is a getter view over a bitmask state private)

also should we hoist this to my new Dispatcher class, so all can share?

packages/flame/lib/src/events/component_mixins/scale_callbacks.dart
```diff
@@ -26,6 +26,32 @@ mixin ScaleCallbacks on Component {
...
+    if (scaleDispatcher == null && multiDragDispatcher == null) {
+      final dispatcher = ScaleDispatcher();
+      game.registerKey(const ScaleDispatcherKey(), dispatcher);
+      game.add(dispatcher);
+    } else if (scaleDispatcher == null && multiDragDispatcher != null) {
+      final dispatcher = MultiDragScaleDispatcher();
+      game.registerKey(const MultiDragScaleDispatcherKey(), dispatcher);
+      game.add(dispatcher);
+      (multiDragDispatcher as MultiDragDispatcher).markForRemoval();
+    }
```
luanpotter: this code is not perfectly parallel with the one on drag_callbacks, which makes me think there are mistakes.

should we have a shared function to handle registering between two callbacks with an upgrade in general?

packages/flame/lib/src/events/tagged_component.dart
```diff
@@ -1,12 +1,11 @@
 import 'package:flame/src/components/core/component.dart';
-import 'package:flame/src/events/flame_game_mixins/multi_drag_dispatcher.dart';
 import 'package:flame/src/events/flame_game_mixins/multi_tap_dispatcher.dart';
 import 'package:meta/meta.dart';
 
 /// [TaggedComponent] is a utility class that represents a pair of a component
 /// and a pointer id.
 ///
-/// This class is used by [MultiTapDispatcher] and [MultiDragDispatcher]
+/// This class is used by [MultiTapDispatcher] and MultiDragScaleDispatcher
```
luanpotter: can we not have a proper comment reference

packages/flame/lib/src/events/flame_game_mixins/scale_drag_dispatcher.dart
```diff
@@ -0,0 +1,329 @@
...
+/// Use [enableDrag] and [enableScale] (called via [addDispatcher]) to control
+/// which event types are forwarded to the underlying
+/// [MultiDragScaleGestureRecognizer].
+class MultiDragScaleDispatcher extends Component
+    implements MultiDragListener, ScaleListener {
+  /// The record of all components currently being touched.
+  final Set<TaggedComponent<DragCallbacks>> _records = {};
+
+  bool _hasDrag = false;
+  bool _hasScale = false;
```
luanpotter: do we need to have copies of these values here instead of just using the recognizer values as src of truth?

packages/flame/lib/src/events/component_mixins/drag_callbacks.dart
```diff
@@ -63,6 +63,10 @@ mixin DragCallbacks on Component {
   @mustCallSuper
   void onMount() {
     super.onMount();
-    MultiDragDispatcher.addDispatcher(this);
+    MultiDragScaleDispatcher.addDispatcher(
+      this,
+      hasDrag: true,
+      hasScale: this is ScaleCallbacks,
```
luanpotter: (for a case with both callbacks on the same component)
if my understanding is correct and how mixins work, will the onMounts be stacked calling super chain so addDispatcher will be called both times, the first adding the dispatcher with one of the bools=true, then the second time the addDispatcher recovers the callback and sets the other bool=false
I understand the current approach where both are handled by this plus the short fallback on Scale, but feels a bit contrived

packages/flame/lib/src/events/flame_game_mixins/scale_drag_dispatcher.dart
```diff
@@ -0,0 +1,329 @@
...
+}
+
+/// Dispatches both drag and scale events to [DragCallbacks] and
+/// [ScaleCallbacks] components. Attached to the [FlameGame] automatically
+/// when either callback type is first mounted.
+///
+/// Use [enableDrag] and [enableScale] (called via [addDispatcher]) to control
+/// which event types are forwarded to the underlying
+/// [MultiDragScaleGestureRecognizer].
+class MultiDragScaleDispatcher extends Component
```
luanpotter: can't this extend Dispatcher?

packages/flame/lib/src/events/multi_drag_scale_recognizer.dart
```diff
@@ -0,0 +1,570 @@
...
+    }
+  }
+
+  double _computeRotationFactor() {
+    var factor = 0.0;
+    if (_scale.initialLine != null && _scale.currentLine != null) {
+      final fx = _scale.initialLine!.pointerStartLocation.dx;
+      final fy = _scale.initialLine!.pointerStartLocation.dy;
+      final sx = _scale.initialLine!.pointerEndLocation.dx;
+      final sy = _scale.initialLine!.pointerEndLocation.dy;
```
luanpotter: nit: I'd save to a variable to avoid the !

packages/flame/lib/src/events/flame_game_mixins/scale_drag_dispatcher.dart
```diff
@@ -0,0 +1,373 @@
+import 'package:flame/components.dart';
+import 'package:flame/events.dart';
+import 'package:flame/src/events/flame_drag_adapter.dart';
+import 'package:flame/src/events/flame_game_mixins/dispatcher.dart';
+import 'package:flame/src/events/interfaces/scale_listener.dart';
+import 'package:flame/src/events/multi_drag_scale_recognizer.dart';
+import 'package:flame/src/events/tagged_component.dart';
+import 'package:flame/src/game/flame_game.dart';
+import 'package:flame/src/game/game_render_box.dart';
+import 'package:flutter/material.dart';
```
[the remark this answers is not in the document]
luanpotter: wow, good catch

packages/flame/lib/src/events/tagged_component.dart
```diff
@@ -1,12 +1,11 @@
 import 'package:flame/src/components/core/component.dart';
-import 'package:flame/src/events/flame_game_mixins/multi_drag_dispatcher.dart';
 import 'package:flame/src/events/flame_game_mixins/multi_tap_dispatcher.dart';
 import 'package:meta/meta.dart';
 
 /// [TaggedComponent] is a utility class that represents a pair of a component
 /// and a pointer id.
 ///
-/// This class is used by [MultiTapDispatcher] and [MultiDragDispatcher]
+/// This class is used by [MultiTapDispatcher] and [MultiDragScaleDispatcher]
```
luanpotter, line 9: you might need one of those docs only imports

luanpotter:
@spydon this is in great shape, I really like this direction! just some final comments and CI above, should be good to go

packages/flame/lib/src/events/deprecated.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:flame/src/events/flame_game_mixins/scale_drag_dispatcher.dart';
+
+@Deprecated('Use MultiDragScaleDispatcher instead.')
+typedef MultiDragDispatcher = MultiDragScaleDispatcher;
+
+@Deprecated('Use MultiDragScaleDispatcherKey instead.')
+typedef MultiDragDispatcherKey = MultiDragScaleDispatcherKey;
```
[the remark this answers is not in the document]
spydon: This is not a goal

packages/flame/lib/src/events/multi_drag_scale_recognizer.dart
```diff
@@ -0,0 +1,563 @@
...
+    if (spanDelta > computeScaleSlop(kind) ||
+        math.max(scaleFactor, 1.0 / scaleFactor) > scaleThreshold) {
+      for (final state in _drag.pointers.values) {
+        if (!state._resolved) {
+          state._arenaEntry?.resolve(GestureDisposition.accepted);
+        }
+      }
+      return true;
+    }
+    return false;
```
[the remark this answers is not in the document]
luanpotter: i'd defer to whatever the flutter native ones do

packages/flame/lib/src/events/multi_drag_scale_recognizer.dart
```diff
@@ -0,0 +1,563 @@
...
+  final _ScaleState _scale = _ScaleState();
+
+  int get pointerCount => _drag.count;
+
+  @override
+  void addAllowedPointer(PointerDownEvent event) {
+    if (!hasDrag && !hasScale) {
+      return;
+    }
+    assert(!_drag.pointers.containsKey(event.pointer));
```
luanpotter: would be good to have a custom message (unless this is trully impossible)

packages/flame/lib/src/events/multi_drag_scale_recognizer.dart
```diff
@@ -0,0 +1,563 @@
...
+      if (_drag != null && !recognizer.hasScale) {
+        _drag!.update(
+          DragUpdateDetails(
+            globalPosition: initialPosition + _pendingDelta,
+            delta: _pendingDelta,
+          ),
+        );
+        _pendingDelta = Offset.zero;
+      }
+    }
```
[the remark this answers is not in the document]
luanpotter: the comment does seem reasonable to me

packages/flame/test/events/component_mixins/scale_drag_callbacks_test.dart
```diff
@@ -0,0 +1,842 @@
...
+      '''adding drag component after scale component 
+    upgrade dispatcher to multiDragMultiDragScaleDispatcher''',
+      (tester) async {
+        final resolution = Vector2(80, 60);
+        final game = FlameGame(
+          camera: CameraComponent.withFixedResolution(
+            width: resolution.x,
+            height: resolution.y,
+          ),
+        );
```
luanpotter: nit: could maybe extract, repeated on all tests and not relevant to the test itself

packages/flame/lib/src/events/multi_drag_scale_recognizer.dart
```diff
@@ -0,0 +1,563 @@
...
+  double currentVerticalSpan = 0.0;
+  Offset localFocalPoint = Offset.zero;
+  _LineBetweenPointers? initialLine;
+  _LineBetweenPointers? currentLine;
+  Matrix4? lastTransform;
+  Offset delta = Offset.zero;
+  VelocityTracker? velocityTracker;
+  Duration? initialEventTimestamp;
+
+  double get scaleFactor => initialSpan > 0.0 ? currentSpan / initialSpan : 1.0;
```
luanpotter, line 428: could these be negative?

packages/flame/lib/src/events/multi_drag_scale_recognizer.dart
```diff
@@ -0,0 +1,563 @@
...
+        _scale.initialVerticalSpan = _scale.currentVerticalSpan;
+      } else if (_drag.count == 2) {
+        // Re-initialize span with actual 2-finger distance so the threshold
+        // check measures movement from this baseline, not from the zero span
+        // of the single-pointer state.
+        _updateScaleFields();
+        _scale.initialSpan = _scale.currentSpan;
+        _scale.initialHorizontalSpan = _scale.currentHorizontalSpan;
+        _scale.initialVerticalSpan = _scale.currentVerticalSpan;
+      }
```
luanpotter, line 88: what if it is more than 2? not possible?

luanpotter: in fact the branches are pretty much identical bar one line, might wanna hoist that and justify

packages/flame/lib/src/events/multi_drag_scale_recognizer.dart
```diff
@@ -0,0 +1,563 @@
...
+    } else if (event is PointerCancelEvent) {
+      assert(event.delta == Offset.zero);
+      state._cancel(event);
+      _removeState(event.pointer);
+      if (hasScale) {
+        _scale.lastTransform = event.transform;
+        _updateScaleFields();
+        _updateLines();
+        _endScaleIfNeeded();
```
luanpotter, line 119: do we need to clear the initials on a cancel?

luanpotter (approved):
LGTM with some nits, awesome!

PR #3785, opened 2025-12-07
feat: Implement padding component inflateChild

### LINKED ISSUE #3784: Add inflateChild property to SingleLayoutComponent
### What could be improved

`inflateChild` would prompt the concrete subclass of `SingleLayoutComponent` to resize its child to a certain size determined by its own requirements. ExpandedComponent grows its child to its own bounds along the main axis of its parent LinearLayoutComponent, while PaddingComponent would grow its child its inset size.

Consider the case where an `ExpandedComponent` wraps some other component X. `ExpandedComponent`, by virtue of its function, will inflate component X to fill up as much of itself as possible along the main axis of its parent `LinearLayoutComponent`. 

There exists the possibility that we want component X to be inset, hence, we reach for the `PaddingComponent`. However, because `PaddingComponent` currently doesn't have its own `inflateChild` property, what happens is `PaddingComponent` is inflated because it is the direct child of `ExpandedComponent` but component X is not, obviating the spirit of `ExpandedComponent`.

With this reasoning, both `PaddingComponent` and `ExpandedComponent` would have `inflateChild`, and since they are the only subclasses of `SingleLayoutComponent`, it makes sense for this superclass to have this common field.

It also makes sense to attach this functionality to SingleLayoutComponent, because the only other subclass of `LayoutComponent` (`LinearLayoutComponent`) can control both sizing and position of its children. it therefore makes sense for all subclasses of `SingleLayoutComponent` to be granted the same abilities. Sizing and positioning are, after all, "layout".

### Risks

Not much. This would be off by default for PaddingComponent, so it wouldn't affect any current users of that component, while ExpandedComponent already uses this field.

### More information

_No response_

### Other

livtanong:
# Description

- Exposes `inflateChild` as a field of `SingleLayoutComponent`
- Implements `inflateChild` for `PaddingComponent`
- Refactors `ExpandedComponent`

## Checklist

## Breaking Change?

## Related Issues

Closes #3784

### DIFF
--- a/packages/flame/lib/src/experimental/padding_component.dart
+++ b/packages/flame/lib/src/experimental/padding_component.dart
@@ -17,7 +17,7 @@ import 'package:flutter/rendering.dart';
 /// cause the layout to refresh.
 ///
 /// If [inflateChild] is true, [resetSize] sets the child's size to fill up
-/// available space via [_syncChildSize]. If the child is a [LayoutComponent]
+/// available space via [syncChildSize]. If the child is a [LayoutComponent]
 /// descendant, then [resetSize] uses the [LayoutComponent.setLayoutSize].
 ///
 /// Example usage:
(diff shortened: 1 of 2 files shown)


luanpotter (approved):
LGTM

PR #3787, opened 2025-12-08
feat: Add TextBoxComponent.resetAnimation

### LINKED ISSUE #3786: Implement "rewind" function for TextBoxComponent
### Problem to solve

Currently, there's no way for a `TextBoxComponent` to reset to its original state. So when employing the typewriter effect, for example, and letting it finish, then changing the text, the typewriter effect doesn't run again.

### Proposal

A rewind/reset method that sets the lifetime back to 0

### More information

Already implemented in https://github.com/Wilhansen/flame/tree/feat-add-rewind-textboxcomponent for our own game.

### Other

Wilhansen:
# Description

Addresses issue #3786 by adding a new function `TextBoxComponent::rewind`.

## Checklist

## Breaking Change?

## Related Issues
Closes #3786 


packages/flame/lib/src/components/text_box_component.dart
```diff
@@ -408,4 +408,15 @@ class TextBoxComponent<T extends TextRenderer> extends TextComponent {
...
     boxConfig = boxConfig.copyWith(timePerChar: 0);
   }
+
+  /// Rewind the typewriter effect to start from the first character again.
+  ///
+  /// Useful for reusing this component when changing lines so that the next
+  /// line does not show immediately.
+  /// Also resets the [onComplete] call state which will be called again
+  /// when the line is finished.
+  void rewind() {
```
spydon: We're using reset as the lingo in the rest of the framework so we should use that here too.

Wilhansen: Also initially called it `reset` initially, but I figured that it can mean many other things like resetting the render cache, width calculation, etc,

 `rewind` imho communicates more clearly what is getting reset.

If you still think I should go for `reset` no prob, lmk.

spydon: `rewind`, to me, sounds like you would run the animation in reverse, so that when text is animating in the `TextBoxComponent` if you call rewind it would remove characters instead of adding them. It could be named something like `resetAnimation` or such.

spydon: @Wilhansen any updates on this? :)

Wilhansen: Figuring out how to write the test for this since the function affects internal state that's not directly probable. (also had other work related things to do)

spydon: `lines` is visible for testing, that should be enough to create s simple test for this

spydon: `lines` is visible for testing, that should be enough to create s simple test for this

spydon: Ah nvm, forgot how `lines` is used. But just expose any getter that you need for testing with the `@visibleForTesting` tag.

Wilhansen: I ended up not using `lines` as it's a value computed only when the actual text to display changes and is unaffected by animation.

Used the `currentChar` and `currentLine` variables instead for the test (and squashed the commits).

Wilhansen:
> Docs for the site is still missing, otherwise it looks good

Added site docs, also included mention of `skip`.

spydon (approved):

PR #3789, opened 2025-12-15
fix: Riverpod 3.0 in flame_riverpod

### LINKED ISSUE #3773: Update flame_riverpod to Support flutter_riverpod 3.0
### Problem to solve

I noticed that the latest version of `flame_riverpod` currently depends on `flutter_riverpod ^2.5.1`, which prevents projects from upgrading to `flutter_riverpod 3.0`.
Our codebase is moving toward Riverpod 3.0 to take advantage of the updated APIs and improvements. However, since `flame_riverpod` hasn’t been updated yet, it blocks the upgrade and creates a dependency conflict.

### Proposal

Investigate the feasibility of upgrading `flame_riverpod` to be compatible with `flutter_riverpod 3.0.3`. 

### More information

Additional Context:

- Current dependency: `flutter_riverpod ^2.5.1`
- This is defined here: https://github.com/flame-engine/flame/blob/main/packages/flame_riverpod/pubspec.yaml

### Other

lunxinfeng:
# Description

fix #3773.

## Checklist

## Breaking Change?

## Related Issues
Closes #3773.

### DIFF
--- a/packages/flame_riverpod/lib/src/widget.dart
+++ b/packages/flame_riverpod/lib/src/widget.dart
@@ -11,7 +11,7 @@ import 'package:riverpod/src/framework.dart';
 /// access to Riverpod [Provider]s.
 ///
 /// The corresponding [State] object ([RiverpodAwareGameWidgetState]) assumes
-/// responsibilities associated with [ConsumerStatefulElement] in
+/// responsibilities associated with ConsumerStatefulElement in
 /// `flutter_riverpod`.
 class RiverpodAwareGameWidget<T extends Game> extends GameWidget<T> {
   RiverpodAwareGameWidget({


packages/flame_devtools/pubspec.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
   flame: ^1.34.0
   flutter:
     sdk: flutter
-  flutter_riverpod: ^2.6.1
+  flutter_riverpod: ^3.0.3
```
spydon, line 17: Have you tried running the devtools with the new version, was there nothing that needed to be updated there?

packages/flame_riverpod/example/pubspec.yaml
```diff
@@ -7,10 +7,11 @@ environment:
   sdk: ">=3.8.0 <4.0.0"
 dependencies:
   flame: ^1.34.0
-  flame_riverpod: ^5.4.21
+  flame_riverpod:
+    path: ../
```
spydon: We never need path dependencies since we are using Melos and pub workspaces.

lunxinfeng: Ok, the path dependencies is only used for test

spydon: You don't need it for testing either, pub workspaces connects it to the local version instead of the released version automatically when you run pub get. :)

packages/flame_devtools/pubspec.yaml
lunxinfeng: I just test flame_riverpod with the example.

packages/flame_riverpod/example/pubspec.yaml
lunxinfeng: OK, has remove the path dependencies, thks!

PR #3796, opened 2026-01-06
fix: Reimplement setLayoutSize to only notify once

livtanong:
# Description

In some cases, like when a PaddingComponent with inflateChild = true wraps a ColumnComponent, we get a stack overflow.
This is because `setLayoutSize` was implemented as `setLayoutAxisLength` on both axes. At first glance, this seems equivalent, but the first invocation would trip all size listeners before the second invocation completes. This meant that the guard statement of `SingleChildLayoutComponent.syncChildSize` would never execute in the case of a `PaddingComponent` with `inflateChild` set to true wrapping the LayoutComponent. Thus, infinite loop.

This PR fixes the implementation of both `setLayoutSize` and `resetSize`.

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/.github/.cspell/dart_dictionary.txt
+++ b/.github/.cspell/dart_dictionary.txt
@@ -1,5 +1,4 @@
 # keywords/terms specific to the Dart/Flutter ecosystem
-canvaskit # CLI option that improves rendering on web https://docs.flutter.dev/platform-integration/web/renderers#command-line-options
 cupertino # Flutter module containing iOS-style widgets
 dartdoc # documentation tool for dart
 dartdocs # plural of dartdoc


packages/flame/test/experimental/linear_layout_component_test.dart
```diff
@@ -483,6 +483,41 @@ void main() {
...
       );
+      testWithFlameGame(
+        'setLayoutSize and setLayoutAxisLength each notify once',
+        (game) async {
+          final layoutComponent = ColumnComponent();
+          // final paddingComponent = PaddingComponent(
+          //   size: Vector2.all(200),
+          //   inflateChild: true,
+          //   child: layoutComponent,
+          // );
```
spydon: Remove?

spydon:
Fixed the failing pipeline step in #3797

packages/flame/test/experimental/linear_layout_component_test.dart
livtanong: Oops, forgot about this. My bad! Fixed.

PR #3816, opened 2026-02-22
feat: Add object pooling support with `ComponentPool`

ufrshubham:
# Description

This PR introduces object pooling functionality to Flame through the new `ComponentPool` class. Object pooling is a performance optimization technique that reduces garbage collection pressure by reusing component instances instead of creating and destroying them repeatedly.

ComponentPool is a generic pool that works with any `Component`. Manages reusable components with configurable sizing, automatic lifecycle handling, and LIFO reuse order. Components are automatically returned to the pool when removed from the game tree via `removeFromParent()`.

## Checklist

## Breaking Change?

## Related Issues

NA

### DIFF
--- a/examples/lib/stories/components/component_pool_example.dart
+++ b/examples/lib/stories/components/component_pool_example.dart
@@ -111,8 +111,7 @@ class _PooledBall extends PositionComponent
       anchor: Anchor.center,
       position: Vector2(0, _maxShadowOffset),
       priority: 0,
-      paint: Paint()
-        ..color = Colors.black.withValues(alpha: 0.6),
+      paint: Paint()..color = Colors.black.withValues(alpha: 0.6),
     );
 
     _ball = CircleComponent(
--- a/packages/flame/test/components/component_pool_test.dart
+++ b/packages/flame/test/components/component_pool_test.dart
@@ -214,8 +214,11 @@ void main() {
           final fresh = pool.acquire();
 
           expect(
-            [component1, component2, component3]
-                .every((c) => !identical(c, fresh)),
+            [
+              component1,
+              component2,
+              component3,
+            ].every((c) => !identical(c, fresh)),
             true,
           );
         },


doc/flame/other/performance.md
```diff
@@ -103,3 +103,153 @@ It is up to the developers to decide which hitboxes can be made passive based on
...
+
+Example:
+
+```dart
+class Bullet extends SpriteComponent with Poolable {
+  Vector2 velocity = Vector2.zero();
+  double damage = 10.0;
+
+  @override
+  void reset() {
```
spydon: Couldn't we just tell the users to do this in `onMount`, then we don't need a new method?

packages/flame/lib/src/components/component_pool.dart
```diff
@@ -0,0 +1,102 @@
...
+/// );
+/// ```
+/// In this example, a pool of `MyBullet` components is created with a maximum
+/// size of 50 and an initial size of 10. You can acquire bullets from the pool
+/// when needed and release them back to the pool when they are no longer in
+/// use.
+///
+/// Note: The pool does not automatically manage the lifecycle of components, so
+/// it is the responsibility of the developer to ensure that components are
+/// properly released back to the pool when they are no longer needed.
```
spydon: Couldn't we just listen to `removed` and add them back to the pool when that happens?

packages/flame/lib/src/components/mixins/poolable.dart
```diff
@@ -0,0 +1,23 @@
...
+/// class MyBullet extends SpriteComponent with Poolable {
+///  @override
+///  void reset() {
+///    // Reset position, velocity, and other properties to initial state
+///  }
+/// }
+/// ```
+/// This mixin is intended to be used with a [ComponentPool] to manage the
+/// lifecycle of poolable components efficiently.
+mixin Poolable on Component {
```
spydon: If we tell the users to use `onMount`, this mixin shouldn't be needed and all components would be able to be poolable

packages/flame/lib/src/components/component_pool.dart
ufrshubham: That is nice idea. I'll give it a try.

ufrshubham:
I've simplified the `ComponentPool` to work with any `Component`. No need for the `Poolable` mixin.

> As a follow-up it would be nice to be able to take a pool as an argument in the `SpawnComponent`

I'll create a follow-up PR to add that support.

PR #3834, opened 2026-03-02
fix: Hitboxes now correctly account for parent scale and rotation

### LINKED ISSUE #1075: `ShapeComponent.containsPoint` does not take `scale` into consideration

spydon:
# Description

globalVertices() previously decomposed the transform hierarchy into absoluteScale * vertex + absoluteTopLeftPosition then rotated by absoluteAngle, which is incorrect when rotation and non-uniform scaling are interleaved at different hierarchy levels. Replace with absolutePositionOf() per vertex which correctly walks the ancestor chain.

Also fix AABB computation for polygon-based hitboxes by adding an overridable computeAabb() method to ShapeHitbox, with PolygonHitbox and RectangleHitbox computing tight AABBs directly from globalVertices().

## Checklist

## Breaking Change?

## Related Issues

Closes #1075 

### DIFF
--- a/doc/tutorials/space_shooter/step_1.md
+++ b/doc/tutorials/space_shooter/step_1.md
@@ -83,6 +83,35 @@ our life way easier if you want to center the component on the screen.
 And that is it for this first part! In this first step, we learned the basics of how to create a
 game class, insert it into the Flutter widget tree, and render a simple component.
 
+
+## Preparing the assets folder
+
+Before we move on, let's prepare our project for the graphics we will be using in the next steps.
+Games need assets such as images, sprites, and animations, and our Space Shooter is no exception.
+
+First, create an `assets/images/` folder at the root of your project. Then tell Flutter about it
+by adding the following lines to your `pubspec.yaml`:
+
+```yaml
+flutter:
+  assets:
+    - assets/images/
+```
+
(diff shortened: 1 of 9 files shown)


ufrshubham:
These changes look good. 

Just had one doubt — introduction of `computeAabb` is not part of the fix, right?

spydon:
> These changes look good.
> 
> Just had one doubt — introduction of `computeAabb` is not part of the fix, right?

Yes it is, because the hitbox implementations need different ways to compute it, so they override it.

ufrshubham (approved):

PR #3841, opened 2026-03-04
fix: Ray direction normalization drift issues

ufrshubham:
# Description

`raytrace()` reuses the same `RaycastResult` objects across bounces, so the reflected direction from one bounce becomes the incident direction for the next. `Vector2.reflect()` is not bit-exact, so each bounce introduces a tiny floating-point rounding error. After enough bounces the direction's length drifts just past the `1e-6` tolerance that `Ray2` enforces, causing a `direction must be normalized` assertion crash.

Two fixes:

- **`CircleHitbox`** and **`PolygonRayIntersection`** (used by `RectangleHitbox` / `PolygonHitbox`): call `normalize()` on the reflected direction before passing it to `Ray2`, so drift is corrected on every bounce.
- **`raytrace_example`**: the initial ray direction was set via a cascade on the `Vector2` getter (`ray.direction..setValues(...)`), which bypasses the `direction=` setter and leaves the cached inverse values stale. Fixed to assign through the setter.

Regression tests are included that inject a direction whose length² is already above the threshold (by writing directly to the underlying `Vector2`, bypassing the setter), then assert that `raycast()` does not throw.

## Checklist

- [NA] I have updated/added relevant examples in `examples` or `docs`.

## Breaking Change?

## Related Issues


examples/lib/stories/collision_detection/raytrace_example.dart
```diff
@@ -142,9 +142,7 @@ bounce on will appear.
     rayPaint.color = _colorTween.transform(0.5 + (sin(_timePassed) / 2))!;
     if (origin != null) {
       _ray.origin.setFrom(origin!);
-      _ray.direction
-        ..setValues(1, 1)
-        ..normalize();
+      _ray.direction = (Vector2(1, 1)..normalize());
```
spydon: Why does this need to create a new vector instead of setting the values of the existing one like we did previously? 🤔

ufrshubham: Creation of new a `Vector2` is not needed here. I'll add a pre-allocated vector in this class and re-use it.

My intention is to invoke the `Ray2.direction` setter instead of directly modifying the direction with something like `Vector2.setValues()` or `Vector2.setFrom()`. Skipping the setter causes skipping `_updateInverses` 

https://github.com/flame-engine/flame/blob/6c6ccf317f3f397edf30a61c90ed00d087a96e9a/packages/flame/lib/src/geometry/ray2.dart#L20-L28

Issue #3854, opened 2026-03-08
Synchronized State Architecture (SSA)


s1r1m1r1:
### Problem to solve

# Synchronized State Architecture (SSA) 
### is a design pattern developed to maximize rendering performance and visual consistency in high-density 2D environments (e.g., Flame Engine).

At its core, SSA decouples the animation/effect logic from the individual component state, moving it into a centralized "State Driver."

## The Core Problem
In traditional architectures, each component manages its own update cycle, animation ticker, and random seeds. When scaling to thousands of objects (e.g., a Bullet Hell game or complex map animations):

###CPU Overhead: 10,000+ independent 

update()
 calls bottleneck the main thread.
GPU Fragmentation: Minor differences in animation phases prevent efficient batching (draw calls) and bitmap caching.
Visual Noise: Subtle desynchronization between similar objects can lead to visual "jitter" or unintended organic patterns.

### Proposal

# The SSA Solution
## SSA solves this by enforcing three main principles:

S — Synchronized: A single global clock or ticker drives all instances of the same type. Every object of a specific group is always in the exact same animation frame or effect phase.
S — State: Components and Decorators become "Pure." They do not store internal timers or progress. Instead, they accept parameters "injected" from an external controller.
A — Architecture: A systemic approach that allows for Frame-Level Caching. Since all objects are synchronized, the engine can rebuild a render batch only once per animation frame, rather than every game tick.
Performance Impact
Real-world benchmarks on Web/Canvas platforms show a massive leap in efficiency:

Standard Approach: web ~3,000 sprites at 32 FPS (bottlenecked by per-component updates).
SSA + SpriteBatch: web 20,000+ sprites at 38-60 FPS (bottlenecked only by raw GPU fill rate).

## Key Use Cases
Massive Battlefields: Synchronized unit animations (marching, idling).
Animated Tilemaps: Global layers for water, lava, or foliage.
Complex Visual Effects: High-density particle-like decorators (glows, outlines, distortions) that share the same pulse or vibration frequency.
Summary: SSA transforms a collection of "independent heavy objects" into a "single high-performance system," enabling professional-grade scalability for Flutter and Flame.

### Other

- [x] Are you interested in working on a PR for this?

PR #3855, opened 2026-03-08
feat: (#3854) add UnifiedAnimationParentComponent , add UnifiedAnimationChildComponent

s1r1m1r1:
# SSA 

## Checklist

## Breaking Change?

## Related Issues

### DIFF
--- a/examples/lib/stories/animations/animations.dart
+++ b/examples/lib/stories/animations/animations.dart
@@ -4,6 +4,7 @@ import 'package:examples/stories/animations/animation_group_example.dart';
 import 'package:examples/stories/animations/aseprite_example.dart';
 import 'package:examples/stories/animations/basic_animation_example.dart';
 import 'package:examples/stories/animations/benchmark_example.dart';
+import 'package:examples/stories/animations/unified_animation_example.dart';
 import 'package:flame/game.dart';
 
 void addAnimationStories(Dashbook dashbook) {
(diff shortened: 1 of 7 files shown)


examples/lib/commons/ssa_sprite_animation_component.dart
```diff
@@ -0,0 +1,56 @@
+import 'dart:ui';
+import 'package:flame/components.dart';
+import 'package:flame/src/sprite_animation_ticker.dart';
+
+/// A version of [SpriteAnimationComponent] that uses an external [animationTicker].
+/// This allows multiple components to share the same ticker, synchronizing
+/// their animations and reducing the number of [update] calls.
+class SynchronizedSpriteAnimationComponent extends SpriteAnimationComponent {
```
spydon: Can you explain what the point of this component is? You can already pass in a ticker to the original component.

s1r1m1r1: The ticker is injected through the constructor. Instead of the component creating its own animation, it is passed a SpriteAnimationTicker that is managed by a parent controller or a global system.

SpriteAnimationTicker that is managed by a parent controller or a global system

Shared State: When render(Canvas canvas) is called, the component asks its _animationTicker for the current sprite to draw. Since the ticker is shared, _animationTicker.getSprite() returns the exact same frame for every component in that frame.

Zero Per-Component Logic: Because playing is false, the expensive update(dt) loop (which calculates time elapsed and frame indexes) is skipped for all 3000 components. Only the single shared ticker is updated once per frame in the parent.

GPU Performance: Because every component is drawing the same region of the same image, the GPU can process these commands much faster (often using a single SpriteBatch call), leading to the jump from ~30 FPS to hundreds or thousands.

s1r1m1r1:
I decided to do some further research.

PR #3869, opened 2026-03-22
fix: Stop stripping trailing digits from sprite names in flame_texturepacker

gnarhard:
# Description

### Bug: `_parseRegion` strips trailing digits from sprite names

The regex `(_?)(\d+)$` in `_parseRegion` stripped trailing digits from all sprite names, assuming they were animation frame indices. This broke any sprite whose name naturally contains trailing digits (e.g. `drohne2` → `drohne`, `cu3` → `cu`, `crystal2` → `crystal`). In 5.0.5, the name was always kept as-is and the index only came from the explicit `index:` field in the atlas.

**Changes:**
- `lib/src/texture_packer_parser.dart`, `_parseRegion`:
  - Removed the regex-based index extraction (the `indexMatch` logic and `nameBeforeIndex`/`extractedIndex` variables)
  - The region index now only comes from the explicit `index:` field in the atlas, defaulting to `-1`
  - The region name is the original name from the atlas (with the file extension stripped), never modified
- Tests updated accordingly: names with trailing digits are preserved as-is, and animations use explicit `index:` fields (the supported mechanism)

Note: this PR originally also fixed a double-prefixing bug in `loadAtlasDataImages`, but that became obsolete when #4015 removed prefixes from the asset loaders on `main`.

## Checklist

## Breaking Change?

## Related Issues


spydon (approved):
Looks sensible to me.
Do you want to have a look @s1r1m1r1, since you made these changes?
So that we don't bounce back and forth,

s1r1m1r1:
I tested this fix,
but when loading from a package, the PNG file for the atlas doesn't load correctly.

Perhaps I should add some fixes for loading images for the atlas from the package.

s1r1m1r1:
If there are too many changes, it's possible to
revert to just one index pattern: image1, image2, image3, image4, etc.
Revert changes only for package integration.

gnarhard:
@s1r1m1r1 Thanks for testing this and flagging the package loading issue.

On the index pattern suggestion though, I think the regex-based index extraction should be fully removed rather than scaled back. TexturePacker already handles frame indexing explicitly via the `index:` field in the atlas file, so the regex was duplicating that behavior and breaking sprite names that naturally end in digits (`drohne2`, `cu3`, `crystal2`, etc.). In 5.0.5 the index only came from the explicit `index:` field, and that worked correctly.

I'd prefer to keep these as two separate concerns — the index stripping fix stands on its own, and the package loading issue can be addressed independently. Happy to collaborate on the package path fix if you want to share what you're seeing there.

spydon:
Just merging in main so that we have the latest pipeline here

PR #3871, opened 2026-03-23
feat: Adding bleed option to SpriteBatch

erickzanardo:
# Description

Adds a `bleed` parameter to `SpriteBatch.add()` and `SpriteBatch.addTransform()` that slightly expands the rendered sprite beyond its source boundaries in all directions. This prevents seam artifacts (visible lines between tiles) that appear when texture filtering samples neighbouring pixels at tile edges in tilemaps.

**Atlas path** (`useAtlas = true`, default): applies a uniform scale to the `RSTransform` via `_computeBleedTransform`, keeping the sprite centre fixed in world space. A uniform scale using `max(bleedScaleX, bleedScaleY)` is used to preserve rotation; for non-square source rects the shorter axis is scaled slightly beyond the requested `bleed` value.

**Non-atlas path** (`useAtlas = false` / web fallback): expands `BatchItem.destination` by `bleed` pixels in every direction, always exactly `bleed` pixels on every side regardless of aspect ratio.

For best results, the atlas should include padding between sprites.

## Checklist

## Breaking Change?

## Related Issues

Addresses seam/ghost-line artifacts commonly reported when using `SpriteBatch` for tilemaps.


packages/flame/lib/src/sprite_batch.dart
```diff
@@ -41,6 +41,7 @@ class BatchItem {
     required this.transform,
     Color? color,
     this.flip = false,
+    this.bleed = 0,
```
wolfenrain, line 44: We should do an assertion to prevent negative values

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -305,6 +312,37 @@ class SpriteBatch {
...
+  /// in all directions while keeping the source sampling region unchanged.
+  /// This helps prevent edge artifacts (seams between tiles in a tilemap).
+  static RSTransform _computeBleedTransform(
+    RSTransform transform,
+    Rect source,
+    double bleed,
+  ) {
+    if (bleed == 0) {
+      return transform;
+    }
```
wolfenrain:
```suggestion
    if (bleed <= 0) {
      return transform;
    }
```

packages/flame/lib/src/sprite_batch.dart
```diff
@@ -305,6 +312,37 @@ class SpriteBatch {
...
+
+    // Scale factors for width and height with bleed
+    final scaleX = (source.width + bleed * 2) / source.width;
+    final scaleY = (source.height + bleed * 2) / source.height;
+
+    // Apply scale to the rotation/scale components
+    final scos = transform.scos * scaleX;
+    final ssin = transform.ssin * scaleY;
```
wolfenrain: We should not scale by x or y but by `math.max` of either so we can preserve rotation (and ensure scale happens uniformly even if width and height are not equal). 

And then the `tx` and `ty` should be multiplied by the scos and ssin (subtract for `tx`, addition for `ty`). That should keep the centerr fixed to bleed across the borders when we are dealing with rotation

spydon:
Any update on this @erickzanardo ?

wolfenrain (approved):

ufrshubham (approved):

PR #3872, opened 2026-03-23
feat(flame_3d)!: Introduce `RenderContext` to make `GraphicsDevice` focused

wolfenrain:
# Description

Introducing a `RenderContext3D` that abstracts away actual rendering from the `GraphicsDevice`. This simplifies both binding, rendering and applying shaders in a way that is more sensible for game developers.

We hide away the complexity of the `GraphicsDevice` while still allowing complexity for those who need it.

## Checklist

## Breaking Change?

## Related Issues


packages/flame_3d/lib/src/graphics/render_context_3d.dart
```diff
@@ -0,0 +1,99 @@
...
+  int _drawCount = 0;
+
+  /// Set camera matrices for this frame.
+  void setCamera(Matrix4 viewMatrix, Matrix4 projectionMatrix) {
+    _viewMatrix.setFrom(viewMatrix);
+    _projectionMatrix.setFrom(projectionMatrix);
+    _cameraPosition.setFrom(
+      Matrix4.inverted(viewMatrix).transform3(Vector3.zero()),
+    );
+  }
```
wolfenrain, line 43: I was thinking of moving this into `CameraComponent3D` but then accessing the `cameraComponent` requires look-ups. That is the current approach that we already have in Flame(2d) so it would align more if we did that, but wanted to know y'all opinion @spydon @luanpotter @erickzanardo @renancaraujo @ufrshubham.

Right now this gets sets when `renderFromCamera` happens, which is in the `World` and if it was in the camera we kinda have to do `world.context.viewMatrix.setFrom` etc in camera which feels low-key disgusting (and would not allow users to build non-component based 3d games)

packages/flame_3d/lib/src/graphics/render_context.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:flame_3d/graphics.dart';
+
+abstract class RenderContext {
+  const RenderContext(this.device);
+
+  final GraphicsDevice device;
```
luanpotter, line 6: this is still tightly coupled with GraphicsDevice - what is the unlock here, are we trying to support new children of RenderContext (but that also bind to GraphicsDevice as per this contract)?
I am not sure what this interface is doing by exposing nothing and just forcing children to be coupled to a GraphicsDevice?

packages/flame_3d/lib/src/graphics/render_context_3d.dart
luanpotter: don't we do a camera "lookup" (except it is tied to a variable in the world) here?
https://github.com/flame-engine/flame/pull/3872/changes#diff-03107aa02261699ef4d9fc57c58b3bb0c6ed63fc0e7ebfb812d3822717d60039R49
so I am not sure how this would become any different

I think forcing users to use the world and camera classes, even though they are components, is perfectly reasonable

packages/flame_3d/lib/src/graphics/render_context.dart
wolfenrain: This simply exists so we can have a `RenderContext2D` in the future that can shim Canvas and allowing us in the longer run to fully switch to `flutter_gpu` for the whole Flame ecosystem. Allowing more complex visualizations with both 2D and 3D games. 

So the render context is a more developer friendly pipeline like the Canvas API while being powered by our GraphicsDevice

packages/flame_3d/lib/src/graphics/render_context_3d.dart
wolfenrain: But by forcing that you could never actually do 3D rendering using Flame abstraction/camera-logic without components, which is something we do support for 2D.

I am not against that perse, a user can of course bring their own camera projection logic and still use the render context for the actual drawing. But having it tight together does make it a lot easier to do so. In the end any camera projection needs 3 element: view, projection and position

packages/flame_3d/lib/src/graphics/render_context.dart
luanpotter: got it, so it is not to decouple from GraphicsDevice but to allow 2D and 3D options. makes sense to me. I wonder if 2D could be implemented as a degenerated case off 3D under the hood (I suspect that is how a lot of engines do it). but I'd love for it to be fully abstracted for the end user - no one should have to care/worry about 3d in a 2d game

packages/flame_3d/lib/src/graphics/render_context_3d.dart
luanpotter: in principle I am totally fine with decoupling, just not fully visualizing rn what exactly it would look like. but general direction makes sense to me as long as it doesn't add much complexity
I am still conflicted on how we changed the 2d camera for example, for the longest time it was significantly harder and more painful to use than the global single camera we had before. realistically most people just want a single camera, which they can change the implementation of by choosing a different child (or making one)
(I also generally overuse the term camera to mean all parts you mentioned (I know on flame 2d we have separate classes))

wolfenrain: Keeping it as is then, this is the best flexibility and allows both component-based and non! And for the component based the users would never know!

packages/flame_3d/lib/src/camera/world_3d.dart
```diff
@@ -14,20 +15,17 @@ import 'package:meta/meta.dart';
 /// The primary feature of this component is that it allows [Component3D]s to
 /// render directly to a [GraphicsDevice] instead of the regular rendering.
 /// {@endtemplate}
-class World3D extends flame.World with flame.HasGameReference {
+class World3D extends flame.World with flame.HasGameReference<FlameGame3D> {
```
luanpotter, line 18: does it make sense for it to declare `T extends FlameGame3D`, `HasGameReference<T>`, in case a user wants to extend World3D and also have a ref to their own game class?

packages/flame_3d/lib/src/graphics/graphics_device.dart
```diff
@@ -1,21 +1,37 @@
+// ignore_for_file: avoid_redundant_argument_values
```
luanpotter, line 1: do we need to disable it? can we just fix

packages/flame_3d/lib/src/graphics/graphics_device.dart
```diff
@@ -1,21 +1,37 @@
...
+  /// No blending, source overwrites destination.
   opaque,
+
+  /// Standard premultiplied-alpha blending:
+  /// `color = src * 1 + dst * (1 - srcAlpha)`
+  alphaBlend,
+
+  /// Additive blending: `color = src * 1 + dst * 1`.
+  additive,
 }
```
luanpotter, line 22: doesn't flutter or dart has these enums somewhere already?

packages/flame_3d/lib/src/model/model_component.dart
```diff
@@ -82,7 +82,7 @@ class ModelComponent extends Object3D {
   }
 
   @override
-  bool shouldCull(CameraComponent3D camera) {
+  bool isVisible(CameraComponent3D camera) {
```
luanpotter, line 85: isn't this the opposite meaning? I guess the name was just wrong before?

packages/flame_3d/lib/src/graphics/render_context_3d.dart
```diff
@@ -0,0 +1,105 @@
...
+  void flush() {
+    for (var i = 1; i < _drawCount; i++) {
+      final entry = _drawPool[i];
+      var j = i - 1;
+      while (j >= 0 && _drawPool[j].distance < entry.distance) {
+        _drawPool[j + 1] = _drawPool[j];
+        j--;
+      }
+      _drawPool[j + 1] = entry;
+    }
```
luanpotter, line 73: isnt there any built-in sort-in-place-by algo on dart?

packages/flame_3d/lib/src/graphics/render_context_3d.dart
```diff
@@ -0,0 +1,105 @@
...
+    final dy = _cameraPosition.y - worldTransform.storage[13];
+    final dz = _cameraPosition.z - worldTransform.storage[14];
+
+    if (_drawCount >= _drawPool.length) {
+      _drawPool.add(_DrawEntry());
+    }
+    _drawPool[_drawCount]
+      ..distance = dx * dx + dy * dy + dz * dz
+      ..object = object;
+    _drawCount++;
```
luanpotter, line 60: I wonder if we could encasulate a generic Pool class that handles minimal allocation while you can just do set/get

luanpotter: also we never de-allocate after increasing the size. a pool class could also handle that through some heuristic

packages/flame_3d/lib/src/graphics/render_context_3d.dart
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /// Submit a deferred draw.
+  void submitDraw(Object3D object, Matrix4 worldTransform) {
+    // Distance for each object in world space. We dont use
+    // Vector3 here as that would be unnecessarily allocating
+    // values.
+    final dx = _cameraPosition.x - worldTransform.storage[12];
+    final dy = _cameraPosition.y - worldTransform.storage[13];
+    final dz = _cameraPosition.z - worldTransform.storage[14];
```
luanpotter: are there no rows/col accessors that return a pre-allocated view over storage? 12-13-14 feel a bit arbitrary but if that is what we need to avoid vector3 allocation then it is what it is

packages/flame_3d/lib/src/camera/world_3d.dart
```diff
@@ -36,44 +34,32 @@ class World3D extends flame.World with flame.HasGameReference {
...
 
-    final image = device.end();
+    final image = game.device.endPass();
     canvas.drawImageRect(
       image,
-      Offset.zero & size,
-      (-viewport.virtualSize / 2).toOffset() &
-          Size(viewport.virtualSize.x, viewport.virtualSize.y),
+      Offset.zero & renderSize,
+      Offset(-size.x / 2, -size.y / 2) & Size(size.x, size.y),
```
luanpotter: this could just be `(-size / 2) & size` right? (assuming size is Vector2)

luanpotter (approved):
LGTM overall. left some considerations!

packages/flame_3d/lib/src/camera/world_3d.dart
wolfenrain: Yes but it would compute another vector so we wouldn't gain much

However we should just do a Rect.fromLTWH here honestly, that's the cheapest

packages/flame_3d/lib/src/graphics/render_context_3d.dart
wolfenrain: you are right, we do have .mNN getters I believe, didn't think about that!

packages/flame_3d/lib/src/model/model_component.dart
wolfenrain: Yep I named it completely wrong 😂 was really confused to see the if logic and be like wait why if isCulled do we draw

packages/flame_3d/lib/src/graphics/graphics_device.dart
wolfenrain: Not that I could find 🤷🏻‍♀️

packages/flame_3d/lib/src/graphics/graphics_device.dart
wolfenrain: No, I decided to disable it be quiet explicit how our render config is setup

If flutter_gpu ever changes the values we wouldn't know on time until rendering goes weird

packages/flame_3d/lib/src/camera/world_3d.dart
wolfenrain: Normally yes but FlameGame has a World generic so it would be self referencing and that might be odd to declare. That is why we didnt do it on the 2D world either.

We should discuss that in general with the team tho!

packages/flame_3d/lib/src/graphics/render_context_3d.dart
wolfenrain: Should be a follow-up, this is also just a temporary solution

PR #3899, opened 2026-04-12
feat: Implement TertiaryTapCallbacks on new event system

luanpotter:
# Description

Following up on [our events master plan](https://docs.google.com/document/d/1nBUup9QCPioVwWL1zs79z1hWF882tAYfNywFMdye2Qc).

This adds `TertiaryTapCallbacks` and necessary infrastructure to make it work.

This replaces the old `TertiaryTapDetector` (which we can deprecate later), e.g.:

```dart
class MyGame extends FlameGame with TertiaryTapDetector {
  @override
  void onTertiaryTapDown(TapDownInfo info) {
    // game-level handling
  }
}
```

With `TertiaryTapCallbacks` following the newer patterns in flame:

```dart
class MiddleClickSquare extends RectangleComponent with TertiaryTapCallbacks {
  @override
  void onTertiaryTapDown(TertiaryTapDownEvent event) {
    // component-level handling, with hit detection
  }

  @override
  void onTertiaryTapUp(TertiaryTapUpEvent event) {
    // tertiary tap finished
  }
}
```

Note that this new detector uses the same dispatcher as secondary, as they share the same underlying flutter recognizer. This is in line with my refactor on https://github.com/flame-engine/flame/pull/3900 - basically we allow for multiple callbacks to register the same detector, handling deduplication and bindings.

Also adds example, tests, and documentation.

## Checklist

## Breaking Change?

### DIFF
--- a/examples/lib/stories/input/tertiary_tap_callbacks_example.dart
+++ b/examples/lib/stories/input/tertiary_tap_callbacks_example.dart
@@ -1,0 +1,70 @@
+import 'package:flame/components.dart';
+import 'package:flame/events.dart';
+import 'package:flame/extensions.dart';
+import 'package:flame/game.dart';
+import 'package:flame/palette.dart';
+import 'package:flutter/material.dart';
+
+class TertiaryTapCallbacksExample extends FlameGame {
+  static const String description = '''
+    In this example we show how to listen to primary (left), secondary (right),
+    and tertiary (middle) tap events at the same time using the `TapCallbacks`,
+    `SecondaryTapCallbacks`, and `TertiaryTapCallbacks` mixins respectively.\n\n
+    The squares will change color depending on which button was used to tap them.
+  ''';
+
+  @override
+  Future<void> onLoad() async {
+    world.add(TappableSquare()..anchor = Anchor.center);
+    world.add(TappableSquare()..y = 350);
+  }
+}
+
+class TappableSquare extends RectangleComponent
+    with TapCallbacks, SecondaryTapCallbacks, TertiaryTapCallbacks {
+  static final Paint _red = BasicPalette.red.paint();
+  static final Paint _blue = BasicPalette.blue.paint();
+  static final Paint _green = BasicPalette.green.paint();
+  static final TextPaint _text = TextPaint(
+    style: TextStyle(color: BasicPalette.white.color, fontSize: 24),
+  );
+
+  int counter = 0;
+
+  TappableSquare({Vector2? position})
+    : super(
+        position: position ?? Vector2.all(100),
+        size: Vector2.all(100),
+        paint: _red,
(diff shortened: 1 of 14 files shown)


packages/flame/lib/src/events/flame_game_mixins/secondary_tap_dispatcher.dart
```diff
@@ -21,29 +21,53 @@ class SecondaryTapDispatcherKey implements ComponentKey {
 /// automatically whenever any [SecondaryTapCallbacks] are mounted into the
 /// component tree.
 class SecondaryTapDispatcher extends Dispatcher<FlameGame> {
```
luanpotter: NOTE: on my consideration around whether to rename this

while it is public, it is not really user facing - I think renaming is totally fine
problem is that it is hard to name because flutter splits primary taps as a completely different recognizer
that means we have (currently) MultiTapDispatcher (for primary)
and this for secondary + tertiary
NonPrimaryTapDispatcher is the best I came up with but isn't perfect

thoughts?

examples/lib/stories/input/tertiary_tap_callbacks_example.dart
```diff
@@ -0,0 +1,70 @@
...
+import 'package:flame/palette.dart';
+import 'package:flutter/material.dart';
+
+class TertiaryTapCallbacksExample extends FlameGame {
+  static const String description = '''
+    In this example we show how to listen to primary (left), secondary (right),
+    and tertiary (middle) tap events at the same time using the `TapCallbacks`,
+    `SecondaryTapCallbacks`, and `TertiaryTapCallbacks` mixins respectively.\n\n
+    The squares will change color depending on which button was used to tap them.
+  ''';
```
spydon, line 14:
```suggestion
  static const String description = '''
In this example we show how to listen to primary (left), secondary (right),
and tertiary (middle) tap events at the same time using the `TapCallbacks`,
`SecondaryTapCallbacks`, and `TertiaryTapCallbacks` mixins respectively.\n\n
The squares will change color depending on which button was used to tap them.
```

packages/flame/lib/src/events/flame_game_mixins/secondary_tap_dispatcher.dart
spydon: `NonPrimaryTapDispatcher` is better than `SecondaryTapDispatcher` at least, so I would go for that.

spydon (approved):
Lgtm, feel free to merge after the rename

examples/lib/stories/input/tertiary_tap_callbacks_example.dart
luanpotter: im very confused by this diff haha

PR #3900, opened 2026-04-13
refactor: Make GestureDetectorBuilder and detector/callback registering more resilient

luanpotter:
# Description

Make GestureDetectorBuilder and detector/callback registering more resilient wrt to multiple registrations (callbacks that use the same detector).

This extracts a base class to dispatchers to share code and remove reundancies.
It also makes the code on *Callbacks themselves trivial, and accounts for duplication.
Before different callbacks could duplicate dispatchers. Now we properly guarantee dispatchers are singleton while allowing callbacks to require and use the same dispatcher. This is important for tertiary callbacks which will use the same underlying dispatcher as secondary.

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame/lib/src/game/flame_game.dart
+++ b/packages/flame/lib/src/game/flame_game.dart
@@ -9,6 +9,7 @@ import 'package:flame/src/devtools/dev_tools_service.dart';
 import 'package:flame/src/effects/provider_interfaces.dart';
 import 'package:flame/src/game/game.dart';
 import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
 // TODO(spydon): Remove this import when flutter version is updated to 3.35.0
 // ignore: unnecessary_import
 import 'package:meta/meta.dart';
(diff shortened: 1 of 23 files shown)


packages/flame/lib/src/game/flame_game.dart
```diff
@@ -66,6 +67,28 @@ class FlameGame<W extends World> extends ComponentTreeRoot
     add(_world);
   }
 
+  /// Registers a dispatcher component for the given [key] if one does not
+  /// already exist.
+  ///
+  /// If no component is registered under [key], creates one via [create],
+  /// registers it under [key], and adds it to the game.
+  void addDispatcher(ComponentKey key, Component Function() create) {
```
spydon: Hmm, it was kind of nice to not have the dispatchers have special treatment, that they were just part of the component tree like any other component. How come we can't do this while mounting the dispatcher instead?

luanpotter: that would be too late, we need to prevent adding it if it exists. (the key registering part could be on mount but then if we have this might as well leave it together for consistency)
this logic was already there, just duplicated on every single dispatcher
I am not 100% happy with it living on the game though, I was considering making a "BaseDispatcher" (though this would have to be a static function, so it is weird)

luanpotter: this PR is a draft btw :P I was still mulling over (I was waiting for the test changes so I could see what was breaking and turns out, nothing)

spydon: >  this logic was already there, just duplicated on every single dispatcher

Yeah, but contained to the dispatchers, it shouldn't leak to the game imho.

>  I am not 100% happy with it living on the game though, I was considering making a "BaseDispatcher" (though this would have to be a static function, so it is weird)

That's a better option I think.

packages/flame/lib/src/game/game_widget/gesture_detector_builder.dart
```diff
@@ -7,42 +7,38 @@ import 'package:flutter/widgets.dart';
...
 
-  void add<T extends GestureRecognizer>(
+  /// Registers a gesture recognizer of type [T].
+  ///
+  /// Throws a [StateError] if [T] is already registered.
+  void register<T extends GestureRecognizer>(
     T Function() constructor,
     void Function(T) initializer,
   ) {
-    final count = _counters[T];
```
luanpotter, line 18: we just have one dispatcher per gesture now, and it can be much better tracked on a single map

spydon (approved):
Good stuff!

PR #3908, opened 2026-05-02
fix: Reset srcPosition to zero on rasterized Sprite

### LINKED ISSUE #3902: Bug: Sprite.rasterize() doesn't reset srcPosition to (0,0), causing drawImageRect assertion failure
## Bug Description

The `Sprite.rasterize()` method creates a new sprite with a rasterized copy of just the sprite region, but it doesn't reset `srcPosition` to `(0, 0)`. This causes the source rectangle to have coordinates that are completely outside the bounds of the new smaller image, leading to a `drawImageRect` assertion failure when the sprite is rendered.

## Reproduction Steps

```dart
// 1. Create a sprite from an atlas at position (8, 7) with tileSize 16
final originalSprite = Sprite(
  atlasImage, // e.g., 336x160 pixels
  srcPosition: Vector2(128, 112), // 8 * 16, 7 * 16
  srcSize: Vector2(16, 16),
);

// 2. Rasterize it - this creates a 16x16 image
final rasterized = await originalSprite.rasterize();

// 3. The rasterized sprite now has:
//    - image: 16x16 (just the sprite region)
//    - src: Rect(128, 112, 144, 128)  <-- PROBLEM!
//    
// The src rect is OUTSIDE the 16x16 image bounds!

// 4. Render it - this will crash:
// Failed assertion: line 7471 pos 12: '<optimized out>': is not true.
// (Canvas.drawImageRect with out-of-bounds src rect)
```

## Root Cause

In `packages/flame/lib/src/sprite.dart` lines 93-96:

```dart
Future<Sprite> rasterize({String? cacheKey, Images? images}) async {
  // ... creates rasterizedImage ...
  
  return Sprite(
    rasterizedImage,
    srcSize: srcSize,  // <-- Only srcSize is passed
    // srcPosition is NOT passed, so it defaults to null
    // and the original srcPosition is retained
  );
}
```

The `Sprite` constructor at lines 23-30 only updates `srcSize`, but `srcPosition` defaults to null, which means the old position is used via the `src` Rect calculation.

## Expected Behavior

After `rasterize()`, the sprite should have `srcPosition` reset to `(0, 0)` since the new image contains only the sprite region starting at the origin.

## Fix

```dart
return Sprite(
  rasterizedImage,
  srcPosition: Vector2.zero(),  // Reset to origin
  srcSize: srcSize,
);
```

## Environment

- Flame version: 1.35.1 (also affects earlier versions)
- Flutter version: 3.x
- Platform: All (macOS, iOS, Android, Web, Windows, Linux)

## Impact

This bug affects:
- `SpriteWidget` with `rasterize: true` parameter
- Any custom code that calls `Sprite.rasterize()` and then renders the result
- Causes hard-to-debug `dart:ui` assertion failures

## Workaround

Until fixed, users can manually create the rasterized sprite:

```dart
final composition = ImageComposition()
  ..add(originalSprite.image, Vector2.zero(), source: originalSprite.src);
final rasterizedImage = await composition.compose();

final fixedSprite = Sprite(
  rasterizedImage,
  srcPosition: Vector2.zero(),  // Explicitly set to origin
  srcSize: originalSprite.srcSize,
);
```

Barba2k2:
## Description

`Sprite.rasterize()` previously forwarded only `srcSize` to the new `Sprite`, leaving the original `srcPosition` in place. When rasterizing a sub-region of an atlas, the resulting `Sprite` is backed by an image already cropped to `srcSize`, but the atlas-relative `srcPosition` was retained — producing an out-of-bounds `src` `Rect` and triggering a `drawImageRect` assertion failure.

This passes `srcPosition: Vector2.zero()` in both the cached and the freshly rasterized branches so the new `src` `Rect` always lies within the rasterized image.

## Checklist

- [x] Bug fix
- [x] Regression test added (covers both cached and uncached branches)
- [x] `flutter analyze` clean on touched files
- [x] All `raster_sprite_test.dart` tests pass

Fixes #3902

### DIFF
--- a/packages/flame/lib/src/sprite.dart
+++ b/packages/flame/lib/src/sprite.dart
@@ -84,6 +84,7 @@ class Sprite {
       final cachedImage = imagesCache.fromCache(key);
       return Sprite(
         cachedImage,
+        srcPosition: Vector2.zero(),
         srcSize: srcSize,
       );
     }
@@ -93,6 +94,7 @@ class Sprite {
 
     return Sprite(
       rasterizedImage,
+      srcPosition: Vector2.zero(),
       srcSize: srcSize,
     );
   }
--- a/packages/flame/test/raster_sprite_test.dart
+++ b/packages/flame/test/raster_sprite_test.dart
@@ -44,6 +44,34 @@ void main() {
       expect(images.keys, hasLength(2));
     });
 
+    test(
+      'resets srcPosition to zero on the rasterized sprite (regression #3902)',
+      () async {
+        final images = Images();
+        final image = await loadImage('flame.png');
+        images.add('flame.png', image);
+
+        final atlasSprite = await Sprite.load(
+          'flame.png',
+          srcPosition: Vector2(10, 10),
+          srcSize: Vector2(20, 20),
+          images: images,
+        );
+
+        final rasterized = await atlasSprite.rasterize(images: images);
+
+        expect(rasterized.srcPosition, Vector2.zero());
+        expect(rasterized.srcSize, Vector2(20, 20));
+        expect(rasterized.src, const Rect.fromLTWH(0, 0, 20, 20));
+
+        // The cached branch must also produce a sprite with srcPosition (0,0)
+        // — the cached image is already cropped to srcSize.
+        final rasterizedAgain = await atlasSprite.rasterize(images: images);
+        expect(rasterizedAgain.srcPosition, Vector2.zero());
+        expect(rasterizedAgain.src, const Rect.fromLTWH(0, 0, 20, 20));
+      },
+    );
+
     group('when using a custom cache key', () {
       test('adds the image to the cache with the custom key', () async {
         final images = Images();


luanpotter (changes requested):
> leaving the original srcPosition in place

I am failing to see where in the code this happens.
When a new sprite is created with a `null` srcPosition, it is set to zero:

```
  set srcPosition(Vector2? position) {
    src = (position ?? Vector2.zero()).toPositionedRect(srcSize);
  }
```

in fact I ran your new test on main w/o the fix and it passes.
am I missing something?

Barba2k2:
You're right — sorry about the noise.

I rebuilt the test on `main` without the patch and it passes. The constructor at `Sprite(image, srcSize: srcSize)` runs `this.srcPosition = srcPosition` with `null`, which goes through the null-coalescing setter you quoted and lands at `Vector2.zero()`. I also wrote a tighter repro that creates a 336×160 atlas, rasterizes the (128, 112)–(16, 16) region and checks `rasterized.src` — it comes out as `Rect(0, 0, 16, 16)` and `render()` does not throw. So the assertion failure described in #3902 isn't reproducible against current `main`.

The "Root Cause" section in the linked issue claims the constructor only updates `srcSize` and leaves `srcPosition` untouched, but that contradicts lines 28-29:
```dart
this.srcSize = srcSize;
this.srcPosition = srcPosition;
```
The patch in this PR is therefore a no-op. Closing this and will leave a comment on #3902 suggesting it be closed unless someone has a reproducer that survives the constructor's null-coalesce.

PR #3909, opened 2026-05-02
fix: Guard against double-dispose of TextBoxComponent cached image

### LINKED ISSUE #3724: TextBoxComponent: A Dart object attempted to access a native peer...
### What happened?

I'm getting these errors through crashlytics, so I don't know the exact extent of these crashes.

```
          Fatal Exception: io.flutter.plugins.firebase.crashlytics.FlutterError: Bad state: A Dart object attempted to access a native peer, but the native peer has been collected (nullptr). This is usually the result of calling methods on a native-backed object when the native resources have already been disposed.
       at Image.dispose(dart:ui)
       at TextBoxComponent.redraw.<fn>(text_box_component.dart:345)

```

### What do you expect?

No error. Disposal can be validated before calling dispose.

I am not entirely sure, because I am not the end user. I do receive a non trivial error reports from crashlytics on this.

"This issue has 86 crash events affecting 39 users" in past 24 hours.

Potentially...(?)
```
if (cachedToRemove.add(cachedImage)) {
   Future.delayed...
}
```

My stacktrace from my reporting tool is too shallow and unfortunately doesn't have the parent caller of the class.

### Relevant log output

```shell
Fatal Exception: io.flutter.plugins.firebase.crashlytics.FlutterError: Bad state: A Dart object attempted to access a native peer, but the native peer has been collected (nullptr). This is usually the result of calling methods on a native-backed object when the native resources have already been disposed.
       at Image.dispose(dart:ui)
       at TextBoxComponent.redraw.<fn>(text_box_component.dart:345)
```

Output of: flutter doctor -v

```
[✓] Flutter (Channel stable, 3.32.7, on macOS 15.1 24B2083 darwin-arm64, locale en-US) [583ms]
    • Flutter version 3.32.7 on channel stable at /Users/macbook/development/flutter
    • Upstream repository git@github.com:flutter/flutter.git
    • Framework revision d7b523b356 (10 weeks ago), 2025-07-15 17:03:46 -0700
    • Engine revision 39d6d6e699
    • Dart version 3.8.1
    • DevTools version 2.45.1

[!] Android toolchain - develop for Android devices (Android SDK version 35.0.0) [2.1s]
    • Android SDK at /Users/macbook/Library/Android/sdk
    • Platform android-35, build-tools 35.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
      This is the JDK bundled with the latest Android Studio installation on this machine.
      To manually set the JDK path, use: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 21.0.3+-79915917-b509.11)
    ! Some Android licenses not accepted. To resolve this, run: flutter doctor --android-licenses

[✓] Xcode - develop for iOS and macOS (Xcode 16.1) [1,334ms]
    • Build 16B40
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [82ms]

[✓] Android Studio (version 2024.2) [82ms]
    • Java version OpenJDK Runtime Environment (build 21.0.3+-79915917-b509.11)

```

### Affected platforms

Android, iOS

### Other information

  flame: 1.32.0

- [ ] I want to work on this

Barba2k2:
# Description

`TextBoxComponent.redraw()` schedules the previous cached `Image` for disposal after a 100 ms delay (to let pending rendering-pipeline frames finish using it). The delayed callback used to call `Image.dispose()` directly, gated only by `isMounted`. Two problems:

1. If the underlying native peer has already been collected (e.g. low-memory scenarios where the engine drops the image), `Image.dispose()` throws and crashes the app — issue #3724 reports this surfacing in production via Crashlytics with 86 events / 39 users in 24 h.
2. The `isMounted` gate **leaked** the cached image whenever the component was removed within the 100 ms window, since `dispose` was simply skipped with no alternative cleanup path.

This PR replaces the gate with a best-effort `_safeDispose` helper that catches `StateError` (release-mode: native peer already collected) and `AssertionError` (debug-mode: already disposed by a parallel path), and uses the same helper from `onRemove` so removal at any point is crash-safe.

A regression test was added in `text_box_component_test.dart` that disposes the cached image manually before removing the component, which without the fix throws and crashes the test runner.

## Checklist

## Breaking Change?

## Related Issues

Fixes #3724


packages/flame/lib/src/components/text_box_component.dart
```diff
@@ -416,15 +416,30 @@ class TextBoxComponent<T extends TextRenderer> extends TextComponent {
...
+  ///
+  /// `Image.dispose` throws a `StateError` if the underlying native peer was
+  /// finalized while the Dart wrapper was still alive (e.g. in low-memory
+  /// scenarios where the engine drops cached image data). Since the image
+  /// resource is already released in that case, the throw provides no value
+  /// to callers and crashes the app — see issue #3724.
+  static void _safeDispose(Image image) {
+    try {
+      image.dispose();
+      // ignore: avoid_catches_without_on_clauses
```
luanpotter: do we know the exact exception that would happen here, and could we catch that specifically?

packages/flame/lib/src/components/text_box_component.dart
```diff
@@ -416,15 +416,30 @@ class TextBoxComponent<T extends TextRenderer> extends TextComponent {
       // See issue #1618 for details.
       Future.delayed(const Duration(milliseconds: 100), () {
```
luanpotter, line 417: this can be for a future thought, but I wonder if we could just register it to be disposed next tick instead of a random future

Barba2k2:
Pushed `f69a420` addressing both review comments:

> do we know the exact exception that would happen here, and could we catch that specifically?

Yes — there are two flavours and I've split the bare catch into explicit `on` clauses for each, with a comment explaining when each path fires:

- **`on StateError`** — the production crash from #3724. When `Image.dispose` accesses the native peer after the engine has dropped its `SkImage` under memory pressure, Flutter throws `"Bad state: A Dart object attempted to access a native peer, but the native peer has been collected (nullptr)"`. This is what Crashlytics in the issue is reporting (86 events / 39 users / 24h).
- **`on AssertionError`** — debug-mode only. `Image.dispose` opens with `assert(!_disposed)`, which trips on a redundant dispose. The regression test runs in debug, so without this clause it would fail; release-mode users never see this path because the assert is stripped.

> this can be for a future thought, but I wonder if we could just register it to be disposed next tick instead of a random future

Agreed that "next tick" is a cleaner mental model than the magic 100 ms — would let the GPU pipeline complete its current frame before we free the image. I left the existing scheduling alone in this PR to keep the change minimal and focused on the crash, but happy to follow up with a separate PR that swaps the `Future.delayed` for a `WidgetsBinding.instance.addPostFrameCallback` (or queues the dispose via the game's own update loop) — let me know if you'd prefer me to fold that in here instead.

spydon:
The PR description needs to follow the PR template.

Barba2k2:
@luanpotter both review threads addressed and resolved — `StateError` / `AssertionError` split landed in f69a420, and the `Future.delayed` → next-tick rework is noted as a follow-up. Also reformatted the description to follow the PR template per @spydon's note across the other PRs.

spydon (approved):

PR #3911, opened 2026-05-02
fix(flame_devtools): Fix component tree growing list and stale snapshot

### LINKED ISSUE #3289: DevTools extension is buggy with constant animations
### What happened?

flame-extension-bug

Notice the constantly growing list that does not take up the entire vertical space of the left hand side and the gray image on the right hand side after selecting a component.

### What do you expect?

I expected the list of components to be usable and the details view to appear as expected.

1. Ensure the DevTools extension assets are built. Run `melos devtools-build`.
2. Run the `examples/` app, and open the Rogue Shooter game.
3. Open DevTools and switch to the Flame tab.

_No response_

_No response_

### Relevant log output

```shell
Exception (minified, so you'll need to repro in debug mode to get a useful stack trace):

![Screenshot 2024-08-27 at 3 10 59 PM](https://github.com/user-attachments/assets/4b23d4dc-0300-4d35-aa03-7953723249e8)
```

Output of: flutter --version

```
Flutter 3.24.1 • channel stable • git@github.com:flutter/flutter.git
Framework • revision 5874a72aa4 (7 days ago) • 2024-08-20 16:46:00 -0500
Engine • revision c9b9d5780d
Tools • Dart 3.5.1 • DevTools 2.37.2
```

### Affected platforms

All

### Other information

This happened from the tip of the main branch: 981bb343c3cfb29644817ed2824af0385d74f148

- [ ] I want to work on this

Barba2k2:
# Description

Two unrelated UI bugs in the DevTools extension surfaced by #3289 (running the extension against a game with constant animations).

### 1. Component tree never fills the panel

`ComponentTreeSection` wrapped `TreeView.simple(shrinkWrap: true)` inside a `SingleChildScrollView`. With `shrinkWrap: true` the tree sized itself to its content and let the outer scroll view be unbounded; it therefore grew on every animation frame and never filled the `Expanded` slot above it — that's the "constantly growing list that does not take up the entire vertical space" in the report.

Dropped the outer `SingleChildScrollView` and the `shrinkWrap: true` flag. `TreeView.simple` now owns its scrolling and fills the available height the parent gives it.

### 2. Selected-component snapshot is stale ("gray image")

`Base64Image` fed snapshot bytes through `Flame.images.fromBase64(imageId, base64)`, keyed by **component id only**. Re-requesting a snapshot for the same component after its game state changed hit the global cache and got the **previous** decoded frame back — that's the gray/stale image in the report.

Converted `Base64Image` to a `StatefulWidget` that decodes the base64 directly via `decodeImageFromList`, owns the resulting `ui.Image`, and disposes it on widget removal or when `base64` / `imageId` changes. Bypasses the global cache entirely, which is the right call for these short-lived debug snapshots.

### How to verify

1. `melos devtools-build`
2. Run `examples/` and open the Rogue Shooter game.
3. Open DevTools → Flame tab.
4. The component tree should fill the left panel and scroll internally instead of growing each frame.
5. Selecting a component repeatedly (or letting time pass) should show a fresh snapshot rather than the first cached frame.

## Checklist

- [-] I have updated/added tests for ALL new/updated/fixed functionality. (`flame_devtools` has no test suite; manual verification steps above)

## Breaking Change?

## Related Issues

Fixes #3289


spydon (commented):
The PR description needs to follow the PR template.

Barba2k2:
@spydon updated the description to follow the PR template — thanks for the heads up.

spydon (approved):
Lgtm!

PR #3914, opened 2026-05-07
feat(flame_3d): Add `UnlitMaterial` support

wolfenrain:
# Description

Add `UnlitMaterial` support

## Checklist

## Breaking Change?

## Related Issues


packages/flame_3d/shaders/unlit_material.frag
```diff
@@ -0,0 +1,19 @@
...
+uniform sampler2D albedoTexture;
+
+uniform Material {
+  vec4 albedoColor;
+} material;
+
+void main() {
+  vec4 texColor = texture(albedoTexture, fragTexCoord);
+  outColor = texColor * material.albedoColor;
+}
```
luanpotter, line 19: does that mean it also does not support animations, joints, etc? should we call that out?

luanpotter (approved):
LGTM

packages/flame_3d/shaders/unlit_material.frag
wolfenrain: They are supported, just not used. Does not affect the material itself

luanpotter: I thought we needed dedicated code on the shader for it to work?

wolfenrain: nuhu, just the inputs and then slap a color on it. We can for sure improve it but imho these stuff should (long term) stay out of material shaders, and be dealt with through a proper shader pipeline

luanpotter: yes, I agree eventually we should have a pipeline.
but rn for these to work, you need extra code on the shader here I believe.
so since you don't have it, it will not work - so we should document it on the class that it doesn't support it.
right?

PR #3934, opened 2026-06-15
fix!: Do not delegate onDragCancel to onDragEnd by default

### LINKED ISSUE #3926: fix: onDragCancel should not delegate to onDragEnd by default
## Problem

`DragCallbacks.onDragCancel` currently delegates to `onDragEnd` by default:

```dart
void onDragCancel(DragCancelEvent event) => onDragEnd(event.toDragEnd());
```

These two events are semantically different:

- **`onDragEnd`**: the user lifted their finger naturally; carries velocity info for fling detection.
- **`onDragCancel`**: the gesture was interrupted (another recognizer won the arena, a second finger caused a scale takeover, a system event, etc.); no meaningful velocity.

Calling `onDragEnd` from `onDragCancel` conflates the two. A drag-to-dismiss component would incorrectly "apply the action" when the drag was actually cancelled. Calls from cancel always produce zero velocity in `DragEndDetails`, which can cause subtle bugs.

The comment in the source calls cancellations "very rare", but with `MultiDragScaleDispatcher` (PR #3782) this is no longer true: every two-finger pinch gesture cancels the individual pointer drags.

## Proposed fix

- `onDragCancel` should reset `isDragged` directly (the same way `onDragEnd` does) without forwarding to `onDragEnd`.
- Users who want identical handling can override `onDragCancel` to call `onDragEnd` themselves.

## Impact

This is a breaking change: existing code that relies on `onDragCancel` triggering `onDragEnd` will need to be updated.

Related to #1938.

rumitvn:
# Description

`DragCallbacks.onDragCancel` used to delegate to `onDragEnd` by default:

```dart
void onDragCancel(DragCancelEvent event) => onDragEnd(event.toDragEnd());
```

But the two events mean different things. `onDragEnd` is the natural end of a drag and carries
velocity for fling detection, while `onDragCancel` means that the gesture was interrupted (another
recognizer won the arena, a second pointer triggered a scale takeover, a system event, etc.) and has
no meaningful velocity. Forwarding a cancel to `onDragEnd` makes a drag-to-dismiss component apply
its action even though the drag was actually cancelled. The old "very rare event" comment is no
longer true either: with `MultiDragScaleDispatcher` every two finger pinch cancels the individual
pointer drags.

`onDragCancel` now only resets `isDragged` (the same state that `onDragEnd` clears) without
forwarding. Anyone who wants the old behavior can override `onDragCancel` and call
`onDragEnd(event.toDragEnd())`, as the updated doc comment and the migration guide describe.

The examples and the Klondike tutorial that relied on the old delegation have been updated to handle
`onDragCancel` explicitly, otherwise a cancelled drag would leave a mouse joint alive, a ball with
its drag paint, or a card floating in the middle of the table.

## Checklist

## Breaking Change?

### Migration instructions

`onDragEnd` is no longer called when a drag is cancelled, only `onDragCancel` is. If you were
relying on the old behavior, forward the event yourself:

```dart
@override
void onDragCancel(DragCancelEvent event) {
  super.onDragCancel(event);
  onDragEnd(event.toDragEnd());
}
```

If a cancelled drag should instead be reverted, put that logic in `onDragCancel` without calling
`onDragEnd`. See `doc/flame/migration.md` for the full entry.

## Related Issues

Closes #3926

### DIFF
--- a/doc/flame/inputs/drag_events.md
+++ b/doc/flame/inputs/drag_events.md
@@ -87,8 +87,11 @@ position associated with this event.
 
 ### onDragCancel
 
-The precise semantics when this event occurs is not clear, so we provide a default implementation
-which simply converts this event into an `onDragEnd`.
+This event is fired when the drag gesture is interrupted before it ends naturally, for example when
+another gesture recognizer wins the gesture arena or a second pointer triggers a scale takeover.
+Unlike `onDragEnd` it carries no velocity information. The default implementation simply resets the
+drag state; override it and call `onDragEnd` yourself if you want a cancellation handled identically
+to a natural drag end.
 
 
 ## Mixins
--- a/doc/flame/inputs/drag_events.md
+++ b/doc/flame/inputs/drag_events.md
@@ -90,8 +90,8 @@ position associated with this event.
 This event is fired when the drag gesture is interrupted before it ends naturally, for example when
 another gesture recognizer wins the gesture arena or a second pointer triggers a scale takeover.
 Unlike `onDragEnd` it carries no velocity information. The default implementation simply resets the
-drag state; override it and call `onDragEnd` yourself if you want a cancellation handled identically
-to a natural drag end.
+drag state; override it and call `onDragEnd(event.toDragEnd())` yourself if you want a cancellation
+handled identically to a natural drag end.
 
 
 ## Mixins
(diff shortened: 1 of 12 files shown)


spydon (commented):
Since this is a breaking change it's not going to be something we pull in now, it is listed among the V2 issues: https://github.com/flame-engine/flame/issues/1938
Specifically: https://github.com/flame-engine/flame/issues/3926

spydon (approved):

PR #3960, opened 2026-07-22
perf!: Flatten the update traversal behind a CustomTraversal seam

### LINKED ISSUE #3957: Make the Flame Component System core more efficient
## Motivation

An investigation into the performance of the Flame Component System (update/render traversal, lifecycle processing, priority changes, hit testing) found that most of the framework overhead came from the children container and from per-frame allocation churn, not from game logic. The parent/children tree structure and the developer experience were to be preserved (this was not an ECS rewrite), but breaking changes were acceptable.

## Implementations considered

**Children container.** Three full replacements for the `SplayTreeMap<num, Set<Component>>`-backed `OrderedSet` were implemented, each passing the entire test suite, and benchmarked head-to-head ([full comparison table](https://github.com/flame-engine/flame/issues/3957#issuecomment-5043618243)):

- **Flat sorted array** (chosen): one contiguous list sorted by `(priority, insertion order)`, with the container and slot index stored intrusively on the component, `null` tombstones for removals (compacted once per tick), and a stable sort on priority changes that is skipped when nothing actually reordered. Best or within noise on nearly every benchmark, and no pathological case.
- **Intrusive doubly-linked list** (`bench/linked-list`): `_prevSibling`/`_nextSibling` pointers, O(1) link/unlink. Fastest pure update walk, but pointer-chasing loses on rendering (contiguous pointer arrays prefetch, dependent loads serialize), priority changes cost about 2.7x more than the array, and any structural mutation must invalidate live iterators.
- **Priority-bucketed arrays** (`bench/priority-buckets`): one dense array per distinct priority, so a priority change is a bucket move instead of a sort. Middling everywhere and pathological under y-sort, where 1000 distinct priorities degenerate into 1000 single-element buckets and it lands slower than the old `OrderedSet`.
- **`SplayTreeMap<(priority, insertOrder), Component>`** (`bench/splay-composite`, proposed in review): removes the per-priority `Set` buckets by giving every component its own composite-keyed tree entry. This turned out to make iteration much worse, not better: in the old design, same-priority components share one splay node and iterate through a hash table, so the inner sets were never the iteration bottleneck; with a node per component, every visit is a scattered pointer chase with iterator workstack bookkeeping (~32ns/visit against ~8 for the old design and ~3 to 6 for the array). Every add/remove also pays an O(log n) splay plus a key-record allocation, so churn is about twice the old cost. It does fix the single-child rebalance pathology (73ms to 19ms), but the array fixes it better (5.7ms), and under y-sort the per-mover re-keying is slower than the old full rebuild.

**Update traversal.** Keeping the recursive `updateTree` was compared against a root-owned flattened pre-order list. The flat list won once two problems were solved: overridden traversals being silently skipped (solved by making `updateTree` non-virtual behind a `CustomTraversal` marker + `updateSubtree`), and rebuild cost under churn (solved by structure versioning with lazy rebuilds that are fused into the same tick's update pass).

**Considered and rejected:**

- Flattening the render pass: rendering is hierarchical through canvas transforms, clips, decorators, and render contexts, and `renderTree` has many legitimate overriders; the render pass stays recursive and virtual. Render-order work continues as the `RenderLayer` proposal in #3967.
- Rewriting engine hit testing as an explicit-stack traversal: `componentsAtLocation` is an overridden extension point (`CameraComponent`, `Route`, ecosystem), so a parallel path would silently skip those overrides. Root-level counters of mounted pointer-event handlers attack the same per-event cost safely instead.
- First-class `timeScale` on `Component`: rarely used; it stays in the opt-in `HasTimeScale` mixin, which now composes through `updateSubtree`.

## Decision

Implemented in #3960:

- `ComponentList` (the flat sorted array) replaces `OrderedSet`; the dependency is removed. `query<T>()`/`register<T>()` are kept, `childrenFactory` is replaced by an overridable `createComponentList()` with an optional comparator for custom orderings such as y-sort.
- The update pass runs over the root-owned flattened list; `CustomTraversal` components are barriers that drive their own subtrees via `updateSubtree`. A new `updatePaused` pauses a subtree's updates entirely (and now backs `Route.stopTime()`).
- Allocation hygiene throughout the hot paths: no per-frame decorator or camera closures, lazy render contexts and debug caches, generator-free removal teardown, an early-out for lifecycle processing and for hit tests in games without pointer handlers, and a near-linear insertion sort in the `Sweep` broadphase.

Golden tests pin lifecycle ordering, hit-test order, and equal-priority ordering across the rewrite. Migration notes are in #3960 and in the migration guide.

## Benchmarks

Container candidates, measured with each backing as a drop-in `children` container (JIT, ms per run, lower is better; the old `OrderedSet` as baseline; all five measured in one same-day run):

| Benchmark | OrderedSet | Flat sorted array | Linked list | Priority buckets | Splay composite key |
|---|---|---|---|---|---|
| Update wide tree (10k x 1) | 8.43 | 5.86 | **3.83** | 6.52 | 32.06 |
| Render wide tree (10k x 1) | 13.21 | **8.68** | 10.97 | 9.12 | 32.68 |
| Update nested tree (1k x 10) | 17.69 | 7.19 | **6.51** | 7.14 | 33.36 |
| Render nested tree (1k x 10) | 23.53 | **10.08** | 11.26 | 10.56 | 43.26 |
| Update deep tree (100 levels) | 18.47 | 6.38 | **5.86** | 7.08 | 34.31 |
| Lifecycle churn (100/tick, 10k pop) | 13.06 | 7.70 | **7.30** | 7.56 | 27.23 |
| Mass add/remove (1k per cycle) | 3.48 | **2.33** | 2.40 | 2.34 | 5.40 |
| Priority change (1 child per parent) | 73.35 | **5.74** | 16.64 | 8.58 | 19.17 |
| Priority change (y-sort, all children) | 22.44 | **6.07** | 9.64 | 34.62 | 39.17 |
| Type-query churn (2 registered queries) | 7.14 | **6.37** | 6.46 | 6.36 | 10.19 |
| Game-like update pass | 191.62 | **117.18** | 118.55 | 130.45 | 291.89 |
| Flat collision detection | 9.68 | 3.96 | **3.61** | 5.19 | 8.72 |
| Nested collision detection | 16.22 | 5.35 | **4.87** | 8.06 | 13.08 |

Final state of #3960 (container plus flattened traversal plus allocation hygiene) versus main:

| Benchmark | main | #3960 | speedup |
|---|---|---|---|
| Update wide tree (10k x 1) | 8.09 | 3.44 | 2.4x |
| Update nested tree (1k x 10) | 16.96 | 3.97 | 4.3x |
| Update deep tree (100 levels) | 18.14 | 3.36 | 5.4x |
| Update barrier tree (10% time-scaled) | 20.45 | 4.47 | 4.6x |
| Render wide tree (10k x 1) | 12.81 | 4.64 | 2.8x |
| Render nested tree (1k x 10) | 23.05 | 8.25 | 2.8x |
| Render deep tree (100 levels) | 27.44 | 6.89 | 4.0x |
| Lifecycle churn (100/tick, 1k pop) | 5.10 | 2.54 | 2.0x |
| Lifecycle churn (100/tick, 10k pop) | 12.56 | 8.63 | 1.5x |
| Mass add/remove (1k per cycle) | 3.43 | 1.31 | 2.6x |
| Priority change (1 child per parent) | 74.46 | 6.92 | 10.8x |
| Priority change (y-sort, all children) | 22.61 | 5.82 | 3.9x |
| Type-query churn (2 registered queries) | 6.95 | 5.16 | 1.3x |
| Game-like update pass | 185.29 | 95.31 | 1.9x |
| Hit test + delivery (cached) | 90.93 | 39.06 | 2.3x |
| Flat collision detection | 9.24 | 3.35 | 2.8x |
| Nested collision detection | 15.41 | 3.52 | 4.4x |

Numbers are `flutter test` JIT on one desktop machine, for relative comparison; AOT device numbers are pending and should show smaller but same-ranked multiples.

Related: #1938 (first item).

spydon:
# Description

Second half of the FCS core rewrite for #3957: replaces the recursive, virtual `updateTree` with a root-owned flattened traversal list.

**Stacked on #4019.** That PR replaces the `OrderedSet` children container with the flat sorted-array `ComponentList` (O(1) remove and contains, tombstone compaction, one stable sort per parent per tick, `createComponentList()` with an optional comparator). This PR is based on that branch and only contains the traversal work on top of it; the diff here shrinks to the traversal commits, and the container change is reviewed over there. Once #4019 lands, this PR retargets to `main` on its own.

## The update pass runs over a flattened traversal list

`Component.updateTree` is now `@nonVirtual`. Components that manage their own subtree traversal mix in `CustomTraversal` and override its `updateSubtree`; the method only exists on the mixin, so it cannot be overridden without also becoming a barrier. Traversal mixins such as `HasTimeScale` are declared `on CustomTraversal` and compose through `super.updateSubtree`, which means their users write `with CustomTraversal, HasTimeScale` (except on `FlameGame`, which already mixes it in).

The root updates everything through a flattened pre-order list that is rebuilt lazily, only on ticks where the tree structure changed (and then fused into that tick's update pass, so the rebuild costs no extra traversal). `CustomTraversal` components act as barriers that drive their own subtrees. The render pass intentionally stays recursive and virtual, since `renderTree` has many legitimate overriders (decorators, visibility, snapshot, cameras).

`Route.stopTime()` and `resumeTime()` keep their public behavior. The gating now lives in `HasTimeScale` itself: a `timeScale` of `0` stops the update pass for the whole subtree instead of updating it with a `dt` of `0`, so `Route` no longer needs an override. `FlameGame` processes lifecycle events before dispatching into the traversal, so components added to a paused game are still mounted.

## Relation to the allocation-hygiene PRs

The tangential hot-path improvements were extracted from this PR and reviewed independently, per [the review feedback](https://github.com/flame-engine/flame/issues/3957#issuecomment-5066267594). Merged: #3978 (lifecycle early-out), #3979 (camera render closures; the decorator tear-off caching was deferred during its review), #3980 (lazy render contexts), #3981 (tree teardown and propagation), #3983 (sweep broadphase). #3982 (pointer-handler counters) was closed in favor of an upcoming events rework. This branch includes all of that, and on top of #4019 it contains only the traversal work. The golden tests added in #4019 (lifecycle-event ordering, hit-test order, equal-priority ordering) run unchanged over the flattened traversal.

## Benchmarks (full stack: #4019 plus this PR, against main)

Measured against current main, which already includes the merged hygiene PRs, so these tables isolate the effect of the two stacked PRs together. The per-PR attribution follows in the next section. ms per run, lower is better.

### JIT (`flutter test`, host machine, repeated best-of runs)

| Benchmark | main | this PR | speedup |
|---|---|---|---|
| Update wide tree (10k x 1) | 8.65 | 3.66 | 2.4x |
| Update nested tree (1k x 10) | 22.45 | 4.00 | 5.6x |
| Update deep tree (100 levels) | 18.81 | 3.41 | 5.5x |
| Update barrier tree (10% time-scaled) | 21.04 | 4.51 | 4.7x |
| Render wide tree (10k x 1) | 12.63 | 7.17 | 1.8x |
| Render nested tree (1k x 10) | 22.66 | 10.80 | 2.1x |
| Render deep tree (100 levels) | 24.96 | 9.47 | 2.6x |
| Lifecycle churn (100/tick, 1k pop) | 4.02 | 2.74 | 1.5x |
| Lifecycle churn (100/tick, 10k pop) | 11.42 | 9.24 | 1.2x |
| Mass add/remove (1k per cycle) | 2.47 | 1.54 | 1.6x |
| Priority change (1 child per parent) | 75.10 | 7.08 | 10.6x |
| Priority change (y-sort, all children) | 23.25 | 6.37 | 3.6x |
| Type-query churn (2 registered queries) | 6.73 | 5.85 | 1.2x |
| Game-like update pass | 202.71 | 107.48 | 1.9x |
| Hit test + delivery (cached) | 84.24 | 77.25 | 1.1x |
| Flat collision detection | 9.57 | 3.35 | 2.9x |
| Nested collision detection | 15.59 | 3.36 | 4.6x |

### AOT (`flutter run --release`, Android arm64 emulator on an Apple Silicon host, single run)

| Benchmark | main | this PR | speedup |
|---|---|---|---|
| Update wide tree (10k x 1) | 6.25 | 1.45 | 4.3x |
| Update nested tree (1k x 10) | 9.53 | 1.13 | 8.4x |
| Update deep tree (100 levels) | 11.28 | 1.67 | 6.8x |
| Update barrier tree (10% time-scaled) | 12.35 | 1.39 | 8.9x |
| Render wide tree (10k x 1) | 6.42 | 3.96 | 1.6x |
| Render nested tree (1k x 10) | 12.44 | 4.92 | 2.5x |
| Render deep tree (100 levels) | 14.74 | 6.61 | 2.2x |
| Lifecycle churn (100/tick, 1k pop) | 1.54 | 1.07 | 1.4x |
| Lifecycle churn (100/tick, 10k pop) | 4.80 | 4.52 | 1.1x |
| Mass add/remove (1k per cycle) | 0.89 | 0.44 | 2.0x |
| Priority change (1 child per parent) | 24.67 | 4.15 | 5.9x |
| Priority change (y-sort, all children) | 11.28 | 3.31 | 3.4x |
| Type-query churn (2 registered queries) | 5.36 | 4.73 | 1.1x |
| Game-like update pass | 79.68 | 21.42 | 3.7x |
| Hit test + delivery (cached) | 43.30 | 37.42 | 1.2x |
| Flat collision detection | 4.26 | 1.01 | 4.2x |
| Nested collision detection | 7.92 | 0.97 | 8.2x |

The AOT numbers come from the `crud_test` AVD (API 34, arm64-v8a, 4 cores), which executes the AOT snapshot natively on the host CPU, so they are representative of release-mode Dart rather than of any particular phone. The earlier guess that AOT would show smaller multiples turned out to be wrong for the update pass: the four update-pass rows and both collision rows gain more under AOT than under JIT (nested tree 8.4x, barrier tree 8.9x, nested collision 8.2x), the render and churn rows land about where JIT put them, and priority change gains less (5.9x against 10.6x). The ordering of the rows is therefore not the same as in the JIT table; the update pass is where AOT rewards the flat list the most.

## Attributing the speedup: how much comes from the container, how much from traversal flattening

The stack bundles two structural changes: the `OrderedSet` to `ComponentList` container rewrite (#4019), and the flattened-traversal update pass built on `CustomTraversal` (this PR). To see how much each contributes on its own, the benchmark suite was run at three points in the branch history, on the same machine.

- `main` (commit `7fd33dee9`): the baseline, before either PR.
- `1a25aea1f`: the container rewrite alone, which is exactly the head of #4019. It has the flat sorted array `ComponentList`, O(1) remove and contains, tombstone compaction, and a single stable sort per parent per tick, with the old recursive, virtual `updateTree` still in place.
- `HEAD`: the full stack, container rewrite plus traversal flattening.

Two of the later commits on this branch were small allocation cleanups to `ComponentList` that built on the traversal-flattening commit, so they could not be isolated from the traversal work and are counted on that side of the split. The query-cache list mirror they introduced has since been dropped again: iterating the map values directly measures the same on the type-query churn benchmark.

### JIT (`flutter test`, host machine, single run)

| Benchmark | main | container only | this PR | container gain | traversal gain |
|---|---|---|---|---|---|
| Update wide tree (10k x 1) | 3.24 | 1.19 | 1.28 | 2.7x | flat |
| Update nested tree (1k x 10) | 8.08 | 2.57 | 1.35 | 3.1x | 1.9x |
| Update deep tree (100 levels) | 8.46 | 3.23 | 1.21 | 2.6x | 2.7x |
| Update barrier tree (10% time-scaled) | 9.15 | 3.85 | 1.69 | 2.4x | 2.3x |
| Render wide tree (10k x 1) | 5.14 | 4.05 | 3.98 | 1.3x | flat |
| Render nested tree (1k x 10) | 9.95 | 4.92 | 4.64 | 2.0x | flat |
| Render deep tree (100 levels) | 12.34 | 6.56 | 5.45 | 1.9x | 1.2x |
| Lifecycle churn (100/tick, 1k pop) | 1.80 | 1.15 | 1.34 | 1.6x | flat |
| Lifecycle churn (100/tick, 10k pop) | 4.86 | 2.82 | 4.64 | 1.7x | flat |
| Mass add/remove (1k per cycle) | 1.15 | 0.64 | 0.66 | 1.8x | flat |
| Priority change (1 child per parent) | 26.01 | 3.44 | 3.73 | 7.6x | flat |
| Priority change (y-sort, all children) | 11.72 | 3.33 | 3.35 | 3.5x | flat |
| Type-query churn (2 registered queries) | 3.41 | 2.83 | 2.94 | 1.2x | flat |
| Game-like update pass | 98.34 | 70.09 | 42.62 | 1.4x | 1.6x |
| Hit test + delivery (cached) | 43.63 | 41.93 | 53.24 | flat | noisy |
| Flat collision detection | 4.95 | 2.04 | 1.70 | 2.4x | 1.2x |
| Nested collision detection | 7.91 | 2.84 | 1.83 | 2.8x | 1.6x |

### AOT (`flutter run --release`, Android arm64 emulator, single run)

| Benchmark | main | container only | this PR | container gain | traversal gain |
|---|---|---|---|---|---|
| Update wide tree (10k x 1) | 6.25 | 4.31 | 1.45 | 1.5x | 3.0x |
| Update nested tree (1k x 10) | 9.53 | 3.91 | 1.13 | 2.4x | 3.5x |
| Update deep tree (100 levels) | 11.28 | 4.27 | 1.67 | 2.6x | 2.6x |
| Update barrier tree (10% time-scaled) | 12.35 | 6.81 | 1.39 | 1.8x | 4.9x |
| Render wide tree (10k x 1) | 6.42 | 3.97 | 3.96 | 1.6x | flat |
| Render nested tree (1k x 10) | 12.44 | 4.70 | 4.92 | 2.6x | flat |
| Render deep tree (100 levels) | 14.74 | 6.53 | 6.61 | 2.3x | flat |
| Lifecycle churn (100/tick, 1k pop) | 1.54 | 0.90 | 1.07 | 1.7x | 0.8x |
| Lifecycle churn (100/tick, 10k pop) | 4.80 | 3.08 | 4.52 | 1.6x | 0.7x |
| Mass add/remove (1k per cycle) | 0.89 | 0.42 | 0.44 | 2.1x | flat |
| Priority change (1 child per parent) | 24.67 | 3.42 | 4.15 | 7.2x | 0.8x |
| Priority change (y-sort, all children) | 11.28 | 3.30 | 3.31 | 3.4x | flat |
| Type-query churn (2 registered queries) | 5.36 | 4.75 | 4.73 | 1.1x | flat |
| Game-like update pass | 79.68 | 50.16 | 21.42 | 1.6x | 2.3x |
| Hit test + delivery (cached) | 43.30 | 38.48 | 37.42 | 1.1x | flat |
| Flat collision detection | 4.26 | 1.47 | 1.01 | 2.9x | 1.5x |
| Nested collision detection | 7.92 | 1.95 | 0.97 | 4.1x | 2.0x |

The container rewrite alone accounts for most of the gain across the board: nearly all of the priority change, lifecycle churn, mass add and remove, and render benchmarks, and roughly half of the update pass benchmarks. The traversal flattening adds most of its own value on trees with real nesting or depth (nested tree, deep tree, barrier tree, both collision detection benchmarks, the game-like update pass), where recursion overhead is what is left to cut once the container is fast. On the flat wide tree and on the simpler priority change and churn benchmarks, the traversal flattening contributes close to nothing on top of the container change, which fits, those do not stress recursive traversal in the first place.

Under AOT the split is the same in shape but the traversal side is worth more: the flattened traversal is a 2.6x to 4.9x gain on top of the container on every update-pass row, including the flat wide tree where it was flat under JIT, and 2.3x on the game-like update pass. The render rows are flat on the traversal side in both modes, as expected since the render pass stays recursive.

One trade-off shows up in both modes and is worth calling out: lifecycle churn at 10k population is slower with the flattened traversal than with the container alone (JIT 2.82 to 4.64 ms, AOT 3.08 to 4.52 ms), and single-child priority changes lose a little (AOT 3.42 to 4.15 ms). Both benchmarks change the tree structure every tick, so the flat list is rebuilt every tick and the rebuild cost is paid without ever being amortized. Against `main` both rows are still ahead or level (churn at 10k: 4.80 to 4.52 ms AOT), and a real game changes structure on a fraction of its ticks, but it is the one workload where this PR gives some of the container's gain back.

These attribution numbers come from a single run per checkpoint, so treat the exact multiples loosely, particularly the rows marked flat or noisy. Hit test and delivery moved within noise across all three checkpoints in both runs and should not be read as a regression.

## Checklist

## Breaking Change?

### Migration instructions

The container migration (`ComponentList`, `createComponentList()`, `ConcurrentModificationError` semantics) is described in #4019. This PR adds:

- `Component.updateTree` is non-virtual. If you overrode it, add `with CustomTraversal` and override its `updateSubtree` instead; call `super.updateSubtree(dt)` for the standard traversal.
- `HasTimeScale` is now `on CustomTraversal`: `extends Component with HasTimeScale` becomes `with CustomTraversal, HasTimeScale` (`extends FlameGame with HasTimeScale` is unchanged, since `FlameGame` already mixes in `CustomTraversal`).

## Related Issues

Closes #3957


packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,551 @@
...
+        element._containerList = null;
+        element._containerIndex = -1;
+      }
+    }
+    elements.clear();
+    _length = 0;
+    _tombstones = 0;
+    _shiftCount++;
+    structureVersion++;
+    _queries?.forEach((_, cache) => cache.data.clear());
```
erickzanardo: I remember reading somewhere that using "normal" `for`s perform better than `forEach`s, since this PR is about performance as well, might be a good micro optimization to add?

erickzanardo (approved):

packages/flame/lib/src/components/core/component_list.dart
spydon: True, fixed!

doc/flame/migration.md
```diff
@@ -25,3 +25,88 @@ GameWidget.managed(
   gameFactory: MyGame.new,
 );
 ```
+
+### `children` is now a `ComponentList` instead of an `OrderedSet`
```
spydon, line 700:
```suggestion

### `children` is now a `ComponentList` instead of an `OrderedSet`
```

erickzanardo (approved):
LGTM!

packages/flame/lib/src/components/router/route.dart
```diff
@@ -90,16 +90,20 @@ class Route extends PositionComponent
...
-  /// Resumes normal time progression for the page, if it was previously slowed
-  /// down or stopped.
-  void resumeTime() => timeScale = 1.0;
+  /// While stopped, neither the page nor any of its descendants are updated
+  /// at all (see [Component.updatePaused]), which saves computational
+  /// resources. The page keeps rendering, and pending lifecycle events
+  /// (adding or removing components) still complete. The [timeScale] is left
+  /// untouched, so a slow-motion factor survives a stop/resume cycle of the
+  /// route below a popup.
+  void stopTime() => updatePaused = true;
```
luanpotter: wouldn't setting updatePaused be cleanly equivalent to override renderSubtree and CustomTraversal? (ie - do we need a separate engine mechanism to pause?)

spydon: Fixed

doc/flame/migration.md
```diff
@@ -533,3 +533,77 @@ if (game.isPaused) {
   game.isPaused = false;
 }
 ```
+
+
+### `children` is now a `ComponentList` instead of an `OrderedSet`
```
luanpotter, line 700: this should be on the other PR

packages/flame/test/components/mixins/has_time_scale_self_test.dart
```diff
@@ -0,0 +1,36 @@
...
+      game,
+    ) async {
+      final component = _ScaledRecorder()..timeScale = 0.5;
+      await game.world.ensureAdd(component);
+
+      game.update(1.0);
+      expect(component.recordedDts, [0.5]);
+    });
+
+    testWithFlameGame('scales its own dt when called via update directly', (
```
luanpotter: should it say "does not scale"?

packages/flame/test/components/core/custom_traversal_test.dart
```diff
@@ -0,0 +1,98 @@
...
+      game.update(0.5);
+      expect(child.recordedDts, [0.5]);
+    });
+
+    testWithFlameGame('updateSubtree can modify the dt for the subtree', (
+      game,
+    ) async {
+      final child = _DtRecorder();
+      final grandChild = _DtRecorder();
+      child.add(grandChild);
```
luanpotter:
```suggestion
       final grandChild = _DtRecorder();
       final child = _DtRecorder()..add(grandChild);
```

packages/flame/lib/src/components/router/route.dart
```diff
@@ -170,9 +170,9 @@ class Route extends PositionComponent
   }
 
   @override
-  void updateTree(double dt) {
+  void updateSubtree(double dt) {
     if (timeScale > 0) {
-      super.updateTree(dt);
+      super.updateSubtree(dt);
     }
```
luanpotter: isn't this something HasTimeScale could handle?
if == 0 -> kill propagation, else * it
then you don't need this layer of override at all

doc/flame/migration.md
```diff
@@ -583,3 +583,38 @@ class YSortedWorld extends World {
...
+}
+
+// After
+class SlowMotionArea extends Component with CustomTraversal {
+  @override
+  void updateSubtree(double dt) => super.updateSubtree(dt / 2);
+}
+```
+
+`HasTimeScale` is now declared `on CustomTraversal`, so components other than `FlameGame` (which
```
luanpotter, line 770: "is now" - archeology

doc/flame/components/components.md
```diff
@@ -146,6 +146,28 @@ class MyComponent extends PositionComponent with TapCallbacks {
...
+```dart
+class SlowMotionArea extends Component with CustomTraversal {
+  @override
+  void updateSubtree(double dt) => super.updateSubtree(dt / 2);
+}
+```
+
+The engine treats every `CustomTraversal` component as a traversal barrier: it appears in the
+flattened list itself and its `updateSubtree` drives its subtree. `updateSubtree` only exists on
+the mixin, so it cannot be overridden without also becoming a barrier. Mixins that build on it
```
luanpotter: "only exists on the mixin" - I think we can just say "the mixin which provides a method" or something, this framing is archeological

doc/flame/migration.md
luanpotter: (sorry this is the migration guide, old habits)

packages/flame/lib/src/components/core/component.dart
```diff
@@ -583,7 +584,77 @@ class Component {
...
+      }
+    }
+  }
+
+  /// The engine's standard update traversal: update this component, then
+  /// update the children in priority order.
+  ///
+  /// This is the default behavior of [CustomTraversal.updateSubtree];
+  /// custom traversals can call it to delegate to the standard behavior.
+  @protected
```
luanpotter: why not internal too? or even private

doc/flame/other/util.md
```diff
@@ -169,7 +169,10 @@ to any Flame `Component` and exposes a simple get/set API for `timeScale`. The d
 `timeScale` is `1`, implying in-game time of the component is running at the same speed as real life
 time. Setting it to `2` will make the component tick twice as fast and setting it to `0.5` will make
 it tick at half the speed as compared to real life time. This mixin also provides `pause` and `resume`
-methods, which can be used instead of manually setting the timeScale to 0 and 1 respectively.
+methods, which can be used instead of manually setting the timeScale to 0 and 1 respectively. A
+`timeScale` of `0` stops the update pass for the component and its whole subtree: `update` is not
+called on any of them until the time scale is raised again. Components added to a paused game are
```
luanpotter: raised -> can it be negative? 🤔

packages/flame/lib/src/components/core/component.dart
```diff
@@ -590,7 +592,72 @@ class Component {
...
+  /// Whether this component manages its own subtree traversal. Evaluated
+  /// once in the constructor, so that the per-frame traversal loops pay a
+  /// plain field load instead of a type check.
+  bool _isTraversalBarrier = false;
+
+  /// Runs one update pass over a flattened traversal list produced by
+  /// [_updateAndFlattenInto].
+  static void _updateFlatList(List<Component> list, double dt) {
+    for (var i = 0; i < list.length; i++) {
+      final component = list[i];
```
luanpotter: for-in

packages/flame/lib/src/components/core/component.dart
```diff
@@ -590,7 +592,72 @@ class Component {
...
+  /// separate pass over the tree.
+  void _updateAndFlattenInto(List<Component> out, double dt) {
+    final children = _children;
+    if (children == null) {
+      return;
+    }
+    children._compact();
+    final elements = children._elements;
+    for (var i = 0; i < elements.length; i++) {
+      final child = elements[i];
```
luanpotter: for-in

doc/flame/other/util.md
luanpotter:
```suggestion
called on any of them until the time scale is changed again. Components added to a paused game are
```

:P

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -62,9 +62,22 @@ class ComponentList extends Iterable<Component> {
...
   static const int _tombstoneCompactionThreshold = 16;
 
-  /// The per-type query caches, created by [register].
+  /// The per-type query caches, created by [register], keyed by type for
+  /// [query] lookups.
   Map<Type, _QueryCache<Component>>? _queries;
 
+  /// The same caches as [_queries], as a list, so that the add/remove hot
+  /// paths can iterate them without allocating a map-values iterator.
+  List<_QueryCache<Component>>? _queryCaches;
```
luanpotter: is this truly necessary 💀

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -173,19 +173,20 @@ class FlameGame<W extends World> extends ComponentTreeRoot
   @mustCallSuper
   void update(double dt) {
     if (parent == null) {
+      // Lifecycle events are processed before the traversal so that they
+      // complete even while the traversal is paused through a time scale.
+      processLifecycleEvents();
       updateTree(dt);
```
luanpotter, line 179: this is really good

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -173,19 +173,20 @@ class FlameGame<W extends World> extends ComponentTreeRoot
...
     }
   }
 
   @override
-  void updateTree(double dt) {
-    processLifecycleEvents();
+  void updateSubtree(double dt) {
     if (parent != null) {
+      processLifecycleEvents();
```
luanpotter, line 186: I don't understand this being added

packages/flame/test/components/mixins/has_time_scale_self_test.dart
```diff
@@ -0,0 +1,81 @@
...
+    ) async {
+      final component = _ScaledRecorder()..timeScale = 0.5;
+      await game.world.ensureAdd(component);
+
+      game.update(1.0);
+      expect(component.recordedDts, [0.5]);
+    });
+
+    testWithFlameGame(
+      'does not scale its own dt when called via update directly',
```
luanpotter: I think this test is misleading, testing something that doesn't actually has any user impact
we want to test that it _actually_ scales itself when the update is called the way the game does

packages/flame/lib/src/components/core/component_list.dart
luanpotter: I think this is such a micro optimization that is not worth the complexity

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -325,7 +347,8 @@ class ComponentList extends Iterable<Component> {
         data.add(element);
       }
     }
-    queries[C] = _QueryCache<C>(data);
+    final cache = _QueryCache<C>(data);
+    queries[C] = cache;
```
luanpotter:
```suggestion
    queries[C] = _QueryCache<C>(data);
```

packages/flame/lib/src/components/core/component_tree_root.dart
```diff
@@ -114,6 +114,38 @@ class ComponentTreeRoot extends Component {
...
+  /// changed since the previous tick.
+  ///
+  /// The visit order is identical to the recursive
+  /// standard traversal: pre-order, children in
+  /// priority order. Components mixing in `CustomTraversal` are treated as
+  /// barriers: they appear in the list themselves, and their `updateSubtree`
+  /// drives their subtree.
+  @internal
+  void updateChildrenFlat(double dt) {
+    if (_flatVersion != ComponentList._structureVersion) {
```
luanpotter, line 137: note to future @spydon - what would it mean for `_flatVersion > ComponentList._structureVersion`?

luanpotter (approved):
LGTM

PR #3969, opened 2026-07-22
feat!: Remove integralSize property from class Svg, enforcing 'always true' behaviour

adario:
# Description

The `Svg.integralSize` property was added in PR #3956: when `true`, it ensures that the keys for `Svg._imageCache` are always integer-sized. The property, however, would _never_ be useful with its current default of `false`, since this would re-introduce floating-point rounding errors in the image cache keys.

# Resolution

[Remove](https://github.com/flame-engine/flame/pull/3956#issuecomment-5045164564) the `integralSize` property, always following its `true` [codepath](https://github.com/flame-engine/flame/pull/3956/changes#diff-ff717ec0f113030b284a5874321c9f93b1bca6b9e3542ec3b3b6ea5307d2f6f9R181).

# Testing

An updated testbed for the changes is [available](https://github.com/adario/flame_extended_svg) — this displays 200+1 SvgComponent objects, all with different rotation/scale values; it now features the **Svg** button, which cycles through a few SVG assets.
For comparison purposes, the testbed starts with the default implementation; the **Mode** and **Size** buttons cycle through several combinations of the `fixedRatio` and `cacheSize` properties..

## Checklist

## Breaking Change?

### Migration instructions

This PR **does** change the internal caching behaviour in class `Svg`, for **all** rendering operations. However, any existing use cases of class `Svg` _should_ build as previously , since the property changes from PR #3956 haven't been published yet. All parameters/properties from PR #3956 have default values, so no build should break.
At runtime, the PR should _not_ break class `Svg` clients either: the actual `Image` objects in the `MemoryCache` _should_ maintain the previous dimensions and contents, since these were _already_ integral prior to PR #3956. A 3-way comparison for the `melos test --scope=flame_svg` results reveals no differences (all tests fail in the same way)...

I apologise if I've missed something, but at present I wouldn't know what migration steps to provide.

## Related Issues

### DIFF
--- a/packages/flame_svg/lib/svg.dart
+++ b/packages/flame_svg/lib/svg.dart
@@ -12,7 +12,8 @@ import 'package:flutter_svg/flutter_svg.dart';
 class Svg {
   /// Creates an [Svg] with the received [pictureInfo].
   /// Default [pixelRatio] is the device pixel ratio.
-  /// Setting [fixedRatio] to `true` ensures the cache uses a single entry.
+  /// Setting [fixedRatio] to `true` ensures the cache uses one entry
+  /// per rendered size/scale.
   /// Default [cacheSize] is [defaultCacheSize], which is 10 as previously;
   /// specifying [unlimitedCacheSize] is the same as using a [Map] instead
   /// of a [MemoryCache].
--- a/packages/flame_svg/lib/svg.dart
+++ b/packages/flame_svg/lib/svg.dart
@@ -12,7 +12,6 @@ import 'package:flutter_svg/flutter_svg.dart';
 class Svg {
   /// Creates an [Svg] with the received [pictureInfo].
   /// Default [pixelRatio] is the device pixel ratio.
-  /// Setting [integralSize] to `true` uses integer dimensions for cache keys.
   /// Setting [fixedRatio] to `true` ensures the cache uses a single entry.
   /// Default [cacheSize] is [defaultCacheSize], which is 10 as previously;
   /// specifying [unlimitedCacheSize] is the same as using a [Map] instead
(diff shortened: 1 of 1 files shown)


spydon (approved):
Thanks for your contribution!

adario:
> Thanks for your contribution!

Thank you. :-)
I'll have a look at Copilot's comments.

adario:
> > Thanks for your contribution!
> 
> Thank you. :-) I'll have a look at Copilot's comments.

@spydon — regarding [this one](https://github.com/flame-engine/flame/pull/3969#discussion_r3637979965), I've updated the [dartdoc](https://github.com/adario/flame/commit/450ebfc0f42f2c10cabd2414a56257be7f8a2a47).

[This one](https://github.com/flame-engine/flame/pull/3969#discussion_r3637979933), while obviously valid, as expected costs _quite_ a bit... the [testbed](https://github.com/adario/flame_extended_svg) shows **~35%** increased cache usage (~170 instead of ~110 entries): after all, `pixelRatio` wasn't used for the cache keys prior to PR #3956 either.

Running image diffs on a few rendering results hasn't revealed any rasterisation improvements; the deltas all appear to be < 1%, and they mostly occur in the alpha channel. Anyways, the cache changes _would_ formally align the keys with the contents' dimensions, so... please let me know whether I should add them to the PR.

(From a 'field usage' POV, the suggested approach would be what the [testbed](https://github.com/adario/flame_extended_svg) does: when `Svg.fixedRatio` is set to `true`with the **Mode** button, and all `SvgComponent` objects share the same `size` property, `Svg._imageCache` only ever contains a single entry.This still allows to use whatever dimensions for any `SvgComponent` via the `scale` property,  and whatever rotation via the `angle` property.)

spydon:
@adario hmm, I'll let you decided, both are fine for me.

adario:
> @adario hmm, I'll let you decided, both are fine for me.

Thanks again: I think we could merge the PR, and revisit the Copilot changes at a later time... I've stashed them just in case.

PR #3978, opened 2026-08-05
perf: Skip lifecycle processing work when the queue is empty

spydon:
# Description

`processLifecycleEvents` now returns immediately when the queue is empty instead of allocating a set and a closure on every tick, the reorder-parents set is only allocated when a priority change is actually queued, and the blocked-set hash lookups are skipped while the set is empty (the common single-pass case).

Extracted from #3960 so the data-structure change there stands alone (as requested in [this comment](https://github.com/flame-engine/flame/issues/3957#issuecomment-5066267594)). Behavior is unchanged; this only removes per-tick allocations and lookups from the game loop.

## Checklist

## Breaking Change?

## Related Issues

Relates to #3957

### DIFF
--- a/packages/flame/lib/src/components/core/component_tree_root.dart
+++ b/packages/flame/lib/src/components/core/component_tree_root.dart
@@ -144,9 +144,11 @@ class ComponentTreeRoot extends Component {
 
   void processLifecycleEvents() {
     if (!hasLifecycleEvents) {
-      // The completer is only ever created while events are queued, so there
-      // is nothing to complete here either.
-      assert(_lifecycleEventsCompleter == null);
+      assert(
+        _lifecycleEventsCompleter == null,
+        'The completer is only ever created while events are queued, so it '
+        'should never exist while the queue is empty',
+      );
       return;
     }
     // reorder events to process later grouped by parent
@@ -189,10 +191,8 @@ class ComponentTreeRoot extends Component {
       _blocked.clear();
     }
 
-    if (reorderParents != null) {
-      for (final parent in reorderParents!) {
-        parent.rebalanceChildren();
-      }
+    for (final parent in reorderParents ?? const <Component>{}) {
+      parent.rebalanceChildren();
     }
 
     if (!hasLifecycleEvents && _lifecycleEventsCompleter != null) {
--- a/packages/flame/lib/src/components/core/component_tree_root.dart
+++ b/packages/flame/lib/src/components/core/component_tree_root.dart
@@ -143,10 +143,16 @@ class ComponentTreeRoot extends Component {
   }
 
   void processLifecycleEvents() {
+    if (!hasLifecycleEvents) {
+      // The completer is only ever created while events are queued, so there
+      // is nothing to complete here either.
+      assert(_lifecycleEventsCompleter == null);
+      return;
+    }
     // reorder events to process later grouped by parent
-    final reorderParents = <Component>{};
+    Set<Component>? reorderParents;
     LifecycleEventStatus handleReorderEvent(Component parent) {
-      reorderParents.add(parent);
+      (reorderParents ??= {}).add(parent);
       return LifecycleEventStatus.done;
     }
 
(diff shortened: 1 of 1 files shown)


packages/flame/lib/src/components/core/component_tree_root.dart
```diff
@@ -176,8 +183,10 @@ class ComponentTreeRoot extends Component {
       _blocked.clear();
     }
 
-    for (final parent in reorderParents) {
-      parent.rebalanceChildren();
+    if (reorderParents != null) {
+      for (final parent in reorderParents!) {
```
luanpotter: I dislike the `!` :P 
could this be folded into `parent in (reorderParents ?? [])` or do you have performance concerns, in which case I am ok with `final reorderParents = this.reorderParents;`

packages/flame/lib/src/components/core/component_tree_root.dart
```diff
@@ -151,7 +157,8 @@ class ComponentTreeRoot extends Component {
       for (final event in queue) {
         final child = event.child!;
         final parent = event.parent!;
-        if (_blocked.contains(child) || _blocked.contains(parent)) {
+        if (_blocked.isNotEmpty &&
```
luanpotter: is the isNotEmpty saving any time here, the contains should be O(0) in that case?

packages/flame/lib/src/components/core/component_tree_root.dart
```diff
@@ -137,10 +137,16 @@ class ComponentTreeRoot extends Component {
   }
 
   void processLifecycleEvents() {
+    if (!hasLifecycleEvents) {
+      // The completer is only ever created while events are queued, so there
+      // is nothing to complete here either.
+      assert(_lifecycleEventsCompleter == null);
```
luanpotter: does dart support assertion messages so the comment could be folded into one? don't want to add an assert that later blows up with no explanation for us

packages/flame/lib/src/components/core/component_tree_root.dart
spydon: Removed by adding `reorderParents ?? const <Component>{})` instead of `?? []` since that would need to do allocation.

packages/flame/lib/src/components/core/component_tree_root.dart
spydon: It saves a little bit since it has to calculate the `identityHashCode` in `contains`, meanwhile `isNotEmpty` is virtually free.

packages/flame/lib/src/components/core/component_tree_root.dart
spydon: True, fixed.

packages/flame/lib/src/components/core/component_tree_root.dart
luanpotter: is this because `_blocked` is almost always empty? that doesn't appear to necessarily be the case here.

I honestly don't see how a hash check could be any sort of bottleneck, in fact if it were wouldn't the set contains implementation do this? according to clanker the dart identityHashCode implementation is cached per object and saved on the 64-bit pointer header.
I am ok with keeping if we have evidence but I am in favour of having commented out justifications for non-obvious micro-optimizations such as this. as is this looks like a silly mistake. a simple one line comment justifying the cost and why this is such a hot path would make it more clear.

thinking more broadly I am now wondering if this repeat with blocked thing is not ideal to begin with, there might be a way to sort the events in a deterministically correct order, but that is out of scope, just a thought.

luanpotter (approved):
LGTM, I am not convinced of the isNotEmpty skip but leave it to you

packages/flame/lib/src/components/core/component_tree_root.dart
```diff
@@ -151,7 +159,8 @@ class ComponentTreeRoot extends Component {
       for (final event in queue) {
         final child = event.child!;
         final parent = event.parent!;
-        if (_blocked.contains(child) || _blocked.contains(parent)) {
+        if (_blocked.isNotEmpty &&
+            (_blocked.contains(child) || _blocked.contains(parent))) {
```
spydon:
```suggestion
        if (_blocked.contains(child) || _blocked.contains(parent)) {
```

PR #3979, opened 2026-08-05
perf: Cache camera render closures instead of allocating per frame

spydon:
# Description

`CameraComponent.render` declared a local `renderWorld` function and, on the post-processing path, an inline context-setter closure. Both were allocated on every frame for every camera. This PR turns them into the instance methods `_renderWorld` and `_updatePostProcessContext`, so the render pass no longer allocates closures per camera per frame and the tear-offs passed to `PostProcess.render` stay stable.

There is no behavior change, only where the functions live.

## Checklist

## Breaking Change?

## Related Issues


packages/flame/lib/src/components/router/route.dart
```diff
@@ -162,10 +162,14 @@ class Route extends PositionComponent
...
 
+  /// Cached `super.renderTree` tear-off, so that the render pass does not
+  /// allocate a fresh closure for [Decorator.applyChain] on every frame.
+  void Function(Canvas)? _superRenderTree;
+
   @override
   void renderTree(Canvas canvas) {
     if (isRendered) {
-      _renderEffect.applyChain(super.renderTree, canvas);
+      _renderEffect.applyChain(_superRenderTree ??= super.renderTree, canvas);
```
luanpotter: it is quite unfortunate that the compiler can't optimize this

luanpotter: an alternative to consider, make applyChain receive a generic component or superclass with renderTree and have it call the function directly, avoiding the lambda entirely

packages/flame/lib/src/components/position_component.dart
```diff
@@ -518,9 +518,13 @@ class PositionComponent extends Component
...
   }
 
+  /// Cached `super.renderTree` tear-off, so that the render pass does not
+  /// allocate a fresh closure for [Decorator.applyChain] on every frame.
+  void Function(Canvas)? _superRenderTree;
+
   @override
   void renderTree(Canvas canvas) {
-    decorator.applyChain(super.renderTree, canvas);
+    decorator.applyChain(_superRenderTree ??= super.renderTree, canvas);
```
luanpotter: curious why all the others can't use the has decorator mixin? is there a way to share this logic upstream?

packages/flame/lib/src/post_process/post_process_component.dart
```diff
@@ -105,22 +105,25 @@ class PostProcessComponent<T extends PostProcess> extends PositionComponent {
...
+    decorator.applyChain(_renderChain ??= _buildRenderChain(), canvas);
+  }
+
+  void Function(Canvas) _buildRenderChain() {
+    final renderTree = super.renderTreeWithoutDecorator;
+    void updateContext(PostProcess? context) {
+      _renderContext.postProcess = postProcess;
+    }
+
+    return (canvas) {
```
luanpotter: you don't want to tear-off-ify this lambda too?

packages/flame/lib/src/rendering/decorator.dart
```diff
@@ -31,16 +31,26 @@ class Decorator {
...
-    apply(
-      _next == null
-          ? draw
-          : (nextCanvas) => _next!.applyChain(draw, nextCanvas),
-      canvas,
-    );
+    if (_next == null) {
+      apply(draw, canvas);
+    } else {
+      if (!identical(_chainedDrawSource, draw)) {
```
luanpotter: the need for this makes me question if the decorator pattern is the best one for the dart language.
the other tear-offs are ok but this is adding many more layers of complexity, state management and potential for mistakes.
but since this is a hot path for all render calls of all components (regardless of having decorators), this is ok. still trying to think more holistically in the meanwhile

luanpotter (approved):
LGTM

PR #3980, opened 2026-08-05
perf: Make render contexts and debug caches lazily allocated

spydon:
# Description

Every `Component` eagerly allocated a `QueueList` for render contexts plus two debug-paint `ValueCache`s. The context stack is now a lazily created plain list (most components never provide or receive a render context) and the debug caches are `late final`, so plain components allocate none of them.

Extracted from #3960 so the data-structure change there stands alone (as requested in [this comment](https://github.com/flame-engine/flame/issues/3957#issuecomment-5066267594)). Stacked on #3979.

## Checklist

## Breaking Change?

## Related Issues

Relates to #3957

### DIFF
--- a/packages/flame/lib/src/components/core/component.dart
+++ b/packages/flame/lib/src/components/core/component.dart
@@ -598,19 +598,16 @@ class Component {
   /// cleans them up afterwards.
   @protected
   void renderChild(Canvas canvas, Component child) {
-    int? originalLength;
-    final hasContext = _renderContexts.isNotEmpty;
-    if (hasContext) {
-      originalLength = child._renderContexts.length;
-      child._renderContexts.addAll(_renderContexts);
+    final contexts = _renderContexts;
+    if (contexts == null || contexts.isEmpty) {
+      child.renderTree(canvas);
+      return;
     }
+    final childContexts = child._renderContexts ??= [];
+    final originalLength = childContexts.length;
+    childContexts.addAll(contexts);
     child.renderTree(canvas);
-    if (hasContext) {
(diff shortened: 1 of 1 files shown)


packages/flame/lib/src/components/core/component.dart
```diff
@@ -612,7 +609,7 @@ class Component {
   void renderTree(Canvas canvas) {
     final context = renderContext;
     if (context != null) {
-      _renderContexts.add(context);
+      (_renderContexts ??= []).add(context);
```
luanpotter: can we do `final renderContexts = _renderContexts ??= [];` at the top to avoid the `!` later?

luanpotter: also wanna do the `const []` trick?

packages/flame/lib/src/components/core/component.dart
```diff
@@ -587,19 +587,16 @@ class Component {
   /// cleans them up afterwards.
   @protected
   void renderChild(Canvas canvas, Component child) {
-    int? originalLength;
-    final hasContext = _renderContexts.isNotEmpty;
-    if (hasContext) {
-      originalLength = child._renderContexts.length;
-      child._renderContexts.addAll(_renderContexts);
+    final contexts = _renderContexts;
```
luanpotter: could also be the same
```
final renderContexts = _renderContexts ?? const [];
```
so you don't have to null check or fallback below?

luanpotter (approved):
Makes sense to me, just a couple comments about structure and avoiding `!`

PR #3981, opened 2026-08-05
perf: Replace generator-based tree teardown and propagation with an explicit collection pass

spydon:
# Description

The removal teardown and `propagateToChildren` walked the subtree through the recursive `descendants` sync* generator, allocating generator frames per tree level on every traversal. The teardown now collects the subtree into a local buffer (same leaves-first order) via `_collectDescendants` and iterates that snapshot, since `onRemove` callbacks may mutate the tree mid-walk. `propagateToChildren` instead walks the tree with direct recursion and unwinds as soon as a handler stops propagation, so it neither allocates a buffer nor visits more components than the lazy generator did. Event delivery (`deliverToComponents`) is routed through `propagateToChildren`, so tap, drag, and keyboard propagation benefit as well.

The public `descendants()` method keeps its documented lazy semantics, since user code relies on early stopping and live iteration there.

Extracted from #3960 so the data-structure change there stands alone (as requested in [this comment](https://github.com/flame-engine/flame/issues/3957#issuecomment-5066267594)).

## Checklist

## Breaking Change?

## Related Issues

Relates to #3957

### DIFF
--- a/packages/flame/lib/src/components/core/component.dart
+++ b/packages/flame/lib/src/components/core/component.dart
@@ -430,10 +430,21 @@ class Component {
     bool Function(T) handler, {
     bool includeSelf = false,
   }) {
-    return descendants(
-      reversed: true,
-      includeSelf: includeSelf,
-    ).whereType<T>().every(handler);
+    final children = _children;
+    if (children != null) {
+      for (final child in children.reversed()) {
+        if (!child.propagateToChildren(handler, includeSelf: true)) {
+          return false;
+        }
+      }
+    }
+    if (includeSelf) {
+      final self = this;
+      if (self is T && !handler(self)) {
+        return false;
+      }
+    }
+    return true;
   }
 
   @internal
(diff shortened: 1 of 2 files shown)


packages/flame/lib/src/components/core/component.dart
```diff
@@ -1117,26 +1117,45 @@ class Component {
...
-          .._clearRemovingBit()
-          .._setRemovedBit()
-          .._removeCompleter?.complete()
-          .._removeCompleter = null
-          .._parent!.onChildrenChanged(component, ChildrenChangeType.removed);
-        return true;
-      },
-      includeSelf: true,
-    );
+    final buffer = _teardownBuffer.isEmpty ? _teardownBuffer : <Component>[];
```
luanpotter: is allocating one empty list per component removal that significant? feels like this static cache is introducing potential future concurrency problems or hard to find headaches. seems that major gain here is the non-generator teardown and not the brittle `[]` cache?

packages/flame/lib/src/components/core/component.dart
```diff
@@ -1117,26 +1117,45 @@ class Component {
...
+    _collectTeardown(buffer);
+    for (var i = 0; i < buffer.length; i++) {
```
luanpotter: please don't say that dart's built-in for each has some hidden performance cost too 😵‍💫

spydon: It does, you've complained about this before 😁

luanpotter: I refuse to believe that

packages/flame/lib/src/components/core/component.dart
```diff
@@ -1183,24 +1183,38 @@ class Component {
...
+    }
     _parent = null;
   }
 
+  /// Collects this component and all its descendants into [out] in teardown
+  /// order: leaves first, siblings in reverse order, ancestors after their
+  /// subtrees. This matches the order that
+  /// `descendants(reversed: true, includeSelf: true)` would produce, without
+  /// allocating generator frames on every removal.
+  void _collectTeardown(List<Component> out) {
```
luanpotter: wondering if descendants should just use this implementation, then they 1 stay in sync and 2 get the perf benefit?

luanpotter (approved):
LGTM, but it is unfortunate that now we have 3 shapes of traversal (before we had 1)

* `descendants` - still kept, lazy using generators, can stop at any time, pays (apparently) huge cost because generators suck?
* `propagateToChildren` - fastest and can stop, but does not collect list
* `_collectDescendants` - fast and creates list, but eager

And we are saying that `_remove` _cannot_ use `propagateToChildren` because of concurrent modifications, but it is not clear to me when that happens.

PR #3982, opened 2026-08-05
perf: Track mounted pointer-event handlers so hit tests can early-out

spydon:
# Description

The root game now counts mounted components that can receive pointer events (`TapCallbacks`, `DragCallbacks`, `DoubleTapCallbacks`, `ScaleCallbacks`, `SecondaryTapCallbacks`), so `containsEventHandlerAt`, which Flutter calls on every hit test, returns immediately without walking the component tree for games without any pointer-event handlers.

Extracted from #3960 so the data-structure change there stands alone (as requested in [this comment](https://github.com/flame-engine/flame/issues/3957#issuecomment-5066267594)). Stacked on #3981.

## Checklist

## Breaking Change?

## Related Issues

Relates to #3957

### DIFF
--- a/packages/flame/lib/src/events/callbacks/double_tap_callbacks.dart
+++ b/packages/flame/lib/src/events/callbacks/double_tap_callbacks.dart
@@ -1,5 +1,6 @@
 import 'package:flame/components.dart';
 import 'package:flame/events.dart';
+import 'package:meta/meta.dart';
 
 /// [DoubleTapCallbacks] adds the ability to receive double-tap events in a
 /// component.
@@ -25,8 +26,17 @@ mixin DoubleTapCallbacks on Component implements PointerInputCallbacks {
   void onDoubleTapCancel(DoubleTapCancelEvent event) {}
 
   @override
+  @mustCallSuper
   void onMount() {
     super.onMount();
     DoubleTapDispatcher.addDispatcher(this);
+    findRootGame()?.adjustPointerEventHandlerCount(1);
+  }
+
+  @override
+  @mustCallSuper
+  void onRemove() {
+    findRootGame()?.adjustPointerEventHandlerCount(-1);
+    super.onRemove();
   }
 }
--- a/packages/flame/lib/src/game/flame_game.dart
+++ b/packages/flame/lib/src/game/flame_game.dart
@@ -243,12 +243,29 @@ class FlameGame<W extends World> extends ComponentTreeRoot
         point.y < canvasSize.y;
   }
 
+  /// The number of currently mounted components (in this game or any nested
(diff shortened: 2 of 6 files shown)


packages/flame/lib/src/game/flame_game.dart
```diff
@@ -243,6 +243,20 @@ class FlameGame<W extends World> extends ComponentTreeRoot
...
+  /// mixins ([TapCallbacks], [DragCallbacks], [DoubleTapCallbacks],
+  /// [ScaleCallbacks], [SecondaryTapCallbacks]) on the root game, so that
+  /// [containsEventHandlerAt] can skip hit testing entirely for games
+  /// without any pointer-event handlers.
+  int _pointerEventHandlerCount = 0;
+
+  @internal
+  void adjustPointerEventHandlerCount(int delta) {
+    _pointerEventHandlerCount += delta;
+    assert(_pointerEventHandlerCount >= 0);
```
luanpotter, line 257: it might not be very actionable but can we still add a message

packages/flame/lib/src/events/callbacks/double_tap_callbacks.dart
```diff
@@ -25,8 +26,17 @@ mixin DoubleTapCallbacks on Component {
   void onDoubleTapCancel(DoubleTapCancelEvent event) {}
 
   @override
+  @mustCallSuper
   void onMount() {
     super.onMount();
     DoubleTapDispatcher.addDispatcher(this);
+    findRootGame()?.adjustPointerEventHandlerCount(1);
```
luanpotter, line 33: is the extra recursive `findRootGame()` on every mount/unmount worth the optmization?

luanpotter (commented):
I am again concerned with the complexity, ad-hoc-ness, and brittleness of these static/global caches and counters, that could lead to impossible to find bugs of undelivered events due to de-syncs. what if a component is removed w/o being mounted due to some edge-case, and then suddenly events stop working reliably.

I am again tempted to think more holistically, do we even need this, can't games just always return true on hitTestSelf within bounds? do we need to make this distinction of games without tap components being touch transparent, I didn't even know that was the case and doesn't feel intuitive. we also used to have and require the game-level mixins, which might be a good alternative, yes a bit more work for the user but more explicit and clear, less error-prone. or just an overridable config to make the game transparent (default false).

is there really an use case for a game that dynamically changes its tap transparency based on current components that we need to bake into the engine?

this also mixes concerns about event delivery even more into the game class instead of the dispatchers, but I am ok with that part as I might be refactoring it again once I remove all the old detectors. right now it is quite the mess already imo.

spydon:
Closed in favor of an upcoming PR from Luan

PR #3983, opened 2026-08-05
perf: Insertion-sort the sweep broadphase and swap-remove its active list

spydon:
# Description

`Sweep.update` re-sorted its items with a full `List.sort` and a closure comparator every tick, although the list is nearly sorted between ticks (hitboxes only move a little per frame); an insertion sort makes that near-linear. `Sweep.query` also pruned its active list with a searching `List.remove`; since active-list order does not matter, that is now an O(1) swap-remove.

Extracted from #3960 so the data-structure change there stands alone (as requested in [this comment](https://github.com/flame-engine/flame/issues/3957#issuecomment-5066267594)). Stacked on #3982.

## Checklist

## Breaking Change?

## Related Issues

Relates to #3957

### DIFF
--- a/packages/flame/lib/src/collisions/broadphase/sweep/sweep.dart
+++ b/packages/flame/lib/src/collisions/broadphase/sweep/sweep.dart
@@ -17,7 +17,21 @@ class Sweep<T extends Hitbox<T>> extends Broadphase<T> {
 
   @override
   void update() {
-    items.sort((a, b) => a.aabb.min.x.compareTo(b.aabb.min.x));
+    // Between two ticks the hitboxes only move a little, so [items] is
+    // always nearly sorted: an insertion sort runs in close to linear time
+    // here, where a general-purpose sort would pay its full O(n log n) on
+    // every tick. This also avoids allocating a comparator closure per tick.
+    final items = this.items;
+    for (var i = 1; i < items.length; i++) {
+      final item = items[i];
+      final minX = item.aabb.min.x;
+      var previousIndex = i - 1;
+      while (previousIndex >= 0 && items[previousIndex].aabb.min.x > minX) {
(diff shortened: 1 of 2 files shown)


packages/flame/lib/src/collisions/broadphase/sweep/sweep.dart
```diff
@@ -17,7 +17,21 @@ class Sweep<T extends Hitbox<T>> extends Broadphase<T> {
 
   @override
   void update() {
-    items.sort((a, b) => a.aabb.min.x.compareTo(b.aabb.min.x));
+    // Between two ticks the hitboxes only move a little, so [items] is
+    // always nearly sorted: an insertion sort runs in close to linear time
+    // here, where a general-purpose sort would pay its full O(n log n) on
+    // every tick. This also avoids allocating a comparator closure per tick.
```
luanpotter: I'd much rather we extracted it to a commons sort util, I know we'd still get the tear-off, but the main benefit is the sort algorithm change anyway? maybe we can use the prefer-inline pragma? or how about creating our own simple data structure class instead of List that wraps the sort and the remove, the comparator can be set at list creator level, no per-tick tear-off

luanpotter: honestly compared to the actual sorting operation (even in the linear case) the comparator tear-off (once per tick) does not feel significant

luanpotter: actually I think the tearoff trick from your other PRs will pair really nicely hear, you don't need the closure at all!

like int compareAabbs(aabb1, aabb2) => ...
and whateverSort(list, compareAabbs)

luanpotter (approved):
Perfect

PR #3988, opened 2026-08-07
chore: Update example, test, and other internal uses to HasGameRef

luanpotter:
# Description

We decided to go with the `Ref` nomenclature, so this updates example, test, and other internal uses to `HasGameRef` from `HasGameReference`, also flipping the deprecated flag.

No API change yet.

## Checklist

## Breaking Change?


spydon (approved):
Lgtm!
There was some talk about also renaming `game` to something more in the lines of `gameRef`, iirc @filiph had a good argument for that, but I don't remember what it was. 😅

filiph:
I don't remember that conversation but I think I can reconstruct my rationale. There's a much larger chance that `game` will conflict with (shadow or be shadowed) an existing field or local variable than if you choose `gameRef`.

`FlameGame` can be "just" the renderer/engine while the "game" (as in, chess game, for example) can be a separate object.

luanpotter:
I honestly do like `gameRef` more because for me it clarifies "ownership"; the flame game owns the components, the components just have a "reference" to the parents/game. I know these are just pointers and there is no such concern wrt memory management in Dart, but helps with my own mental model.

The shadowing argument is a good one too, I can definitely see that happening. Since we are now willing to break change, should we consider bringing it back to `HasGameRef` / `gameRef`? I know @spydon you don't like abbreviations but I think ref is pretty clear, widely used, and `gameReference` would be a mouthful. Could also explore other names.

wdyt?

spydon:
> I honestly do like `gameRef` more because for me it clarifies "ownership"; the flame game owns the components, the components just have a "reference" to the parents/game. I know these are just pointers and there is no such concern wrt memory management in Dart, but helps with my own mental model.
> 
> The shadowing argument is a good one too, I can definitely see that happening. Since we are now willing to break change, should we consider bringing it back to `HasGameRef` / `gameRef`? I know @spydon you don't like abbreviations but I think ref is pretty clear, widely used, and `gameReference` would be a mouthful. Could also explore other names.

Yeah, I think that one is okay. The reason we created the new one from the beginning was because it worked a little bit differently from `HasGameRef`, but they have both worked the same way for probably 2 years now. :smile:

luanpotter:
I will do the switch!

packages/flame/test/experimental/has_world_test.dart
```diff
@@ -70,7 +70,7 @@ void main() {
       expect(c.world.calledFoo, isTrue);
     });
 
-    testWithGame<_MyGame>('gameRef can be mocked', _MyGame.new, (game) async {
+    testWithGame<_MyGame>('game can be mocked', _MyGame.new, (game) async {
```
luanpotter, line 73: nice catch?

packages/flame/test/components/mixins/has_world_test.dart
```diff
@@ -20,7 +20,7 @@ class _ParentComponent extends Component {
 class _ChildComponent extends _ParentComponent with HasWorldReference<World> {}
 
 void main() {
-  group('HasGameRef', () {
+  group('HasWorldReference', () {
```
luanpotter, line 23: we will want to rename this later @spydon

packages/flame/test/components/mixins/has_game_ref_test.dart
```diff
@@ -1,30 +1,92 @@
-// ignore_for_file: deprecated_member_use_from_same_package
-
 import 'package:flame/components.dart';
```
luanpotter, line 1: of note

spydon (approved):
Lgtm!

luanpotter:
Updated the PR and description to go the other - ptal.

PR #3994, opened 2026-08-12
fix: Recognize every input callbacks mixin when hit testing

luanpotter:
# Description

Follow up from https://github.com/flame-engine/flame/pull/3986#discussion_r3733520433

`FlameGame.containsEventHandlerAt` backs `GameRenderBox.hitTestSelf`, which decides (for now) whether the game takes part in the hit test when the `GameWidget` uses a `HitTestBehavior` other than `opaque`. It enumerated some callbacks mixins and was missing five others: `LongPressCallbacks`, `TertiaryTapCallbacks`, `ScrollCallbacks`, `PointerMoveCallbacks` and `HoverCallbacks`.

Before, components using any of those would silently received no events under `deferToChild` or `translucent`.

Rather than extending the list and the problem, this adds two marker interfaces:

- `InputCallbacks`: anchors the whole family, much like `Event` does for the event classes.
- `PointerInputCallbacks implements InputCallbacks`: the positional/pointer subset, i.e. the mixins whose events carry a position and can therefore take part in hit testing.

This serves this and other future purposes, and is public API for users to access and write generic code on top.

I will still explore (after this) following the direction I proposed in #3982: `FlameGame` stops overriding `containsEventHandlerAt` altogether - the game reports a hit anywhere within its bounds, and per-component transparency becomes an explicit opt-in user override written against `PointerInputCallbacks` using this interface:

```dart
@override
bool containsEventHandlerAt(Vector2 position) =>
    componentsAtPoint(position).any((c) => c is PointerInputCallbacks);
```

Regardless, this is a great shape to have the callbacks in, so I am extracting this first.

## Checklist

## Breaking Change?


packages/flame/lib/src/game/flame_game.dart
```diff
@@ -247,18 +247,16 @@ class FlameGame<W extends World> extends ComponentTreeRoot
...
         this is MultiTouchDragDetector) {
       return true;
     }
     for (final component in super.componentsAtPoint(position)) {
-      if (component is TapCallbacks ||
-          component is DragCallbacks ||
-          component is DoubleTapCallbacks ||
-          component is ScaleCallbacks ||
-          component is SecondaryTapCallbacks) {
+      if (component is PointerInputCallbacks) {
```
luanpotter, line 257: this is the fix

packages/flame/lib/src/events/callbacks/pointer_input_callbacks.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:flame/src/events/callbacks/input_callbacks.dart';
+
+/// Marker interface implemented by every input callbacks mixin whose events
+/// carry a position (taps, drags, scrolls, hover) and thus participate in
+/// hit-testing.
```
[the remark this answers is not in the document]
luanpotter: it is an example list, not exhaustive

packages/flame/lib/src/events/callbacks/input_callbacks.dart
```diff
@@ -0,0 +1,4 @@
+/// Marker interface implemented by every input callbacks mixin.
+///
+/// See `PointerInputCallbacks` for the positional subset.
+abstract interface class InputCallbacks {}
```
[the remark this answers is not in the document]
luanpotter: actually it is not imported, so this is right

erickzanardo, line 4: Shouldn't this and the other interface define methods?

erickzanardo (approved):
Left a question, but LGTM either way

packages/flame/lib/src/events/callbacks/input_callbacks.dart
luanpotter: they don't actually share any methods or fields atm - this is just a marker interface for type-checking purposes

PR #3998, opened 2026-08-14
fix!: Wait for the whole component tree to load before starting the game

### LINKED ISSUE #3997: The whole component tree should be waited for in the initial `onLoad` in the `FlameGame`.

spydon:
# Description

Previously the `GameWidget` only awaited the game's own `onLoad` before removing the loading widget and starting the game. Components added during the initial load (the world, the camera, and anything added in `onLoad`) were still loading in the lifecycle queue when the first frames ran, so a `loadingBuilder` disappeared too early and the game started half-materialized.

With this PR the `GameWidget` waits for the whole initial component tree to be loaded and mounted before the game attaches and starts ticking:

- `Game` gained a `ready()` method that the `GameWidget` awaits after `load()` and `mount()`, before the first `update`. It is a no-op by default and overridden by `FlameGame`.
- `FlameGame.ready()` no longer polls with zero-delay timers. It now drains the lifecycle queue event-driven, sleeping on a new internal `Component.loadSettled` future (which completes on load success or failure, without interfering with how load errors are reported through `Component.loaded` or the current `Zone`) and on a new internal queue mutation notification, so it also wakes up when a stuck component is removed while loading. This also makes `ready()` work under `testWidgets`' fake async, where zero-delay timer polling either stalls or spins.
- A component that fails to load still does not block the game start, its error reporting is unchanged.
- The `loaderFuture` in `GameWidget` now stops early if the widget is disposed or the game instance is swapped out while the game is still loading, instead of continuing to mount components into a disposed game.
- `dequeueAdd`/`dequeueRemove` now traverse the queue with `forEachWhere` instead of the shared iterator, since they can be called from user code that runs while `processLifecycleEvents` is iterating (the queue only supports one iteration at a time).
- `flame_test`'s `initializeGame` and the benchmark helper boot games the same way, so tests and benchmarks match the production startup path.

Note that a component whose initial `onLoad` never completes will now keep the game on the loading widget instead of starting without it, which is the intent of this change. Removing such a component unblocks the start.

## Checklist

## Breaking Change?

### Migration instructions

The game no longer starts while components added during the initial load are still loading:

- If you relied on the game rendering and updating while slow components were still loading (for example an in-game loading indicator implemented as a component), move that indicator to the `GameWidget.loadingBuilder`, or add the slow components after the initial load instead of during `onLoad`.
- A component whose `onLoad` awaits something that only happens once the game is running (an update tick, an overlay interaction) will now keep the game on the loading widget forever; restructure such loads so they complete on their own.
- Widget tests that pump a `GameWidget` may need an extra `tester.pump()` before the game is attached, since the loader now also waits for the component tree.
- If your `Game` subclass defines its own `ready()` member, it now overrides `Game.ready()` and is awaited by the `GameWidget` before the game starts.

## Related Issues

Closes #3997


packages/flame_test/lib/src/test_flame_game.dart
```diff
@@ -99,6 +99,8 @@ Future<T> initializeGame<T extends FlameGame>(CreateFunction<T> create) async {
   await game.load();
   // ignore: invalid_use_of_internal_member
   game.mount();
+  // The same startup sequence as the GameWidget uses; see [FlameGame.ready].
+  await game.ready();
```
luanpotter, line 103: does this mean we can now remove a lot of ad-hoc code from many tests?
or what is the consequence of having this now?

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -216,21 +216,67 @@ class FlameGame<W extends World> extends ComponentTreeRoot
...
+  /// The `GameWidget` awaits this future when the game is first shown, so
+  /// that the game only starts, and the loading widget is only removed, once
+  /// the whole initial component tree has been loaded and mounted.
+  ///
+  /// A component that fails to load does not block this future; its error is
+  /// reported through its [Component.loaded] future, or the current [Zone] if
+  /// nothing is awaiting that future.
+  ///
+  /// Warning: awaiting on a game that was not fully connected will result in
+  /// an infinite loop. For example, this could occur if you run `x.add(y)` but
```
luanpotter: what are x and y here, is x the game, or x and y are just two random components that hadn't been added to the game?

erickzanardo (approved):

packages/flame/lib/src/components/core/component_tree_root.dart
```diff
@@ -22,12 +22,29 @@ class ComponentTreeRoot extends Component {
...
+  /// changed by something other than a component finishing its load, for
+  /// example when a component is removed while it is still loading.
+  @internal
+  Future<void> get nextLifecycleEventMutation =>
+      (_lifecycleEventMutationCompleter ??= Completer<void>()).future;
+
+  void _notifyLifecycleEventMutation() {
+    _lifecycleEventMutationCompleter?.complete();
+    _lifecycleEventMutationCompleter = null;
+  }
```
luanpotter, line 56: this is a slightly non-intuitive pattern for a "reusable completer" that `_lifecycleEventsCompleter` has exactly the same but not even extracted to functions - wondering if we could add class to encapsulate it?

packages/flame/lib/src/components/core/recycled_queue.dart
```diff
@@ -209,6 +209,28 @@ class RecycledQueue<T extends Disposable> extends Iterable<T>
     }
   }
 
+  /// Returns the first element matching [test], or null if there is none,
+  /// by directly traversing the internal storage. Unlike iteration, this can
+  /// be safely called while another iteration is in progress.
```
luanpotter: why iterating the same queue twice cause issues? each should create its own iterator no?
for example, lower below the queue is iterated on the game-level await to check for loading children, which surely happens at the same time as the actual queue consumer:
```
for (final event in queue) {
```

packages/flame/lib/src/game/flame_game.dart
```diff
@@ -216,21 +216,73 @@ class FlameGame<W extends World> extends ComponentTreeRoot
...
+      }
+      if (hasLoadingChildren) {
+        // Sleep until a load settles, or until the event queue is changed
+        // from the outside, for example by a component being removed while
+        // it is still loading.
+        await Future.any([wake.future, nextLifecycleEventMutation]);
+      } else {
+        // The queue is blocked on something other than loading, give other
+        // futures a chance to execute and try again.
+        await Future<void>.delayed(Duration.zero);
```
luanpotter: is this some expected kind of "other than loading" things, or the unexpected "spanish inquisition" kind?

luanpotter (approved):
LGTM!

packages/flame/lib/src/game/flame_game.dart
spydon: Did this in a bit of a cleaner way, and updated the comment.

PR #4003, opened 2026-08-16
feat!: Return a list instead of a set from the collision detection methods

### LINKED ISSUE #4001: Return a list instead of a set from the collision detection methods.

spydon:
# Description

The collision detection system previously passed the intersection points around as a
`Set<Vector2>`, which meant that every collision check in the hot loop had to hash
`Vector2`s and allocate hash sets. This PR changes all of these APIs to use
`List<Vector2>` instead.

The returned points are still guaranteed to be unique: the polygon and circle-polygon
intersection accumulation now deduplicates with a linear `contains` check (the point
lists are tiny, so this is cheaper than hashing), and tangent circles now explicitly
return a single point instead of relying on the set to collapse the two identical
solutions.

Changed APIs (all from `Set<Vector2>` to `List<Vector2>`):

- `CollisionCallbacks.onCollision`/`onCollisionStart` and the corresponding methods on
  `GenericCollisionCallbacks`, `ShapeHitbox` and `CollisionPassthrough`, as well as the
  `CollisionCallback` typedef.
- `CollisionDetection.intersections`, `handleCollisionStart` and `handleCollision`
  (and the `StandardCollisionDetection` overrides).
- `Hitbox.intersections` and `ShapeHitbox.intersections`.
- `Intersections.intersect`/`unorderedIntersect` and the top-level `intersections`
  function in the geometry package.
- The experimental `Rectangle.intersections`.

Benchmarked the accumulation strategies against each other with the real intersection
workloads (200k iterations per case): the list version is roughly 2x faster for
circle-circle intersections, faster for overlapping axis-aligned rectangles (which
produce collinear duplicate candidates), and equal within noise for tilted polygons.

`flame_behaviors` and all examples, docs and tutorials have been updated accordingly.

Migration: change the `Set<Vector2> intersectionPoints` parameter type to
`List<Vector2>` in `onCollision`/`onCollisionStart` overrides (and in custom
`CollisionDetection`/`Hitbox` implementations).

## Checklist

## Breaking Change?

### Migration instructions

Replace `Set<Vector2>` with `List<Vector2>` in your `onCollision` and `onCollisionStart`
overrides and collision callbacks:

```dart
@override
void onCollision(List<Vector2> intersectionPoints, PositionComponent other) {
  // ...
}
```

The same applies to custom `CollisionDetection`, `Hitbox` and `Intersections`
implementations, which now return `List<Vector2>` from their intersection methods.
The points in the list are still unique.

## Related Issues

Closes #4001

### DIFF
--- a/doc/bridge_packages/flame_behaviors/collision-detection.md
+++ b/doc/bridge_packages/flame_behaviors/collision-detection.md
@@ -33,7 +33,7 @@ class MyEntityCollisionBehavior
     extends CollisionBehavior<MyCollidingEntity, MyParentEntity> {
   @override
   void onCollisionStart(
-    Set<Vector2> intersectionPoints,
+    List<Vector2> intersectionPoints,
     MyCollidingEntity other,
   ) {
     // We are starting colliding with MyCollidingEntity
--- a/doc/flame/collision_detection.md
+++ b/doc/flame/collision_detection.md
@@ -81,7 +81,7 @@ Example:
 ```dart
 class MyCollidable extends PositionComponent with CollisionCallbacks {
   @override
-  void onCollision(Set<Vector2> points, PositionComponent other) {
+  void onCollision(List<Vector2> points, PositionComponent other) {
     if (other is ScreenHitbox) {
       //...
     } else if (other is YourOtherComponent) {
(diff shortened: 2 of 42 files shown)


luanpotter (approved):
I am not happy, but I accept

PR #4004, opened 2026-08-16
fix!: Remove middle intersection point on shared line segments in collision detection

### LINKED ISSUE #4002: Remove middle intersection point on line for collision detection.

spydon:
# Description

When two line segments were collinear and overlapping, `LineSegment.intersections` returned a single synthetic point: the average of the overlap's end points. For shape collision detection this meant that two shapes sharing an edge (for example two axis-aligned rectangles standing side by side) reported both the real corner points and an extra point in the middle of the shared edge.

This PR changes the collinear-overlap case of `LineSegment.intersections` to return the actual end points of the overlapping section instead of their average. Segments that only touch in a single point still return that one point. As a result, collision detection on shared edges now only reports the end points of the shared segment, without the synthetic middle point.

The dartdoc on `LineSegment.intersections` and `PolygonPolygonIntersections.intersect` has been updated to describe the new behavior, and the affected tests now assert the overlap end points.

## Checklist

## Breaking Change?

### Migration instructions

`LineSegment.intersections` now returns the end points of the overlapping section when the two segments are collinear and overlap, instead of a single averaged point in the middle of the overlap. Consequently, `Hitbox.intersections` and the top-level `geometry.intersections` no longer report a middle point for shapes that share an edge, only the end points of the shared segment. If you relied on the averaged point, compute it from the returned end points, for example with `LineSegment(points.first, points.last).midpoint`.

## Related Issues

Closes #4002

### DIFF
--- a/packages/flame/lib/src/geometry/line.dart
+++ b/packages/flame/lib/src/geometry/line.dart
@@ -27,7 +27,7 @@ class Line {
     final determinant = a * otherLine.b - otherLine.a * b;
     if (determinant == 0) {
       //The lines are parallel (potentially coincides) and have no intersection
-      return [];
+      return const [];
     }
     return [
       Vector2(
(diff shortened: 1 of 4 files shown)


luanpotter (approved):
LGTM

PR #4008, opened 2026-08-16
chore: Add more scroll-to-zoom examples and a widget-level scroll test

luanpotter:
# Description

Add more scroll-to-zoom examples and a widget-level scroll test; no src changes.

`fixed_resolution_example` and `static_components_example` both used to declare the legacy `ScrollDetector` without ever overriding `onScroll`; I removed it on the previous cleanup but I think the intent was valid, so I am adding a zoom-on-scroll behaviour that was never there other than in spirit.

Also adds a widget-level scroll test missing for the callbacks; the legacy path had one but the exact shape was never added to the new system.

## Checklist

## Breaking Change?

### DIFF
--- a/packages/flame/test/events/component_mixins/scroll_callbacks_test.dart
+++ b/packages/flame/test/events/component_mixins/scroll_callbacks_test.dart
@@ -2,6 +2,7 @@ import 'package:flame/components.dart';
 import 'package:flame/events.dart';
 import 'package:flame/game.dart';
 import 'package:flame_test/flame_test.dart';
+import 'package:flutter/gestures.dart' show PointerDeviceKind;
 import 'package:flutter_test/flutter_test.dart';
 
 void main() {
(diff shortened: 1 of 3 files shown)


packages/flame/test/events/component_mixins/scroll_callbacks_test.dart
```diff
@@ -76,6 +77,25 @@ void main() {
       },
     );
 
+    testWidgets(
+      'scroll events reach the game through the widget tree',
+      (tester) async {
+        final game = _ScrollCallbacksGame();
+        await tester.pumpWidget(GameWidget(game: game));
+        await tester.pump();
+        await tester.pump();
```
luanpotter, line 86: apparently basically all tests do this nonsense, so just replicating. can investigate further later

spydon (approved):

PR #4009, opened 2026-08-16
refactor!: Make descendants return an eagerly collected snapshot list

spydon:
# Description

`descendants()` was a lazy `sync*` generator, which allocates an iterator per tree level and pays a delegation chain walk per element. After #3981 it was the last generator-shaped traversal of the component tree; this PR makes it delegate to the same `_collectDescendants` collection pass, so there is only one traversal shape left to care about.

The method now returns a `List<Component>` that is an eagerly collected snapshot of the tree at the moment of the call. This means the whole subtree is always traversed (early stopping no longer saves traversal work), and iterating while adding or removing components is safe instead of surfacing a `ConcurrentModificationError`.

## Checklist

## Breaking Change?

### Migration instructions

Most call sites need no changes, since `List` implements `Iterable`. Overrides of `descendants` must update their return type from `Iterable<Component>` to `List<Component>`. Code that relied on the lazy iteration semantics (early stopping to avoid traversing large trees, or catching `ConcurrentModificationError` on mutation during iteration) now gets a full-tree snapshot instead; query methods such as `firstWhere` still stop early, they just iterate an already collected list.

## Related Issues

Relates to #3957

### DIFF
--- a/packages/flame/lib/src/components/core/component.dart
+++ b/packages/flame/lib/src/components/core/component.dart
@@ -393,26 +393,15 @@ class Component {
   /// on every game cycle. The optional parameter [reversed] allows iterating
   /// through the same set of descendants in reverse order.
   ///
-  /// The [Iterable] produced by this method is "lazy", which means it will only
-  /// traverse the component tree when required. This allows efficient chaining
-  /// of various iterable methods, such as filtering, early stopping, folding,
-  /// and so on -- see the documentation of the [Iterable] class for details.
-  Iterable<Component> descendants({
+  /// The returned list is an eagerly collected snapshot of the tree at the
+  /// moment of the call. This means that the whole subtree is always
+  /// traversed, but also that it is safe to add or remove components while
+  /// iterating over the result.
+  List<Component> descendants({
     bool includeSelf = false,
     bool reversed = false,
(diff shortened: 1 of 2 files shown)


packages/flame/lib/src/components/core/component.dart
```diff
@@ -1194,7 +1183,7 @@ class Component {
 
   void _remove(Component parent) {
     parent._internalChildren.remove(this);
-    for (final component in _collectDescendants()) {
+    for (final component in _collectDescendants(reversed: true)) {
```
luanpotter, line 1186: why not use the public descendants here?

packages/flame/test/components/component_test.dart
```diff
@@ -1485,7 +1485,7 @@ void main() {
         },
       );
 
-      testWithFlameGame('descendants() iterator is lazy', (game) async {
+      testWithFlameGame('descendants() supports early stopping', (game) async {
```
luanpotter, line 1488: this is no longer a feature of descendants but rather just how dart streams work, this test is misleading as the list is now eagerly collected (arguably it was wrong to begin with bc even before it was not testing what it was intending to)

PR #4019, opened 2026-08-26
perf!: Replace the OrderedSet children container with a flat sorted-array ComponentList

spydon:
# Description

First half of the FCS core rewrite for #3957: replaces the `OrderedSet` children container with a Flame-owned `ComponentList`. Split out of #3960 so that the container change can be reviewed on its own; #3960 is stacked on top of this branch and adds the flattened update traversal.

## Children live in a Flame-owned `ComponentList`

The `ordered_set` dependency is gone. Children are stored in a single flat array sorted by `(priority, insertion order)`, and each component intrusively stores its container and slot index:

- `remove` and `contains` are O(1), with no hashing or tree walks; `add` is an O(1) append in the common case.
- Removals leave `null` tombstones that are compacted once per parent per tick, so removing k children among n costs O(k + n) instead of O(k*n).
- A priority change triggers at most one stable sort per parent per tick, skipped entirely when nothing actually reordered.
- The `register<T>()`/`query<T>()` cache surface is kept, and results are now always in priority order.
- `Component.childrenFactory` is replaced by an overridable `createComponentList()`, which accepts an optional `Comparator<Component>` for custom orderings such as y-sort.

The update, render, and hit-test passes iterate the backing array directly instead of going through an iterator, but the traversal itself is unchanged: `updateTree` is still recursive and virtual here. `component_tree_ordering_test.dart` pins lifecycle-event ordering, hit-test order, and equal-priority ordering across the rewrite.

Three backing designs were implemented and benchmarked before settling on this one; see [the comparison in the issue](https://github.com/flame-engine/flame/issues/3957#issuecomment-5043618243).

## Benchmarks

Measured against `main` (commit `7fd33dee9`, which already includes the merged allocation-hygiene PRs), so both tables isolate this PR's own effect. The benchmark suite is in `packages/flame/benchmark`; ms per run, lower is better.

### JIT (`flutter test`, host machine)

| Benchmark | main | this PR | speedup |
|---|---|---|---|
| Update wide tree (10k x 1) | 3.24 | 1.19 | 2.7x |
| Update nested tree (1k x 10) | 8.08 | 2.57 | 3.1x |
| Update deep tree (100 levels) | 8.46 | 3.23 | 2.6x |
| Update barrier tree (10% time-scaled) | 9.15 | 3.85 | 2.4x |
| Render wide tree (10k x 1) | 5.14 | 4.05 | 1.3x |
| Render nested tree (1k x 10) | 9.95 | 4.92 | 2.0x |
| Render deep tree (100 levels) | 12.34 | 6.56 | 1.9x |
| Lifecycle churn (100/tick, 1k pop) | 1.80 | 1.15 | 1.6x |
| Lifecycle churn (100/tick, 10k pop) | 4.86 | 2.82 | 1.7x |
| Mass add/remove (1k per cycle) | 1.15 | 0.64 | 1.8x |
| Priority change (1 child per parent) | 26.01 | 3.44 | 7.6x |
| Priority change (y-sort, all children) | 11.72 | 3.33 | 3.5x |
| Type-query churn (2 registered queries) | 3.41 | 2.83 | 1.2x |
| Game-like update pass | 98.34 | 70.09 | 1.4x |
| Hit test + delivery (cached) | 43.63 | 41.93 | flat |
| Flat collision detection | 4.95 | 2.04 | 2.4x |
| Nested collision detection | 7.91 | 2.84 | 2.8x |

### AOT (`flutter run --release`, Android arm64 emulator on an Apple Silicon host)

| Benchmark | main | container only | speedup |
|---|---|---|---|
| Update wide tree (10k x 1) | 6.25 | 4.31 | 1.5x |
| Update nested tree (1k x 10) | 9.53 | 3.91 | 2.4x |
| Update deep tree (100 levels) | 11.28 | 4.27 | 2.6x |
| Update barrier tree (10% time-scaled) | 12.35 | 6.81 | 1.8x |
| Render wide tree (10k x 1) | 6.42 | 3.97 | 1.6x |
| Render nested tree (1k x 10) | 12.44 | 4.70 | 2.6x |
| Render deep tree (100 levels) | 14.74 | 6.53 | 2.3x |
| Lifecycle churn (100/tick, 1k pop) | 1.54 | 0.90 | 1.7x |
| Lifecycle churn (100/tick, 10k pop) | 4.80 | 3.08 | 1.6x |
| Mass add/remove (1k per cycle) | 0.89 | 0.42 | 2.1x |
| Priority change (1 child per parent) | 24.67 | 3.42 | 7.2x |
| Priority change (y-sort, all children) | 11.28 | 3.30 | 3.4x |
| Type-query churn (2 registered queries) | 5.36 | 4.75 | 1.1x |
| Game-like update pass | 79.68 | 50.16 | 1.6x |
| Hit test + delivery (cached) | 43.30 | 38.48 | 1.1x |
| Flat collision detection | 4.26 | 1.47 | 2.9x |
| Nested collision detection | 7.92 | 1.95 | 4.1x |

Both tables are single runs per checkpoint (the JIT one is the `container only` column of the attribution run in #3960), so treat the exact multiples loosely. The AOT numbers come from the `crud_test` AVD (API 34, arm64-v8a, 4 cores), which executes the AOT snapshot natively on the host CPU, so they are representative of release-mode Dart rather than of any particular phone.

### Update pass with real game logic (JIT and AOT)

@luanpotter asked whether the traversal gains above still matter once components do real work, so `update_workload_benchmark.dart` reuses the wide and nested trees from the traversal suite but gives every component a bog-standard, allocation-free `update` body:

- **light**: bullet or particle style movement, `position.addScaled(velocity, dt)` plus bouncing off the world bounds.
- **heavy**: enemy AI style steering, scanning 32 moving targets for the nearest one, steering towards it with clamped acceleration, `atan2` facing, and a cooldown timer.

The no-op rows are the traversal benchmarks re-run in the same session for reference. Light rows do 500k component visits per run and heavy rows 100k (the no-op rows do 1M), so compare each row against its no-op counterpart rather than across rows. Median of 3 alternating runs, ms per run, same machines as the tables above (`main` at `b4578fc6a`).

| Benchmark | JIT main | JIT this PR | speedup | AOT main | AOT this PR | speedup |
|---|---|---|---|---|---|---|
| Wide tree (10k x 1), no-op | 3.28 | 1.21 | 2.7x | 4.51 | 2.38 | 1.9x |
| Wide tree (10k x 1), light logic | 21.20 | 19.62 | 1.1x | 12.50 | 11.22 | 1.1x |
| Wide tree (10k x 1), heavy logic | 13.41 | 13.41 | flat | 10.86 | 10.47 | flat |
| Nested tree (1k x 10), no-op | 8.57 | 2.55 | 3.4x | 8.98 | 2.66 | 3.4x |
| Nested tree (1k x 10), light logic | 22.23 | 19.58 | 1.1x | 21.41 | 17.59 | 1.2x |
| Nested tree (1k x 10), heavy logic | 15.50 | 14.64 | 1.1x | 12.57 | 11.36 | 1.1x |

Per component visit in AOT, the framework overhead on `main` is about 4.5ns in the wide tree and 9ns in the nested tree, against roughly 25-45ns for the light update body (most of which is the `NotifyingVector2` change notification that marks the transform dirty) and 110-130ns for the heavy one. So on `main` the traversal accounts for around 18% (wide) to 21% (nested) of a light-logic update pass and 4-7% of a heavy-logic one, and this PR removes most of that: the whole update pass gets 10-20% cheaper with light logic and 4-10% cheaper with heavy logic. The nested shape, which is what real games look like (entities holding a handful of child components), is where the gain is largest, and the heavy-logic rows are the floor: the container change cannot speed up the game's own code, so a game that is bound by its own `update` bodies sees single-digit percent gains from this PR on its own. The traversal flattening in #3960 stacks on top of these numbers.

## Checklist

## Breaking Change?

### Migration instructions

- `children` is now a `ComponentList` instead of an `OrderedSet`. The iterable surface, `query<T>()`, and `register<T>()` are unchanged, so most code compiles as is. The `ordered_set` dependency is gone.
- `children.reversed()` is now a getter: `children.reversed`.
- `Component.strictQueryMode` is removed. Strict mode is off by default; to enable it for a component, override `createComponentList()` to return `ComponentList(strictMode: true)`.
- `Component.childrenFactory` is removed. Override `createComponentList()` on the component instead; it accepts an optional `Comparator<Component>` for custom orderings (for example y-sort).
- Mutating `children` while iterating it now tolerates removals and tail appends; only position-shifting operations (mid-list insertion, reorder, compaction) throw `ConcurrentModificationError`.

## Related Issues

Part of #3957 (closed by #3960, which is stacked on this PR).


packages/flame/lib/src/components/core/component.dart
```diff
@@ -302,53 +302,53 @@ class Component {
   ///
   /// This makes it possible to have lighter components that don't have any
   /// children.
-  OrderedSet<Component>? _children;
+  ComponentList? _children;
+
+  /// The [ComponentList] that currently stores this component, if any.
+  /// Maintained by [ComponentList].
+  ComponentList? _containerList;
```
luanpotter, line 318: is this equivalent to `parent._children`? would be good to document and/or explain the difference

packages/flame/lib/src/components/core/component.dart
```diff
@@ -302,53 +302,53 @@ class Component {
...
-      _componentPriorityMapper,
-      strictMode: strictQueryMode,
-    );
-  };
-
-  /// Whether OrderedSet's strict mode mode should be enabled for all children
-  /// sets.
+  /// Whether the strict query mode should be enabled for all children lists;
+  /// see [ComponentList.strictMode].
   static bool strictQueryMode = false;
```
luanpotter: we can prob move the static there now that we own the ComponentList class (and we allow breaking changes)!

packages/flame/lib/src/components/core/component.dart
```diff
@@ -639,8 +643,12 @@ class Component {
     render(canvas);
     final children = _children;
     if (children != null) {
-      for (final child in children) {
-        renderChild(canvas, child);
+      final elements = children._elements;
+      for (var i = 0; i < elements.length; i++) {
```
luanpotter: for each

packages/flame/lib/src/components/core/component.dart
```diff
@@ -885,8 +893,10 @@ class Component {
   ) sync* {
     nestedContexts?.add(locationContext);
     if (_children != null) {
-      for (final child in _children!.reversed()) {
-        if (child is IgnoreEvents && child.ignoreEvents) {
+      final elements = _children!._elements;
+      for (var i = elements.length - 1; i >= 0; i--) {
```
luanpotter: is `for (element in elements.reversed())` just as cheap here?

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,529 @@
...
+
+/// A fast, ordered container for a [Component]'s children.
+///
+/// The children are stored in a single flat [List], sorted by
+/// [Component.priority] (or by a custom comparator, see
+/// [Component.createComponentList]); components that compare equal keep the
+/// order in which they were added. This is the same ordering that the old
+/// `OrderedSet`-based container provided, but backed by a contiguous array
+/// instead of a splay tree of hash sets, which makes iteration, additions,
+/// removals and reorders significantly faster and allocation-free.
```
luanpotter: this is an archeological reliquary - this should live on the changelog & migration guide, not here. new users should not have to understand past states to understand this class. comment should refer to the current state

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,529 @@
...
+///
+/// Performance characteristics:
+///  - iteration: O(n) over a contiguous array, without any allocations on the
+///    internal engine paths;
+///  - [add]: O(1) when the component sorts at or after the end, which is the
+///    common case, and O(n) for a mid-list insertion;
```
luanpotter: can we clarify the justification for common case here (ie since no priority is set)

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,529 @@
...
+/// The children are stored in a single flat [List], sorted by
+/// [Component.priority] (or by a custom comparator, see
+/// [Component.createComponentList]); components that compare equal keep the
+/// order in which they were added. This is the same ordering that the old
+/// `OrderedSet`-based container provided, but backed by a contiguous array
+/// instead of a splay tree of hash sets, which makes iteration, additions,
+/// removals and reorders significantly faster and allocation-free.
+///
+/// Performance characteristics:
+///  - iteration: O(n) over a contiguous array, without any allocations on the
```
luanpotter: this is a contiguous array of, basically, pointers right. so not truly a contiguous memory access pattern which would justify the performance gains (unlike the allocations portion which I think does)

packages/flame/lib/src/components/core/component_list.dart
luanpotter: also - we still do "add later" right so this is O(1) per add because it is amortized per tick? assuming many adds per tick, which if not the case then this barely matters anyway

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,529 @@
...
+///
+/// A removal does not shift the array, it leaves a `null` "tombstone" in
+/// place. Tombstones are invisible to iteration and are compacted away at the
+/// start of the next update tick, or earlier if they grow to dominate the
+/// array.
+///
+/// Mutating the container while iterating it is allowed in the ways that the
+/// component lifecycle needs: removals take effect immediately (the removed
+/// component is simply not visited), and additions at the end may or may not
+/// be visited by an ongoing iteration. Operations that shift the positions of
```
luanpotter: > and additions at the end may or may not be visited by an ongoing iteration

I wonder if this is something we want to codify deterministically, but seems unlikely that anyone would (should?) care

packages/flame/lib/src/components/core/component.dart
luanpotter: actually we can prob just inline as a default param now (there is no longer a package barrier)

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,529 @@
...
+  @override
+  bool get isNotEmpty => _length != 0;
+
+  @override
+  Iterator<Component> get iterator => _ComponentListIterator(this);
+
+  /// The elements of this set in reverse order.
+  ///
+  /// Unlike the rest of the [Iterable] interface, this is a method and not a
+  /// getter, for historical reasons. The returned iterable is a lazy view: it
```
luanpotter: No reason to keep any historical reasons; is a getter a better shape?

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,529 @@
...
+  @override
+  bool contains(Object? element) {
+    return element is Component && identical(element._containerList, this);
+  }
+
+  @override
+  Component get first {
+    final elements = _elements;
+    for (var i = 0; i < elements.length; i++) {
+      final element = elements[i];
```
luanpotter: for each

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,529 @@
...
+
+  /// Inserts [component] before all existing elements that sort after it,
+  /// and after all elements that sort the same or before it.
+  void _insertSorted(Component component) {
+    _compact();
+    final elements = _elements;
+    var lo = 0;
+    var hi = elements.length;
+    while (lo < hi) {
+      final mid = (lo + hi) >>> 1;
```
luanpotter: cute, but please don't tell me dart can't optimize `~/ 2` as well :P 
(every time I see your PRs is another slight disappointment with dart)

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,529 @@
...
+      if (_compareOrder(elements[mid]!, component) <= 0) {
+        lo = mid + 1;
+      } else {
+        hi = mid;
+      }
+    }
```
luanpotter: wondering if the binary search can be extracted to a generic commons function, or - if like before-, even better, there is already one

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,529 @@
...
+        break;
+      }
+    }
+    if (isSorted) {
+      return;
+    }
+    _shiftCount++;
+    // Ties are broken by the pre-sort index, which makes the (unstable)
+    // built-in sort behave as a stable sort.
+    elements.sort((a, b) {
```
luanpotter, line 277: on this deep treasure trove of micro optimizations, we were happy with dart's default (unstable) sort algo?

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,529 @@
...
+  }
+
+  /// Restores the priority ordering after one or more elements have changed
+  /// their [Component.priority].
+  ///
+  /// This is invoked automatically, at most once per parent per game tick,
+  /// when priorities of mounted components change. The sort is stable:
+  /// components with equal priority keep their relative order.
+  void rebalance() {
+    _compact();
```
luanpotter, line 259: I'd add a comment here: `// removes all tombstones making the element! access safe`

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,529 @@
...
+
+  /// All elements of type [C], in priority order, in O(1).
+  ///
+  /// The type [C] must have been [register]ed beforehand, unless [strictMode]
+  /// is false, in which case the registration happens on the first query.
+  Iterable<C> query<C extends Component>() {
+    final cache = _queries?[C];
+    if (cache == null) {
+      if (strictMode) {
+        throw StateError('Cannot query unregistered query $C');
```
luanpotter: could include a note on to why (and how to turn off)
this is not pendantic - it is a performance optmization for games to pre-register that avoids frame drops

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,529 @@
...
+    var lo = 0;
+    var hi = list.length;
+    while (lo < hi) {
+      final mid = (lo + hi) >>> 1;
+      if (list[mid]._containerIndex < index) {
+        lo = mid + 1;
+      } else {
+        hi = mid;
+      }
+    }
```
luanpotter: another binary search to unify with the other

luanpotter (commented):
This is great and _super_ clean @spydon thanks for breaking it up.

I left some comments but have 3 overarching comments:

1. as we agreed before, dart's for each (I mean the language construct- not the lambda) compiles just the same as the indexed for

2. I am considering if we can extract a mixin, base class, whatever from Component to hold the "bears children" logic - it would include all the 3 related fields (the children, the parent list ref, and the index). these are the only things `ComponentList` cares about for Component specifically (other than the default priority comparator that can move to `createComponentList`. that way `ComponentList` becomes truly component agnostic, and the component class can remove some of this related index tracking and children logic. not sure it would be as clean as I am envisioning, but worth investigating.

3. I am curious about the efficacy of our benchmarks in real life:

    > All components are plain [Component]s with no-op [Component.update] and [Component.render], so any time measured here is framework overhead rather than game logic.

    while yes that isolate engine overhead, I wonder if the magnitude is even relevant on real cases (I suspect yes, but not 100% sure). would be awesome to run the same bench with some bog-standard possible implementations of updates, like a light and a heavy one, to see how the optimizations fare into a real game.


but overall, all things being said - I am very happy with this direction :)

packages/flame/lib/src/components/core/component_list.dart
luanpotter: (realized at the end we might no longer necessarily be doing the add later - wonder if we considered whether to keep that or not)

packages/flame/lib/src/components/core/component_list.dart
spydon: I think it should be fine like it is now, it's always been like that.
We could do it in a follow-up if we find that we'd need it.

packages/flame/lib/src/components/core/component.dart
spydon: Agree, it's for-in that you mean everywhere though 😄 (since forEach is the higher order method in Dart)

spydon:
> This is great and _super_ clean @spydon thanks for breaking it up.
> 
> I left some comments but have 3 overarching comments:
> 
>     1. as we agreed before, dart's for each (I mean the language construct- not the lambda) compiles just the same as the indexed for

Yes, for-in, not for each. I'll fix that up. :)

>     2. I am considering if we can extract a mixin, base class, whatever from Component to hold the "bears children" logic - it would include all the 3 related fields (the children, the parent list ref, and the index). these are the only things `ComponentList` cares about for Component specifically (other than the default priority comparator that can move to `createComponentList`. that way `ComponentList` becomes truly component agnostic, and the component class can remove some of this related index tracking and children logic. not sure it would be as clean as I am envisioning, but worth investigating.

It sounds like overengineering to me... It's not that much overhead in the Component, and it really is the core of the Component.

>     3. I am curious about the efficacy of our benchmarks in real life:
>        > All components are plain [Component]s with no-op [Component.update] and [Component.render], so any time measured here is framework overhead rather than game logic.
>        
>        
>        while yes that isolate engine overhead, I wonder if the magnitude is even relevant on real cases (I suspect yes, but not 100% sure). would be awesome to run the same bench with some bog-standard possible implementations of updates, like a light and a heavy one, to see how the optimizations fare into a real game.

Setting up more "real life" benchmark too and adding the results to the description.

packages/flame/lib/src/components/core/component.dart
spydon: Almost, but not always. `_containerList` is the list this component is *physically stored in right now*, while `_parent` is set as soon as an add is requested. The two go out of sync in both directions:

- on `add` to a mounted parent, `_parent` is assigned immediately but the insertion into `_children` happens later, when the lifecycle queue is processed, so `_containerList` is still `null` in between;
- on removal, the component leaves the list first and `_parent` is cleared afterwards.

Keeping a direct reference (plus `_containerIndex`) is also what makes `contains` and `remove` O(1): the list checks `identical(component._containerList, this)` and nulls out `_elements[component._containerIndex]`, without going through the parent at all. I've expanded the doc comment on the field to spell this out (0728c30).

(Clanker-written, but I think it explained it well)

erickzanardo (approved):

packages/flame/test/components/component_tree_ordering_test.dart
luanpotter, line 1: this is not a golden test!

packages/flame/lib/src/components/core/component.dart
```diff
@@ -885,7 +893,7 @@ class Component {
   ) sync* {
     nestedContexts?.add(locationContext);
     if (_children != null) {
-      for (final child in _children!.reversed()) {
+      for (final child in _children!.reversed) {
```
luanpotter: (followp) we should extract a local var to avoid !

packages/flame/lib/src/components/core/component.dart
```diff
@@ -302,53 +302,56 @@ class Component {
...
+
+  /// This component's index within [_containerList]'s backing list, or -1
+  /// when the component is not in any list. Maintained by [ComponentList].
+  int _containerIndex = -1;
 
   /// This field should be used internally for functionality when you need to
   /// make sure that the component set is created if it doesn't already exist.
-  OrderedSet<Component> get _internalChildren =>
-      _children ??= createComponentSet();
+  ComponentList get _internalChildren => _children ??= createComponentList();
```
luanpotter: why do we have _children, _internalChildren AND children?

luanpotter: (it was because of the ReadOnly OS - we can remove now, just use children)

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,546 @@
...
+  }
+
+  /// Adds [component] to this list, keeping the priority ordering; returns
+  /// whether the component was added (`false` if it already was in the list).
+  ///
+  /// This is internal machinery: adding a component here does not make it go
+  /// through the component lifecycle. Use [Component.add] instead.
+  @internal
+  bool add(Component component) {
```
luanpotter: ReadOnlyComponentList and ComponentList as separate would ensure this can be proper private
(that is what we did before, but we don't need to keep it necessarily, if `@internal` does the job)

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,543 @@
...
+/// the start of the next update tick, or earlier if they grow to dominate the
+/// list.
+///
+/// Mutating the container while iterating it is allowed in the ways that the
+/// component lifecycle needs: removals take effect immediately (the removed
+/// component is simply not visited), and additions at the end may or may not
+/// be visited by an ongoing iteration. Operations that shift the positions of
+/// existing elements (a mid-list insertion, [rebalance], or tombstone
+/// compaction) invalidate live iterators, which will then throw a
+/// [ConcurrentModificationError].
```
luanpotter: q: do we forbid these operations at a public API level by exposing the "do later" versions? (add -> is just an add-later, rebalance later (on next tick)) - essentially it should be impossible for the user to get this Concurrent exception (but this can be a followup hardening if we don't)

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,543 @@
...
+///
+/// The contents of this container are managed by the [Component] lifecycle:
+/// use [Component.add], [Component.remove] and their related methods to
+/// modify which components are in it.
+class ComponentList extends Iterable<Component> {
+  ComponentList({this.strictMode = false, this.comparator});
```
luanpotter, line 21: you don't want strictMode to be true by default? :P

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,543 @@
...
+  int _compareOrder(Component a, Component b) {
+    final comparator = this.comparator;
+    if (comparator == null) {
+      final ap = a._priority;
+      final bp = b._priority;
+      return ap < bp
+          ? -1
+          : ap > bp
+          ? 1
+          : 0;
```
luanpotter: no built in compareTo to delegate? :cry:
or even ap-bp

( btw glad you recovered from your past crusades, allowing ap and bp var names 😆 )

packages/flame/lib/src/components/core/component_list.dart
spydon: No, but we should add it as a note in the performance improval hint docs.

packages/flame/lib/src/components/core/component_list.dart
```diff
@@ -0,0 +1,543 @@
...
+    }
+  }
+
+  /// Adds [component] to this list, keeping the priority ordering; returns
+  /// whether the component was added (`false` if it already was in the list).
+  ///
+  /// Adding a component here does not make it go through the component
+  /// lifecycle, which is why this is only reachable from [Component.add].
+  bool _add(Component component) {
+    if (identical(component._containerList, this)) {
```
luanpotter: could call the contains

luanpotter (approved):
LGTM

packages/flame/lib/src/components/core/component_list.dart
luanpotter: aligned

packages/flame/lib/src/components/core/component.dart
```diff
@@ -302,53 +302,52 @@ class Component {
...
-      _children!.rebalanceAll();
-    }
-  }
+  /// This component's index within [_containerList]'s backing list, or -1
+  /// when the component is not in any list. Maintained by [ComponentList].
+  int _containerIndex = -1;
+
+  /// Restores the priority ordering of the [children], after one or more of
+  /// them have changed their [priority].
+  void rebalanceChildren() => _children?.rebalance();
```
luanpotter: we should followup on this - it's a footgun
we can either remove or I have some ideas, but it can be a followup

packages/flame/lib/src/components/core/component.dart
```diff
@@ -799,11 +803,11 @@ class Component {
         // already marked as removed during ancestor removal propagation.
         // The child is now being explicitly removed by user code (e.g.
         // via removeAll(children) in onRemove), so detach it.
-        _internalChildren.remove(child);
+        children._remove(child);
```
luanpotter, line 813: for later - maybe we could consider queueing removals as well, just for consistency and clarity (even though it doesn't cause concurrent issues)

luanpotter (approved):
:shipit:

packages/flame/lib/src/components/core/component_tree_root.dart
```diff
@@ -1,8 +1,4 @@
-import 'dart:async';
-
-import 'package:flame/components.dart';
-import 'package:flame/src/components/core/recycled_queue.dart';
-import 'package:meta/meta.dart';
+part of 'component.dart';
```
luanpotter, line 1: lovely

PR #4041, opened 2026-09-08
fix: Keep TextComponent glyphs in place when its paint changes

### LINKED ISSUE #4014: TextComponent glyphs jump up by one ascent when an OpacityEffect starts
## Description

The first time an `OpacityEffect` (or any `HasPaint` change) updates a `TextComponent`, the rendered glyphs teleport upward by one font ascent and stay there. Combined with a movement effect this reads as a sudden vertical jump in the middle of an otherwise smooth animation.

Related to #4013 but a distinct defect: that one is about shadow alpha, this one is about glyph positioning.

## Cause

`TextComponent.updateBounds` translates the freshly formatted element down by its ascent:

```dart
@internal
void updateBounds() {
  _updateElement();
  final measurements = _textElement.metrics;
  _textElement.translate(0, measurements.ascent);
  size.setValues(measurements.width, measurements.height);
}
```

but `TextComponent.onChanged`, which runs on every paint mutation, rebuilds the element without that translation:

```dart
@override
void onChanged() {
  _textRenderer = _textRenderer.copyWithPaint(paint) as T;
  _updateElement();
}
```

So the element created before the first paint change is ascent-translated, and every element created after a paint change is not, shifting the drawn text up by `metrics.ascent` for the rest of the component's life.

## Steps to reproduce

```dart
final text = TextComponent(
  text: '+1 kr',
  textRenderer: TextPaint(style: const TextStyle(fontSize: 24)),
)
  ..add(MoveByEffect(Vector2(0, -60), EffectController(duration: 0.8)))
  ..add(OpacityEffect.fadeOut(EffectController(duration: 0.5, startDelay: 0.3)));
```

The text rises smoothly for 0.3 seconds, then jumps up by roughly the ascent when the fade begins.

## Expected behavior

Paint changes only affect paint; the glyphs stay where they are.

## Suggested fix

Have `onChanged` apply the same ascent translation as `updateBounds` (or call a shared helper), for example:

```dart
@override
void onChanged() {
  _textRenderer = _textRenderer.copyWithPaint(paint) as T;
  _updateElement();
  _textElement.translate(0, _textElement.metrics.ascent);
}
```

Observed on the `perf/component-set-backing` branch; the implementation is the same on `main`.

Yasser-Ameur:
# Description

The first time an `OpacityEffect` (or any paint change through `HasPaint`) updates a `TextComponent`, the glyphs jump up by one font ascent and stay there (#4014).

`updateBounds` translates the freshly formatted element down by its ascent, but `onChanged` rebuilt the element through `_updateElement` without that translation, so every element created after a paint change was drawn one ascent too high.

The translation now lives in `_updateElement`, which both paths share, so an element is always translated exactly once. `updateBounds` still reads the width and height after the update; `LineMetrics.translate` moves only `left` and `baseline`, so the size is unchanged.

The new test uses a recording `TextRenderer` and asserts that the element created by the constructor and the element created by `setOpacity` both receive the ascent translation. It fails on `main` and passes here.

## Checklist

## Breaking Change?

## Related Issues

Closes #4014

### DIFF
--- a/packages/flame/lib/src/components/text_component.dart
+++ b/packages/flame/lib/src/components/text_component.dart
@@ -43,13 +43,13 @@ class TextComponent<T extends TextRenderer> extends PositionComponent
 
   void _updateElement() {
     _textElement = _textRenderer.format(_text);
+    _textElement.translate(0, _textElement.metrics.ascent);
   }
 
   @internal
   void updateBounds() {
     _updateElement();
     final measurements = _textElement.metrics;
-    _textElement.translate(0, measurements.ascent);
     size.setValues(measurements.width, measurements.height);
   }
 
--- a/packages/flame/test/components/text_component_test.dart
+++ b/packages/flame/test/components/text_component_test.dart
@@ -1,6 +1,7 @@
-import 'package:flame/src/components/text_component.dart';
+import 'package:flame/components.dart';
+import 'package:flame/text.dart';
+import 'package:flutter/painting.dart';
 import 'package:test/test.dart';
-import 'package:vector_math/vector_math.dart';
 
 void main() {
   group('TextComponent', () {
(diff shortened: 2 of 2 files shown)


packages/flame/lib/src/components/text_component.dart
```diff
@@ -43,13 +43,13 @@ class TextComponent<T extends TextRenderer> extends PositionComponent
 
   void _updateElement() {
     _textElement = _textRenderer.format(_text);
+    _textElement.translate(0, _textElement.metrics.ascent);
```
spydon, line 46: I initially thought this would run for every update when it is in the base-case, but I have verified that is not the case.

spydon (approved):

PR #4042, opened 2026-09-08
fix: Fade TextComponent shadows and fill together under OpacityEffect

### LINKED ISSUE #4013: TextComponent shadows are not affected by OpacityEffect
## Description

When an `OpacityEffect` runs on a `TextComponent` whose `TextPaint` style defines `shadows`, the glyph color fades but the shadows keep their original alpha. The shadows stay at full strength for the whole animation, so a fading text leaves a non-fading shadow or glow behind, which reads as a visual glitch.

There is also a related gotcha in the same code path: the first time the effect updates the paint, the style's fill color is replaced entirely by the component's `HasPaint` paint (white by default), so a colored text visibly snaps to white when the fade begins unless the user remembers to seed `paint.color` with the same color as the style.

## Cause

`TextComponent.onChanged` applies opacity through `TextRenderer.copyWithPaint`, and `TextPaint.copyWithPaint` only swaps the style's `foreground`:

```dart
@override
TextRenderer copyWithPaint(Paint paint) {
  return copyWith(
    (style) {
      return style.copyWith(
        foreground: paint,
      );
    },
  );
}
```

`style.shadows` is left untouched, so shadow alpha never follows the component's opacity.

## Steps to reproduce

```dart
final text = TextComponent(
  text: '+1 kr',
  textRenderer: TextPaint(
    style: const TextStyle(
      color: Color(0xFF2E9940),
      shadows: [Shadow(color: Color(0x99FFFFFF), blurRadius: 4)],
    ),
  ),
)..add(OpacityEffect.fadeOut(EffectController(duration: 1)));
```

## Expected behavior

The text and its shadows fade out together, in the text's own color.

## Actual behavior

The text snaps to white when the effect starts (foreground paint replaces the style color) and fades, while the white shadow stays at constant alpha until the component is removed.

## Suggested fix

In `TextPaint.copyWithPaint`, scale each shadow's color alpha by the incoming paint's alpha (and consider multiplying the paint color into the existing style color instead of replacing it, so `OpacityEffect` composes with a styled color the way it does for sprites).

Observed on the `perf/component-set-backing` branch, but the `copyWithPaint` implementation is the same on `main`.

Yasser-Ameur:
# Description

When an `OpacityEffect` runs on a `TextComponent` whose `TextPaint` style defines `shadows`, the glyphs fade but the shadows keep their original alpha, and the first paint change replaces the text colour with the component's default white paint, so coloured text snaps to white as the fade begins (#4013).

`TextPaint.copyWithPaint` only swapped the style's `foreground` for the component's paint, so `style.shadows` never followed the paint's alpha. The component also derived each new renderer from the previous derived one, so any per-update scaling would have compounded.

This PR:

- scales every shadow's alpha by the paint's alpha in `TextPaint.copyWithPaint`, keeping the paint itself as the foreground so colour filters and other paint properties still apply;
- derives every paint change from the renderer the user gave, so the alpha is always relative to the original style and never compounds. `textRenderer` now returns that renderer rather than the last derived copy, which also makes `textRenderer = textRenderer.copyWith(...)` safe after an effect has run; the copy with the paint applied is the new `@internal` `paintedTextRenderer`, used for drawing by `TextComponent` and `TextBoxComponent`;
- seeds the component's paint colour from the style's colour at construction and when `textRenderer` is set, so the fade starts from the text's own colour instead of white. Setting `paint.color` afterwards still overrides it; note that assigning a new `textRenderer` reseeds it.

The new tests run `OpacityEffect.fadeOut` for half its duration and check that the fill keeps its RGB at half alpha and the shadow keeps its RGB at half its own alpha, then that both reach zero; and that a `textRenderer` round-trip between two opacity changes keeps the original colour and shadow and scales only once. Both fail on `main` and pass here.

## Checklist

## Breaking Change?

## Related Issues

Closes #4013

### DIFF
--- a/packages/flame/lib/src/components/text_box_component.dart
+++ b/packages/flame/lib/src/components/text_box_component.dart
@@ -401,7 +401,7 @@ class TextBoxComponent<T extends TextRenderer> extends TextComponent {
         line = line.substring(0, nChars);
       }
 
-      final textElement = textRenderer.format(line);
+      final textElement = paintedTextRenderer.format(line);
       final metrics = textElement.metrics;
 
       final position = Vector2(
--- a/packages/flame/lib/src/components/text_component.dart
+++ b/packages/flame/lib/src/components/text_component.dart
@@ -20,6 +20,8 @@ class TextComponent<T extends TextRenderer> extends PositionComponent
     super.key,
   }) : _text = text ?? '',
        _textRenderer = textRenderer ?? TextRendererFactory.createDefault<T>() {
+    _paintedTextRenderer = _textRenderer;
+    _seedPaintColor();
     updateBounds();
   }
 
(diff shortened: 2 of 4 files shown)


spydon:
This solution doesn't work for the following cases:
- Outlined or gradient text. A style that uses foreground instead of color gets that paint replaced by the component's plain white fill the moment the fade begins. The outline or shader disappears and the text snaps to white. Seeding does nothing here because style.color is null.
- Sprite fonts. SpriteFontRenderer.copyWithPaint still drops scale, letter spacing and the tint, so scaled or colored sprite font text shrinks and loses its color on the first paint change.
- Reassigning the renderer mid-fade shows one frame at full opacity and silently overwrites any paint color the user had set.

And since we already had an agent running for this issue, this PR has been superseded by #4043 which takes these things into consideration.
Thanks for your contribution though.

PR #4045, opened 2026-09-13
perf: Only test polygon edges inside the hitbox overlap and drop allocations from segment intersection

spydon:
# Description

Two changes to the collision narrow phase that reduce the cost of polygon hitboxes, especially ones with many vertices.

**Only test the edges inside the overlapping area.** The polygon intersection systems already accept an `overlappingRect` and filter edges with it through `possibleIntersectionVertices`, but nothing ever passed one in, so every edge of one polygon was tested against every edge of the other. `ShapeHitbox.intersections` now computes the overlap of the two AABBs, inflated by the existing 0.01 tolerance so shapes that only touch still report their shared edge, and hands it to the intersection system. Any intersection point lies inside both AABBs, so edges that do not reach into the overlap cannot contribute one.

**Allocation-free segment intersection with a bounding box early out.** `LineSegment.intersections` built two `Line` objects, a result list, and several temporary vectors for every edge pair. It now rejects pairs whose bounding boxes do not overlap before doing any math and computes the same line intersection inline, allocating only when there is a hit. The parallel and collinear handling is unchanged.

Measured with a collision benchmark of 50 moving, rotating hitboxes of one kind against 50 of another, milliseconds per 20 ticks in JIT:

| Scene | Before | After |
|---|---|---|
| 10-vertex polygon vs circle | 6.7 | 5.4 |
| 10-vertex polygon vs rectangle | 6.2 | 5.1 |
| 10-vertex polygon vs 10-vertex polygon | 10.3 | 5.9 |
| 37-vertex polygon vs circle | 17.5 | 10.2 |
| 37-vertex polygon vs rectangle | 12.4 | 6.8 |
| 37-vertex polygon vs 10-vertex polygon | 16.0 | 5.9 |
| 37-vertex polygon vs 37-vertex polygon | 92.0 | 11.4 |

The 37-vertex polygons come from sampling a `Path` contour, which is what the upcoming path hitbox work produces, but the change is independent of that. #4046 is stacked on this PR and continues from these numbers.

## Checklist

## Breaking Change?


ufrshubham (approved):
LGTM!

PR #4046, opened 2026-09-13
perf: Compose one transform for polygon global vertices and hoist the circle center out of the edge loop

spydon:
# Description

Follow-up to #4045, and stacked on it since the circle change touches the code that PR added. Two more narrow phase costs that showed up once the edge pair work was gone.

**Compose one transform for the global vertices.** `PolygonComponent.globalVertices` called `absolutePositionOf` for every vertex, which walks the ancestor chain and allocates a vector per level per vertex, and then decided whether its cache was valid by comparing a list of cloned vectors from three more ancestor walks. It now walks the ancestors once, composes a single 2D affine transform, writes every vertex into the preallocated list, and uses the six transform values plus the size as the cache key. Whether the polygon is mirrored is read from the determinant instead of the scale signs, which is the same condition for scale and rotation transforms. One call for a 37-vertex polygon under a rotating parent went from 1.24 to 0.19 microseconds.

**Hoist the circle center and radius out of the edge loop.** `CirclePolygonIntersections` called `absoluteCenter` and `scaledRadius` for every polygon edge, each walking the ancestors with allocations. They are now computed once per pair and passed into `lineSegmentIntersections` through new optional parameters, and an edge whose bounding box lies farther than the radius from the center is skipped before the quadratic solve.

Measured with the same benchmark as #4045, milliseconds per 20 ticks in JIT with 50 hitboxes of each kind, starting from that PR:

| Scene | #4045 | Global vertices | Plus circle hoist |
|---|---|---|---|
| 10-vertex polygon vs circle | 5.4 | 4.7 | 3.4 |
| 10-vertex polygon vs rectangle | 5.1 | 3.0 | 2.9 |
| 10-vertex polygon vs 10-vertex polygon | 5.9 | 3.4 | 3.3 |
| 37-vertex polygon vs circle | 10.2 | 8.7 | 4.7 |
| 37-vertex polygon vs rectangle | 6.8 | 4.2 | 4.0 |
| 37-vertex polygon vs 10-vertex polygon | 5.9 | 3.5 | 3.4 |
| 37-vertex polygon vs 37-vertex polygon | 11.4 | 7.8 | 7.7 |

## Checklist

## Breaking Change?


packages/flame/lib/src/collisions/hitboxes/shape_hitbox.dart
luanpotter: this seems like could use a compose aabb method (maybe an ext or something from https://github.com/flame-engine/flame/pull/4048 even)

packages/flame/lib/src/geometry/circle_component.dart
```diff
@@ -116,26 +116,39 @@ class CircleComponent extends ShapeComponent {
   List<Vector2> lineSegmentIntersections(
```
luanpotter, line 117: it is weird to me that a circle.lineSegmentIntersections would need to receive an override for the center and radius, ie the entire identity of a circle provided by [this]

packages/flame/lib/src/geometry/circle_component.dart
```diff
@@ -116,26 +116,39 @@ class CircleComponent extends ShapeComponent {
...
+    Vector2? center,
+    double? radius,
   }) {
+    final effectiveCenter = center ?? absoluteCenter;
+    final effectiveRadius = radius ?? scaledRadius;
+    final from = lineSegment.from;
+    final to = lineSegment.to;
+    // The segment can only reach the circle if its bounding box comes within
+    // the radius of the center.
+    if (min(from.x, to.x) > effectiveCenter.x + effectiveRadius ||
```
luanpotter: another make_aabb candidate

packages/flame/lib/src/geometry/polygon_component.dart
```diff
@@ -187,30 +185,87 @@ class PolygonComponent extends ShapeComponent {
...
-      size,
-      scale,
-      angle,
-    ])) {
+    _composeAbsoluteTransform();
+    final m = _absoluteTransform;
+    final cache = _globalVerticesCacheKey;
+    var isCacheValid = _hasGlobalVertices;
+    if (isCacheValid) {
+      for (var i = 0; i < 6; i++) {
```
luanpotter: there has to be a better way to do this, or at least encapsulate it

spydon:
@luanpotter thanks for the review, all comments should be fixed now :)

PR #4048, opened 2026-09-15
feat: Add support to use convex/concave Path shapes in PolygonComponent, PolygonHitbox and Polygon

### LINKED ISSUE #4040: PolygonRayIntersection failing on non-Euclidean/concave/hollow polygons
### What happened?

When using a `PolygonHitbox` for non-Euclidean/concave/hollow polygons, the `PolygonRayIntersection` mixin fails when a ray traverses multiple hollow parts of a concave polygon.

### What do you expect?

As long as a `PolygonHitbox` defines a closed polygon correctly, the actual topology should not confuse `PolygonRayIntersection`.

While I no longer have steps to reproduce the issue, there is [a testbed](https://github.com/adario/flame_path_shapes) to verify the refactoring result in `PolygonRayIntersection`, which uses my Flame fork mentioned below.

These are the [relevant changes](https://github.com/adario/flame/commit/29755b2defe512c150406b66db70326f2888c82e) in my Flame fork; eventually I plan to submit a PR.

Without [my changes](https://github.com/adario/flame/commit/29755b2defe512c150406b66db70326f2888c82e), the bug used to occur with **all** the new `Path` shapes of the `RaysInShapeExample` in [the testbed](https://github.com/adario/flame_path_shapes); the pre-existing shapes (circle, rectangle, quadrilateral) used to work, but changing the quad into a sort of 'U' shape also failed.

### Relevant log output

```shell

```

Output of: flutter doctor -v

[✓] Flutter (Channel stable, 3.47.2, on macOS 26.6.2 25G83 darwin-arm64, locale en-IT) [881ms]
    • Flutter version 3.47.2 on channel stable at /Users/d.accornero/Applications/flutter
    • Framework revision d3b14c8769 (12 days ago), 2026-08-26 16:07:51 -0700
    • Engine revision a804b26164
    • Dart version 3.13.2
    • DevTools version 2.60.0
    • Feature flags: enable-web, no-enable-linux-desktop, enable-macos-desktop, no-enable-windows-desktop, enable-android, enable-ios, cli-animations, enable-native-assets, enable-record-use,
      enable-swift-package-manager, omit-legacy-version-file, enable-lldb-debugging, enable-uiscene-migration

[!] Android toolchain - develop for Android devices (Android SDK version 36.1.0) [1,658ms]
    • Android SDK at /Users/d.accornero/Library/Android/sdk
    • Emulator version 37.1.11.0 (build_id 15917651) (CL:N/A)
    • Platform android-36.1, build-tools 36.1.0
    • ANDROID_HOME = /Users/d.accornero/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
      This is the JDK bundled with the latest Android Studio installation on this machine.
      To manually set the JDK path, use: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 25.0.3+-15898627-b508.16)
      See https://flutter.dev/to/macos-android-setup for more details.

[✓] Xcode - develop for iOS and macOS (Xcode 26.6) [1,716ms]
    • Build 17F113
    • CocoaPods version 1.17.0

[✓] Chrome - develop for the web [6ms]

[✓] Connected device (4 available) [6.6s]
    • Undici (mobile)     • 00008030-001065060C42802E            • ios            • iOS 26.6.1 23G83
    • iPhone 17e (mobile) • FD5942B7-8265-4568-B74E-698F1FB20941 • ios            • com.apple.CoreSimulator.SimRuntime.iOS-26-5 (simulator)
    • macOS (desktop)     • macos                                • darwin-arm64   • macOS 26.6.2 25G83 darwin-arm64
    • Chrome (web)        • chrome                               • web-javascript • Google Chrome 152.0.7977.83
    ! Error: Browsing on the local area network for iPad di Sviluppo (2). Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
      The device must be opted into Developer Mode to connect wirelessly. (code -27)
    ! Error: Browsing on the local area network for Eleven. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
      The device must be opted into Developer Mode to connect wirelessly. (code -27)
    ! Error: Browsing on the local area network for iPad di Julie. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
      The device must be opted into Developer Mode to connect wirelessly. (code -27)

[✓] Network resources [285ms]

### Affected platforms

iOS, Android, Web

### Other information

_No response_

- [x] I want to work on this

adario:
# Description

The purpose of this PR is to add support for using both convex and concave `Path` shapes when creating `PolygonComponent`, `PolygonHitbox` and `Polygon` objects: the contours of a `Path` object are sampled and converted to a list of polygon vertices. Support for concave polygons, described below, is available in [`PolygonRayIntersection`](https://github.com/adario/flame/blob/feat/path-shapes/packages/flame/lib/src/geometry/polygon_ray_intersection.dart).

A complete testbed for all changes is [available here](https://github.com/adario/flame_path_shapes); the relevant examples have been ported here too.

The changes add to and introduce a few extensions:

- [`PathExtension`](https://github.com/adario/flame/blob/feat/path-shapes/packages/flame/lib/src/extensions/path.dart) on `Path` gets `contours`, `walkContours`, `walkContourAt`, `resizeTo`, `toOrigin` and `centered`
- [`PathMetricExtension`](https://github.com/adario/flame/blob/feat/path-shapes/packages/flame/lib/src/extensions/path.dart) on `PathMetric` has `walkContour`
- [`PathMetricListExtension`](https://github.com/adario/flame/blob/feat/path-shapes/packages/flame/lib/src/extensions/path.dart) on `List<PathMetric>` has `contoursLength`
- [`OffsetListExtension`](https://github.com/adario/flame/blob/feat/path-shapes/packages/flame/lib/src/extensions/offset.dart) on `List<Offset>` has `vertices` and `rectangle`
- [`RectExtension.fromOffsets`](https://github.com/adario/flame/blob/74046f1da478efc8e840506cccf9f3907ecf9282/packages/flame/lib/src/extensions/rect.dart#L62) and [`Aabb2Extension.fromVertices`](https://github.com/adario/flame/blob/74046f1da478efc8e840506cccf9f3907ecf9282/packages/flame/lib/src/extensions/aabb.dart#L8) create the bounds of a list of offsets and of vertices

The `walkContours`, `walkContourAt` and `walkContour` methods extract the vertex lists from the contours of a `Path`:

- The contour is walked with strides that follow how much it bends, so straight stretches cost a few lookups and tight curves are followed below the step. The `granularity` parameter is that step, which defaults to `1.0`.
- The corners between straight lines are placed exactly, and the points where the contour reaches its bounds are kept, so that the polygon has the size of the contour.
- The samples that are not needed to stay within the `tolerance` are removed, which defaults to half of the `granularity`. This also works across the point where a closed contour starts and ends.

Direct construction from `Path` objects is available via:

- [`PolygonComponent.fromPath`](https://github.com/adario/flame/blob/feat/path-shapes/packages/flame/lib/src/geometry/polygon_component.dart)
- [`PolygonHitbox.fromPath`](https://github.com/adario/flame/blob/feat/path-shapes/packages/flame/lib/src/collisions/hitboxes/polygon_hitbox.dart)
- [`Polygon.fromPath`](https://github.com/adario/flame/blob/feat/path-shapes/packages/flame/lib/src/experimental/geometry/shapes/polygon.dart)

All three take the index of the `contour` to use, the first one by default, and the `granularity`, and only walk that contour of the path.

As long as the `Path` contours represent valid polygons, the `PolygonRayIntersection` mixin now supports both convex and concave polygons. Whether the ray starts inside of the polygon is decided by the parity of the edges that it crosses, instead of their absolute quantity. The hits and the crossings come from the same test of which side of the ray each vertex is on, which both edges of a vertex share, so a ray through or next to a vertex is not misjudged and can not slip through between two edges.

`isSolid` is now an argument of the `ShapeComponent` constructor, which the polygon, rectangle and circle components and their hitboxes pass on instead of assigning it in their constructor bodies.

The shape components and collision detection pages of the documentation describe the new constructors, and no longer say that polygon hitboxes have to be convex.

In terms of performance, @spydon was kind enough to provide the new [`path_collision`](https://github.com/adario/flame/blob/feat/path-shapes/packages/flame/benchmark/path_collision_benchmark.dart) and [`path_contour`](https://github.com/adario/flame/blob/feat/path-shapes/packages/flame/benchmark/path_contour_benchmark.dart) benchmarks (thank you very much, Lukas.) There is also a new [`ray_intersection`](https://github.com/adario/flame/blob/feat/path-shapes/packages/flame/benchmark/ray_intersection_benchmark.dart) benchmark for rays against concave and convex contour hitboxes.

## Checklist

## Breaking Change?

## Related Issues

Closes #4040


packages/flame/lib/src/collisions/hitboxes/polygon_hitbox.dart
```diff
@@ -55,6 +55,22 @@ class PolygonHitbox extends PolygonComponent
...
+    super.position,
+    super.angle,
+    super.anchor,
+    super.contour,
+    super.granularity,
+    bool isSolid = false,
+    CollisionType collisionType = CollisionType.active,
+  }) : super.contour() {
+    this.isSolid = isSolid;
+    this.collisionType = collisionType;
```
spydon: How come these need to be assigned in the constructor body?

packages/flame/lib/src/extensions/offset.dart
```diff
@@ -18,3 +18,58 @@ extension OffsetExtension on Offset {
...
+    }
+    return false;
+  }
+
+  /// Returns itself as a fixed list of [Vector2] objects.
+  List<Vector2> get vertices =>
+      map((o) => o.toVector2()).toList(growable: false);
+
+  /// Returns the approximate enclosing rectangle.
+  Rect get rectangle {
```
spydon: I wonder if [AABB](https://pub.dev/documentation/vector_math/latest/vector_math/Aabb2-class.html) should be used here instead? 🤔

packages/flame/lib/src/extensions/offset.dart
```diff
@@ -18,3 +18,58 @@ extension OffsetExtension on Offset {
...
+      }
+    });
+    return r;
+  }
+}
+
+extension VerticesList on List<List<Offset>> {
+  /// Returns the given subcontour as a vertices list.
+  List<Vector2> getVertices([int index = 0]) {
+    assert(index >= 0 && index < length, 'Ivalid subcontour index $index');
```
spydon:
```suggestion
    assert(index >= 0 && index < length, 'Invalid subcontour index $index');
```

spydon (commented):
Very exited for this!

adario:
@spydon thank you for your comments: I'll address them after reviewing the clanker's verdict, which at least in [this case](https://github.com/flame-engine/flame/pull/4048#discussion_r4016053033) warrants a debugger trip...

adario:
> @spydon thank you for your comments: I'll address them after reviewing the clanker's verdict, which at least in [this case](https://github.com/flame-engine/flame/pull/4048#discussion_r4016053033) warrants a debugger trip...

Aye, looks like a false flag... the suggestion is valid (should use `Rect.zero` here), but `r.isEmpty` on first run _is_ `true`...


r isEmpty

packages/flame/lib/src/extensions/offset.dart
```diff
@@ -18,3 +18,58 @@ extension OffsetExtension on Offset {
...
+  }
+
+  /// Returns itself as a fixed list of [Vector2] objects.
+  List<Vector2> get vertices =>
+      map((o) => o.toVector2()).toList(growable: false);
+
+  /// Returns the approximate enclosing rectangle.
+  Rect get rectangle {
+    const epsilon = 1e-6;
+    var r = Rect.fromCenter(center: .zero, width: 0, height: 0);
```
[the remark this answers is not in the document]
adario: This is wrong, since a debugger trip reveals that in the first iteration `r.isEmpty` is indeed `true`, but I'll apply the suggested change since it's definitely cleaner.

packages/flame/lib/src/extensions/path.dart
```diff
@@ -9,4 +12,202 @@ extension PathExtension on Path {
...
+    );
+    final box = getBounds();
+    final scale = Vector2(size.width / box.width, size.height / box.height);
+    if (keepRatio) {
+      if (box.height > box.width) {
+        scale.setValues(scale.y, scale.y);
+      } else {
+        scale.setValues(scale.x, scale.x);
+      }
+    }
```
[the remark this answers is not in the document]
adario: Fixed in commit https://github.com/adario/flame/commit/76d8f0034fbc8234d6bba4284c8e220168cf29fb

packages/flame/lib/src/geometry/polygon_component.dart
```diff
@@ -93,6 +93,48 @@ class PolygonComponent extends ShapeComponent {
...
+    Anchor? anchor,
+    int? priority,
+    Paint? paint,
+    List<Paint>? paintLayers,
+    ComponentKey? key,
+    List<Component>? children,
+  }) : this(
+         pathContourToVertices(path, granularity, contour),
+         position: position,
+         size: path.getBounds().size.toVector2(),
```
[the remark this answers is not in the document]
adario: The quick fix (calling `pathContourToVertices` twice, the second time to compute the contour vertices size) would be quite expensive as-is... on the other hand, stashing vertices in a static map wouldn't be safe when using `isolate`. I'm working on an alternative approach, which won't be free but shouldn't be as expensive as calling `pathContourToVertices` twice.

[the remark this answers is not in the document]
adario: The safest/cheapest solution seems to be defaulting to compute the contour size [automatically](https://github.com/adario/flame/commit/d4318dd1c42eaa85c08193a894dcea9608196a9b).

packages/flame/lib/src/collisions/hitboxes/polygon_hitbox.dart
adario: AFAICT, this occurs everywhere in [`PolygonHitbox`](https://github.com/adario/flame/blob/d4318dd1c42eaa85c08193a894dcea9608196a9b/packages/flame/lib/src/collisions/hitboxes/polygon_hitbox.dart#L19)... looks like `PolygonComponent` [isn't exposing](https://github.com/adario/flame/blob/d4318dd1c42eaa85c08193a894dcea9608196a9b/packages/flame/lib/src/geometry/polygon_component.dart#L40) `isSolid` either: I'll add it.

packages/flame/benchmark/path_contour_benchmark.dart
```diff
@@ -0,0 +1,401 @@
...
+      final suffix = contours.length > 1
+          ? '  (${contours.length} contours)'
+          : '';
+      print(
+        '${pathContourShapeNames[index].padRight(11)} '
+        '${granularity.toStringAsFixed(1).padLeft(11)}  '
+        '${polygon.length.toString().padLeft(8)}  '
+        '${tangentCalls.toString().padLeft(13)}  '
+        '${microseconds.toStringAsFixed(0).padLeft(22)}  '
+        '${_maxError(path, polygon).toStringAsFixed(2).padLeft(14)}  '
```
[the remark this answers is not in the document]
adario: @spydon will you look at this, or shall I?

packages/flame/lib/src/collisions/hitboxes/polygon_hitbox.dart
adario: I've [added](https://github.com/adario/flame/commit/e7fc5202a9db73b64c135993bc7dde6f307efb42) `isSolid` with propagation; however, since the whole `PolygonHitbox` is currently declaring the `collisionType` parameter everywhere the same, I'll leave that decision to you.
Honestly, I'm not sure how to make the constructors look cleaner, what with the current `collisionType` property structure ...

packages/flame/lib/src/extensions/offset.dart
adario: > I wonder if [AABB](https://pub.dev/documentation/vector_math/latest/vector_math/Aabb2-class.html) should be used here instead? 🤔

Aye, you're right: my approach is admittedly 15th century, but I haven't managed to get the same results with `Aabb2`...
I've tried this:

```
    final box = Aabb2();
    forEach((offset) => box.hullPoint(offset.toVector2()));
    return Rect.fromLTRB(box.min.x, box.min.y, box.max.x, box.max.y);
```

I'm sure I've missed something... 

alien2
setup
recycle

adario:
@spydon apart from the earlier (open) comments, in the benchmarks there's a couple of [spellcheck issues](https://github.com/flame-engine/flame/actions/runs/35034779273/job/104601296192?pr=4048), and a failure in [polygon_component_contour_test](https://github.com/flame-engine/flame/actions/runs/35034779248/job/104601295959?pr=4048)... please let me know how to proceed.

packages/flame/lib/src/extensions/offset.dart
spydon: But do you need a Rect at all? Can't you just use the Aabb2?

adario: > But do you need a Rect at all? Can't you just use the Aabb2?

Just need the size of the area encompassing all points: I agree that `rectangle` looks geriatric at the moment, but from the screenshots there's a difference with the `Aabb2` results...

Maybe my `Aabb2` calculation is wrong, but I can't spot the problem here:

```
    final box = Aabb2();
    forEach((offset) => box.hullPoint(offset.toVector2()));
    return Rect.fromLTRB(box.min.x, box.min.y, box.max.x, box.max.y);

```

packages/flame/lib/src/collisions/hitboxes/polygon_hitbox.dart
adario: If you have a cleaner solution for passing `collisionType` throughout `PolygonHitbox`, I'd be more than happy to use it... otherwise, perhaps we could resolve this issue?

packages/flame/lib/src/extensions/offset.dart
adario: Hmm, while I've fixed the obvious bits (`subcontour` and `ivalid`), the spellcheck pipeline phase still complains about `Ramer-Douglas-Peucker` in both the `Path` extensions as well as in your benchmarks...

adario: Closing this: I had to use `RDP` instead of `Ramer-Douglas-Peucker`.

adario:
@spydon pipeline's almost clean, but there's still one failure in [polygon_component_contour_test](https://github.com/adario/flame/blob/3dca59eb8ab7efca7d74807b2f1bf34b3986ea15/packages/flame/test/geometry/polygon_component_contour_test.dart#L18):

```
00:01 +0 -1: contour preserves the path bounds [E]
  Expected: a numeric value within <1e-10> of <64>
    Actual: <63.997955322265625>
     Which:  differs by <0.002044677734375>
  
  package:matcher                                         expect
  test/geometry/polygon_component_contour_test.dart 18:5  main.<fn>
```

This is due to RDP simplification: without RDP the test passes, even with large `granularity` values (e.g. 4). Honestly, I'm not sure how this test is supposed to pass if the polyline gets simplified: using RDP implies accepting some deformation in the topology... previously, it was passing only because `PolygonComponent.contour` explicitly (and erroneously) assigned the whole `Path` size to the contour, whereas currently this occurs via `shrinkToBounds`.

Please let me know how to proceed: I've got other `Path`-related changes I'd like to show you... ;-)

packages/flame/lib/src/extensions/offset.dart
adario: > > But do you need a Rect at all? Can't you just use the Aabb2?
> 
> Just need the size of the area encompassing all points: I agree that `rectangle` looks geriatric at the moment, but from the screenshots there's a difference with the `Aabb2` results...
> 
> Maybe my `Aabb2` calculation is wrong, but I can't spot the problem here:
> 
> ```
>     final box = Aabb2();
>     forEach((offset) => box.hullPoint(offset.toVector2()));
>     return Rect.fromLTRB(box.min.x, box.min.y, box.max.x, box.max.y);
> ```

@spydon My bad, obviously my fault... the `Aabb2` must be [inited properly](https://github.com/adario/flame/commit/394d182cbda7b88345d0838580048f180fc45844).

Not sure whether this change warrants a performance check, but the [`Aabb2.fromOffsets`](https://github.com/adario/flame/blob/394d182cbda7b88345d0838580048f180fc45844/packages/flame/lib/src/extensions/aabb.dart#L8) approach might end up being cheaper than the equivalent [`RectExtension.getBounds`](https://github.com/adario/flame/blob/394d182cbda7b88345d0838580048f180fc45844/packages/flame/lib/src/extensions/rect.dart#L108): clearly they're both linear, but `RectExtension.getBounds` iterates several times...

spydon: Yeah, definitely warrants a performance check, because Aabb2 might be more expensive to initialize.

packages/flame/benchmark/path_contour_benchmark.dart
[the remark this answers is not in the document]
spydon: Fixed!

packages/flame/lib/src/collisions/hitboxes/polygon_hitbox.dart
spydon: Aah, right, `collisionType` has a setter on it, that makes sense then.
 I fixed up so that `isSolid` lives directly in the constructor at least.

adario: Looking good! Closing this.

packages/flame/lib/src/extensions/path.dart
```diff
@@ -102,107 +100,176 @@ extension Contour on PathMetric {
       tolerance == null || (tolerance.isFinite && tolerance >= 0),
       'The tolerance can not be negative: $tolerance',
     );
+    if (!(length > 0)) {
```
adario: Maybe reverse this to `if (length <= 0)`...?

packages/flame/lib/src/extensions/path.dart
```diff
@@ -273,6 +265,393 @@ extension Contour on PathMetric {
...
+/// Samples a contour so that the polyline through [points] stays within
+/// [_maxDeviation] of it, using as few tangent lookups as possible.
+///
+/// The stride to the next sample is predicted from how much the contour bent
+/// over the previous one, so that straight stretches are crossed in a few
+/// lookups while tight curves are followed below the step. A sample that turns
+/// out to be too far ahead is kept in [_ahead] until the walk reaches it. The
+/// indices of the corners and of the points where the contour reaches its
+/// bounds are collected in [anchors].
+class _ContourSampler {
```
adario, line 328: Love this!

packages/flame/lib/src/extensions/path.dart
```diff
@@ -75,6 +75,26 @@ extension PathContours on Path {
         metric.walkContour(granularity, tolerance),
     ];
   }
+
+  /// Walk only the contour at [index], without sampling the other contours of
+  /// the path.
+  ///
+  /// See [Contour.walkContour] for the [granularity] and [tolerance]
+  /// parameters.
+  OffsetList walkContourAt(
```
adario: Beat me to this... looking great!

packages/flame/lib/src/extensions/rect.dart
```diff
@@ -58,12 +57,30 @@ extension RectExtension on Rect {
...
 
-  /// Create a [Rect] from an [offsets] list.
+  /// Create the smallest [Rect] that has all of the [offsets] within it or on
+  /// its edges.
   static Rect fromOffsets(List<Offset> offsets) {
     if (offsets.isEmpty) {
       return .zero;
     }
-    return Aabb2Extension.fromVertices(offsets.vertices).toRect();
+    var left = offsets.first.dx;
```
adario, line 66: Looking good, basically the same approach as `computeAabb()`.

packages/flame/lib/src/extensions/path.dart
adario: [Done](https://github.com/adario/flame/commit/f43c556a649ce33dd8eaeb980c5d8fc363f23468), closing this.

packages/flame/lib/src/extensions/path.dart
```diff
@@ -132,7 +132,7 @@ extension PathMetricExtension on PathMetric {
       tolerance == null || (tolerance.isFinite && tolerance >= 0),
       'The tolerance can not be negative: $tolerance',
     );
-    if (!(length > 0)) {
+    if (length <= 0) {
```
spydon, line 142: Haha, much better!

adario:
@spydon I've updated all affected examples, and added path shapes in `RaytraceExample`.

adario:
Marking ready for review after @spydon gave the green light.

erickzanardo (approved):

adario:
@erickzanardo thank you very much for approving the PR: I imagine @spydon will merge this...

spydon:
I'd love a review from @luanpotter on this one too before merging :)

packages/flame/test/geometry/polygon_ray_intersection_test.dart
```diff
@@ -0,0 +1,286 @@
...
+Path _roundRectPath(Size size) {
+  return Path()..addRRect(
+    RRect.fromRectAndRadius(
+      Rect.fromLTWH(0, 0, size.width, size.height),
+      Radius.circular(size.shortestSide * 0.25),
+    ),
+  );
+}
+
+Path _flamePath() {
```
luanpotter: is this the flame logo? 🤯
wondering if we could use one from the test fixtures or move this one there as well

packages/flame/test/extensions/path_test.dart
```diff
@@ -0,0 +1,355 @@
+import 'dart:math';
+import 'dart:ui';
+
+import 'package:flame/extensions.dart';
+import 'package:test/test.dart';
+
+double _area(List<Offset> polygon) {
```
luanpotter, line 7: is this something we want to expose as an extension as well?
otherwise weird to have a test to test a private test impl

packages/flame/test/extensions/aabb_test.dart
```diff
@@ -55,6 +55,34 @@ void main() {
...
+          Offset(
+            aabbSize.width * r.nextDouble(),
+            aabbSize.height * r.nextDouble(),
+          ),
+      ];
+
+      final vertices = offsets.vertices;
+
+      final aabb2 = Aabb2Extension.fromVertices(vertices);
+      final verticesBounds = RectExtension.getBounds(vertices);
```
luanpotter: I find this kind of test a bit confusing, is it basically asserting the `Aabb2Extension.fromVertices` implementation is the same as `RectExtension.getBounds`? instead if could use fixed values and expect the exact response.

packages/flame/test/collisions/screen_hibox_test.dart
```diff
@@ -8,6 +11,32 @@ import 'collision_test_helpers.dart';
 
 void main() {
   group('ScreenHitbox', () {
+    runCollisionTestRegistry({
+      'does not collide when a contour is fully contained':
```
luanpotter: would be good to then move it so they do collide and check

packages/flame/test/collisions/screen_hibox_test.dart
```diff
@@ -8,6 +11,32 @@ import 'collision_test_helpers.dart';
...
+            final game = hasCollisionDetection as FlameGame;
+            final path = Path()
+              ..addRRect(
+                RRect.fromRectAndRadius(
+                  const Rect.fromLTWH(0, 0, 64, 64),
+                  const Radius.circular(10),
+                ),
+              );
+            final block = TestBlock(
+              Vector2(100, 100),
```
luanpotter: nit: Vector2.all

packages/flame/lib/src/geometry/polygon_component.dart
```diff
@@ -89,6 +91,40 @@ class PolygonComponent extends ShapeComponent {
...
          paintLayers: paintLayers,
          shrinkToBounds: shrinkToBounds,
+         isSolid: isSolid,
+         key: key,
+         children: children,
+       );
+
+  /// With this constructor you create a [PolygonComponent] from the given
+  /// [contour] (the first by default) of a [Path], with an optional
+  /// [granularity].
```
luanpotter: maybe briefly explain what [granularity] is, like a sampling rate from the Path definition?

packages/flame/lib/src/extensions/rect.dart
```diff
@@ -55,6 +57,32 @@ extension RectExtension on Rect {
     ];
   }
 
+  /// Create the smallest [Rect] that has all of the [offsets] within it or on
+  /// its edges.
+  static Rect fromOffsets(List<Offset> offsets) {
```
luanpotter: I found the name to be confusing at first, how about smallestContaining or something similar?

packages/flame/lib/src/extensions/offset.dart
```diff
@@ -18,3 +19,12 @@ extension OffsetExtension on Offset {
...
   Rect toRect() => Rect.fromLTWH(0, 0, dx, dy);
 }
+
+extension OffsetListExtension on List<Offset> {
+  /// Returns itself as a fixed list of [Vector2] objects.
+  List<Vector2> get vertices =>
+      map((o) => o.toVector2()).toList(growable: false);
+
+  /// Returns the enclosing area as a [Rect].
+  Rect get rectangle => RectExtension.fromOffsets(this);
```
luanpotter: how about enclosingRectangle, or something with aabb

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
```diff
@@ -35,6 +34,18 @@ class Polygon extends Shape {
     }
   }
 
+  /// Constructs the polygon from the given [contour] of a [Path]
+  /// (the first by default).
+  /// The [granularity] parameter controls the amplitude of the sampling step:
+  /// higher values produce fewer samples.
+  factory Polygon.fromPath(
```
luanpotter, line 48: since a path is an arbitrary curve, I think the [granularity] param is doing some heavy lifting here - would approximate or sample Path be a more clear name?

packages/flame/benchmark/path_contour_benchmark.dart
```diff
@@ -0,0 +1,324 @@
...
+double _maxError(PathMetric contour, List<Offset> polygon) {
+  var worst = 0.0;
+  for (var distance = 0.0; distance < contour.length; distance += 0.25) {
+    final position = contour.getTangentForOffset(distance)!.position;
+    worst = max(worst, _distanceToPolygon(position, polygon));
+  }
+  return worst;
+}
+
+double _distanceToPolygon(Offset point, List<Offset> polygon) {
```
luanpotter: again it is weird to me a bench would just re-implement these (if they are already provided on src) or care about it (if they are not)

examples/lib/stories/collision_detection/rays_in_shape_example.dart
```diff
@@ -28,28 +60,298 @@ shape it will be gray.
...
+    _rotateButton.isDisabled = isCircle;
+    camera.viewport.addAll([_rotateButton, _raysButton, _shapeButton]);
+  }
+
+  Vector2 get halfSize => size * 0.5;
+  bool get isCircle => world.isCircle;
+
+  ButtonColors _getColorsFor(Color color) {
+    final disabledColor = color.withValues(alpha: 0.5);
+    Color downColor;
```
luanpotter: could be a switch-expression or a Map

examples/lib/commons/rounded_rect_component.dart
```diff
@@ -0,0 +1,15 @@
+import 'package:flame/components.dart';
+import 'package:flame/extensions.dart';
+import 'package:flutter/material.dart';
+
+/// A rectangle whose corners are rounded with half of its height, so that
+/// its short sides are semicircles.
+class RoundedRectComponent extends PositionComponent with HasPaint {
```
luanpotter, line 7: does it make sense for it be a ShapeComponent so it inherits some shared goodies (like hitbox)

examples/lib/commons/paths.dart
```diff
@@ -0,0 +1,135 @@
...
+  ..style = PaintingStyle.stroke
+  ..strokeWidth = 3
+  ..strokeCap = .round
+  ..strokeJoin = .bevel;
+
+/// The stroke paints of something that can be hovered and dragged.
+///
+/// Paints are expensive to create, so the three of them are created once
+/// and picked by state with [forState] whenever they are needed.
+class StatePaints {
```
luanpotter: how about InteractableStatePaints ?

examples/lib/commons/path_component.dart
```diff
@@ -0,0 +1,129 @@
...
+  final bool loadHitboxes;
+
+  /// Whether the hitboxes are rendered or not (the default).
+  final bool renderHitboxes;
+
+  /// Whether the hitboxes are filtered to include only disjoint ones.
+  final bool filterHitboxes;
+
+  /// The paint used to render the hitboxes.
+  final Paint? hitboxesPaint;
```
luanpotter, line 56: these are a lot of fields, some seem duplicative, like what does renderHitboxes=false with hitboxesPaint=null mean? others make me question if they are needed, ie to do we need to be able to control when hitboxes are loaded or added? can't we be opinionated about it?

luanpotter (approved):
Overall, LGTM, nice addition! Left a few comments and questions, but nothing blocking

packages/flame/test/geometry/polygon_ray_intersection_test.dart
adario: I couldn't find our "official" logo as an .svg, but I'd love to use it if available...

packages/flame/test/extensions/path_test.dart
adario: I would love expose this from `Polygon`, could be useful outside of testing, but here it is used directly on path contours, so it would have to be exposed from `OffsetListExtension`...

packages/flame/test/extensions/aabb_test.dart
adario: Sure, so it should use a fixed set of points instead of random ones?

packages/flame/test/collisions/screen_hibox_test.dart
adario: Sounds good but this is @spydon's territory... ;-)

packages/flame/test/collisions/screen_hibox_test.dart
adario: Addressed [here](https://github.com/flame-engine/flame/pull/4048/changes/ea20b0aa66562223ba7c1a1544facb15e7575238).

packages/flame/lib/src/geometry/polygon_component.dart
adario: Addressed [here](https://github.com/flame-engine/flame/pull/4048/changes/5fff3dbf22566a9bdcd0d7d79e9e292206f17927).

packages/flame/lib/src/extensions/rect.dart
adario: Addressed [here](https://github.com/flame-engine/flame/pull/4048/changes/cd93bac521eb2314ec200ad55c513819ac2ff069).

packages/flame/lib/src/extensions/offset.dart
adario: Addressed [here](https://github.com/flame-engine/flame/pull/4048/changes/8342f75867b82fb18eafd6f403e54ded0a24d365).

packages/flame/lib/src/experimental/geometry/shapes/polygon.dart
adario: I have [renamed](https://github.com/flame-engine/flame/pull/4048/changes/99b73e165dab0cffc98d88688058dcfb4a20f108) `granularity` to `sampling` everywhere; did you also mean to rename the `Polygon` constructor itself...?

packages/flame/benchmark/path_contour_benchmark.dart
adario: 'tis @spydon's turf...

examples/lib/stories/collision_detection/rays_in_shape_example.dart
adario: Of course, my bad... addressed [here](https://github.com/flame-engine/flame/pull/4048/changes/bb587c1b8c9afa543f706c87bcda158d1d44ebc9).

examples/lib/commons/rounded_rect_component.dart
adario: It could certainly be useful; currently, this is only used for button display in `RaysInShapeExample` and `AdvancedButtonExample`.

examples/lib/commons/paths.dart
adario: @spydon's turf...

examples/lib/commons/path_component.dart
adario: Right, so... we could definitely simplify this:

- with the default rendering parameters (`renderHitboxes = false` and `hitboxesPaint = null`), only the `Path` itself is rendered: the hitboxes are used just for collision detection
- it's true that using a valid `hitboxesPaint` with `renderHitboxes = false` would not be meaningful
- in the opposite situation, the hitboxes would be rendered with a default `Paint`
- with the default hitboxes parameters (`addHitboxes = false` and `loadHitboxes = true`), the hitboxes are created and added in `onLoad`
- with `addHitboxes = true` the hitboxes would be added at construction time: nothing would happen in `onLoad`

I understand this appears to suffer from 'featuritis': the parameters try to cater to whatever the users' preferences are, but we can remove all those looking intrusive and/or misguided: please let me know and I'll make the required changes.

adario:
> Overall, LGTM, nice addition! Left a few comments and questions, but nothing blocking

Thank you very much, Luan. :-)

packages/flame/test/geometry/polygon_ray_intersection_test.dart
adario: Addressed [here](https://github.com/flame-engine/flame/pull/4048/changes/f7164809dfad866cdf02455adea6853760b59916), but would still love to get me hands on the .svg file containing our official logo, if there is such a file.

examples/lib/commons/path_component.dart
adario: This should probably be discussed  in PR #4050...

adario:
Apologies for the latest erroneous merge and subsequent revert... *blush*

examples/lib/commons/rounded_rect_component.dart
spydon: I think it is overkill, since it is just a seven line example helper that only draws button backgrounds in two examples.

spydon (approved):
Thanks for this massive undertaking, this will be really useful!

